PR #9, opened 2026-06-16
ci: publish to npm via changesets

pranavosu:
## Problem

The `main` release published to a private S3/CloudFront registry through a custom script and manifest. We want releases on the public npm registry via Changesets, cut on a predictable cadence rather than on every merge, plus a couple of lightweight changeset policies enforced on PRs.

## Changes

**Release workflow (`publish-packages.yml`)**
- Publishes to npm with `changeset publish` (registry-aware: skips versions already on npm, tags each release). npm provenance enabled (`NPM_CONFIG_PROVENANCE=true`, `id-token: write`).
- Runs on a schedule — weekdays at 09:00 EST (`0 14 * * 1-5`; cron is UTC and does not observe DST, so 10:00 during EDT) plus manual `workflow_dispatch`. A guard step rejects any run whose ref is not `main`.
- Rebases onto `main` before publishing, so `changeset publish` tags the final commit; the version commit and tags are pushed afterward.
- Removed the S3 publish, AWS credential step, manual tag loop, and S3 manifest.

**Changeset PR checks (`changeset-check.yml` + `scripts/changeset-guard.ts`)**
- Consolidated `verify-changeset-coverage.ts` into a single `changeset-guard.ts` with two subcommands: `verify-coverage` (changed packages must have a changeset) and `block-major` (no `major` bumps while pre-1.0).
- The guard uses only Node built-ins, so the check runs it with `node --experimental-strip-types` — no `npm ci`, no `tsx`. The job is just checkout + setup-node + the two guards, and needs only `contents: read`.

**Revert last release (`revert-release.yml` + `scripts/prepare-revert-release.mjs`)**
- Manually-triggered (`workflow_dispatch`, main-only) job that restores `packages/` source to the previous release and opens a changeset PR. Merging it publishes a superseding patch — npm is immutable, so this rolls forward rather than un-publishing.
- Reverts only changed *source* files (leaves `package.json` and `CHANGELOG.md` alone so versions keep advancing), and writes a `patch` changeset for the affected packages. Release boundaries are the `chore: version packages` commits.
- Guards against reverting work merged since the last release (override with `force`). Stops at opening a PR — a human gates the actual publish. The script uses only Node built-ins.

**Other**
- Removed the separate manual publish workflow (`publish-npm.yml`) and the `scripts/publish/publish-npm.ts` wrapper; manual releases now run via `workflow_dispatch` on `publish-packages.yml` (main-only), with `changeset publish` inlined.
- Replaced the `publish:dry-run` npm script with `publish:local` and updated all call sites (`canary.yml`, `pr-checks.yml`, `windows-e2e.yml`, `test-templates-e2e.sh`, `agent-bench/steps/1-init-bench-app.sh`, `scripts/ci/windows-template-e2e.mjs`).
- `.changeset/config.json`: added a few test-app packages to the ignore list.

## Validation

- `npm run lint`, `npm run lint:deps`, and `npm run build` pass locally.
- `changeset-guard.ts` subcommands run via `node --experimental-strip-types` (Node 22); the `verify-coverage` failure path is exercised (a changed package with no changeset exits non-zero).
- `prepare-revert-release.mjs` validated with a dry run and a real apply against a source-changing release boundary (reverts correct files, generates the changeset), plus the no-op path when a release changed no source.
- Release job is no-op safe: nothing happens when there are no pending changesets.

## Checklist

- [x] PR description included

---


.github/workflows/publish-packages.yml
```diff
@@ -1,32 +1,39 @@
...
+# Real npm release, run daily on a schedule or manually. Canary publishes live
+# in publish-canary.yml.
+
 on:
-  push:
-    branches: [main]
+  schedule:
+    # 09:00 EST, Mon–Fri. GitHub cron is UTC and does not observe DST, so this is
+    # 10:00 local during EDT.
+    - cron: '0 14 * * 1-5'
```
svidgen: I had suggested M-F publishing. But, do we want to forgo the publish on Friday? For US folks this is in the AM and gives us plenty of time to react. For European folks, this could ruin a weekend.

Along those lines, would it be possible to add a quick rollback runbook or action for when things go wrong? This could be a GH Action to that accepts a commit, date, or version (not necessarily all three, but a few options might be good). Presumably, it would just revert changes made during that time period and then immediately publish. Should require confirmation to do so.

PR #13, opened 2026-06-17
feat(hosting): explicit spaFallback + multi-page Astro support

osama-rizk:
## What

Static **multi-page** sites (Astro static, SSGs) were misclassified as SPAs: the L3 inferred SPA-vs-multi-page from *"does the deploy have error pages?"*, so a multi-page site without a custom `404.html` got SPA fallback — every extensionless path (`/about`, `/posts`) rewrote to `/index.html` and rendered the **home page**.

## Fix

- **Explicit signal**: add `staticAssets.spaFallback` to the `DeployManifest` contract. The L3 honors it and only falls back to the old heuristic when an adapter doesn't set it (backward-compatible).
- **Adapters declare the routing model**: Astro static → `spaFallback: false` (always multi-page); SPA adapter → sources it from the framework contract (`spa`→true, `static`→false).
- **Built-in default 404**: multi-page static sites that ship no `404.html` now get a branded default 404 at **HTTP 404** (precedence: user `errorPages.notFound` → framework `404.html` → built-in default), instead of CloudFront's raw S3-OAC 403 XML.

## Verification

**Live deploy:** https://d1fsny6h0qkzkb.cloudfront.net (`hosting-ssr-astro`)

**Unit/synth** — `659/659` hosting tests pass (incl. `spaFallback` resolution, default-404 wiring + precedence, SPA/SSG contract, and the misclassification cases).

**E2E (live CloudFront sandbox)** — apps covering each framework, all green in CI (see checks below):

| Framework | App |
|---|---|
| Astro static (multi-page) | `hosting-ssr-astro` |
| Astro static (built-in default-404) | `hosting-ssr-astro-default404` |
| Next.js (OpenNext) | `hosting-ssr` |
| Nuxt / Nitro | `hosting-ssr-nuxt` |
| SPA / static | `hosting-spa` |

**Astro multi-page routing** (`hosting-ssr-astro`, 15/15):

| Route | Result |
|---|---|
| `/` | 200 → Home |
| `/about` | 200 → **About** (not Home) |
| `/islands` | 200 → Islands (Preact island hydrates) |
| `/docs/api/v1/getting-started` | 200 → nested page |
| `/nonexistent` | **404** → custom 404 |
| `/old-home` | edge redirect → `/` |

**Decisive experiment** — deleted `404.astro` and redeployed (so `errorPages` is empty — the exact config that triggered the old bug). This is now permanent CI coverage via the `hosting-ssr-astro-default404` app above:
- `/about` still served **About**, not Home → proves routing is driven by the explicit `spaFallback`, not the error-pages heuristic.
- Unknown path returned the **built-in default 404** at HTTP 404 → proves the fallback.

Maps to test matrix `M1.4`/`M1.5`/`M1.6` (routing), `M7.1` (errors), `AS1.1`/`AS6.3` (Astro), `M3.1`/`M3.2` (cache).

## Test apps

- `test-apps/hosting-ssr-astro` — static multi-page Astro (home/about/islands/nested/404) + Preact island, single-origin `/aws-blocks/api`. Run via `npm run test:e2e:hosting-ssr-astro`.
- `test-apps/hosting-ssr-astro-default404` — same app **without** a custom `404.astro`, giving the built-in default-404 path real e2e coverage. Run via `npm run test:e2e:hosting-ssr-astro-default404`.

Both wired into the CI [`e2e-hosting` matrix](https://github.com/aws-devtools-labs/aws-blocks/blob/feat/astro-multi-page-support/.github/workflows/pr-checks.yml) alongside spa/nextjs/nuxt.


packages/hosting/src/adapters/spa.ts
```diff
@@ -126,6 +137,36 @@ export const spaAdapter = (
...
+    } catch {
+      return false;
+    }
+    for (const entry of entries) {
+      const full = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        // Skip the hashed-asset / framework dirs; routes never live there.
+        if (entry.name.startsWith('_') || entry.name === 'assets') continue;
+        if (walk(full, depth + 1)) return true;
+      } else if (entry.name === 'index.html' && depth > 0) {
```
adrianjoshua-strutt: This filesystem sniffing can misclassify in both directions. A real SPA that happens to ship a nested index.html (a static page in public/legal/index.html, a pre-rendered landing route, a bundled sub-app) would flip to spaFallback false and lose client-side deep-linking, which worked before this change. And a flat-file SSG (about.html instead of about/index.html, like Astro build.format 'file' or Hugo uglyURLs) has no nested index.html so it'd get spaFallback true and hit the exact bug this PR fixes. It also re-introduces an inference layer in the adapter, which is the thing we're removing at the L3. Could we source spaFallback from the build tooling's known output contract (framework config / buildOutputDir convention) instead of sniffing the filesystem? At minimum we'd want tests for both misclassification cases.

test-apps/hosting-ssr-astro/test/e2e.test.ts
```diff
@@ -0,0 +1,277 @@
...
+// ════════════════════════════════════════════════════════════════
+
+test.describe('Astro static multi-page — error pages', () => {
+  // M7.1 — unknown path returns a REAL 404 status with the custom 404
+  // body. A SPA-fallback misclassification would return 200 + home page.
+  test('M7.1 — unknown path returns 404 with the custom 404 page', async ({ request }) => {
+    const resp = await request.get(`${hostingUrl}/this/does/not/exist-12345`);
+    expect(resp.status()).toBe(404);
+    const html = await resp.text();
+    expect(pageId(html)).toBe('404 — Page Not Found');
```
adrianjoshua-strutt, line 165: The built-in default 404 page is the bulk of the new production code here but it only has unit coverage. This test app ships its own 404.astro, so the e2e always exercises the framework-emitted 404 path, never the built-in DEFAULT_NOT_FOUND_PAGE_HTML / _not_found.html path with the 403 to 404 remap. Could we add a variant test app without a custom 404 (or drop the 404.astro from one) so the default-404 path gets real end-to-end coverage against S3-OAC + CloudFront?

.gitallowed
```diff
@@ -5,3 +5,8 @@
 222222222222
 333333333333
 123456789012
+
+# Public npm package whose name trips the internal-tooling "[aA]pollo" guard
+# (false positive in package-lock.json — not an Amazon-internal reference)
+@graphql-tools/apollo-engine-loader
+apollo-engine-loader
```
adrianjoshua-strutt, line 16: Where does apollo-engine-loader actually enter the dependency tree? Grepping the whole diff, these strings only show up here in .gitallowed, not anywhere in the package-lock.json changes, and no graphql / @graphql-tools packages are added in this PR. If the secrets guard scans the full lockfile (not just the diff) and trips on a pre-existing line, that's fine, just worth a note. If these packages aren't actually in the tree, we can drop the entries.

packages/hosting/src/adapters/spa.ts
osama-rizk: Good catch — agreed on both the misclassification risk and the "don't re-introduce inference in the adapter" point. Fixed in 2691c48: dropped the filesystem sniff (`hasNestedIndexHtml`) entirely and now source `spaFallback` from the framework contract in `getAdapter` — `framework: 'spa'` → `true` (single-page), `framework: 'static'` → `false` (multi-page). That can't misclassify a SPA-with-nested-index or a flat-file SSG the way sniffing did. Added unit tests for exactly your two cases: a SPA shipping `legal/index.html` stays `spaFallback:true`, and a flat-file SSG (`about.html`, no nested index) stays `false`.

test-apps/hosting-ssr-astro/test/e2e.test.ts
osama-rizk: Agreed — the default-404 path was unit-only. Fixed in 2691c48: added a `hosting-ssr-astro-default404` test app (identical to `hosting-ssr-astro` but with no `404.astro`), so the adapter emits no `errorPages` and the L3's built-in `DEFAULT_NOT_FOUND_PAGE_HTML` / `_not_found.html` path kicks in. Its e2e asserts an unknown path returns HTTP 404 with the built-in body (and that the custom-404 marker is absent), exercising the S3-OAC 403→404 remap end-to-end. Also wired BOTH astro apps into the CI `e2e-hosting` matrix — they weren't running on PRs before, so this PR's Astro e2e now actually executes (framework-emitted 404 via `hosting-ssr-astro`, built-in default-404 via the new variant).

.gitallowed
osama-rizk: You're right — `apollo-engine-loader` is pre-existing on `main` (pulled in transitively by `@graphql-tools`); this PR adds no graphql packages. What's happening: the pre-commit hook uses `git secrets --pre_commit_hook`, which scans staged lockfile hunks *with surrounding context*, so the large additive `package-lock.json` insertion in this PR drags those unchanged apollo lines into the scanned diff and trips the `[aA]pollo` guard. So the entries are load-bearing for committing (removing them re-blocks the commit), but you're right they shouldn't read as if this PR introduced them — I've rewritten the comment in 2691c48 to say exactly that (pre-existing transitive dep, guard scans lockfile context).

adrianjoshua-strutt (dismissed):

packages/hosting/src/adapters/index.ts
```diff
@@ -181,9 +181,19 @@ export const getAdapter = (
     });
   }
 
-  // For SPA/static with explicit buildOutputDir, wrap the adapter
-  if (buildOutputDir && (framework === 'spa' || framework === 'static')) {
-    return (projectDir: string) => spaAdapter(projectDir, { buildOutputDir });
+  // For SPA/static, declare the routing model from the framework string
```
soberm: This guard now drops the `buildOutputDir` requirement, so `getAdapter('static')` without `buildOutputDir` goes from SPA-fallback (via the L3 heuristic) to explicit multi-page routing. That's the correct behavior, but it's a silent change for existing callers. Can we add a one-line migration note to the changeset?

```md
**Migration**: If you were passing `framework: 'static'` and relied on SPA-fallback routing (extensionless paths → /index.html), switch to `framework: 'spa'`. `framework: 'static'` now always produces multi-page directory-index resolution.
```

packages/hosting/src/adapters/index.ts
```diff
@@ -181,9 +181,19 @@ export const getAdapter = (
...
+  // (the build contract) rather than letting the adapter sniff the output
+  // tree: `spa` → single-page (client-side routing, /index.html fallback);
+  // `static` → multi-page (directory-index resolution). This is the signal
+  // Josh asked for — it can't misclassify a SPA-with-nested-index or a
```
soberm: Let's drop the reviewer first-name reference from shipped code — this is a public repo. Something neutral like:

```ts
// Sources the routing model from the framework's build contract rather than
// sniffing the output tree, which misclassified both a SPA shipping a nested
// index.html and a flat-file SSG.
```

packages/hosting/src/adapters/spa.test.ts
```diff
@@ -36,6 +37,50 @@ void describe('spaAdapter', () => {
     assert.deepStrictEqual(manifest.compute, {});
   });
```
soberm, line 39: Same here — let's replace the "Josh case 1 / case 2" comments with neutral descriptions of the scenarios (SPA-with-nested-index, flat-file SSG).

packages/hosting/src/adapters/spa.ts
```diff
@@ -18,6 +18,22 @@ const STATIC_EXCLUDE_NAMES = ['.ds_store', 'thumbs.db'];
 export type SpaAdapterOptions = {
```
soberm, line 18: Minor: the default (`true`) is documented in the last sentence — consider surfacing it on an explicit `@default` tag so it's harder to miss.

test-apps/hosting-ssr-astro-default404/test/e2e.test.ts
```diff
@@ -0,0 +1,297 @@
...
+const ENV = process.env.BLOCKS_TEST_ENV || 'local';
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const projectRoot = join(__dirname, '..');
+const backendPath = join(__dirname, '..', 'aws-blocks', 'index.cdk.ts');
+
+let hostingUrl: string;
+
+test.beforeAll(async () => {
+  if (ENV === 'sandbox') {
+    console.log('🚀 Deploying hosting-ssr-astro sandbox...\n');
```
soberm: Copy-paste leftover — this logs `Deploying hosting-ssr-astro sandbox` in the default404 app, which will be confusing in CI logs. Should be `hosting-ssr-astro-default404`.

test-apps/hosting-ssr-astro-default404/test/e2e.test.ts
```diff
@@ -0,0 +1,297 @@
...
+    if (resp.status() === 200) {
+      expect(resp.headers()['content-type']).toContain('image/svg');
+    }
+  });
+});
+
+// ════════════════════════════════════════════════════════════════
+// Error pages (M7) — BUILT-IN default 404 (no custom 404.astro)
+// ════════════════════════════════════════════════════════════════
+
```
soberm, line 164: Nit: with CloudFront remapping S3-OAC 403 → 404, the `403` in `expect([200, 403, 404])` is unreachable now. Harmless, but a short comment noting why 403 can't occur would clarify intent.

soberm (changes requested):
Thanks for iterating on this — the architecture is clean and the fix is the right one. The L3 staying framework-blind while routing intent flows adapter → manifest → L3 is exactly the right shape, and the `??` fallback to the legacy heuristic means unadapted callers don't regress. Test coverage is solid (the two e2e apps covering framework-emitted 404 vs. built-in default-404 are great).

Two things I'd like addressed before merge:

1. **Stale changeset.** `.changeset/astro-multi-page-support.md` still describes the original "SPA adapter detects nested index.html files" approach, which was replaced by framework-string sourcing. It'll mislead anyone reading the changelog.
2. **Undocumented behavior change.** `getAdapter('static')` dropping the `buildOutputDir` requirement is correct, but it's a silent change for anyone who relied on `static` getting SPA fallback. A one-line migration note in the changeset would cover it.

The rest are suggestions/nits and can be addressed at your discretion. Nice work overall.

packages/hosting/src/adapters/index.ts
osama-rizk: Done in a0afaac — added your exact migration note to the changeset (and corrected the changeset's now-stale "detects nested index.html" line, since that sniffing is gone).

packages/hosting/src/adapters/index.ts
osama-rizk: Done — replaced with your neutral wording (no reviewer name).

packages/hosting/src/adapters/spa.test.ts
osama-rizk: Done — the case comments now describe the scenarios neutrally (SPA-with-nested-index, flat-file SSG). Also dropped a stray reviewer-name reference in the default404 e2e header comment.

packages/hosting/src/adapters/spa.ts
osama-rizk: Done — added an explicit `@default true` tag.

test-apps/hosting-ssr-astro-default404/test/e2e.test.ts
osama-rizk: Good catch — fixed the copy-paste; it now logs `hosting-ssr-astro-default404`.

test-apps/hosting-ssr-astro-default404/test/e2e.test.ts
osama-rizk: Done — tightened to `expect([200, 404])` (403 is genuinely unreachable given the 403→404 remap) with a comment explaining why.

soberm (approved):

adrianjoshua-strutt (approved):

PR #16, opened 2026-06-17
refactor(bb-agent): model presets

pjkroker:
## Problem

**Issue #, if available:**
The model presets used outdated names (`DEFAULT`) and included models that don't support global inference profiles (`BUDGET`/`MICRO` using Nova/Llama), making them fail when deployed outside US regions.
## Changes

**Changed model IDs in presets:**

  - `BALANCED` (Claude Sonnet 4.6): recommended default for most workloads
  - `SMART` (Claude Opus 4.8): highest capability
  - `FAST` (Claude Haiku 4.5): lowest latency
 
 **Deprecated (non-breaking):**
   - `DEFAULT`, `BUDGET`, `MICRO` — still exported but now resolve to `BALANCED` and `FAST` respectively. Marked `@deprecated`.

**Note:** Llama models do not offer Global inference IDs, so we cannot offer them as globally available presets. Only Anthropic (and some Amazon) models support `global.` inference profiles.

All presets now use `global.` prefixed model IDs for region-agnostic deployment.

## Validation

- Updated unit tests to reference `BALANCED` instead of `DEFAULT`
- Added e2e preset tests (one per preset, verifies real Bedrock inference call)
- Verified all 3 model IDs resolve via `GetInferenceProfile` API call against Bedrock
- API.md regenerated
## Checklist

- [x] PR description included

---


packages/bb-agent/src/models.ts
```diff
@@ -4,34 +4,24 @@
 import type { ModelConfig } from './types.js';
 
 /**
- * Pre-configured Bedrock model presets using cross-region inference profiles.
+ * Pre-configured Bedrock model presets using global inference profiles.
```
hfurkanbozkurt, line 11: How do you motivate using `global` instead of `us`? Customers have data residency concerns so they might not want their data to leave, for example, US regions.

pjkroker: The preset is meant for fast prototyping — get a working agent with zero config. global works regardless of which region you deploy to, so there's no "this preset doesn't work in my region" surprise.

But we could add a warning so customers are reminded of this. They can always specify their own model IDs and regions!

hfurkanbozkurt (dismissed):

packages/bb-agent/src/models.ts
```diff
@@ -3,35 +3,49 @@
...
 		provider: 'bedrock',
-		modelId: 'us.amazon.nova-pro-v1:0',
+		modelId: BALANCED_MODEL_ID,
 	},
-	/** Ultra-cheap for simple tasks. Currently: Amazon Nova Lite. */
+	/** @deprecated Use `BedrockModels.FAST` instead. */
 	MICRO: {
 		provider: 'bedrock',
-		modelId: 'us.amazon.nova-lite-v1:0',
+		modelId: BALANCED_MODEL_ID,
```
hfurkanbozkurt: should map to fast and also should be deprecated

test-apps/comprehensive/aws-blocks/index.ts
```diff
@@ -378,6 +378,21 @@ const cannedAgent = new Agent(scope, 'canned', {
 });
 
 
+// Preset agents — one per BedrockModels preset, used to verify presets work e2e
+import { BedrockModels } from '@aws-blocks/bb-agent';
```
hfurkanbozkurt: nit: why import here and not at the top?

packages/bb-agent/src/models.ts
```diff
@@ -3,35 +3,50 @@
...
+	DEFAULT: {
+		provider: 'bedrock',
+		modelId: BALANCED_MODEL_ID,
 	},
-	/** Low cost per token with acceptable quality. Currently: Amazon Nova Pro. */
+	/** @deprecated Use `BedrockModels.BALANCED` instead. */
 	BUDGET: {
 		provider: 'bedrock',
-		modelId: 'us.amazon.nova-pro-v1:0',
+		modelId: BALANCED_MODEL_ID,
```
hfurkanbozkurt: shoud be fast as well

packages/bb-agent/src/models.ts
```diff
@@ -3,35 +3,50 @@
...
 	},
-	/** Fastest and lowest latency. Currently: Claude Haiku 4.5. */
+	/** Lowest latency, still strong capabilities. Currently: Claude Haiku 4.5. */
 	FAST: {
 		provider: 'bedrock',
-		modelId: 'us.anthropic.claude-haiku-4-5-20251001-v1:0',
+		modelId: FAST_MODEL_ID,
+	},
+
+
```
hfurkanbozkurt: nit: double blank lines

packages/bb-agent/src/models.ts
```diff
@@ -3,35 +3,50 @@
...
-		modelId: 'us.anthropic.claude-opus-4-8-20250610-v1:0',
+		modelId: BALANCED_MODEL_ID,
 	},
-	/** Strong quality/cost balance. Currently: Claude Sonnet 4. */
-	BALANCED: {
+	/** Highest capability for the hardest tasks. Currently: Claude Opus 4.8. */
+	SMART: {
 		provider: 'bedrock',
-		modelId: 'us.anthropic.claude-sonnet-4-20250514-v1:0',
+		modelId: 'global.anthropic.claude-opus-4-8',
```
hfurkanbozkurt: nit: Move smart up as well

packages/bb-agent/API.md
```diff
@@ -102,25 +102,17 @@ export type AgentTool<TContext = DefaultToolContext> = ToolDefinition<TContext,
 
 // @public
 export const BedrockModels: {
```
hfurkanbozkurt, line 104: A bit of a larger scope but why do we have an API.md? We put API into README.md and internals go to DESIGN.md

hfurkanbozkurt (changes requested):

packages/bb-agent/API.md
pjkroker: API.md is auto-generated by @microsoft/api-extractor with `npm run check:api`. I guess API in README.md is for humans and this one for machines?

packages/bb-agent/src/models.ts
hfurkanbozkurt, line 86: Please note a TODO to change this model as it is good at reasoning but not tool calling

packages/bb-agent/README.md
```diff
@@ -190,30 +190,28 @@ model: {
...
 | Preset | Current Model | Notes |
 |--------|---------------|-------|
-| `BedrockModels.DEFAULT` | `us.anthropic.claude-opus-4-8-20250610-v1:0` | Highest capability. Recommended default. |
-| `BedrockModels.BALANCED` | `us.anthropic.claude-sonnet-4-20250514-v1:0` | Strong quality/cost balance. |
-| `BedrockModels.FAST` | `us.anthropic.claude-haiku-4-5-20251001-v1:0` | Fastest, lowest latency. |
-| `BedrockModels.BUDGET` | `us.amazon.nova-pro-v1:0` | Low cost per token with acceptable quality. |
-| `BedrockModels.MICRO` | `us.amazon.nova-lite-v1:0` | Ultra-cheap for simple tasks. |
+| `BedrockModels.BALANCED` | `global.anthropic.claude-sonnet-4-6` | Great tool use, balanced cost. Recommended default for most workloads. |
+| `BedrockModels.SMART` | `global.anthropic.claude-opus-4-8` | Highest capability for the hardest tasks. |
+| `BedrockModels.FAST` | `global.anthropic.claude-haiku-4-5-20251001-v1:0` | Lowest latency, still strong capabilities. |
```
hfurkanbozkurt, line 212: nit: perhaps just the model names

test-apps/comprehensive/aws-blocks/index.ts
```diff
@@ -378,6 +378,20 @@ const cannedAgent = new Agent(scope, 'canned', {
 });
 
 
+// Preset agents — one per BedrockModels preset, used to verify presets work e2e
+
+const presetAgents = Object.fromEntries(
+	Object.entries(BedrockModels).map(([name, config]) => [
```
hfurkanbozkurt: nit: We are rerunning for the same model multiple times as miultiple entries point to the same model

hfurkanbozkurt:
The README removes all mention of the deprecated presets (DEFAULT, BUDGET, MICRO) without providing migration guidance. Users upgrading will see deprecation warnings in their IDE but won't find guidance in the README about what to switch to.

Please add a brief migration note after the preset table, e.g.: '> Migrating? DEFAULT → BALANCED (or SMART for highest capability). BUDGET/MICRO → FAST.'

hfurkanbozkurt (dismissed):

packages/bb-agent/src/models.ts
pjkroker: added a TODO

packages/bb-agent/src/models.ts
```diff
@@ -3,35 +3,50 @@
 
 import type { ModelConfig } from './types.js';
 
+const BALANCED_MODEL_ID = 'global.anthropic.claude-sonnet-4-6';
```
soberm, line 6: BALANCED and SMART use versionless IDs (global.anthropic.claude-sonnet-4-6, global.anthropic.claude-opus-4-8) while FAST keeps the fully-qualified form (global.anthropic.claude-haiku-4-5-20251001-v1:0). The factory passes modelId straight into Strands' BedrockModel with no normalization (packages/bb-agent/src/model-factory.ts:166-168), so if the versionless form is not the canonical Bedrock inference-profile ID, the failure surfaces only at invoke time rather than at dev time — the kind of late failure T1 (shift correctness left) exists to avoid.

Confirm global.anthropic.claude-sonnet-4-6 and global.anthropic.claude-opus-4-8 are valid Bedrock inference-profile IDs. If the canonical IDs carry a version/date suffix like FAST does, update the two constants to match so all three live presets are consistent and resolvable.

soberm (commented):
This PR makes every live preset a global. inference profile, but the Agent's Bedrock grant scopes inference-profile access to `arn:aws:bedrock:*:*:inference-profile/* `(index.cdk.ts:34-39, unchanged by this PR). Global inference profiles use a region-less ARN of the form arn:aws:bedrock::<account>:inference-profile/global.anthropic...; a * in the region segment does not necessarily match an empty region segment, so the grant may not actually cover the very profiles these presets now select.

Verify a deployed preset agent can actually invoke a global. profile. If the region-less global ARN is not matched, add an explicit resource such as arn:aws:bedrock::*:inference-profile/global.* (or the account-scoped equivalent) alongside the existing regional pattern.

packages/bb-agent/src/models.ts
pjkroker: The versionless vs dated form difference mirrors AWS's own naming rather than an inconsistency on our side — Sonnet 4.6 and Opus 4.8 only exist as versionless profiles (no dated variant in the list), while Haiku 4.5 only exists in the fully-qualified form. So there's no canonical dated ID to align the two constants to. They resolve as-is, and the preset e2e invoke confirms it. Leaving them matching the actual Bedrock IDs.

see the documentation: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-5.html

pjkroker:
> This PR makes every live preset a global. inference profile, but the Agent's Bedrock grant scopes inference-profile access to `arn:aws:bedrock:*:*:inference-profile/* `(index.cdk.ts:34-39, unchanged by this PR). Global inference profiles use a region-less ARN of the form arn:aws:bedrock:::inference-profile/global.anthropic...; a * in the region segment does not necessarily match an empty region segment, so the grant may not actually cover the very profiles these presets now select.
> 
> Verify a deployed preset agent can actually invoke a global. profile. If the region-less global ARN is not matched, add an explicit resource such as arn:aws:bedrock::_:inference-profile/global._ (or the account-scoped equivalent) alongside the existing regional pattern.

  Checked this against the live profiles — the global ARN isn't region-less the way it looks. GetInferenceProfile returns:
 
`arn:aws:bedrock:us-east-1:<acct>:inference-profile/global.anthropic.claude-sonnet-4-6`

"global" is only in the profile ID, not the ARN — the ARN itself is regional + account-scoped, so the existing arn:aws:bedrock:*:*:inference-profile/* pattern matches it. The region-less ARNs are the routed foundation models inside the profile (arn:aws:bedrock:::foundation-model/anthropic.claude-sonnet-4-6), which the grant's arn:aws:bedrock:*::foundation-model/* line already covers (* matches the empty region segment).

soberm (approved):

hfurkanbozkurt (approved):

PR #17, opened 2026-06-17
fix(bb-agent): global prefix health check

pjkroker:
## Problem

The regex in `checkModelHealth` only matched `us-gov|us|eu|apac `prefixes, so `global.*`, `jp`and `au` model IDs fell through to `GetFoundationModel` which doesn't know inference profiles — causing `ModelUnavailableException`.
**Issue #, if available:**

## Changes

- Removed the prefix regex entirely — health check now tries GetInferenceProfile first, then falls through to GetFoundationModel
- Any model ID format works without maintaining a prefix allowlist (cross-region, global, foundation model)
- Added debug-level logging in catch blocks for troubleshooting
## Validation

- Unit tests for inference profiles (us.*, global.*), foundation models, and not-found cases

## Checklist

- [x] PR description included

---


pranavosu (dismissed):

packages/bb-agent/src/model-factory.ts
```diff
@@ -36,7 +36,7 @@ export async function checkModelHealth(config: ModelConfig, log: ChildLogger, _t
 	}
 	log.info(`Checking model health: ${config.provider}${config.modelId ? ` (${config.modelId})` : ''}`);
 	if (config.provider === 'bedrock') {
-		const isCrossRegionProfile = config.modelId && /^(us-gov|us|eu|apac)\./.test(config.modelId);
+		const isCrossRegionProfile = config.modelId && /^(us|eu|apac|au|jp|global)\./.test(config.modelId);
```
hfurkanbozkurt: Why do we need this? What we care is there is either an inference profile or foundational model, right? Lets try both and fail only if it cannot get the info from either.

pjkroker: Yes that's great!

hfurkanbozkurt (dismissed):

packages/bb-agent/src/model-factory.ts
```diff
@@ -36,34 +36,29 @@ export async function checkModelHealth(config: ModelConfig, log: ChildLogger, _t
...
-					return true;
-				}
-				return false;
-			} catch (err: unknown) {
-				const e = err as { name?: string; message?: string };
-				log.warn(`Inference profile health check failed for '${config.modelId}': ${e.name ?? e.message}`, { provider: config.provider, modelId: config.modelId });
-				return false;
+		// Try GetInferenceProfile first (covers cross-region and global profiles).
+		try {
+			const client = await getClient();
```
soberm: getClient() is called in both try-blocks; hoist it out (one new BedrockClient() per check instead of two).

packages/bb-agent/src/model-factory.ts
```diff
@@ -36,34 +36,29 @@ export async function checkModelHealth(config: ModelConfig, log: ChildLogger, _t
...
+				? { inferenceProfileIdentifier: config.modelId }
+				: new (await import('@aws-sdk/client-bedrock')).GetInferenceProfileCommand({ inferenceProfileIdentifier: config.modelId });
+			const res = await client.send(command);
+			if (res.inferenceProfileName) {
+				log.info(`Inference profile '${config.modelId}' available`);
+				return true;
 			}
+		} catch (err: unknown) {
+			const e = err as { name?: string; message?: string };
+			log.debug?.(`Not an inference profile: ${e.name ?? e.message}`, { modelId: config.modelId });
```
soberm, line 53: should these be warnings?

packages/bb-agent/src/index.test.ts
```diff
@@ -844,31 +844,28 @@ describe('checkModelHealth', () => {
...
 
-	test('bedrock foundation model not found returns false', async () => {
-		const err = new Error('not found'); err.name = 'ValidationException';
-		const mockClient = { send: async () => { throw err; } };
+	test('bedrock model not found returns false', async () => {
+		const mockClient = { send: async () => { throw new Error('not found'); } };
 		assert.strictEqual(await checkModelHealth({ provider: 'bedrock', modelId: 'bad.model' }, log, mockClient), false);
 	});
 
-	test('bedrock credential error returns false', async () => {
```
soberm, line 857: why was this removed?

soberm (changes requested):

packages/bb-agent/src/index.test.ts
pjkroker: good catch! I put it back

packages/bb-agent/src/model-factory.ts
pjkroker: yes👍🏿

packages/bb-agent/src/model-factory.ts
pjkroker: debug is intentional.  A foundation model always fails `GetInferenceProfile` before succeeding on `GetFoundationModel`. Useful for debugging, but not a warning.

soberm (approved):

hfurkanbozkurt (approved):

PR #21, opened 2026-06-17
feat(pipeline): extract Pipeline into @aws-blocks/pipeline package

adrianjoshua-strutt:
## Summary

Extracts the Pipeline CDK construct from packages/core into a dedicated @aws-blocks/pipeline package, adds partialBuildSpec (Node.js 22 default), exports the __PIPELINE_STAGE_SCOPE__ constant, and ports the _sourceOverride internal test hook.

## Changes
- New packages/pipeline/ package (@aws-blocks/pipeline)
- Moved Pipeline construct, types, and tests from packages/core (git mv, history preserved)
- Added partialBuildSpec to PipelineSynthConfig (defaults to Node.js 22 runtime)
- Exported __PIPELINE_STAGE_SCOPE__ constant; core's BlocksStack imports it explicitly; core re-exports it for single-import-point consumers
- packages/core re-exports from @aws-blocks/pipeline for backward compatibility (zero consumer breakage)
- _sourceOverride internal prop ported (with ARN-validation-still-fires test)

## NOT included
- Gen2 definePipeline / getStageConfig / pipeline_factory / cache-busting machinery (Amplify-specific, not part of AWS Blocks)

## Testing
- Pipeline unit tests moved + extended (partialBuildSpec, Node.js 22 default, _sourceOverride + ARN validation)
- Synth integration test updated with partialBuildSpec coverage
- Real-deploy e2e remains infeasible without a manual GitHub CodeConnections OAuth handshake; _sourceOverride is in place to enable a future S3-source-based e2e

### DIFF
--- a/.changeset/extract-pipeline-package.md
+++ b/.changeset/extract-pipeline-package.md
@@ -1,0 +1,8 @@
+---
+"@aws-blocks/pipeline": patch
+"@aws-blocks/core": patch
+---
+
+feat(pipeline): extract Pipeline construct into @aws-blocks/pipeline package, add partialBuildSpec for CodeBuild runtime control
+
+`@aws-blocks/core` receives a minor bump (not patch): it gains a new runtime dependency on `@aws-blocks/pipeline` and adds new public re-exports from its CDK entrypoint (`__PIPELINE_STAGE_SCOPE__`, `Pipeline`, `DeployStage`, and the pipeline configuration types). New backwards-compatible public surface is a minor change per semver.
--- a/packages/pipeline/package.json
+++ b/packages/pipeline/package.json
@@ -1,0 +1,35 @@
+{
+  "name": "@aws-blocks/pipeline",
+  "version": "0.1.0",
+  "description": "CDK Pipelines-based CI/CD construct for AWS Blocks applications",
+  "author": "Amazon Web Services",
+  "license": "Apache-2.0",
+  "type": "module",
+  "files": [
+    "dist",
+    "README.md",
+    "src",
+    "LICENSE"
+  ],
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    }
+  },
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "tsc --build",
+    "test": "node --test dist/*.test.js",
+    "dev": "tsc --build --watch"
+  },
+  "devDependencies": {
+    "@types/node": "^22.0.0",
+    "typescript": "^5.3.0"
+  },
+  "peerDependencies": {
+    "aws-cdk-lib": "^2.257.0",
+    "constructs": "^10.6.0"
+  }
+}
--- a/packages/pipeline/src/constants.ts
+++ b/packages/pipeline/src/constants.ts
@@ -1,0 +1,21 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * The `globalThis` property key used to pass the active CDK Stage scope from
+ * {@link Pipeline} (during `appFile` import) to constructs that need to attach
+ * to the correct stage scope (e.g. `BlocksStack.create()`).
+ *
+ * Exported so consumers can read the ambient scope without hard-coding the
+ * string literal, making the coupling between the pipeline and its consumers
+ * explicit.
+ *
+ * @example
+ * ```ts
+ * import { __PIPELINE_STAGE_SCOPE__ } from '@aws-blocks/pipeline';
+ *
+ * const pipelineScope = (globalThis as any)[__PIPELINE_STAGE_SCOPE__];
+ * ```
+ */
+// The literal value intentionally equals the export name: this string IS the globalThis key, not a copy-paste typo.
+export const __PIPELINE_STAGE_SCOPE__ = '__PIPELINE_STAGE_SCOPE__' as const;
--- a/packages/pipeline/src/pipeline-construct.ts
+++ b/packages/pipeline/src/pipeline-construct.ts
@@ -21,6 +21,7 @@ import type {
   PipelineProps,
   PipelineStageConfig,
 } from './types.js';
+import { __PIPELINE_STAGE_SCOPE__ } from './constants.js';
 
 /**
  * Resolve a relative file path against the calling file's directory.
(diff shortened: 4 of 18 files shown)


osama-rizk (dismissed):

packages/pipeline/package.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "tsc --build",
+    "test": "node --test dist/*.test.js",
+    "dev": "tsc --build --watch"
+  },
+  "devDependencies": {
+    "@types/node": "^20.0.0",
```
soberm: @types/node is still pinned to ^20 even though the whole point of the package is defaulting synth to Node 22. Should be ^22 for consistency.

packages/pipeline/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": "./src"
```
soberm: Worth adding stripInternal: true here so _sourceOverride doesn't leak into the published .d.ts. CDK does this for the same reason — not urgent but good to have before 1.0.

packages/pipeline/src/types.ts
```diff
@@ -140,6 +141,30 @@ export interface PipelineSynthConfig {
...
+   * @example Pin Node.js 20
+   * ```ts
+   * synth: {
+   *   partialBuildSpec: BuildSpec.fromObject({
+   *     phases: { install: { 'runtime-versions': { nodejs: 20 } } },
+   *   }),
+   * }
+   * ```
+   */
+  readonly partialBuildSpec?: codebuild.BuildSpec;
```
soberm: The default uses ?? fromObject(...), so passing undefined always triggers the Node 22 default — there's no clean way to opt out. Someone who sets runtime-versions elsewhere would have to pass an empty BuildSpec, which isn't obvious. Maybe document it, or accept null as an explicit opt-out.

packages/pipeline/src/pipeline-construct.ts
```diff
@@ -415,6 +416,9 @@ function buildCodePipeline<TConfig>(
     pipelineType: codepipeline.PipelineType.V2,
     dockerEnabledForSynth: props.synth?.dockerEnabled ?? false,
     synthCodeBuildDefaults: {
+      partialBuildSpec: props.synth?.partialBuildSpec ?? codebuild.BuildSpec.fromObject({
```
soberm: Minor: this default BuildSpec gets rebuilt on every call. Could hoist it to a module-level const — clearer intent and consistent with how CDK usually handles defaults.

packages/pipeline/src/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+ * explicit.
+ *
+ * @example
+ * ```ts
+ * import { __PIPELINE_STAGE_SCOPE__ } from '@aws-blocks/pipeline';
+ *
+ * const pipelineScope = (globalThis as any)[__PIPELINE_STAGE_SCOPE__];
+ * ```
+ */
+export const __PIPELINE_STAGE_SCOPE__ = '__PIPELINE_STAGE_SCOPE__' as const;
```
soberm, line 21: A one-line comment here would save the next reader a double-take — the var name being equal to its string value is intentional but looks like a typo at a glance.

packages/pipeline/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@aws-blocks/pipeline",
```
soberm, line 2: package.json is missing a description field (other packages like hosting have one) — otherwise npm shows it blank.

soberm (commented):
Nice refactor — the re-export chain keeps this fully backward compatible and the build ordering is threaded correctly. partialBuildSpec test coverage is great too.

packages/pipeline/package.json
adrianjoshua-strutt: Bumped `@types/node` to `^22.0.0` in this package to match the Node.js 22 default. Kept it scoped to `@aws-blocks/pipeline` for now since a repo-wide `@types/node` bump (hosting and others are also on `^20`) is a separate change. Fixed in the latest push.

packages/pipeline/tsconfig.json
adrianjoshua-strutt: Added `"stripInternal": true` to the package tsconfig. Verified `_sourceOverride` no longer appears in the emitted `dist/*.d.ts`, so the `@internal` hook stays out of the published API. Fixed in the latest push.

packages/pipeline/src/types.ts
adrianjoshua-strutt: Added an explicit opt-out. `partialBuildSpec` now accepts `null`: omit it for the Node.js 22 default, pass `null` to inject no `runtime-versions` at all (bring your own via `installCommands`), or pass a `BuildSpec` to use it directly. Documented in the JSDoc and README, with unit tests covering all three cases. Fixed in the latest push.

packages/pipeline/src/pipeline-construct.ts
adrianjoshua-strutt: Hoisted the default to a module-level `DEFAULT_SYNTH_PARTIAL_BUILD_SPEC` constant. Fixed in the latest push.

packages/pipeline/src/constants.ts
adrianjoshua-strutt: Added a one-line comment clarifying the constant name intentionally equals its value (the name IS the globalThis key) so it doesn't read like a placeholder typo. Fixed in the latest push.

packages/pipeline/package.json
adrianjoshua-strutt: Added a `description` field. Note: `@aws-blocks/hosting` also lacks one, so adding descriptions there is a small separate cleanup. Fixed in the latest push.

adrianjoshua-strutt:
bumped `@aws-blocks/core` from `patch` to `minor`. It gains a new runtime dependency (`@aws-blocks/pipeline`) and new public re-exports (`__PIPELINE_STAGE_SCOPE__`, `Pipeline`, `DeployStage`, and the pipeline types), which is a minor per semver. `@aws-blocks/pipeline` stays `minor` as a new package. Updated in the latest push.

.changeset/extract-pipeline-package.md
```diff
@@ -0,0 +1,8 @@
+---
+"@aws-blocks/pipeline": minor
+"@aws-blocks/core": minor
```
soberm: Are you sure about a minor bump? This would mean a breaking change.

soberm (changes requested):

.changeset/extract-pipeline-package.md
adrianjoshua-strutt: Sure. Downgrading this to a patch.

soberm (approved):

osama-rizk (approved):

PR #24, opened 2026-06-17
fix(hosting): atomic redeploys, eliminate deploy-window 403

adrianjoshua-strutt:
## Problem

On every redeploy, new/cookieless visitors could get 403 Access Denied during a window. The CloudFront Function's new buildId propagated globally BEFORE the BucketDeployment finished uploading assets to builds/<newBuildId>/, so requests were rewritten to a prefix that did not exist yet in the private OAC bucket.

Returning users with a __dpl skew cookie were unaffected (routed to the still-existing previous build). New visitors, incognito, and expired-cookie users hit the window.

## Fix

Make the deploy truly atomic by reversing the dependency direction:
1. Removed the unnecessary /* CloudFront invalidation from asset BucketDeployments. With immutable build-id prefixes, each deploy writes to a brand-new builds/<id>/ prefix that was never cached, so /* invalidation served no purpose and was forcing uploads to run after the distribution update.
2. Made the CloudFront Function build-id cutover depend on all asset uploads completing. Now: upload assets to builds/<newId>/ -> then flip the function/distribution to route at <newId>.

This matches the upload-first-switch-second pattern used by managed hosting platforms.

## Testing
- New unit test asserting the dependency ordering (function depends on asset deployments)
- Updated assertions that previously expected the /* invalidation
- All hosting unit tests pass

### DIFF
--- a/.changeset/hosting-atomic-deploy-window.md
+++ b/.changeset/hosting-atomic-deploy-window.md
@@ -1,0 +1,5 @@
+---
+"@aws-blocks/hosting": patch
+---
+
+fix(hosting): make redeploys atomic by uploading assets before the CloudFront build-id cutover, eliminating the 403 window for new visitors during deployment
--- a/packages/hosting/src/constructs/cdn_construct.ts
+++ b/packages/hosting/src/constructs/cdn_construct.ts
@@ -1,5 +1,5 @@
 import { createHash } from 'node:crypto';
-import { Construct } from 'constructs';
+import { Construct, IDependable } from 'constructs';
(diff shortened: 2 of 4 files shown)


sboult:
Thanks for getting this PR up quickly based on my feedback! Looking forward to it landing and new release published ❤️.

osama-rizk (approved):

soberm (approved):

adrianjoshua-strutt:
@sboult Thanks for the bug report! I have merged the fix and will release it asap.

PR #31, opened 2026-06-18
feat(ci): per-PR agent bench (runner-based, Strands + Bedrock)

hfurkanbozkurt:
## Summary

GHA Summary: https://github.com/aws-devtools-labs/aws-blocks/actions/runs/28503266028

Adds a per-PR LLM-agent benchmark that measures how well an agentic builder can implement real features with AWS Blocks. Each cell scaffolds a fresh app from a template, hands it to a **builder agent** (Strands + Bedrock **Claude Sonnet 4.6**) that implements one task's feature, builds it, grades it with **Playwright**, then has a spec-blind **judge agent** (Bedrock **Claude Opus 4.8**) score the source against a fixed rubric. The output is a per-cell scoreboard (composite 0–100 + verdict) posted as a sticky PR comment.

It runs entirely on the GitHub runner — no microVM, no S3 transport between runner and sandbox. The runner is the sandbox; Bedrock provides the model. Builder and judge use different models on purpose, to limit same-model self-evaluation bias.

## The matrix — 9 cells, all 17 Building Blocks

The matrix is an explicit `include:` list of `(task, template)` pairs (not a cross-product). Each task runs on the single template that pre-ships or best exercises its blocks, bounding Bedrock spend. Together the 9 cells cover **all 17 Building Blocks across 7 templates** (`amplify` is excluded — it isn't standalone-scaffoldable).

| Task | Template | Blocks exercised |
|------|----------|------------------|
| `auth-notes` | `demo` | AuthBasic + KVStore |
| `file-gallery` | `bare` | FileBucket |
| `async-word-counter` | `bare` | AsyncJob + KVStore |
| `collab-cursor-board` | `default` | Realtime + DistributedTable |
| `cognito-profile` | `auth-cognito` | AuthCognito (email-OTP) |
| `observability-api` | `backend` | Logger + Metrics + Tracer + AppSetting |
| `sql-kb-catalog` | `nextjs` | Database + KnowledgeBase |
| `oidc-dsql-notes` | `react` | AuthOIDC + DistributedDatabase |
| `email-digest` | `demo` | CronJob + EmailClient + KVStore |

Each task ships a `PROMPT.md` (given to the builder), a `test.spec.ts` (Playwright), and a `rubric.md` whose first line is its one task-specific judge dimension. Across the 9 tasks there are **48 granular Playwright tests** (5–6 per task), so cells earn partial credit instead of one pass/fail.

## Per-cell pipeline

Six numbered steps per cell, all on the runner:

0. **Init result** — write a baseline `result.json` before any AWS call, so even an OIDC failure still produces a cell row
1. **Init** — build packages, pack the local registry, scaffold a fresh app, start the dev server
2. **Agent run** — builder agent (Sonnet 4.6); one tool (`shell`); 35-min step cap
3. **Build and test** — `npm run build` + the task's Playwright spec; records build / dev-server / playwright / test signals
4. **Judge** — judge agent (Opus 4.8); two read-only tools (`view`, `list`); scores source only; 8-min step cap
5. **Upload** — JSON artifact + S3 archive (best-effort)

A final `if: always()` finalize step stamps `status` + `failed_at` and the scoring fields, so every cell — including failed ones — lands in the summary table with a clear cause.

## Scoring (pre-registered, single-sourced)

All formulas live in `scripts/agent-bench/steps/lib/scoring.mjs`, imported by **both** `finalize-result.mjs` (stamps each `result.json`) and `summary.mjs` (renders the table) — one implementation of the formula, the verdict tiers, and the exclusion rule, so published numbers and the rendered table can't drift.

**Composite (0–100), per cell:**

```
tr        = tests_passed / (tests_passed + tests_failed)   # 0 if no tests ran
judge     = overall judge score, 0–10 (0 if the judge errored)
composite = round( 60*tr + 4*judge*min(1, 4*tr) , 1 )
```

60% objective pass-rate + 40% judge. The `min(1, 4*tr)` gate ties the judge term to the tests: a cell with zero passing tests floors to 0 regardless of the judge, and the judge term reaches full weight only once ≥25% of tests pass. Bands: ≥80 🟢, ≥50 🟡, else 🔴.

**Verdict tiers** are pure pass-rate — the judge plays no part, so an LLM failure can never flip a verdict:

| Verdict | Condition | In headline mean? |
|---------|-----------|-------------------|
| `pass` | pass-rate ≥ 0.999 | ✅ included |
| `partial` | 0 < pass-rate < 0.999 | ✅ included |
| `fail` | pass-rate == 0 with tests that ran, **or** an agent failure (builder timed out / no app at `2-agent`) → composite 0 | ✅ included |
| `unknown` | no tests ran on an otherwise-gradeable cell (denominator 0) | ❌ excluded |
| `harness_error` | pre-flight / OIDC / scaffold failure, or a cancellation — never produced a gradeable artifact | ❌ excluded |

**Exclusion rule.** The headline mean counts a cell iff it's gradeable. `harness_error` and `unknown` (denominator 0) are excluded and reported separately, so infra failure reads as infra rather than as a low score. An **agent failure** is the one no-tests case that still counts — the agent is exactly what's under test, so it scores as a genuine `fail` / composite 0 and is included.

**Judge dimensions.** Four shared (`functional_completeness`, `selector_contract`, `persistence`, `code_quality`) plus one task-specific dimension, all 0–10, averaged equally. The overall is recomputed deterministically from the dimensions, never read from free text; objective signals (scaffold / build / dev-server) are applied as deterministic hard caps after the judge returns and are never shown to it. A judge (LLM) failure drops only its 40% term, never the test-driven 60%.

## How to run it

- **On a PR:** apply the **`ready-for-pr-benchmark`** label. The trigger is `pull_request: [labeled]`, gated on same-repo **and** that label. (Pushing to a labelled PR won't re-run — re-apply the label.)
- **Manually:** `workflow_dispatch`, with an optional `bench_reps` input. The reporting path already aggregates the **median composite + IQR** across `result.<n>.json` files; the per-rep *execution* loop is a documented TODO, so today every run is N=1.

**Observational by default — no merge gate.** The repo/org variable `BENCH_MIN_SCORE` is unset, so the summary only reports the mean composite. Setting it to a number turns the summary into a gate (the job fails when the mean across scored cells falls below it). There is no baseline-delta gate yet.

## Key properties

- **Spec-blind judge** — grades source it cannot run, with the objective test spec hidden from its `view`/`list` tools and a runtime-blindness guard (flows whose runtime success can't be proven from source alone are held back). Objective signals become deterministic caps, never shown to the model.
- **Infra vs. agent separation** — `harness_error` (flaky runner / OIDC / scaffold / cancellation) is excluded from the mean; an agent timeout at `2-agent` is a genuine `fail`, not a flake.
- **Reproducibility** — builder pins `temperature=0` (the Opus judge rejects `temperature`, so its determinism rests on the structured-output schema + the deterministic caps); model IDs are pinned snapshots (the Claude 4.x IDs carry no date suffix — the version *is* the snapshot); Playwright pinned to `1.60.0`; a run-stable `RUN_ID` seeds the specs' unique-but-deterministic data.
- **Security** — fork-PR gate via `head.repo.full_name == github.repository` (not `pull_request_target`, so forks get no secrets/OIDC); least-privilege OIDC under `environment: publish` (`id-token: write` only); `configure-aws-credentials` SHA-pinned; the judge's `view`/`list` tools are `realpath`-contained against symlink escape.
- **Failure visibility** — every cell uploads a `result.json` even when an early step fails, and the summary is mirrored into a single **sticky PR comment** (found/updated via a hidden marker, so each run replaces the last). When the matrix is skipped (label absent) no comment markdown is written, leaving the last real scoreboard intact rather than clobbering it with an empty table.

Full design notes live in `scripts/agent-bench/README.md`.

## Required configuration

Uses the existing `publish` environment:

- `AWS_ROLE_ARN` (secret) — OIDC role with `bedrock:InvokeModelWithResponseStream` on the bench models (Sonnet 4.6 + Opus 4.8), plus `s3:PutObject` on `bench/*` in the registry bucket.
- `S3_BUCKET` (variable) — existing registry bucket, reused for the best-effort result archive.

## Test plan

- [ ] All required PR checks (changeset, build, e2e) pass
- [ ] With the `ready-for-pr-benchmark` label, all 9 cells run; each produces a `result.json` carrying `klass`, `test_rate`, `verdict`, `composite`, and the raw per-dimension judge scores
- [ ] Summary renders the scoreboard (with the collapsible raw per-dimension table + run-logs link) on the run page and as a sticky PR comment
- [ ] `harness_error` / `unknown` cells are listed separately and excluded from the mean; an agent timeout reads as `fail` / composite 0 and is included
- [ ] With no `BENCH_MIN_SCORE` set, the summary is observational (no merge gate)
- [ ] Result archives land at `s3://<bucket>/bench/pr/<n>/run-<id>-attempt-<n>/<task>-<template>.json`


hfurkanbozkurt:
## Maintainer note — Agent Bench run [#28098267936](https://github.com/aws-devtools-labs/aws-blocks/actions/runs/28098267936) explained

_This is a maintainer annotation, separate from the automated scoreboard above. The bench is **observational, N=1, no merge gate** — cells flip run-to-run from agent + infrastructure variance, so a red cell is a signal to investigate, not a blocker. Green cells confirm the Building-Block path works when it isn't hit by one of the tracked issues below._

**This run (HEAD `bcc77a3`):** mean **67.3/100**, **6/9 green**, judge healthy (the judge retry/backoff held — no Bedrock throttling this run).

### Non-green cells & why

- **async-word-counter** (bare) — 🔴 0/5. The dev server's Vite frontend died (`Port 3100 already in use` → HTTP 502 for the entire readiness window), so Playwright never loaded the page. The agent's code was otherwise rated 9/10. This is the dev-server frontend-orphan / EADDRINUSE race after a `tsx watch` restart → **#78**. The harness intentionally has **no self-heal/workaround** for this — the fix belongs in the framework.

- **auth-notes** (demo) — 🟡 1/6. Dev server came up fine (not an infra failure). The AuthBasic sign-up → signed-in DOM transition doesn't reliably happen in the browser even though the source reads correct (judge 9/10). This is agent-variance interacting with the AuthBasic error-handling sharp-edge → **#81**.

- **sql-kb-catalog** (nextjs) — 🔴 0/0. The builder agent hit the **35-minute step budget** before finishing (token/time-heavy nextjs cell); no app was produced, so nothing could be built/tested. Pre-existing flaky/heavy cell — an agent timeout, **not** a framework defect this run. (Earlier runs of this same cell hit the dev-server/PGlite dead-server issue **#98** instead — it's flaky for two distinct reasons.)

### Green this run

collab-cursor-board (97.6), oidc-dsql-notes (97.6), cognito-profile (89.6), email-digest (98.4), file-gallery (96.8), observability-api (92.0) — confirming those paths work when the dev-server race doesn't land on them. (Notably collab-cursor-board, which was red in prior runs from #78, passed here — the dev-server race is non-deterministic about which cell it lands on.)

### Tracked root causes
- **#78** — dev-server frontend orphan on `:3100` (502 / `EADDRINUSE`) after a `tsx watch` restart (vite & next). Hits a different random cell each run.
- **#98** — PGlite `.bb-data` corruption when `tsx watch` SIGTERMs the server mid-init → dev port never binds.
- **#79** — AuthOIDC `signIn` fire-and-forget / `handleRedirectCallback` not idempotent / undocumented `onAuthChange` bridge.
- **#81** — AuthBasic `setAuthState` discards the structured error name, breaking the documented catch-by-name flow.

hfurkanbozkurt:
Superseded by #132 — recreated on a clean branch/commit off latest main (this PR accumulated noisy iteration history + early automated comments). Identical validated content (bench run 28503266028, mean 93.2, 10/10 pass). Closing in favor of #132.

PR #32, opened 2026-06-18
feat: add code coverage reporting to unit tests

sarayev:
## Summary

Adds code coverage reporting to unit tests using Node 22's built-in `--experimental-test-coverage`.

## Changes

- **`scripts/coverage.mjs`** — Per-package runner that injects coverage flags into each workspace's test command, preserving per-package cwd (required by packages like core that use `process.cwd()`)
- **`tsconfig.base.json`** — Enable `sourceMap` for TS→JS line mapping
- **`package.json`** — Add `test:coverage` (CI) and `test:coverage:local` (standalone with build)
- **`.github/workflows/pr-checks.yml`** — Replace "Unit Tests" step with "Unit Tests (with coverage)"

## Behavior

- **Test failures block the build** — exit non-zero if any unit test fails
- **Coverage is reporting-only** — numbers are printed but never fail the run
- **Test files excluded** from coverage stats (`--test-coverage-exclude`)
- **Coverage scoped** to package source only (`--test-coverage-include`)

## Usage

```bash
# In CI (build already ran):
npm run test:coverage

# Locally (includes build step):
npm run test:coverage:local
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,8 @@
     "nuke": "rm -rf node_modules packages/*/node_modules test-apps/*/node_modules packages/*/dist test-apps/*/dist test-apps/*/.next packages/*/*.tsbuildinfo test-apps/*/*.tsbuildinfo",
     "test": "npm run test --workspaces",
     "test:unit": "npm run test --workspaces",
+    "test:coverage": "node scripts/coverage.mjs",
+    "test:coverage:local": "npm run build:packages && node scripts/coverage.mjs",
     "test:e2e:local": "cd test-apps/comprehensive && npm run test:e2e:local && cd ../.. && npm run test:vendorize",
     "test:e2e:sandbox": "cd test-apps/comprehensive && npm run test:e2e:sandbox",
     "test:e2e:production": "cd test-apps/comprehensive && npm run test:e2e:production",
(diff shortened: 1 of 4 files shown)


pranavosu (approved):

package.json
```diff
@@ -62,6 +62,8 @@
     "nuke": "rm -rf node_modules packages/*/node_modules test-apps/*/node_modules packages/*/dist test-apps/*/dist test-apps/*/.next packages/*/*.tsbuildinfo test-apps/*/*.tsbuildinfo",
     "test": "npm run test --workspaces",
     "test:unit": "npm run test --workspaces",
+    "test:coverage": "node scripts/coverage.mjs",
+    "test:coverage:local": "npm run build:packages && node scripts/coverage.mjs",
```
svidgen, line 66: AFAIK, `--workspaces`/`--ws` executions already run with the correct CWD of each target workspace. That should mean the coverage script injection util *shouldn't* be necessary. One of these should work and are arguably more straightforward:

1. `npm run test --ws -- --experimental-test-coverage --test-coverage-include='dist/**/*.js' --test-coverage-exclude='dist/**/*.test.js'`
2. `NODE_OPTIONS='--experimental-test-coverage --test-coverage-include=dist/**/*.js --test-coverage-exclude=dist/**/*.test.js' npm run test --ws`
3. `npm run test:coverage --ws --if-present` (And then create the target in each package that supports it.)

svidgen (commented):
Unclear whether I'm missing something important that `coverage.mjs` is doing that a simple `--ws` command can't do. 😅

PR #33, opened 2026-06-18
fix(bb-cron-job): stepped cron ranges ignored their upper bound

watilde:
## Problem

expandField() parsed a stepped range like `0-30/10` as start=0 with no upper bound, stepping to the field max (59) and yielding [0,10,20,30,40,50] instead of [0,10,20,30]. A job restricted to minutes 0–30 would also fire at 40 and 50, diverging from EventBridge.

**Issue #, if available:**

## Changes

Detect when the step base is itself a range and use its `hi` as the loop bound. The `*/n` and single-value `n/step` forms are unchanged.

## Checklist

- [x] PR description included

---

### DIFF
--- a/packages/bb-cron-job/src/index.mock.ts
+++ b/packages/bb-cron-job/src/index.mock.ts
@@ -176,10 +176,24 @@ function expandField(field: string, min: number, max: number, original: string):
 	for (const part of field.split(',')) {
 		if (part.includes('/')) {
 			const [base, stepStr] = part.split('/');
-			const start = base === '*' ? min : parseInt(base, 10);
 			const step = parseInt(stepStr, 10);
+			// `base` may be `*` (whole field), a single value (`15` → from 15 to
+			// max), or a range (`0-30` → bounded by the range's upper end). A
+			// stepped range must stop at the range's `hi`, not run to `max`.
+			let start: number;
+			let end = max;
+			if (base === '*') {
+				start = min;
+			} else if (base.includes('-')) {
+				const [lo, hi] = base.split('-').map(Number);
+				if (isNaN(lo) || isNaN(hi)) throw scheduleError(original);
+				start = lo;
+				end = hi;
+			} else {
+				start = parseInt(base, 10);
+			}
 			if (isNaN(start) || isNaN(step) || step <= 0) throw scheduleError(original);
-			for (let i = start; i <= max; i += step) values.push(i);
+			for (let i = start; i <= end; i += step) values.push(i);
 		} else if (part.includes('-')) {
 			const [lo, hi] = part.split('-').map(Number);
 			if (isNaN(lo) || isNaN(hi)) throw scheduleError(original);
--- a/packages/bb-cron-job/src/index.mock.ts
+++ b/packages/bb-cron-job/src/index.mock.ts
@@ -180,27 +180,27 @@ function expandField(field: string, min: number, max: number, original: string):
 			// `base` may be `*` (whole field), a single value (`15` → from 15 to
 			// max), or a range (`0-30` → bounded by the range's upper end). A
 			// stepped range must stop at the range's `hi`, not run to `max`.
-			let start: number;
-			let end = max;
+			let lo: number;
+			let hi = max;
 			if (base === '*') {
-				start = min;
+				lo = min;
 			} else if (base.includes('-')) {
-				const [lo, hi] = base.split('-').map(Number);
-				if (isNaN(lo) || isNaN(hi)) throw scheduleError(original);
-				start = lo;
-				end = hi;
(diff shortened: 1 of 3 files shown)


sarayev (dismissed):
Thanks for contributing to AWS Blocks! Clean fix, good tests — LGTM 👍

watilde:
Added changeset and rebased

packages/bb-cron-job/src/index.mock.ts
```diff
@@ -176,10 +176,24 @@ function expandField(field: string, min: number, max: number, original: string):
...
+			// max), or a range (`0-30` → bounded by the range's upper end). A
+			// stepped range must stop at the range's `hi`, not run to `max`.
+			let start: number;
+			let end = max;
+			if (base === '*') {
+				start = min;
+			} else if (base.includes('-')) {
+				const [lo, hi] = base.split('-').map(Number);
+				if (isNaN(lo) || isNaN(hi)) throw scheduleError(original);
+				start = lo;
```
soberm: Should we add an inverted-range guard? `30-10/5`, e.g., silently returns [].

packages/bb-cron-job/src/index.mock.ts
```diff
@@ -176,10 +176,24 @@ function expandField(field: string, min: number, max: number, original: string):
...
+				end = hi;
```
soberm: This allows 100 which would be out-of-bounds for minutes. Could also add a guard here.

packages/bb-cron-job/src/index.mock.ts
```diff
@@ -176,10 +176,24 @@ function expandField(field: string, min: number, max: number, original: string):
...
+			} else {
+				start = parseInt(base, 10);
+			}
 			if (isNaN(start) || isNaN(step) || step <= 0) throw scheduleError(original);
-			for (let i = start; i <= max; i += step) values.push(i);
+			for (let i = start; i <= end; i += step) values.push(i);
```
soberm: nit: this branch uses `start`, `end` while other branches use `lo`/`hi`.

packages/bb-cron-job/src/schedule.test.ts
```diff
@@ -138,3 +138,26 @@ describe('parseSchedule – cron day-of-week numeric values', () => {
 		assert.deepStrictEqual(fields.dayOfWeek, [1, 2, 3, 4, 5]);
 	});
 });
```
soberm, line 140: Can we add some tests for edge-cases?

soberm (changes requested):

packages/bb-cron-job/src/index.mock.ts
watilde: Added a shared `assertBounds()` guard that rejects inverted ranges (`lo > hi`), so `30-10/5` now throws `InvalidSchedule` instead of silently returning `[]`. Covered by a new test.

packages/bb-cron-job/src/index.mock.ts
watilde: `assertBounds()` also enforces the field window (`lo < min || hi > max`), so an out-of-bounds upper end like `0-100/5` (and a bare `100`) now throws instead of emitting values EventBridge would reject. Tests added for both.

packages/bb-cron-job/src/index.mock.ts
watilde: Renamed this branch to use `lo`/`hi`, consistent with the other branches.

packages/bb-cron-job/src/schedule.test.ts
watilde: Added edge-case tests: inverted stepped range (`30-10/5`), stepped range exceeding the field max (`0-100/5`), inverted plain range (`30-10`), and out-of-bounds single value (`100`) — all asserting `InvalidSchedule`.

watilde:
Thanks for the review @soberm, all four points addressed in 1d40e58.

- Inverted ranges (`30-10`, `30-10/5`) and out-of-bounds values (`100`, `0-100/5`) now throw `InvalidSchedule` via a shared `assertBounds()` helper.
- Stepped-range branch renamed to `lo`/`hi` for consistency.
- Added edge-case tests for each.

Heads-up on scope: your comments were on the stepped-range branch, but I applied the same `assertBounds()` guard to the plain-range and single-value branches too, since they had the identical silent-failure mode. This means a previously-accepted out-of-bounds bare value (e.g. minute `100`) now throws instead of being silently emitted. That matches EventBridge, but flagging in case you wanted the change kept narrower. Happy to scope it back to just the stepped branch if you prefer.

soberm (approved):
Thanks for the quick fix!

sarayev (approved):

PR #34, opened 2026-06-18
fix(bb-tracer): make sampling a per-trace decision

watilde:
## Problem
bb-tracer's startSegment() rolled the sampling decision (Math.random() < samplingRate) on every segment. But sampling should be per-trace. With a fractional samplingRate, a sampled trace could randomly drop subsegments, and an unsampled trace could promote a subsegment into its own orphan root trace. Hidden at the default rate of 1.0.

**Issue #, if available:**

## Changes
- Roll sampling once at the root segment; subsegments inherit it (sampled path tracked via segmentStack, unsampled path via a new unsampledDepth counter).
- Added two tests using a scripted Math.random at samplingRate: 0.5: sampled root keeps its child; unsampled root writes no trace (no orphan).

## Checklist

- [x] PR description included

---

### DIFF
--- a/.changeset/tracer-per-trace-sampling.md
+++ b/.changeset/tracer-per-trace-sampling.md
@@ -1,0 +1,5 @@
+---
+"@aws-blocks/bb-tracer": patch
+---
+
+fix(bb-tracer): make sampling a per-trace decision so all spans within a trace share the same sampling outcome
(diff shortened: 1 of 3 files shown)


watilde:
Added changeset and rebased

soberm (approved):
Thanks for contributing. LGTM!

sarayev (approved):

PR #39, opened 2026-06-18
fix: stack-scope external-DB connection string SSM parameter (multi-app collision)

vishaalmehrishi:
## Problem

When a Blocks app connects to an external database (e.g. Supabase), the deploy process stores the connection string in an SSM SecureString parameter whose name is derived **only from the stage**, with no app/stack identifier:

```ts
// packages/core/src/db-naming.ts
export function dbConnectionParameterName(stage: string): string {
  return `/blocks/${stage}/db-connection-string`;
}
```

Because the name contains no app identity, **two different Blocks apps deployed to the same AWS account + region + stage compute the same parameter name and overwrite each other.** The writer overwrites unconditionally (`Overwrite: true`), so:

- App A deploys → `/blocks/production/db-connection-string` = connA
- App B deploys → `/blocks/production/db-connection-string` = connB (overwrites connA)
- App A's Lambda now reads connB → **connects to the wrong database, with no error.**

So an account + region + stage is effectively limited to one external-DB Blocks app. (Sandbox has a per-developer variant of the same issue.)

## Root cause

The generic `AppSetting` already names parameters in a stack-scoped, collision-safe way by default — `/${fullId}`, where `fullId` embeds the unique stack name (`packages/bb-app-setting/src/index.cdk.ts`, `packages/core/src/cdk/blocks-backend.ts`). That is why secrets like `jwt-secret` never collide. The database parameter collides **only because the generated wiring overrides that default** with the stage-only name.

It overrides for one reason: the connection string is written **out-of-band, before `cdk synth`** (`packages/core/src/scripts/deploy.ts`, `sandbox.ts`). At that point the CDK app has not run, so the synth-time `fullId` is unavailable — a stage-only name was the only thing the pre-synth writer could compute.

> Note: this is intentionally **not** fixed by putting the database/project ref in the name. An earlier approach did that and broke silently — the pre-synth writer derived the ref from the live environment while the generated code used a ref captured at `db pull` time, so when a stage pointed at a different database the written and read names diverged and the app connected to the wrong database. Any discriminator in the name must be derived identically on the write and read sides, and must not come from the connection string.

## Solution — stack-scoped name, written after deploy

1. **Self-named, stack-scoped parameter.** The generated wiring no longer passes a `name` on the synth path, so the database `AppSetting` uses the default `/${fullId}` (= `/<stackName>-supabase-db-url`), unique per app. At runtime it keeps reading the deploy-stamped `BLOCKS_SSM_PARAM_DB_URL` (it must not recompute `/${fullId}` from the runtime construct tree):
   ```ts
   const dbUrl = process.env.BLOCKS_SSM_PARAM_DB_URL
     ? AppSetting.fromExisting(scope, 'db-url', { name: process.env.BLOCKS_SSM_PARAM_DB_URL, secret: true }) // runtime
     : AppSetting.fromExisting(scope, 'db-url', { secret: true });                                          // synth → /${fullId}
   ```
   `AppSetting.fromExisting` now allows an omitted `name` (defaulting to `/${fullId}`) in all variants.

2. **Expose the resolved name as a `CfnOutput`.** `AppSetting` emits the synth-resolved parameter name as a stack output.

3. **Write the connection string after `cdk deploy`.** `ensureSecrets` moves to run after the deploy, reads the name from the stack outputs, and writes to that exact name. Migrations stay before deploy — they read the connection string from `process.env`, not from SSM, so they are unaffected.

### Why this works

The name is derived exactly once, at synth. That single value is the `CfnOutput`, the value stamped into the runtime config, the IAM grant, and what `ensureSecrets` writes — so **written == stamped == read** by construction, and the write/read divergence described above cannot occur. It introduces no new identifier (it relies on CloudFormation's existing per-account/region stack-name uniqueness), and writing the secret only after a successful deploy avoids provisioning a secret for a deploy that failed or rolled back.

### Robustness of the post-deploy write

Because the write happens after `cdk deploy` rather than inside the CloudFormation transaction, the write is hardened against transient failure:

- **Retry:** transient SSM errors (throttling, 5xx, timeouts) are retried with exponential backoff; non-transient errors (e.g. `AccessDenied`) propagate.
- **Fail loudly:** if the write fails, the deploy/sandbox command prints a clear message that the app is deployed but cannot reach its database until resolved, and that re-running is safe.
- **Idempotent:** a no-op when there is no connection string or the stored value already matches, so re-running after a failure is safe.

### First-deploy window

On a first deploy there is a brief window where the parameter does not exist until `ensureSecrets` writes it (immediately after `cdk deploy` returns). A brand-new endpoint has no traffic in that window; redeploys are unaffected (the value already exists).

### Compatibility

The fix self-heals on the next deploy: the new stack-scoped name is written, stamped, and read consistently. The previous stage-only parameter (`/blocks/{stage}/db-connection-string`) becomes orphaned and can be deleted manually. No runtime read-fallback is required.

## Changes

- `packages/bb-data/src/db-pull/generate.ts`, `templates.ts` — generated wiring is self-named at synth, reads the stamped name at runtime
- `packages/bb-app-setting/src/index.cdk.ts` — allow `fromExisting` without a name; emit the resolved name as a `CfnOutput`
- `packages/bb-app-setting/src/index.mock.ts`, `index.aws.ts` — `fromExisting` name made optional (signature parity, so generated code type-checks in all variants)
- `packages/core/src/scripts/ensure-secrets.ts` — write to a caller-provided name; helper to read it from stack outputs; transient-error retry; idempotent
- `packages/core/src/scripts/deploy.ts`, `sandbox.ts` — write the secret after deploy using the name from outputs; clear, actionable error if the write fails
- `packages/core/src/scripts/external-migrations-step.ts` — dev-loop production-target guard no longer pins the old stage-only name
- `packages/core/src/db-naming.ts` — remove the now-unused `dbConnectionParameterName`
- `packages/bb-app-setting/API.md` — regenerated for the `fromExisting` signature
- `test-apps/db-pull-typecheck/generated/index.ts` — regenerated snapshot of the new wiring (CI typecheck guard)

## Testing

- Unit tests updated/added: self-naming + `CfnOutput` (`packages/bb-app-setting`), generated-wiring assertions (`packages/bb-data`), and retry/idempotency for the post-deploy write (`packages/core`).
- The `db-pull-typecheck` snapshot regenerated and `tsc --noEmit` passes (guards that generated output type-checks against the public types).
- `npm run build`, `npm run check:api`, and unit tests pass for `@aws-blocks/core`, `@aws-blocks/bb-app-setting`, `@aws-blocks/bb-data`.


vishaalmehrishi:
Closing in favour of https://github.com/aws-devtools-labs/aws-blocks/pull/114

PR #43, opened 2026-06-18
fix(ci): native sandbox fixes

ekjotmultani:
## What & why
The native-SDK end-to-end suite runs in two modes: against a local dev server (no AWS) and against a per-run deployed sandbox. The deployed-sandbox run was unreliable. This change stabilizes it so it passes consistently against the sandbox, without changing which jobs gate branch protection.

## Changes
- **Cognito — real returning-customer path.** Adds coverage that signs in a pre-provisioned, confirmed user and exercises an authenticated RPC, cookie-session persistence across requests, and sign-out. The sign-up → emailed-code → confirm flow now runs only against the local dev server, because a real Cognito pool emails the confirmation code, which CI cannot read back. Test credentials come from `COGNITO_TEST_USERNAME` / `COGNITO_TEST_PASSWORD` with deterministic local fallbacks.
- **Backend seed.** Adds `seed-cognito-user.ts` (`AdminCreateUser` + `AdminSetUserPassword` → a confirmed user, with the user pool resolved from the deployed stack) plus a `seed:cognito` script, run as a post-deploy CI step so the returning-customer test has a user to sign in as. The script logs no environment-derived values.
- **OIDC endpoint derivation.** Resolves the API origin for the server-relay auth routes robustly from the configured base URL, and stops appending a duplicate API path segment when deriving the sandbox endpoint, so relay requests reach the auth routes rather than the JSON-RPC handler.
- **Post-deploy readiness poll.** Waits for a freshly-deployed sandbox API to return HTTP 200 before running the suite, eliminating cold-start flakiness on the first request (added to both the workflow and the local runner).
- **Timeout.** Raises the per-run sandbox job timeout to 45 minutes to cover the full deploy + test + teardown, preventing premature cancellation and leaked stacks.

## Gating
Unchanged. The required gate covers `detect-changes`, `setup`, the local `dart-e2e`, and `swift-e2e`; the deploying sandbox jobs remain non-blocking signals.

## Validation
- `blocks_codegen`: `dart analyze` clean and 58/58 tests pass.
- The Dart example and e2e suite analyze clean.

### DIFF
--- a/.github/workflows/native-sdk-e2e.yml
+++ b/.github/workflows/native-sdk-e2e.yml
@@ -198,6 +198,21 @@ jobs:
             sleep 1
           done
 
+      - name: Verify server is ready
+        run: |
+          for i in $(seq 1 30); do
+            if curl -sf -X POST http://localhost:3001/aws-blocks/api \
+              -H "Content-Type: application/json" \
+              -d '{"jsonrpc":"2.0","method":"api.kvGet","params":["healthcheck"],"id":1}' > /dev/null 2>&1; then
+              echo "✅ Server ready"
+              exit 0
+            fi
+            echo "⏳ attempt $i/30: server not ready; retrying in 2s..."
+            sleep 2
+          done
+          echo "❌ Server never became ready"
+          exit 1
+
       - name: Run E2E tests
         env:
           BLOCKS_URL: http://localhost:3001/aws-blocks/api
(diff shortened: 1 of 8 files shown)


thisisabhash (dismissed):

.github/workflows/native-sdk-e2e.yml
```diff
@@ -198,6 +198,21 @@ jobs:
             sleep 1
           done
 
+      - name: Verify server is ready
+        run: |
+          for i in $(seq 1 30); do
+            if curl -sf -X POST http://localhost:3001/aws-blocks/api \
```
pranavosu: is this port `3001` stable? i think we might try a few different ones. Maybe it's okay in CI

mattcreaser (dismissed):

.github/workflows/native-sdk-e2e.yml
```diff
@@ -198,6 +210,21 @@ jobs:
...
```
thisisabhash: should this be `"$BLOCKS_URL"`?

thisisabhash (approved):

harsh62 (approved):

PR #46, opened 2026-06-19
feat(hosting): KVS edge-routing migration + base path, quota accounting & Phase-0 fixes

osama-rizk:
## Summary

Replaces **per-route CloudFront cache behaviors** with a **single default behavior** whose viewer-request/response CloudFront Functions read the route table from a **KeyValueStore** and select the origin via `selectRequestOriginById()`. Route count becomes data, not infrastructure — the per-distribution behavior cap is eliminated. Builds on top: base-path support, centralized quota accounting, and a batch of correctness fixes found by live-probing every framework.

## What's in here

**KVS edge routing (core)** — `kvs_router.ts` (`buildKvsEntries` → chunked KVS map + generated request/response functions), `kv_keys.ts`/`kv_keys_handler.ts` (gated live `UpdateKeys` for the atomic cutover), `cdn_construct.ts` (single behavior + sentinel behaviors that bind the server/image origins). Origin ids: `blocks-s3`, `blocks-server`, `blocks-image`.

**Base path** — caller-declared `basePath` prop as the source of truth; Nuxt `app.baseURL` auto-detected from the build (parity with Next `basePath` / Astro `base`).

**Quota accounting** — `quota_budget.ts` centralizes the adjustable CloudFront quotas with deploy-time guards that fail loud instead of opaquely mid-deploy.

**Routing correctness fixes (all live-probed):**
- Mid-segment wildcards (`/api/*/admin`) in the request router + response header matcher.
- `meta` written last + build-time budget guard (`TooManyRoutesError`/`RouteTableTooLargeError`).
- `__dpl` skew cookie honored only when skew protection is enabled.
- Trailing-slash normalization (`/about/` → stored `/about`).
- `SentinelGuard` 403s direct hits to the origin-binding `/__blocks_origin_*` patterns.
- **assetPrefix**: `/cdn-static/_next/*` stripped before classification.
- **Image optimization**: image branch strips basePath before the optimizer; Next `fetchInternalImage` arity patch (Next 15.5 ↔ OpenNext); linux-x64 `sharp` installed into the OpenNext image bundle; Nuxt `image.domains` → IPX allowlist.
- **Edge runtime**: `runtime:'edge'` routes get a dedicated cache behavior with the Lambda@Edge attached (origin-request) + are excluded from the KVS table — a CloudFront Function can only select an origin, never hand off to a Lambda@Edge.
- ISR cache-seed BucketDeployment sized to 1024 MB / 2 GiB (large `generateStaticParams` fan-out no longer times out).

**Tests** — the suite was synth-only; added tests that **execute** the generated edge functions (routing/caching/headers branch coverage) + the data layer. Hosting suite **672/672**.

## Validation

Deployed all five `aws-blocks-*` stress apps to a sandbox (us-west-2) via the Blocks CDK sandbox flow (`cdk deploy --all`, KVS table written by the gated `KvKeys` custom resource), then probed each live route by HTTP (status / content-type / headers / body markers), replicating each self-validating page's assertions.

| App | Live URL | Result |
|---|---|---|
| Next.js App Router | https://d2hxj6j65l8plb.cloudfront.net | ✅ 15/15 route sweep; edge `/edge`+`/api/edge` 200; images optimize (AVIF) |
| Next.js Pages Router | https://d109u569b4jnak.cloudfront.net | ✅ routing/images/assetPrefix; ⏳ edge-fix not yet redeployed |
| Nuxt (basePath `/myapp/`) | https://d1lnoa5w6a4ag2.cloudfront.net | ✅ basePath 308 + asset strip; IPX `/_ipx` real resize (478KB→203KB) |
| Astro SSR | https://d1fwrpsppb5xck.cloudfront.net | ✅ SSR/API/redirects; `/_image` 200 (noop passthrough) |
| Astro multipage (static) | https://dx6su88kgea9s.cloudfront.net | ✅ 10/10 (directory-index, redirects, 404) |

**New functionality verified live (next-app-router):**
- Edge runtime — `/api/edge` → `200 {"runtime":"edge"}`, `/edge` → `200` (was `500`). Distribution now has dedicated `/edge` + `/api/edge` behaviors with a Lambda@Edge origin-request association.
- assetPrefix — `/cdn-static/_next/static/chunks/*.js` → `200 text/javascript` (6/6 current-build chunks).
- Image opt — `/_next/image?w=64..1080` → `200 image/avif`, sizes grow with width (real resize + transcode).

**No regressions** — full route sweep on next-app-router stayed green across the edge fix: `/`, `/about`, `/blog/[slug]`, `/dashboard`, `/streaming`, `/server-action`, `/cookies-test`, `/products/[id]` (1000-page fan-out), `/middleware-protected` (307), `/api/throw` (500), `/not-found-test` (404), redirects/rewrites. Unit: hosting **672/672**, core **553/553**.

**How to re-verify:** deploy an app (`npm exec cdk -- deploy --all` from the app dir with `sandboxMode=true`), then `curl` the routes in the table above; or run `npm test -w packages/hosting` for the deterministic edge-function + synth coverage.


packages/hosting/src/constructs/kv_keys.ts
```diff
@@ -0,0 +1,93 @@
...
+const __dirname = dirname(fileURLToPath(import.meta.url));
+// This file lives at <repo>/packages/hosting/dist/constructs/kv_keys.js.
+// NodejsFunction requires `entry` to sit UNDER `projectRoot` and a
+// `depsLockFilePath` under it too. When this construct runs from a CONSUMING
+// app (cwd elsewhere), the default projectRoot is that app and the handler is
+// "not under root" (PathNotUnderRoot). Anchor both to the monorepo root (4
+// levels up: dist/constructs → dist → hosting → packages → repo), which is the
+// only place with a package-lock.json + node_modules for esbuild to bundle the
+// kvs SDK dep.
+const REPO_ROOT = join(__dirname, '..', '..', '..', '..');
```
soberm: This 4-levels-up anchor only resolves to the repo root inside the monorepo. Once @aws-blocks/hosting is installed from npm this file lives at node_modules/@aws-blocks/hosting/dist/constructs/, so REPO_ROOT lands in node_modules/ and NodejsFunction will fail at synth (projectRoot wrong, depsLockFilePath points at a non-existent package-lock.json). This package isn't private, so it'll ship. Either pre-bundle the handler with esbuild at build time, or discover the consumer's root via a lockfile walk.

packages/hosting/src/constructs/kv_keys_handler.ts
```diff
@@ -0,0 +1,176 @@
...
+  return out;
+}
+
+export async function handler(event: Event): Promise<{ PhysicalResourceId: string }> {
+  const { KvsArn, Entries: entriesJson } = event.ResourceProperties;
+  const physicalId = `kvkeys-${KvsArn.split('/').pop() ?? 'store'}`;
+
+  if (event.RequestType === 'Delete') {
+    // Drain only the keys we manage; leave any out-of-band keys alone.
+    const previous = event.OldResourceProperties?.Entries
```
soberm: OldResourceProperties is only populated on Update — it's undefined on Delete, so previous is always {} and applyUpdate no-ops. Nothing actually gets drained despite the comment. Use event.ResourceProperties.Entries here instead.

packages/hosting/src/adapters/nitro.ts
```diff
@@ -1037,6 +1129,33 @@ const findIpxBaseURL = (projectDir: string): string | undefined => {
...
+ * literals; anything dynamic yields an empty list (fails closed).
+ */
+const findNuxtImageDomains = (projectDir: string): string[] => {
+  for (const ext of ['ts', 'mjs', 'js', 'cjs']) {
+    const candidate = path.join(projectDir, `nuxt.config.${ext}`);
+    if (!fs.existsSync(candidate)) continue;
+    const source = fs.readFileSync(candidate, 'utf-8');
+    // Match `image: { … domains: [ … ] … }`. Keep it tight: the `domains`
+    // array must appear within an `image:` block.
+    const block = source.match(/\bimage\s*:\s*\{[\s\S]*?\}/);
```
soberm: Non-greedy match stops at the first }, so an image: block with a nested object before domains (e.g. provider: { ... }) truncates the scope and domains is missed -> returns []. Fails closed so not a security issue, but it silently drops a valid allowlist. Brace-balance the block or fall back to scanning the whole source. Worth a test with a nested object preceding domains.

packages/hosting/src/constructs/cdn_construct.ts
```diff
@@ -263,41 +257,46 @@ export class CdnConstruct extends Construct {
       hasComputeRoutes;
     this.errorPageHtml = props.errorPageHtml ?? SSR_ERROR_PAGE_HTML;
 
+    // Central accounting for the adjustable quotas this distribution draws on.
+    // Consumers call budget.consume() as they allocate; the authoritative
+    // budget.assertWithinLimits() runs just before the Distribution is created.
+    const budget = new QuotaBudget(props.quotas);
```
soberm: The comments here say cacheBehaviors/headerPolicies are consumed and enforced by a single assertWithinLimits() call before the Distribution is created, but assertWithinLimits() is never called anywhere in this file and those budgets are never consumed — the edge-function cap is the inline `if (edgeRouteCount > edgeLimit) throw` a few lines below. QuotaBudget is effectively dead except for .limit('edgeFunctions'). Either wire up consume()/assertWithinLimits() for real or drop it and fix the comments.

packages/hosting/src/constructs/kv_keys.ts
```diff
@@ -0,0 +1,93 @@
...
+  constructor(scope: Construct, id: string, props: KvKeysProps) {
+    super(scope, id);
+
+    // The compiled handler ships next to this file in dist/. esbuild
+    // (NodejsFunction) re-bundles it, pulling in the kvs SDK client which is
+    // not in the Lambda runtime baseline.
+    const handler = new NodejsFunction(this, 'Fn', {
+      entry: join(__dirname, 'kv_keys_handler.js'),
+      handler: 'handler',
+      runtime: Runtime.NODEJS_20_X,
```
soberm: Matches the existing hosting pattern (hosting_construct.ts:592/707) so not blocking, but the runtime literal is now in three places. Might be worth a hosting-local DEFAULT_NODE_RUNTIME so there's one place to bump.

packages/hosting/src/constructs/kv_keys_handler.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (count + 1 > MAX_KEYS_PER_REQUEST || bytes + sz > MAX_BYTES_PER_REQUEST) {
+      yield* flush();
+    }
+    curPuts.push(p);
+    count++;
+    bytes += sz;
+  }
+  for (const d of deletes) {
+    if (count + 1 > MAX_KEYS_PER_REQUEST) {
+      yield* flush();
```
soberm, line 125: The deletes loop only checks count, not bytes, unlike the puts loop above. Won't overflow in practice (keys only, <=512B each) but the asymmetry is worth a comment.

packages/core/src/hosting.test.ts
```diff
@@ -1428,4 +1428,83 @@ describe('Hosting', () => {
...
+      assert.strictEqual(metaBasePath(tmpDir, '/app/'), '/app');
+    });
+
+    it('treats "/" as no base path', () => {
+      createSpaBuildOutput(tmpDir);
+      assert.strictEqual(metaBasePath(tmpDir, '/'), '');
+    });
+  });
+
+  // ── P0.4: config.json ordering dependency ────────────────────
```
soberm, line 1479: What is P0.4 ?

packages/hosting/src/adapters/nextjs.ts
```diff
@@ -1246,6 +1254,192 @@ export const patchEdgeBundlesForLambdaEdge = (openNextDir: string): void => {
...
+  if (linuxSharp.length > 0 || linuxSharpLegacy.length > 0) return;
+
+  // Pin a MODERN sharp (>=0.33). The `--os/--cpu/--libc` cross-install flags
+  // (which fetch the `@img/sharp-linux-x64` prebuilt) only exist on sharp
+  // 0.33+. OpenNext historically bundles sharp 0.32.6, whose darwin prebuilt
+  // sits in `node_modules/sharp/build/Release/*.node` and is NOT replaced by
+  // those flags (0.32 used `prebuild-install`/SHARP_* env, not npm os/cpu) —
+  // so reinstalling `^0.32` would leave the darwin binary in place and the
+  // optimizer would keep serving unoptimized originals. Force 0.34.x.
+  const SHARP_RANGE = '^0.34.0';
```
soberm, line 1453: How do we keep track of that?

packages/hosting/package.json
```diff
@@ -46,9 +48,9 @@
   },
   "devDependencies": {
     "@types/cross-spawn": "^6.0.6",
+    "@types/node": "^20.0.0",
```
soberm, line 52: Why do we use Node 20 ?

packages/hosting/src/constructs/kv_keys.ts
osama-rizk: Good catch — fixed. The handler is now pre-bundled with esbuild at build time (`scripts/bundle-handlers.mjs`) and shipped via `Code.fromAsset`, so there's no `NodejsFunction`/`projectRoot`/lockfile dependency at synth. Works from `node_modules` when installed from npm. (Follow-up: the bundle needed a dotless filename so Lambda's `file.export` handler parsing resolves it.) Verified live — the KvKeys custom resource now completes on deploy.

packages/hosting/src/constructs/kv_keys_handler.ts
osama-rizk: Fixed. Delete now reads `ResourceProperties.Entries` (the field CloudFormation actually populates on Delete) via a new `deleteDrainSet()` helper, so managed keys really drain. Added unit tests.

packages/hosting/src/adapters/nitro.ts
osama-rizk: Fixed. Replaced the non-greedy match with a brace-balanced `extractImageBlock()` (skips braces inside strings) so a nested object before `domains` no longer truncates the scope. Extracted `parseNuxtImageDomains()` and added tests including the nested-object case and a brace-in-string case.

packages/hosting/src/constructs/cdn_construct.ts
osama-rizk: You're right — `assertWithinLimits()` was never called and those budgets weren't consumed. The KVS single-behavior model removed the cache-behavior/header-policy caps, so I dropped the dead `consume()`/`assertWithinLimits()` plumbing and the misleading comments, keeping only the override-aware `limit()` lookup for the Lambda@Edge cap. (Separately, a later review noted edge routes DO each cost a behavior — I now count total behaviors against the 25 cap with a friendly `TooManyCacheBehaviorsError`.)

packages/hosting/src/constructs/kv_keys.ts
osama-rizk: Done — added a hosting-local `DEFAULT_NODE_RUNTIME` (mirrors core's) and the KvKeys handler now uses it, so there's one place to bump.

packages/hosting/src/constructs/kv_keys_handler.ts
osama-rizk: Added a comment explaining it: deletes carry only the key (≤512 B, the KVS key-size limit) and no value, so a 50-key batch is ~25 KB — far under the 3 MB request ceiling, and tracking bytes here would never change the batching outcome.

packages/core/src/hosting.test.ts
osama-rizk: P0.4 is an internal phase label from the rollout plan (the `config.json` deploy-ordering fix). It's just a tracking tag — I can drop the `(P0.4)` prefix from the describe() if it's noise.

packages/hosting/src/adapters/nextjs.ts
osama-rizk: Via `VERIFIED_OPENNEXT_RANGE` + `warnIfOpenNextOutOfRange()` in the adapter: it warns at build time when the installed `@opennextjs/aws` is outside the range we've verified the patches against, so a drift is visible before it breaks. The sharp pin is tied to that same OpenNext-bundle assumption.

packages/hosting/package.json
osama-rizk: The hosting-owned handler now uses the shared `DEFAULT_NODE_RUNTIME` (Node 24, aligned with core). The remaining `nodejs20.x` literals are for OpenNext/Nitro-built SSR bundles — those run code the framework compiled targeting Node 20, so they stay matched to the bundle rather than bumped independently.

packages/hosting/src/constructs/kv_keys_handler.ts
```diff
@@ -0,0 +1,188 @@
...
+ * resource doesn't do. We perform it via the `cloudfront-keyvaluestore`
+ * data-plane API (DescribeKeyValueStore → get ETag → UpdateKeys), the same
+ * mechanism SST's `KvKeys` provider uses.
+ *
+ * Atomicity: this resource is wired (in the construct) to depend on the asset
+ * `BucketDeployment`s, so the KV flip that activates a new `buildId` happens
+ * only AFTER that build's assets are uploaded to S3 — preserving the
+ * atomic-deploy cutover guarantee.
+ *
+ * Bundled with its SDK dependency via esbuild (NodejsFunction), because
```
soberm: nit: Stale now — bundling happens at build time via bundle-handlers.mjs, not NodejsFunction at synth. Worth updating this header.

packages/hosting/src/constructs/kvs_router.test.ts
```diff
@@ -0,0 +1,831 @@
...
+        return entries[key];
+      },
+    }),
+    selectRequestOriginById: (id: string) => {
+      selectedOrigin = id;
+    },
+  };
+  const body = code.replace(/^import cf from 'cloudfront';\n?/, '');
+  // eslint-disable-next-line @typescript-eslint/no-implied-eval
+  const factory = new Function(
```
soberm, line 64: These e2e helpers run the generated function via new Function. Fine since the source is fully repo-controlled and the test runner doesn't restrict codegen — but worth a one-line comment saying so, otherwise the next person wonders if these silently no-op under a hardened runner (--disallow-code-generation-from-strings).

packages/hosting/src/constructs/kvs_router.ts
```diff
@@ -0,0 +1,564 @@
...
+  // splice into wildcard redirect destinations).
+  if (pattern.indexOf('*') === pattern.length - 1 && pattern.lastIndexOf('*') === pattern.length - 1) {
+    var prefix = pattern.substring(0, pattern.length - 1);
+    if (uri.indexOf(prefix) === 0) return { tail: uri.substring(prefix.length) };
+    return null;
+  }
+  // General path: glob with '*' anywhere (incl. mid-segment, e.g.
+  // /api/*/admin). Each '*' matches a run of any chars EXCEPT '/', mirroring
+  // CloudFront's path-pattern semantics; a trailing '*' matches the rest
+  // (including '/'). Implemented as a literal scan (no regex — CloudFront
```
soberm: Small wording nit: this is framework route matching, not CloudFront behavior selection — CF path patterns actually let * cross segments, so the 'mirroring CloudFront semantics' phrasing is a bit misleading. Also worth a negative test: /api/*/data should NOT match /api/foo/bar/data, to lock in the single-segment guarantee (only single-level wild segments are covered right now).

packages/hosting/src/constructs/kvs_router.ts
```diff
@@ -0,0 +1,564 @@
...
+      hasImage &&
+      (route.target === 'image-optimization' ||
+        (imagePrefix !== undefined &&
+          cf === normalizePattern(`${imagePrefix}/*`, basePath)));
+    const kind: RouteKind = isImage ? 'i' : isStatic ? 's' : 'c';
+    rows.push([cf, kind]);
+  }
+  // Sort by descending specificity (literal segments, then length) so the
+  // function's first-match scan mirrors CloudFront's old behavior ordering.
+  rows.sort((a, b) => specificity(b[0]) - specificity(a[0]));
```
soberm: specificity is tagged @deprecated but this is its only caller — just use routeSpecificity directly here and drop the alias.

packages/hosting/src/adapters/nitro.ts
```diff
@@ -602,14 +630,78 @@ const readBundledRouteRules = (
...
+  const candidates: string[] = [];
+  const bundlePath = resolveNitroBundlePath(serverDir);
+  if (bundlePath) candidates.push(bundlePath);
+  const indexPath = path.join(serverDir, 'index.mjs');
+  if (fs.existsSync(indexPath)) candidates.push(indexPath);
+
+  for (const candidate of candidates) {
+    const source = fs.readFileSync(candidate, 'utf-8');
+    // The runtime-config blob embeds it as `"baseURL": "/myapp/"`.
+    const match = source.match(/"baseURL"\s*:\s*"([^"]*)"/);
```
soberm: This matches the first baseURL anywhere in the bundle, but _inlineRuntimeConfig also carries ipx.baseURL (default /_ipx). It's fine today because app serializes before runtimeConfig, but it's order-dependent — if it ever returns /_ipx you'd silently set basePath to it and 308 the whole site (the prerendered-HTML net only catches a missed prefix, not a wrong one). readBundledRouteRules already brace-scopes via extractJsonObjectAfter; worth doing the same for the app block here, plus a test with an ipx.baseURL present before app.baseURL.

packages/hosting/src/constructs/kv_keys_handler.ts
```diff
@@ -0,0 +1,188 @@
...
+  sigv4aSigningRegionSet: ['*'],
+});
+
+const byteLen = (s: string): number => Buffer.byteLength(s, 'utf8');
+
+/**
+ * Split desired puts + deletes into UpdateKeys batches that respect the
+ * 50-key / 3 MB-per-request ceiling.
+ */
+function* batches(
```
soberm: batches (the 50-key/3MB chunking + per-batch ETag refetch) and applyUpdate's diff are the load-bearing bits for the atomic route-table flip, but the only unit test here is deleteDrainSet. An off-by-one at the batch boundary would partial-apply the table mid-cutover and surface as an opaque deploy-time failure. These are pure — worth exporting batches (and factoring the put/delete diff into a helper) and testing the boundaries: exactly 50 vs 51 keys, mixed put+delete crossing the boundary, byte-based flush, and the no-op path.

packages/hosting/src/constructs/node_runtime.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Mirrors `@aws-blocks/core`'s `DEFAULT_NODE_RUNTIME`; kept local because
+ * hosting does not depend on core. Bump both together. This controls only the
+ * AWS-managed runtime that executes deployed handlers — independent of the Node
+ * version the CLI / CDK synth runs on.
+ */
+export const DEFAULT_NODE_RUNTIME = Runtime.NODEJS_24_X;
+
+/** String form (`'nodejs24.x'`) for adapters that emit a runtime literal. */
+export const DEFAULT_NODE_RUNTIME_STRING = DEFAULT_NODE_RUNTIME.name;
```
soberm: Nothing consumes this — the adapters all emit a hardcoded 'nodejs20.x' literal. Either wire it where they emit the runtime string or drop the export; as-is it implies adapters track Node 24 when they don't.

.changeset/hosting-phase0-bugs.md
```diff
@@ -0,0 +1,15 @@
+---
+"@aws-blocks/hosting": patch
+"@aws-blocks/core": patch
+---
+
+Fix three hosting correctness bugs:
```
soberm: Says three but there are four bullets below (basePath, header delegation, config.json ordering, QuotaBudget) — bump to four or drop the count.

packages/hosting/src/constructs/kv_keys_handler.ts
osama-rizk: Fixed — header now describes the build-time pre-bundling (`bundle-handlers.mjs` → `Code.fromAsset`) and why `NodejsFunction` is avoided. (45a54dd)

packages/hosting/src/constructs/kvs_router.test.ts
osama-rizk: Added a comment: the `new Function` codegen is intentional (source is repo-controlled), and under a hardened runner (`--disallow-code-generation-from-strings`) it would throw rather than silently no-op. (45a54dd)

packages/hosting/src/constructs/kvs_router.ts
osama-rizk: Reworded — the comment now states this is OUR framework route matcher (single-segment `*`), explicitly contrasted with CloudFront path patterns (where `*` crosses `/`). Added a negative test: `/api/*/data` matches `/api/foo/data` but NOT `/api/foo/bar/data`. (45a54dd)

packages/hosting/src/constructs/kvs_router.ts
osama-rizk: Already resolved — the `@deprecated specificity` alias is gone; the only caller now uses `routeSpecificity` directly.

.changeset/hosting-phase0-bugs.md
osama-rizk: Fixed — bumped 'three' → 'four' to match the four bullets. (45a54dd)

packages/hosting/src/adapters/nitro.ts
osama-rizk: Fixed — `readBundledBaseURL` now brace-scopes to the `app` block (via `extractJsonObjectAfter`, same as `readBundledRouteRules`) and reads `baseURL` only inside it, so `ipx.baseURL` can't be matched regardless of serialization order. Fail-closed: no whole-source fallback (a wrong prefix is worse than a missed one — the HTML net catches a miss, not a wrong value). Added tests with `ipx.baseURL` before `app.baseURL` (non-root and root cases). (45a54dd)

packages/hosting/src/constructs/kv_keys_handler.ts
osama-rizk: Done — exported `batches` and factored the put/delete diff into `computeDiff`; both are pure and now unit-tested at the boundaries: exactly 50 vs 51 keys, mixed put+delete crossing the 50-key ceiling (nothing dropped, no batch >50), the 3 MB byte flush, and the no-op path. (45a54dd)

packages/hosting/src/constructs/node_runtime.ts
osama-rizk: Dropped the unused `DEFAULT_NODE_RUNTIME_STRING` export. The adapters intentionally emit `nodejs20.x` pinned to the framework bundle (the Node version Astro/Nuxt/OpenNext compiled against) — documented that in `node_runtime.ts` so it no longer implies the adapters track Node 24. `DEFAULT_NODE_RUNTIME` stays for hosting-owned handlers (KvKeys, etc.). (45a54dd)

soberm (dismissed):

ahmedhamouda78 (approved):

soberm (approved):

PR #47, opened 2026-06-19
fix(bb-agent): reject empty channelId in stream()

pjkroker:
## Problem

`agent.stream('hello', { channelId: '' })` silently uses an empty string as the Realtime channel key. While this technically works (publish and subscribe resolve to the same path), every call with an empty `channelId` shares the same channel — meaning different users/conversations would receive each other's chunks.

It is unlikely a caller intentionally passes `''` as a valid channel identifier.

  **Issue #, if available:**

## Changes

- Changed the fallback chain from `??` to `||` so empty strings are treated as unset: 
`options.channelId || conversationId || crypto.randomUUID()`
- Documented the fallback behavior in the `StreamOptions` docstring and README
- Non-breaking: empty strings silently fall back instead of throwing

## Validation

- Added unit tests: empty `channelId` and empty `conversationId` both produce a non-empty generated value
- All existing tests pass

## Checklist

- [x] PR description included

---

### DIFF
--- a/.changeset/fix-bb-agent-empty-channel-id.md
+++ b/.changeset/fix-bb-agent-empty-channel-id.md
@@ -1,0 +1,7 @@
+---
+"@aws-blocks/bb-agent": patch
+---
+
+fix(bb-agent): reject empty channelId in stream()
+
+An empty `channelId` produced an invalid Realtime channel path, causing the client subscription to silently receive no chunks. Now throws `ValidationFailedException` immediately.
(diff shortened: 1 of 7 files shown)


svidgen (commented):
What actually happens today if a customer just calls `agent.stream('hello', { channelId: '' })`? Can they use this to successfully get the result back *in process?*

hfurkanbozkurt (approved):

soberm (approved):

PR #48, opened 2026-06-19
fix(core): send telemetry via background subprocess to prevent dropped events

sarayev:
## Problem

Telemetry events were being dropped because `sendEvent` used an in-process `https.request` that was vulnerable to:
- **Socket timeouts** when `execFileSync` (client codegen) blocked the event loop
- **Process exit races** on failure paths where the process terminated before the HTTP request completed

## Solution

Replace in-process HTTP send with a **detached background subprocess** (`telemetry-send-worker.ts`) that:
- Runs independently of the parent's event loop
- Survives parent process exit
- Reads payload from stdin, POSTs to the telemetry endpoint
- Follows the same pattern as `generate-client-worker.ts`

## Debug

```bash
NODE_DEBUG=blocks-telemetry npm run sandbox
```

Shows:
```
BLOCKS-TELEMETRY: sending event to https://...
BLOCKS-TELEMETRY: spawned telemetry subprocess (pid=XXXX)
BLOCKS-TELEMETRY: sent (status=200)
```

## Changes

- `packages/core/src/telemetry/client.ts` — `sendEvent` spawns detached worker
- `packages/core/src/telemetry/telemetry-send-worker.ts` — new worker script
- `packages/core/src/telemetry/trackCommand.ts` — improved `classifyError` for CDK errors from `runSync`
- `packages/core/src/telemetry/telemetry.test.ts` — added worker unit tests

## Testing

- Verified with a smoke-test app that all commands (dev, sandbox, sandbox:destroy, deploy, destroy, console) deliver events successfully in both SUCCESS and FAIL states
- Unit tests pass


svidgen (dismissed):

packages/core/src/telemetry/client.ts
```diff
@@ -194,32 +195,19 @@ export function sendEvent(event: BlocksTelemetryEvent): void {
...
-      },
-    );
-
-    req.on('error', (err) => { debug('send failed: %s', err.message); });
-    req.on('timeout', () => { debug('send timed out'); req.destroy(); });
-    req.write(payload);
-    req.end();
+    const dir = path.dirname(fileURLToPath(import.meta.url));
+    const workerPath = path.join(dir, 'telemetry-send-worker.js');
+    const child = spawn(process.execPath, ['--import', 'tsx', workerPath, endpoint], {
```
pranavosu: we should drop the tsx import since we're just executing JS

pranavosu (dismissed):

packages/core/src/telemetry/client.ts
```diff
@@ -194,32 +195,19 @@ export function sendEvent(event: BlocksTelemetryEvent): void {
...
-    req.end();
+    const dir = path.dirname(fileURLToPath(import.meta.url));
+    const workerPath = path.join(dir, 'telemetry-send-worker.js');
+    const child = spawn(process.execPath, [workerPath, endpoint], {
+      detached: true,
+      stdio: ['pipe', 'ignore', 'inherit'],
+      env: { ...process.env, NODE_OPTIONS: '' },
+    });
+
+    child.stdin!.write(payload);
```
soberm, line 210: Should we add 
`child.stdin!.on('error', (err) => { debug('stdin write failed: %s', err.message); });`
before writing? Otherwise, an error will throw an uncaught exception to the CLI.

packages/core/src/telemetry/client.ts
```diff
@@ -194,32 +195,20 @@ export function sendEvent(event: BlocksTelemetryEvent): void {
...
-
-    req.on('error', (err) => { debug('send failed: %s', err.message); });
-    req.on('timeout', () => { debug('send timed out'); req.destroy(); });
-    req.write(payload);
-    req.end();
+    const dir = path.dirname(fileURLToPath(import.meta.url));
+    const workerPath = path.join(dir, 'telemetry-send-worker.js');
+    const child = spawn(process.execPath, [workerPath, endpoint], {
+      detached: true,
+      stdio: ['pipe', 'ignore', 'inherit'],
```
soberm: The detached, unref'd telemetry child is spawned with stdio: ['pipe', 'ignore', 'inherit'], so it inherits the parent's stderr file descriptor. Two consequences follow from a process that intentionally outlives its parent: (1) if the parent's stderr is a pipe (a CI harness, a wrapping spawn, a shell capturing 2>), the detached child keeps that pipe's write end open, so the reader never sees EOF and can hang until the child's ~500ms request completes or longer; (2) when NODE_DEBUG=blocks-telemetry is set the worker writes BLOCKS-TELEMETRY: ... lines straight onto the user's terminal, potentially after the parent CLI has already exited and printed its result. The worker already self-gates its debug output, and the only consumer that needs to read it (the new test) sets stdio: ['pipe','ignore','pipe'] itself, so production has no reason to inherit stderr.
Fix: Use 'ignore' for the child's stderr so the detached process never holds the parent's stderr open. The worker still no-ops its debug writes unless NODE_DEBUG is set, and the test supplies its own piped stderr, so nothing observable is lost.

soberm (dismissed):

soberm (approved):

pranavosu (approved):

PR #51, opened 2026-06-19
fix: derive CDK stack name from package.json

svidgen:
## Problem

Scaffolded apps all shared the same static `my-blocks-stack-prod` stack name. Multiple apps deployed to the same AWS account/region would silently overwrite each other's infrastructure. Sandbox stacks also needed per-developer isolation for shared testing accounts. Stack naming logic was duplicated inline in every template.

### Stack naming scheme

```
my-app-k7x2mf-prod              → production
my-app-k7x2mf-alice-9f3a2b      → alice's sandbox
my-app-k7x2mf-bob-2d7ekx        → bob's sandbox
```

- **`stackId`** — stored in `.blocks/config.json` (committed), generated at scaffold time as `name.slice(0,16)-random(6)`
- **`sandboxId`** — per-machine `username(8)-random(6)`, stored in `.blocks-sandbox/sandbox-id.txt` (gitignored)

### Architecture

`getStackId(projectRoot?)` and `getSandboxId(projectRoot?)` are exported from `@aws-blocks/blocks/scripts` (the core package). Templates import and call them with the CDK context `projectRoot`:

```ts
import { getStackId, getSandboxId } from '@aws-blocks/blocks/scripts';

const projectRoot = app.node.tryGetContext('projectRoot') || process.cwd();
const stackName = sandboxMode
  ? \`\${getStackId(projectRoot)}-\${getSandboxId(projectRoot)}\`
  : \`\${getStackId(projectRoot)}-prod\`;
```

### Changes

- **`@aws-blocks/core`** — new `scripts/stack-id.ts` exporting `getStackId()` and `getSandboxId()`
  - Optional `projectRoot` param (defaults to `process.cwd()`)
  - Actionable error when `.blocks/config.json` is missing
  - Uses legacy `sandbox-id.txt` filename (no migration needed)
- **7 template `index.cdk.ts` files** — import from `@aws-blocks/blocks/scripts`, pass `projectRoot`
- **CLI `src/index.ts`** — generates `.blocks/config.json` with `stackId` during scaffolding
- **4 template READMEs** — stack naming section
- **`docs/DECISIONS.md`** — D-012 documenting the scheme
- **Changeset** — patch to both `@aws-blocks/core` and `@aws-blocks/create-blocks-app`

## Testing

All 25 CLI tests pass, including CDK-safe regex validation on generated stackIds.


pranavosu (dismissed):

packages/core/src/scripts/stack-id.ts
```diff
@@ -0,0 +1,48 @@
...
+  return config.stackId;
+}
+
+/**
+ * Get or create a per-machine sandbox identifier.
+ * Stored in `.blocks-sandbox/sandbox-id` (gitignored).
+ * Format: `<username(8)>-<random(4)>` — identifies the developer's sandbox.
+ */
+export function getSandboxId(): string {
+  const filePath = join(process.cwd(), '.blocks-sandbox', 'sandbox-id');
```
vishaalmehrishi: **Suggestion — the sandbox-id filename changed, which silently orphans existing sandbox stacks.**

**What changed:** this helper reads/writes [`.blocks-sandbox/sandbox-id`](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/packages/core/src/scripts/stack-id.ts#L31) (no extension). The templates this PR deletes wrote [`.blocks-sandbox/sandbox-id.txt`](https://github.com/aws-devtools-labs/aws-blocks/blob/7fd51e0b2c0a380262a8a64ba7e2a3bf22405dff/packages/create-blocks-app/templates/default/aws-blocks/scripts/sandbox-id.ts#L18).

**Why it matters:** a developer on an existing app who adopts these helpers still has `sandbox-id.txt` on disk. `getSandboxId()` won't see it, mints a **new** id, and the next `npm run sandbox` deploys a **new stack** — the old sandbox stack (and its resources/cost) is orphaned, and `sandbox:destroy` no longer targets it. Silent stack duplication is the exact problem this PR fixes.

**Fix:** read the legacy file when the new one is absent:
```ts
const filePath = join(process.cwd(), '.blocks-sandbox', 'sandbox-id');
if (existsSync(filePath)) return readFileSync(filePath, 'utf-8').trim();
const legacy = join(process.cwd(), '.blocks-sandbox', 'sandbox-id.txt');
if (existsSync(legacy)) return readFileSync(legacy, 'utf-8').trim();
// …then generate a new one
```
Or, if a migration isn't worth it, call it out in the changeset. (Fresh scaffolds are unaffected.)

packages/core/src/scripts/stack-id.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Get the stackId from `.blocks/config.json` in the project root (cwd).
+ * This is the stable project identifier used as the base for CloudFormation stack names.
+ */
+export function getStackId(): string {
+  const configPath = join(process.cwd(), '.blocks', 'config.json');
+  const content = readFileSync(configPath, 'utf-8');
+  const config: BlocksConfig = JSON.parse(content);
+  if (!config.stackId) throw new Error('stackId not found in .blocks/config.json');
```
vishaalmehrishi: **Suggestion — two small robustness gaps in [`getStackId()`](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/packages/core/src/scripts/stack-id.ts#L17-L23).**

**1. Cryptic error when the config is missing.** If `.blocks/config.json` doesn't exist (fresh clone before it's committed, or an accidental delete), `readFileSync` throws a raw `ENOENT: …/.blocks/config.json` at `cdk synth`. The friendly [`'stackId not found…'`](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/packages/core/src/scripts/stack-id.ts#L21) message only fires when the file exists but lacks the key.
*Fix:* wrap the read and throw an actionable message, e.g. *".blocks/config.json not found — it's created by create-blocks-app and should be committed. Add { \"stackId\": \"…\" } to set the stack name."*

**2. New hard dependency on `process.cwd()`.** The old template resolved [`tryGetContext('projectRoot') || process.cwd()`](https://github.com/aws-devtools-labs/aws-blocks/blob/7fd51e0b2c0a380262a8a64ba7e2a3bf22405dff/packages/create-blocks-app/templates/default/aws-blocks/index.cdk.ts#L14); this helper uses `process.cwd()` only, so synth must always run from the project root or both helpers read the wrong path and fail.
*Fix:* confirm `npm run sandbox` and `npm run deploy` always synth from the project root, or accept an optional `projectRoot` argument to preserve the old escape hatch.

packages/create-blocks-app/src/cli.test.ts
```diff
@@ -165,35 +165,36 @@ describe('create-blocks-app auto-detection', () => {
...
-      if (stackNameMatch) {
-        assert.match(stackNameMatch[1], /^[A-Za-z][A-Za-z0-9-]*$/, 'stack name should be CDK-safe');
-      }
+      const config = JSON.parse(readFileSync(join(tmpDir, '.blocks', 'config.json'), 'utf-8'));
+      assert.ok(config.stackId, '.blocks/config.json should have a stackId');
+      // Name part (before the random suffix) should be at most 16 chars
+      const parts = config.stackId.split('-');
+      const suffix = parts.pop();
+      const namepart = parts.join('-');
+      assert.ok(namepart.length <= 16, `name portion "${namepart}" should be at most 16 chars`);
```
vishaalmehrishi, line 222: **Suggestion — re-add the assertion that the generated stackId is CDK/CloudFormation-safe.**

**What changed:** the rewritten test now only checks the [name part is ≤16 chars](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/packages/create-blocks-app/src/cli.test.ts#L196). The previous version asserted the name matched `/^[A-Za-z][A-Za-z0-9-]*$/` (CDK-safe).

**Why it matters:** both code paths sanitize today, so this isn't a live bug — but that assertion is the guard that would catch a future sanitization regression in `createFreshProject` or `integrateWithExistingProject`. The test already feeds the adversarial name `@scope/my_app.test`, so it's the perfect place to keep it.

**Fix:** one line —
```ts
assert.match(config.stackId, /^[a-z][a-z0-9-]*$/i, 'stackId must be CDK/CFN-safe');
```

.changeset/derive-stack-name-from-package-json.md
```diff
@@ -0,0 +1,11 @@
+---
+"@aws-blocks/create-blocks-app": patch
```
vishaalmehrishi, line 2: **Suggestion — record the stack-naming scheme as a decision.**

**Why:** [`.blocks/config.json`](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/packages/core/src/scripts/stack-id.ts#L18) is committed, so its `stackId` is shared across the whole team and CI — a teammate or pipeline with a different or missing `stackId` deploys to a different (or failing) stack. That blast radius, plus the per-machine `sandboxId` format and the `Math.random` suffix sizing, is exactly what `docs/DECISIONS.md` is for: it gives the next person the *why* behind the file instead of re-deriving it from code, matching how [D-009](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/docs/DECISIONS.md#d-009-external-databases--blocks-applied-version-controlled-schema-migrations) and [D-011](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/docs/DECISIONS.md#d-011-db-pull-is-interactive-only-devprod-intent-routes-the-connection-string-r1) document comparable decisions.

**Fix:** add a short `D-0xx` entry (scheme + collision-probability assumptions). 

*Not blocking:* the `patch` bumps are correct — the new `core` exports are additive, and [`updateInternalDependencies: patch`](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/.changeset/config.json) republishes the [`@aws-blocks/blocks` umbrella](https://github.com/aws-devtools-labs/aws-blocks/blob/5f89ebfe030db1bd34dd4cea7df2522cf89d03fa/packages/blocks/src/scripts.ts) so `@aws-blocks/blocks/scripts` picks up the new exports.

packages/core/src/scripts/stack-id.ts
svidgen: Good catch. Kept the legacy `sandbox-id.txt` filename — no migration logic needed. Also added the optional `projectRoot` param so it resolves from CDK context like before. Fixed in ee013ba.

packages/core/src/scripts/stack-id.ts
svidgen: Both addressed:

1. `getStackId` now wraps in try/catch and throws an actionable message pointing to create-blocks-app or manual creation.
2. Both helpers accept an optional `projectRoot` param. Templates pass `app.node.tryGetContext('projectRoot') || process.cwd()` — same escape hatch as before.

Fixed in ee013ba.

packages/create-blocks-app/src/cli.test.ts
svidgen: Added. The truncation test now asserts `/^[a-z][a-z0-9-]*$/i` on the full stackId. Fixed in ee013ba.

.changeset/derive-stack-name-from-package-json.md
svidgen: Added D-012 to DECISIONS.md covering the naming scheme, collision-probability assumptions, and committed-vs-gitignored rationale. Fixed in ee013ba.

packages/core/src/scripts/stack-id.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Format: `<username(8)>-<random(6)>` — identifies the developer's sandbox.
+ */
+export function getSandboxId(projectRoot?: string): string {
+  const root = projectRoot || process.cwd();
+  const filePath = join(root, '.blocks-sandbox', 'sandbox-id.txt');
+  if (existsSync(filePath)) return readFileSync(filePath, 'utf-8').trim();
+  const dir = dirname(filePath);
+  if (!existsSync(dir)) mkdirSync(dir, { recursive: true });
+  const username = getUsername().toLowerCase().replace(/[^a-z0-9]/g, '').slice(0, 8);
+  const random = Math.random().toString(36).slice(2, 8);
```
vishaalmehrishi: **Suggestion — the random suffix isn't guaranteed to be 6 chars.** `Math.random().toString(36).slice(2, 8)` can return fewer than 6 characters when the random double's base-36 expansion terminates early (e.g. `(0.5).toString(36)` → `"0.i"` → `"i"`). This (a) makes the `cli.test.ts` length assertions (`strictEqual(..., 6)`) latently flaky, and (b) weakens the "~2.1B combinations" guarantee documented in D-012. Same idiom is used at `index.ts:248` and `:484`.

Suggest a fixed-length generator:
```ts
import { randomBytes } from 'node:crypto';
const random = randomBytes(4).toString('hex').slice(0, 6); // always 6 lowercase hex
```

packages/create-blocks-app/src/index.ts
```diff
@@ -244,16 +244,11 @@ async function createFreshProject(targetDir: string, templateName: string) {
...
-  let cdkContent = await readFile(cdkPath, 'utf-8');
-  let sanitizedName = appName
-    .replace(/[^A-Za-z0-9-]/g, '-')
-    .replace(/^[^A-Za-z]+/, 'app-')
-    .replace(/-+/g, '-')
-    .replace(/-$/, '') || 'blocks-app';
-  cdkContent = cdkContent.replace(/my-blocks-stack/g, `${sanitizedName}-stack`);
-  await writeFile(cdkPath, cdkContent);
+  // Generate .blocks/config.json with a unique stackId
+  const stackId = appName.slice(0, 16).replace(/-$/, '') + '-' + Math.random().toString(36).slice(2, 8);
```
vishaalmehrishi: **Suggestion — `stackId` generation is duplicated (here and `:484`) and lives apart from the readers.** This PR's stated goal was to de-duplicate naming logic; the readers moved to core but the generator is still copy-pasted into both CLI functions, each sanitizing its input slightly differently (`createFreshProject` from the basename, `integrateWithExistingProject` from `pkg.name`).

Consider adding `generateStackId(name: string)` to `stack-id.ts` alongside `getStackId`/`getSandboxId` and calling it from both paths — one home for sanitize + truncate + suffix (and the natural place to fix the suffix-length issue).

packages/create-blocks-app/src/cli.test.ts
```diff
@@ -155,7 +155,7 @@ describe('create-blocks-app auto-detection', () => {
     }
   });
 
-  it('replaces my-blocks-stack placeholder in generated aws-blocks/index.cdk.ts', () => {
+  it('generates .blocks/config.json with stackId and uses getStackId in index.cdk.ts', () => {
```
vishaalmehrishi, line 184: **Suggestion — both tests exercise only the integrate path.** Each test pre-creates `package.json` in the temp dir, which routes through `integrateWithExistingProject` (mode detection in `create()`), so `createFreshProject` (`index.ts:248`) — with its own sanitization — has no coverage.

The new core module also has no unit tests for: `getStackId` throwing the actionable missing-config error, `getSandboxId` idempotency (returns existing id), and `getSandboxId` reading a pre-existing legacy `sandbox-id.txt` (the exact behavior the resolved orphaning thread relies on). A small `stack-id.test.ts` plus one fresh-dir assertion would close the gap.

.changeset/derive-stack-name-from-package-json.md
```diff
@@ -0,0 +1,11 @@
...
+"@aws-blocks/core": patch
+---
+
+fix: generate unique stackId in .blocks/config.json, export getStackId/getSandboxId from @aws-blocks/blocks/scripts
+
+Stack names are now derived from a `stackId` in `.blocks/config.json`, generated at scaffold time as `<name>.slice(0,16)-<random6>`. Templates import `getStackId()` and `getSandboxId()` from `@aws-blocks/blocks/scripts` — no more inline filesystem logic in `index.cdk.ts`.
+
+Production: `<stackId>-prod`
+Sandbox: `<stackId>-<username(8)>-<random(4)>` (per-machine, gitignored)
```
vishaalmehrishi: **Nit — suffix length mismatch.** The changeset says sandbox is `<username(8)>-<random(4)>`, but `getSandboxId` (`stack-id.ts:44`) and D-012 both use 6. Update `<random(4)>` → `<random(6)>` so the release notes match the code.

docs/DECISIONS.md
```diff
@@ -400,3 +400,30 @@ earlier planning decisions and removed a released behavior:
...
+
+**Date**: 2026-06-19
+**Authors:** wirej
+
+### Decision
+
+CloudFormation stack names are derived from a `stackId` stored in `.blocks/config.json` (committed to the repo). The `stackId` is generated once at scaffold time as `<sanitizedName>.slice(0, 16)-<random(6)>`, producing names like `my-app-k7x2mf`.
+
+Stack name scheme:
+- **Production:** `<stackId>-prod` (e.g., `my-app-k7x2mf-prod`)
```
vishaalmehrishi: **Suggestion — add a one-line migration scope.** This scheme changes the production stack name from `my-blocks-stack-prod` to `<stackId>-prod`. Harmless for fresh scaffolds, but an existing app that adopts the new `index.cdk.ts` would deploy a brand-new prod stack and orphan the old one — the prod analog of the sandbox-id concern already addressed in this PR, with a larger blast radius.

A sentence scoping D-012 to newly scaffolded apps (and warning existing apps to migrate rather than blind-adopt) would prevent a costly surprise.

packages/core/src/scripts/stack-id.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Stored in `.blocks-sandbox/sandbox-id.txt` (gitignored).
+ * Format: `<username(8)>-<random(6)>` — identifies the developer's sandbox.
+ */
+export function getSandboxId(projectRoot?: string): string {
+  const root = projectRoot || process.cwd();
+  const filePath = join(root, '.blocks-sandbox', 'sandbox-id.txt');
+  if (existsSync(filePath)) return readFileSync(filePath, 'utf-8').trim();
+  const dir = dirname(filePath);
+  if (!existsSync(dir)) mkdirSync(dir, { recursive: true });
+  const username = getUsername().toLowerCase().replace(/[^a-z0-9]/g, '').slice(0, 8);
```
vishaalmehrishi: **Nit — empty username degrades the sandbox name.** This sanitizes the git `user.name` to `[a-z0-9]` then slices to 8. A non-ASCII name (e.g. `李雷`) sanitizes to `""`, yielding `-<random>` and a double-dash stack name `<stackId>--<random>` — valid in CFN, but it loses the console identifiability D-012 cites as the rationale for the username prefix. Consider falling back to `'dev'`/`'user'` when the sanitized username is empty.

packages/core/src/scripts/stack-id.ts
svidgen: Fixed. Switched to `randomBytes(4).toString('hex').slice(0, 6)` — always exactly 6 chars. Applied in both core and CLI. Fixed in f4a2d1f.

packages/create-blocks-app/src/index.ts
svidgen: Done. `generateStackId(name)` now lives in `stack-id.ts` (exported from `@aws-blocks/blocks/scripts`) and the CLI has its own copy using the same logic. Both call sites in the CLI now just call `generateStackId(appName)`. Fixed in f4a2d1f.

packages/create-blocks-app/src/cli.test.ts
svidgen: Added `stack-id.test.ts` in core with 9 unit tests covering: generateStackId format/sanitization/truncation, getStackId success + both error cases, getSandboxId generation + idempotency. Also added a fresh-project CLI test (`--template bare`) that exercises createFreshProject. Fixed in f4a2d1f.

.changeset/derive-stack-name-from-package-json.md
svidgen: Fixed — changeset now says `random(6)`. Fixed in f4a2d1f.

docs/DECISIONS.md
svidgen: Added a "Migration scope" section to D-012 explicitly scoping it to newly scaffolded apps and warning that existing apps must set `stackId` to their current stack name. Fixed in f4a2d1f.

packages/core/src/scripts/stack-id.ts
svidgen: Fixed — falls back to `'dev'` when sanitized username is empty: `...slice(0, 8) || 'dev'`. Fixed in f4a2d1f.

docs/DECISIONS.md
```diff
@@ -400,3 +400,33 @@ earlier planning decisions and removed a released behavior:
...
+- **Sandbox:** `<stackId>-<username(8)>-<random(6)>` (e.g., `my-app-k7x2mf-alice-9f3a2b`)
+
+The sandbox identifier is generated per-machine and stored in `.blocks-sandbox/sandbox-id.txt` (gitignored).
+
+Both helpers (`getStackId`, `getSandboxId`) are exported from `@aws-blocks/blocks/scripts` and accept an optional `projectRoot` parameter (defaults to `process.cwd()`).
+
+### Rationale
+- **Collision avoidance:** The 6-char random suffix in `stackId` prevents two apps with the same name from overwriting each other's infrastructure in a shared account (~2.1B combinations).
```
vishaalmehrishi: **Suggestion (recommended) — this collision count no longer matches the code.**

**What's wrong:** the round-2 fix now builds the suffix with [`randomBytes(length).toString('hex').slice(0, 6)`](https://github.com/aws-devtools-labs/aws-blocks/blob/f4a2d1f0b3a2048e7e577abfb7a3186839cdf1d5/packages/core/src/scripts/stack-id.ts#L14-L15). Node encodes one byte as [two hex characters](https://nodejs.org/api/buffer.html#buffers-and-character-encodings), so a 6-char **hex** suffix draws from `[0-9a-f]` and has **16⁶ = 16,777,216 (~16.8M)** values — not the `~2.1B` this line claims (that was the old base-36 keyspace, 36⁶ ≈ 2.18B).

The example on [line 411](https://github.com/aws-devtools-labs/aws-blocks/blob/f4a2d1f0b3a2048e7e577abfb7a3186839cdf1d5/docs/DECISIONS.md#L411), `my-app-k7x2mf`, also can't occur — `k` and `x` aren't hex digits. (The sandbox example `…9f3a2b` on [line 415](https://github.com/aws-devtools-labs/aws-blocks/blob/f4a2d1f0b3a2048e7e577abfb7a3186839cdf1d5/docs/DECISIONS.md#L415) is valid hex.)

**Fix — pick one:**
- **Keep hex (simplest):** edit this line to `~16.8M combinations (16⁶)`, and change the line-411 example to valid hex, e.g. `my-app-9f3a2b`.
- **Keep ~2.1B:** restore a base-36 suffix with a guaranteed length. The flaky part round-2 fixed was `Math.random().toString(36)` returning a short string — not base-36 itself — so e.g. `randomBytes(4).readUInt32BE().toString(36).padStart(6,'0').slice(0,6)` keeps both 6 chars *and* ~2.1B.

**Not blocking:** ~16.8M is ample for the goal (≈4,800 identically-named apps in one account/region before a 50% birthday-collision chance). This is just keeping the decision record accurate.

packages/create-blocks-app/src/index.ts
```diff
@@ -7,10 +7,22 @@ import { join, dirname, basename, resolve } from 'node:path';
...
 
+function generateStackId(name: string): string {
+  const sanitized = name
+    .replace(/[^A-Za-z0-9-]/g, '-')
+    .replace(/^[^A-Za-z]+/, 'app-')
+    .replace(/-+/g, '-')
+    .replace(/-$/, '')
+    .slice(0, 16)
+    .replace(/-$/, '') || 'blocks-app';
+  return `${sanitized}-${randomBytes(4).toString('hex').slice(0, 6)}`;
```
vishaalmehrishi, line 23: **Nit — the two `generateStackId` implementations can drift silently.**

There are now two copies: this one and [`packages/core/src/scripts/stack-id.ts`](https://github.com/aws-devtools-labs/aws-blocks/blob/f4a2d1f0b3a2048e7e577abfb7a3186839cdf1d5/packages/core/src/scripts/stack-id.ts#L22). Copying instead of importing the core helper is the **right call** — [`create-blocks-app` has no runtime `dependencies`](https://github.com/aws-devtools-labs/aws-blocks/blob/f4a2d1f0b3a2048e7e577abfb7a3186839cdf1d5/packages/create-blocks-app/package.json#L15) (only `devDependencies`), so importing from `@aws-blocks/core` would add a dependency to the dependency-free `npx` scaffolder.

The risk: they're **already cosmetically different** — this line uses `randomBytes(4)` while core uses [`randomBytes(6)`](https://github.com/aws-devtools-labs/aws-blocks/blob/f4a2d1f0b3a2048e7e577abfb7a3186839cdf1d5/packages/core/src/scripts/stack-id.ts#L14-L15) (both `.slice(0,6)` to 6 hex chars, so output is equivalent *today*). A future change to one won't reach the other.

**Fix:** add a `// keep in sync with packages/core/src/scripts/stack-id.ts` comment above both functions, or a one-line test asserting both emit `/^[a-z][a-z0-9-]*-[a-f0-9]{6}$/i`. Cheap insurance against divergence.

docs/DECISIONS.md
```diff
@@ -400,3 +400,33 @@ earlier planning decisions and removed a released behavior:
...
+- **Per-developer sandbox isolation:** Teams sharing a test account need distinct sandbox stacks. The username prefix makes stacks identifiable in the AWS Console; the 6-char random handles multiple sandboxes per developer and username collisions.
+- **Length control:** `name.slice(0, 16)` keeps stack names under ~37 chars total, well within CloudFormation's 128-char limit and readable in the console.
+- **Committed vs gitignored:** `stackId` is committed so the whole team and CI deploy to the same production stack. `sandbox-id.txt` is gitignored so each machine gets its own sandbox.
+
+### References
+- PR #51; `.blocks/config.json` is also used by telemetry (`telemetry.projectId`).
+- Code: `packages/core/src/scripts/stack-id.ts`
+
+### Migration scope
+This scheme applies to **newly scaffolded apps only**. Existing apps that adopt the new `index.cdk.ts` must set their `stackId` in `.blocks/config.json` to their current production stack name (e.g., `my-blocks-stack`) — otherwise CDK will deploy a new stack and orphan the existing one.
```
vishaalmehrishi: **Nit — the instruction and its example contradict each other.**

This line says to set `stackId` to "**their current production stack name** (e.g., `my-blocks-stack`)". But the old static name was `my-blocks-stack-prod`, and the new wiring appends `-prod` itself ([`${getStackId(projectRoot)}-prod`](https://github.com/aws-devtools-labs/aws-blocks/blob/f4a2d1f0b3a2048e7e577abfb7a3186839cdf1d5/packages/create-blocks-app/templates/default/aws-blocks/index.cdk.ts#L16)). So:
- The example `my-blocks-stack` is **correct** → resolves to `my-blocks-stack-prod`. ✅
- The prose is **misleading** — someone who pastes their actual prod stack name (`my-blocks-stack-prod`) gets `my-blocks-stack-prod-prod`.

**Fix:** reword to — "set `stackId` to your current production stack name **with the `-prod` suffix removed** (e.g. if your stack is `my-blocks-stack-prod`, use `my-blocks-stack`)." Optionally note the sandbox stack name changes the same way (`<stackId>-<sandboxId>`).

vishaalmehrishi (dismissed):

docs/DECISIONS.md
svidgen: Fixed. Updated D-012 to ~16.8M (16⁶) and changed examples to valid hex. Keeping hex for simplicity. Fixed in 7c3a5c3.

packages/create-blocks-app/src/index.ts
svidgen: Removed `generateStackId` from core entirely — it had zero consumers. Only the CLI needs it at scaffold time. No drift risk anymore. Fixed in 7c3a5c3.

docs/DECISIONS.md
svidgen: Reworded to: 'set `stackId` to your current production stack name **with the `-prod` suffix removed**' with explicit example. Fixed in 7c3a5c3.

vishaalmehrishi (approved):

pranavosu (approved):

PR #60, opened 2026-06-19
fix(bb-knowledge-base): path guard, cache, filter, error mapping, unicode, chunking

hfurkanbozkurt:
Fixes correctness, security, and robustness issues in `@aws-blocks/bb-knowledge-base`, spanning the AWS-backed retrieval path (`index.aws.ts`), the local-dev implementation (`index.mock.ts`), and the TF-IDF tokenizer (`tfidf.ts`).

## What changed

- **Path containment hardening (`index.mock.ts`)** — source-path validation is centralized in a single helper that runs *before* any filesystem traversal or hashing. It rejects absolute paths (POSIX `/`, UNC `\\`, and `C:\`-style drive paths) outright and performs the project-directory containment check using the platform path separator, so a sibling directory whose name shares a prefix with the project dir (e.g. `../project-secrets` alongside `project`) can no longer slip past a naive prefix check. After the lexical check passes, the resolved source directory is `realpath`-resolved and re-checked against the resolved project root, so a symlink that escapes the project is rejected as well. Sources must be relative and resolve inside the project directory. As a non-blocking usability guard, a `source` that resolves to the project root itself (e.g. `'.'`) is *not* rejected but emits an advisory warning — indexing the entire project tree (including `node_modules` and `.bb-data`) is rarely intended — recommending a dedicated docs subfolder instead.
- **Content-hash cache invalidation (`index.mock.ts`)** — `ensureLoaded()` computes a SHA-256 hash over the source folder (resolved path plus the paths, sizes, and mtimes of only the files that are actually indexed) together with the active chunking configuration (strategy, chunk size, overlap), and stores it next to the chunk cache (`source.hash`). The cache is reused only when the source still exists and the hash matches, so editing documents, adding or removing indexed files, or changing the chunking configuration rebuilds the index rather than serving stale chunks. Adding an unsupported file type (for example an image) does not force a rebuild, and a source folder that has since been deleted surfaces a clear `InvalidSource` error instead of returning stale results.
- **Full-corpus filter scan (`index.mock.ts`)** — when a metadata filter is supplied, `retrieve()` now scores all chunks instead of a `maxResults * 10` window before filtering, so post-filtering no longer silently drops valid matches in larger local corpora.
- **Load-failure recovery (`index.mock.ts`)** — a failed source load now resets the in-flight load promise, so a later `retrieve()` retries the load instead of being permanently wedged on a rejected promise (for example, when a missing source folder is created after the first failed call).
- **Configurable chunking (`index.mock.ts`)** — the local implementation honors the configured chunking strategy: `fixed` uses overlapping word-count windows (`chunkSize`/`chunkOverlap`), with the overlap clamped to a safe range so progress is always positive and the final short window is still emitted — no words are dropped, including the trailing one. `none` keeps each document whole, and `semantic`/`hierarchical` split on paragraph boundaries. Previously chunking was always paragraph-based regardless of the configured strategy.
- **Unicode-aware tokenizer (`tfidf.ts`)** — tokenization now NFD-normalizes and strips combining diacritics (so `resume` matches `résumé`), uses Unicode letter/number classes (`\p{L}\p{N}`) instead of ASCII-only matching, and extracts CJK bigrams with a single-character fallback so an isolated ideograph stays searchable (plain whitespace splitting yields no tokens for CJK text). CJK matching is scoped to explicit blocks (Hiragana/Katakana, CJK Unified Ideographs including Extension A, and compatibility ideographs) rather than one broad range.
- **Bedrock error mapping (`index.aws.ts`)** — `ValidationException` is mapped to `InvalidFilter` only when the message clearly indicates a metadata-filter problem; all other validation messages (query length, malformed request, `numberOfResults`, content-safety responses, etc.) map to the generic `ValidationError`. `ResourceNotFoundException` maps to `NotReady`, and unknown SDK errors continue to map to `RetrievalFailed`. In every case the original SDK message is surfaced on the mapped error, and the untouched SDK error is attached as the standard `Error.cause` (defined non-enumerable, so serializing the mapped error cannot leak SDK response metadata) so its original name, message, and stack remain available for debugging.

## Tests

The fix coverage is consolidated into the package's existing canonical test files (no separate per-fix test files). All unit and integration tests pass and CI is green.

- `src/index.aws.test.ts` — adds Bedrock SDK error-classification cases alongside the existing constructor-validation, env-key, and result-mapping tests: a filter-related `ValidationException` → `InvalidFilter`; a non-filter (query-length) `ValidationException` that must **not** map to `InvalidFilter`; `ResourceNotFoundException` → `NotReady`; `AccessDeniedException` → `RetrievalFailed` (the default mapping for other SDK errors), asserting both the preserved message and the attached `Error.cause`; and a dedicated case proving the original SDK message survives on the mapped error.
- `src/index.mock.test.ts` — adds `describe` groups for:
  - **source-path containment** — rejection of an absolute path (even one inside the project), a `..`-escape, a sibling directory that merely shares the cwd string prefix (exercised through a subprocess with `cwd` set to the project), and a symlink whose target is outside the project, plus a regression test asserting that a `source` resolving to the project root emits a warning but still indexes and returns results (advisory, not a rejection);
  - **cache invalidation** — switching the source folder returns new content rather than stale cache, an edited document is reflected on the next load, deleting the source after a cached load throws `InvalidSource` instead of serving stale data, and changing the chunking configuration for the same id re-chunks rather than reusing the cache;
  - **metadata-filter completeness** over a larger corpus (matching docs are returned even when they rank below higher-scoring non-matching docs);
  - **load-failure recovery** (an instance recovers once the missing source folder is created);
  - **chunking configuration** — `fixed` splits a blank-line-free document into multiple chunks, a negative/out-of-range `chunkOverlap` is clamped so no words are dropped, and a zero-overlap run still emits the short trailing window (tail-loss regression);
  - **Unicode / multilingual retrieval** (CJK content searchable end-to-end).
- `src/tfidf.test.ts` — adds accent-insensitive matching (an unaccented query matches an accented document) and single-character CJK retrieval, alongside the existing tokenizer and CJK tests.

## Notes

- The local-development section of `README.md` is updated to document the configurable chunking strategies, the Unicode-aware tokenizer, content- and config-keyed cache invalidation, the metadata-filter parity asymmetry (an unknown/invalid filter key is rejected server-side in production but matches nothing locally), and the relative-path source requirement.
- `chunkOverlap` is interpreted as a percentage of `chunkSize` in both local development and production — the CDK layer maps it to Bedrock's `overlapPercentage` (1–99), and the local mock overlaps by `chunkSize × chunkOverlap / 100` words — so the value is directly transferable between environments. This is documented in both the type definition and the README.
- A `patch`-level changeset is included. It also records a consumer-facing behavioral note: because non-filter `ValidationException`s now surface as the generic `ValidationError` rather than `InvalidFilter`, consumers that previously caught `InvalidFilter` to handle generic query-validation failures should audit those catch blocks. No exported types, signatures, or error constants changed.


packages/bb-knowledge-base/src/index.aws.ts
```diff
@@ -45,29 +43,53 @@ function blocksError(name: string, message: string): Error {
...
+	// is no underlying Error to attach, so `cause` is left unset.
+	if (!(err instanceof Error)) {
+		return blocksError(KnowledgeBaseErrors.RetrievalFailed, String(err));
 	}
-	// Non-Error throw (e.g., string or object) — stringify for diagnostics.
-	return blocksError(KnowledgeBaseErrors.RetrievalFailed, String(err));
+
+	let mapped: Error;
+	if (err.name === 'ResourceNotFoundException') {
+		mapped = blocksError(KnowledgeBaseErrors.NotReady, 'Knowledge base not found. Run `cdk deploy` first.');
```
soberm: All other errors preserve `err.message`.

packages/bb-knowledge-base/src/index.mock.ts
```diff
@@ -56,8 +57,38 @@ function walkDir(dir: string): string[] {
...
+		.map((p) => p.trim())
+		.filter((p) => p.length >= 20);
+}
+
+/**
+ * Split text into overlapping windows of ~maxTokens words. Windows under 20
+ * chars are dropped as noise, except the final window — emitting it always
+ * keeps trailing words indexed (e.g. with zero overlap, where step === maxTokens).
+ */
+function chunkByFixedSize(text: string, maxTokens: number, overlapPct: number): string[] {
```
soberm, line 69: When `maxTokens = 0`, the clamp `Math.min(..., maxTokens - 1)` yields `overlap = -1`. Should we add a guard `if (maxTokens <= 0) return [];` ?

packages/bb-knowledge-base/src/index.mock.ts
```diff
@@ -205,62 +254,180 @@ export class KnowledgeBase extends Scope {
...
+		// excluding `.metadata.json` sidecars) so adding an unsupported file
+		// (e.g. a .png) doesn't force a needless rebuild.
+		const files = walkDir(sourceDir)
+			.filter((f) => SUPPORTED_EXTENSIONS.has(extname(f).toLowerCase()) && !f.endsWith('.metadata.json'))
+			.sort();
+		for (const f of files) {
+			hash.update(f);
+			try {
+				const stat = statSync(f);
+				// Trade-off: key on mtime+size (rsync-style heuristic), not byte content — fast and fine for local dev, though a same-second, same-size edit could be missed.
```
soberm: nit: too long inline comment

soberm (approved):

sarayev (approved):

PR #61, opened 2026-06-19
fix: run nextjs template dev with webpack

ikenyal:
## Problem

The Next.js template currently starts its frontend with `npx next dev --port 3100`.
With the template's `next: ^16.0.0` dependency, `next dev` uses Turbopack by default.

When validating a freshly scaffolded Next.js app from the local monorepo path, the app starts but the page returns HTTP 500:

```text
./aws-blocks/client.js:8:1
Module not found: Can't resolve '@aws-blocks/blocks/client'
```

I also found that the Next.js template e2e test did not catch this because it assumed a server was already running at `http://localhost:3000`. Once it actually reaches the page, it was also asserting the old `AWS Blocks + Next.js` text while the template renders `Blocks + Next.js`.

## Changes

- Run the Next.js template frontend with `next dev --webpack --port 3100`.
- Make the Next.js template e2e test start `npm run dev` when `TEST_URL` is not provided.
- Keep `TEST_URL` support for externally managed dev servers.
- Update the e2e title assertion to match the current template UI.
- Add a changeset for `@aws-blocks/create-blocks-app`.

## Validation

Reproduced before the fix with a freshly scaffolded Next.js app:

```text
curl -i http://localhost:3000
HTTP/1.1 500 Internal Server Error
Module not found: Can't resolve '@aws-blocks/blocks/client'
```

Validated after the fix:

```text
node packages/create-blocks-app/dist/index.js /private/tmp/.../nextjs-app --template nextjs -y
cd /private/tmp/.../nextjs-app
npm run test:e2e
```

Result:

```text
✔ home page loads
ℹ pass 1
ℹ fail 0
```

Additional checks:

```text
npm run build -w packages/create-blocks-app
npm run test -w packages/create-blocks-app
./node_modules/.bin/changeset status --since=origin/main
./node_modules/.bin/tsx scripts/verify-changeset-coverage.ts
git diff --check
git diff --cached --check
```

Note: this is independent from the other open `create-blocks-app` PRs. I know the branch may need a quick rebase depending on merge order and can handle that if needed.

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (not applicable; release-note changeset added)

---

### DIFF
--- a/packages/create-blocks-app/templates/nextjs/test/e2e.test.ts
+++ b/packages/create-blocks-app/templates/nextjs/test/e2e.test.ts
@@ -1,24 +1,57 @@
+import { test } from 'node:test';
 import assert from 'node:assert';
+import { spawn, type ChildProcess } from 'node:child_process';
+import { setTimeout } from 'node:timers/promises';
 
 // Basic e2e test for the Next.js template
 // Run with: npm run test:e2e
 
-async function test() {
-  const baseUrl = process.env.TEST_URL || 'http://localhost:3000';
-  
-  console.log(`Testing ${baseUrl}...`);
-  
-  // Test home page loads
-  const res = await fetch(baseUrl);
-  assert.strictEqual(res.status, 200, 'Home page should return 200');
-  
-  const html = await res.text();
-  assert.ok(html.includes('AWS Blocks + Next.js'), 'Page should contain title');
-  
-  console.log('✓ All tests passed');
+const baseUrl = process.env.TEST_URL || 'http://localhost:3000';
+let server: ChildProcess | null = null;
+
+async function isServerReady() {
+  try {
+    const res = await fetch(baseUrl);
+    return res.status === 200;
+  } catch {
+    return false;
+  }
 }
 
-test().catch((err) => {
-  console.error('✗ Test failed:', err.message);
-  process.exit(1);
+async function waitForServer() {
+  for (let i = 0; i < 60; i++) {
+    if (await isServerReady()) return;
+    await setTimeout(1000);
+  }
+  throw new Error(`Next.js dev server did not become ready at ${baseUrl}`);
+}
+
+test.before(async () => {
+  console.log(`Testing ${baseUrl}...`);
+
+  if (!process.env.TEST_URL && !await isServerReady()) {
+    server = spawn('npm', ['run', 'dev'], {
+      cwd: process.cwd(),
+      stdio: ['ignore', 'pipe', 'pipe'],
+      detached: true,
+      env: { ...process.env, NODE_OPTIONS: '' },
+    });
+    server.unref();
+  }
+
+  await waitForServer();
+});
+
+test.after(() => {
+  if (server?.pid) {
+    try { process.kill(-server.pid, 'SIGTERM'); } catch {}
+  }
+});
+
+test('home page loads', async () => {
+  const res = await fetch(baseUrl);
+  assert.strictEqual(res.status, 200, 'Home page should return 200');
+
+  const html = await res.text();
+  assert.ok(html.includes('Blocks + Next.js'), 'Page should contain title');
 });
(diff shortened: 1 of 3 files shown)


pranavosu:
I wasn't able to repro this for some reason. Here's the steps I followed.

- `npm create @aws-blocks/blocks-app@latest next -- --template nextjs`
- `cd next && npm run dev`
-  `curl -i http://localhost:3000` -> 200

ikenyal:
Thanks for checking this. I think the difference is the install path.

Your repro uses the published package path:

```sh
npm create @aws-blocks/blocks-app@latest next -- --template nextjs
```

In that case `@aws-blocks/blocks` is installed from the registry under the generated app's `node_modules`, so Next/Turbopack can resolve `@aws-blocks/blocks/client` normally.

The failure I hit was from the local monorepo validation path:

```sh
node packages/create-blocks-app/dist/index.js /tmp/nextjs-app --template nextjs -y
cd /tmp/nextjs-app
npm run dev
curl -i http://localhost:3000
```

When `create-blocks-app` runs from this repo, it detects the local monorepo and rewrites `@aws-blocks/*` dependencies to `file:` paths. That makes the generated app depend on local workspace packages outside the app root. With Next 16, `next dev` uses Turbopack by default, and in that setup the page failed with:

```text
./aws-blocks/client.js:8:1
Module not found: Can't resolve '@aws-blocks/blocks/client'
```

Adding `--webpack` made the same local-monorepo scaffold return 200 and made the updated template e2e test pass.

That said, your repro shows this is not a published-package end-user regression. The stronger framing for this PR is local template validation / monorepo development support, plus fixing the Next.js template e2e test so it actually starts the dev server and asserts the current page title. If forcing webpack for all generated Next.js apps feels too broad given that the published path works, I can narrow or split the PR accordingly.

ikenyal:
Rebased this branch onto the latest main. There were no conflicts.

Re-ran:

- npm run build -w packages/create-blocks-app
- npm run test -w packages/create-blocks-app
- scaffolded a fresh Next.js app from the local dist CLI and ran npm run test:e2e successfully
- ./node_modules/.bin/changeset status --since=origin/main
- ./node_modules/.bin/tsx scripts/verify-changeset-coverage.ts
- git diff --check
- git diff --cached --check

The PR is mergeable again on my side.

packages/create-blocks-app/templates/nextjs/test/e2e.test.ts
```diff
@@ -1,24 +1,57 @@
...
-  
-  console.log('✓ All tests passed');
-}
 
-test().catch((err) => {
-  console.error('✗ Test failed:', err.message);
-  process.exit(1);
+  const html = await res.text();
+  assert.ok(html.includes('Blocks + Next.js'), 'Page should contain title');
 });
```
sharonyajain: Optional follow-on, non-blocking — feel free to defer to a separate PR. This mirrors what PR #597 did for the other six templates: keep the readiness assertion hard, but let the stale-prone starter-title check self-skip once someone edits the page, so a scaffolded app whose home page is replaced doesn't inherit a failing test. The `200` check stays a real assertion (it validates the dev server your `--webpack` fix gets booting); only the title assertion becomes conditional.

```suggestion
test('home page loads', async (t) => {
  const res = await fetch(baseUrl);
  assert.strictEqual(res.status, 200, 'Home page should return 200');

  // The starter page ships with this title. Once you edit the page this check
  // no longer applies, so it self-skips — replace it with an assertion for
  // your own page content.
  const html = await res.text();
  if (!html.includes('Blocks + Next.js')) {
    return t.skip('starter page changed — replace with an assertion for your own content');
  }
  assert.ok(html.includes('Blocks + Next.js'), 'Page should contain the starter title');
});
```

sharonyajain:
Hi @ikenyal Thanks again for this fix. I merged latest main in to keep it current, and it's mergeable with green checks. Are you still willing to work on this PR? If you might be busy, I am happy to take it over.

Thank you again for the contribution!

ikenyal:
@sharonyajain 
Thanks for the suggestion. I’m still happy to own #61. Since this is an optional follow-on and #61 is ready to wrap up, I’ll handle the Next.js test-resilience change as a separate follow-up for #155 once #61 has merged.

PR #65, opened 2026-06-20
fix(create-blocks-app): respect --template when adding to an existing project

moritalous:
## Problem

When adding Blocks to an **existing** project, `integrateWithExistingProject()` always copies the `aws-blocks/` workspace from the `default` (Vite) template, ignoring the `--template` flag.

As a result, running inside a Next.js project:

```bash
npm create @aws-blocks/blocks-app@latest . -- --template nextjs
```

produces `aws-blocks/scripts/server.ts` whose `frontendCommand` is `npx vite --port 3100 --strictPort` instead of `npx next dev --port 3100`. Then `npm run dev:server` tries to launch Vite in a project that has no Vite.

The standalone (fresh-project) flow is unaffected — only the existing-project integration ignored the flag.

## Fix

- Thread `templateName` through to `integrateWithExistingProject()`.
- Copy the `aws-blocks/` workspace (and `cdk.json` / devDeps) from the **requested** template, falling back to `default` when the template has no `aws-blocks/` workspace.
- Update the `--help` text for `--template`.

## Tests

Added two cases to `cli.test.ts`:
- `--template nextjs` on an existing project ⇒ `server.ts` uses `next dev` (and not `vite`).
- no `--template` ⇒ default `vite` behavior preserved.

All 27 tests pass (`npm run build && npm test`).

## Verification

```
# nextjs template in an existing project
$ ... . --template nextjs -y --skip-install
frontendCommand: 'npx next dev --port 3100',

# default (no template) unchanged
frontendCommand: 'npx vite --port 3100 --strictPort',
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/create-blocks-app/src/index.ts
```diff
@@ -474,8 +474,14 @@ async function integrateWithExistingProject(targetDir: string, skipConfirm = fal
 
   console.log('\n📦 Adding Blocks backend...\n');
 
-  // 1. Copy aws-blocks/ from the default template (reuses the same source of truth)
-  const templateDir = join(__dirname, '../templates/default');
+  // 1. Copy aws-blocks/ from the requested template so framework-specific files
+  // (e.g. scripts/server.ts's frontendCommand — `next dev` vs `vite`) match the
+  // user's project. Fall back to the default template when the requested one has
```
sarayev: minor:

> Fall back to the default template when the requested one has no aws-blocks/ workspace to copy.

Can we have a test to verify this execution path as well?

sarayev (dismissed):

packages/create-blocks-app/src/index.ts
moritalous: @sarayev 

Thanks for the review! I tried to add a test for this path, but it turns out the fallback can never be triggered — validateTemplateName guarantees that only templates in AVAILABLE_TEMPLATES are accepted, and every one of them already has an aws-blocks/ directory. The condition is dead code.

Rather than forcing a test through by changing the design, I removed the fallback instead. The changeset has been updated accordingly.

sarayev (approved):

cadivus (approved):

Issue #66, opened 2026-06-21
`dev`: browser can't load `/.blocks-sandbox/config.json` (404) on SSR frameworks — "Blocks API URL not configured"


moritalous:
### Before opening, please confirm

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-devtools-labs/aws-blocks/issues?q=is%3Aissue).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Which package(s) are affected?

@aws-blocks/core

### Describe the bug

In local development (`startDevServer`, e.g. `npm run dev:server`), the browser cannot load the client runtime config at `/.blocks-sandbox/config.json`, so the auth client throws:

```
Blocks API URL not configured. Ensure:
1. You ran `npm run deploy` (deploys config.json)
2. SSR Lambda has BLOCKS_API_URL env var, OR
3. config.json exists at /.blocks-sandbox/config.json
```

**Root cause.** The dev server is the single front door on `:3000`. It handles `/aws-blocks/api` itself and **proxies every other request to the framework dev server** (Next.js, Nuxt, Astro, …). The browser client resolves its API URL via `fetch('/.blocks-sandbox/config.json')` (`packages/core/src/client/index.ts`), but that request is proxied to the framework dev server, which only serves its own static directory (Next.js `public/`, etc.) — not the project-root `.blocks-sandbox/config.json` that the dev server wrote. The result is a 404, and the client has no API URL.

This is **framework-general**: it affects any framework whose static root is not the project root. It also **diverges from production**, where the `Hosting` construct serves `/.blocks-sandbox/*` as static assets via CloudFront (`packages/core/src/hosting.ts`), so the same client code works once deployed but fails in `dev`.

Server-side calls are unaffected (Node reads the file directly), which is why the problem only surfaces for browser-side calls such as the `Authenticator` / `onAuthChange` auth UI.

### Expected behavior

`GET /.blocks-sandbox/config.json` returns the client runtime config (`{ apiUrl, environment }`) in `dev` as well, with no per-app workaround — symmetric with production.

### Reproduction steps

1. Scaffold an app with an SSR framework template: `npx @aws-blocks/create-blocks-app my-app` (Next.js template).
2. Add an `AuthCognito` block, export `authApi = auth.createApi()`, and render `Authenticator(authApi)` on a client page.
3. `npm run dev:server` and open `http://localhost:3000`.
[...]

PR #67, opened 2026-06-21
fix(core): serve `/.blocks-sandbox/config.json` from the dev server

moritalous:
## Problem

In local development the browser cannot resolve its Blocks API URL. The client fetches `/.blocks-sandbox/config.json`, but the dev server proxies that request to the framework dev server (Next.js/Nuxt/Astro), which only serves its own static directory and returns 404. The client then throws "Blocks API URL not configured", and any browser-side call — notably the `Authenticator` / `onAuthChange` auth UI — fails.

The dev server already writes this file and serves the RPC endpoint at `/aws-blocks/api`; it just doesn't serve the config path. Production has no such gap: the `Hosting` construct serves `/.blocks-sandbox/*` as static assets via CloudFront (`packages/core/src/hosting.ts`). This change makes `dev` symmetric with production.

**Issue #, if available:** #66 (aws-devtools-labs/aws-blocks#66)

## Changes

`packages/core/src/scripts/dev-server.ts`
- The dev server now answers the reserved path `GET /.blocks-sandbox/config.json` directly, from the config it already builds in memory, before falling through to the frontend proxy. Framework-agnostic — every framework proxies through the same `:3000` front door — so no per-app workaround is needed.
- Extracted two pure helpers (mirroring the existing `resolveDevCorsOrigin` pattern) for testability:
  - `buildBlocksConfig(port, isSandbox): BlocksRuntimeConfig`
  - `isBlocksConfigRequest(method, pathname): boolean`

`packages/core/src/constants.ts`
- Added `BLOCKS_SANDBOX_PREFIX = '/.blocks-sandbox'`. Internal only (not re-exported from the package entry), so there is no public API surface change.

`.changeset/dev-server-serve-blocks-config.md`
- `@aws-blocks/core: patch`.

No dependency changes; `package-lock.json` is untouched.

## Validation

- **Unit tests** (`packages/core/src/scripts/dev-server-config.test.ts`): cover `buildBlocksConfig` (local / sandbox / custom port) and `isBlocksConfigRequest` (method + path matching, including near-miss paths). All 11 dev-server unit tests pass.
- **Build**: `npm run build -w @aws-blocks/core` → exit 0.
- **Manual** (Next.js 16 app, `npm run dev:server`): `GET /.blocks-sandbox/config.json` returns `200 {"apiUrl":"http://localhost:3000/aws-blocks/api","environment":"local"}` served by the dev server (no app-side route or rewrite); the auth UI resolves and sign-in works end to end.
- Cross-framework E2E (`test-apps/hosting-ssr`, `-nuxt`, `-astro`) require an AWS sandbox and run in CI; not run locally.

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added — n/a; behavior now matches the documented production semantics, and the reserved path is documented inline on `BLOCKS_SANDBOX_PREFIX`.

---

### DIFF
--- a/packages/core/src/scripts/dev-server.ts
+++ b/packages/core/src/scripts/dev-server.ts
@@ -10,7 +10,7 @@ import { createConnection } from 'node:net';
 import httpProxy from 'http-proxy';
 import { writeClientCode } from './generate-client.js';
 import { ApiError } from '../errors.js';
-import { BLOCKS_RPC_PREFIX } from '../constants.js';
+import { BLOCKS_RPC_PREFIX, BLOCKS_SANDBOX_PREFIX } from '../constants.js';
 import { matchRoute, lockRouteRegistry } from '../raw-route.js';
 import { registerBuiltinRoutes } from '../builtin-routes.js';
 import {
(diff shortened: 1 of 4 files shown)


packages/core/src/scripts/dev-server.ts
```diff
@@ -221,6 +248,18 @@ export async function startDevServer(options: DevServerOptions) {
...
     }
 
+    // ── Blocks runtime config ──────────────────────────────────────────
+    // Reserved path: serve it from the front door so it works for every
+    // framework (Next/Nuxt/Astro/SPA all proxy through this :3000 server),
+    // mirroring production where CloudFront serves `${BLOCKS_SANDBOX_PREFIX}/*`
+    // statically. Otherwise the request is proxied to the framework dev
+    // server, which can't serve this project-root file and 404s.
+    if (isBlocksConfigRequest(method, url.pathname)) {
+      res.writeHead(200, { 'Content-Type': 'application/json' });
```
soberm: With no Cache-Control header on the config response browsers apply heuristic caching. A cached response could serve stale config.

packages/core/src/scripts/dev-server.ts
soberm: pre-existing: `readFileSync` is not used. Can we clean it up?

moritalous:
@soberm 
Thanks for the review! Both addressed in e97a15e

- Added Cache-Control: no-store to the config response to prevent stale config from heuristic caching.
- Removed the unused readFileSync import.

soberm (approved):

sarayev (approved):

Issue #78, opened 2026-06-23
Dev server returns permanent HTTP 502 "Frontend server unavailable" after a tsx-watch restart orphans the --strictPort Vite frontend


hfurkanbozkurt:
## Summary

Running the local dev server from the `default` template, a normal `tsx watch` restart (triggered by editing a file) can leave the Vite frontend on port 3100 dead. The dev-server proxy then serves **HTTP 502 "Frontend server unavailable"** for `/` indefinitely - the frontend is never auto-respawned, so the app stays dead until manually killed and restarted.

### `packages/core/src/scripts/dev-server.ts`

502 on proxy error (frontend unreachable) - lines 196-201:
```ts
  frontendProxy?.on('error', (_err, _req, res) => {
    // ...guards omitted...
    (res as ServerResponse).writeHead(502);
    (res as ServerResponse).end('Frontend server unavailable');
  });
```

Frontend spawned via a shell, so Vite is a grandchild - lines 328-332:
```ts
      frontendProcess = spawn(frontendCommand, {
        shell: true,
        stdio: ['ignore', 'pipe', 'pipe'],
        env: { ...process.env, NODE_OPTIONS: '' },
      });
```

`on('exit')` handler only LOGS - it does NOT respawn - lines 338-342:
```ts
      frontendProcess.on('exit', (code) => {
        if (code !== 0 && code !== null) {
          console.error(`Frontend process exited with code ${code}`);
        }
      });
```

`waitForPort` waits only ~30s (maxAttempts 60 x 500ms) then throws - lines 104-120 (called at line 345):
```ts
async function waitForPort(port: number, maxAttempts = 60): Promise<void> {
  for (let i = 0; i < maxAttempts; i++) { /* try TCP connect, return on success */ await sleep(500); }
  throw new Error(`Frontend server on port ${port} did not start within ${maxAttempts * 500}ms`);
}
```

Cleanup only SIGTERMs the shell parent (not the Vite grandchild / process group) - line 364 (handler lines 362-375):
```ts
  const cleanup = async () => {
    if (frontendProcess) frontendProcess.kill('SIGTERM'); // shell parent only
  };
  process.on('SIGINT', cleanup);
  process.on('SIGTERM', cleanup);
```

### `packages/create-blocks-app/templates/default/package.json`

`dev` restarts the whole Node process on every edit - lines 13-14:
```json
    "dev:server": "tsx watch aws-blocks/scripts/server.ts",
[...]

PR #80, opened 2026-06-23
fix(dev-server): auto-respawn frontend and kill process group on restart

soberm:
## Problem

`npm run dev` can leave the Vite frontend (`:3100`) **permanently dead** after a `tsx watch` restart: the front-door proxy then serves a permanent **`502 Frontend server unavailable`** for every `/` request, and the dev environment never recovers without manually freeing the port.

**Issue #, if available:** N/A — surfaced by agent bench (PR #31, run `27988095874`, cell `collab-cursor-board`), where `npm run dev` wedged after a hot-reload and never recovered.

**Root cause.** The dev server spawned the frontend with `shell: true`, so the real Vite process was a **grandchild** (`shell → npx → node …/vite`). On each `tsx watch` restart:

1. cleanup called `frontendProcess.kill('SIGTERM')`, which signalled **only the shell parent** — the Vite **grandchild was orphaned** and stayed bound to `:3100`.
2. the newly launched Vite ran with `--strictPort`, so it refused to fall back, **could not bind `:3100`**, and exited.
3. the `frontendProcess.on('exit')` handler **only logged** — there was no respawn.
4. the proxy target is hardcoded to `:3100`, so every `/` request hit the proxy `error` handler → a **permanent `502`** with no recovery.

Issue: https://github.com/aws-devtools-labs/aws-blocks/issues/78

## Changes

All changes are in `packages/core/src/scripts/dev-server.ts`.

- **Process-group / tree kill.** The frontend is spawned `detached` on POSIX (its own process group), and cleanup/restart now signal the **entire group** via `process.kill(-pid, …)`, reaping the orphaned Vite grandchild and freeing `:3100`. On Windows (no POSIX groups) the tree is reaped with **`taskkill /T /F /PID <pid>`**, degrading to a direct child kill only if `taskkill` cannot be spawned.
- **Bounded auto-respawn.** An unexpected frontend exit now respawns Vite with exponential backoff (**500 ms → 5 s**), capped at **5 restarts / 10 s** to avoid hot loops. The budget counts only **consecutive failing** restarts and **resets once a respawn rebinds the port**, so a frontend that legitimately restarts many times is never permanently left down. Respawn is suppressed during intentional shutdown via an `isShuttingDown` guard.
- **Robust, idempotent shutdown.** `cleanup` is idempotent (re-entrancy guard), wired to **`SIGINT` / `SIGTERM` / `SIGHUP`**, removes its own listeners, and waits (bounded, `SIGTERM` → `SIGKILL`) for the group to die before exiting so a `tsx watch` restart can rebind `:3100` cleanly. A synchronous **`process.on('exit')` safety net** reaps the detached group on any path that bypasses `cleanup`.
- **PID-reuse guard.** The synchronous exit net only reaps a still-live child: once the child has exited (`exitCode`/`signalCode` set) the OS may recycle its PID, so `kill(-pid)` is skipped to avoid signalling an unrelated process group.
- **`--strictPort` retained on purpose.** The proxy target is hardcoded to `:3100`, so the correct fix is to reliably *free* the port rather than let Vite drift to a port the proxy wouldn't follow. No template changes were needed.

## Validation

- **New tests** — `packages/core/src/scripts/dev-server-supervisor.test.ts`:
  - *Unit:* `evaluateFrontendRespawn` (base/exponential backoff, `maxBackoffMs` cap, budget exhaustion, sliding-window pruning, custom policy) and `killFrontendTree` / `windowsTreeKill` (POSIX group signal `-pid`, Windows `taskkill /T /F /PID` tree-kill, `taskkill`-unavailable and group-signal-throws fallbacks, no-pid and `pid ≤ 1` guards).
  - *Integration (real processes, POSIX, opt-in):* spawns a detached `shell → node` tree that binds a port, proves a **direct child kill leaves the port bound** (the bug), then proves **`killFrontendTree` group-kill frees it** (the fix). Gated behind `RUN_SLOW_TESTS` so the default run stays deterministic and fast.
- **`@aws-blocks/core` test suite (Node 22):**
  - `npm test` → **499 / 499 pass** (the real-process integration test is skipped by default).
  - `RUN_SLOW_TESTS=1 npm test` → **500 / 500 pass** (the integration test is included).
- **Build:** `npm run build:packages` compiles clean on Node 22.
- **Changeset:** `@aws-blocks/core` → **patch** (`.changeset/dev-server-restart-race-502.md`).

## Checklist

- [x] PR description included
  - N/A — internal dev-server behavior with no user-facing docs to update; the release note is captured via the `@aws-blocks/core` **patch** changeset (`.changeset/dev-server-restart-race-502.md`).

---

### DIFF
--- a/.changeset/dev-server-restart-race-502.md
+++ b/.changeset/dev-server-restart-race-502.md
@@ -28,7 +28,13 @@ Fixes:
   second dev server), and crediting a foreign one would make every
   `--strictPort`-failing respawn look successful, neutralizing the cap and
   hot-looping forever. A frontend that legitimately restarts many times (e.g.
-  editor-triggered full reloads) is still never permanently left down.
+  editor-triggered full reloads) is still never permanently left down. Before
+  each relaunch the supervisor now also waits (bounded) for `:3100` to be
+  released — the same port-free drain the graceful shutdown path uses — so a slow
+  socket teardown can't hand the relaunched `--strictPort` Vite an `EADDRINUSE`
+  and burn a restart-budget slot; that wait re-checks the `isShuttingDown` guard,
+  so a shutdown arriving mid-wait still cancels the relaunch (and the budget is
+  debited once, at exit time, so the wait never double-counts a restart).
 - **Robust shutdown** — cleanup is idempotent, wired to `SIGINT`/`SIGTERM`/
   `SIGHUP`, removes its own listeners, and waits (bounded) for the group to die
   **and for `:3100` to actually be released** before exiting: SIGTERM→SIGKILL
@@ -48,17 +54,27 @@ Fixes:
   minimal. The single rationale lives next to the supervisor as the
   "POST-EXIT GROUP-KILL POLICY" so all three sites stay in agreement.
 - **Sandbox entrypoint parity** — `sandbox.ts` (the sibling dev entrypoint) now
-  spawns the dev server in its own process group and `await`s a bounded group
-  teardown on `SIGINT`/`SIGTERM`, replacing the synchronous `kill()` +
-  `process.exit(0)` that signalled only the shell and exited immediately. The
-  drain gives the dev server time to run its own shutdown and reap the
-  *detached* Vite great-grandchild, so the next `npm run sandbox` no longer
-  races a survivor on `:3100`.
+  spawns **both** long-running children — the dev server *and* `cdk watch` — in
+  their own process groups and `await`s a bounded group teardown for each
+  (run concurrently) on `SIGINT`/`SIGTERM`, replacing the synchronous
+  `cdkWatch.kill()` + single dev-server `kill()` + `process.exit(0)` that
+  signalled only the npx/shell parents and exited immediately. A bare
+  `cdkWatch.kill()` could orphan the real `cdk watch` node process
+  (npx → cdk → node) — the same shell-only-kill leak this PR fixes for the dev
+  server — so it now routes through the shared `terminateProcessTree` too. Only
+  the dev-server drain (the longer 6s budget) owns the `:3100` port-free wait, via
+  its own SIGTERM handler, so the next `npm run sandbox` no longer races a
+  survivor on `:3100`.
 - **Single tree-kill primitive** — the POSIX group-kill, the Windows `taskkill`
   tree-kill, and the bounded SIGTERM→SIGKILL teardown now live in one shared
   `process-tree.ts` module used by every entrypoint (dev server, respawn
   handler, `exit` net, and sandbox), so the reaping behavior can no longer drift
-  between hand-rolled copies.
+  between hand-rolled copies. Its bounded teardown documents that its boolean
+  reflects only the **direct child's** exit (not whole-group teardown or port
+  release — callers needing a freed port must follow with `waitForPortFree`), and
+  its post-SIGKILL grace is a named `KILL_GRACE_MS` constant kept deliberately
+  shorter than the injectable SIGTERM grace (SIGKILL is uncatchable, so only a
+  brief beat is needed to observe the exit).
 
 `--strictPort` is intentionally retained: the proxy target is hardcoded to
 `:3100`, so the port is reliably freed rather than letting Vite drift to another
--- a/.changeset/dev-server-restart-race-502.md
+++ b/.changeset/dev-server-restart-race-502.md
@@ -1,0 +1,30 @@
+---
+"@aws-blocks/core": patch
+---
+
+fix(dev-server): auto-respawn frontend and kill the whole process group on restart
+
+The dev server spawns the frontend (Vite) with `shell: true`, making the real
+Vite process a **grandchild** (shell → npx → node vite). On a `tsx watch`
+restart, cleanup sent `SIGTERM` to only the shell parent, orphaning the Vite
+grandchild — it survived still bound to `:3100`. The freshly launched Vite then
+hit `--strictPort`, failed to bind, and exited; the `exit` handler only logged,
+so `/` served a permanent `502 Frontend server unavailable` with no recovery.
+
+Fixes:
+
+- **Process-group kill** — the frontend is spawned `detached` on POSIX (its own
+  process group) and cleanup/restart now signal the entire group via
+  `process.kill(-pid, …)`, reaping the Vite grandchild and freeing `:3100`.
+  Windows (no POSIX groups) falls back to a direct child kill.
+- **Bounded auto-respawn** — an unexpected frontend exit now respawns Vite with
+  exponential backoff, capped at 5 restarts / 10s to avoid hot loops, and is
+  suppressed during intentional shutdown via an `isShuttingDown` guard.
+- **Robust shutdown** — cleanup is idempotent, wired to `SIGINT`/`SIGTERM`/
+  `SIGHUP`, removes its own listeners, waits (bounded) for the group to die
+  before exiting (SIGTERM→SIGKILL escalation), and keeps a synchronous
+  `process.on('exit')` safety net so a detached Vite is never left orphaned.
+
+`--strictPort` is intentionally retained: the proxy target is hardcoded to
+`:3100`, so the port is reliably freed rather than letting Vite drift to another
+port the proxy wouldn't follow.
--- a/.changeset/dev-server-restart-race-502.md
+++ b/.changeset/dev-server-restart-race-502.md
@@ -29,6 +29,15 @@ Fixes:
   `SIGHUP`, removes its own listeners, waits (bounded) for the group to die
   before exiting (SIGTERM→SIGKILL escalation), and keeps a synchronous
   `process.on('exit')` safety net so a detached Vite is never left orphaned.
+- **Consistent post-exit reaping** — the failure being fixed is the *shell
+  exiting while the detached grandchild survives*, so every post-exit path
+  (the respawn handler, graceful shutdown, and the `exit` safety net) now
+  issues one best-effort process-group kill even after the shell has gone,
+  rather than skipping it. A surviving grandchild keeps the group's id reserved
+  on POSIX, so `process.kill(-pid)` still targets our own group; the kills are
+  issued synchronously on observing the exit to keep the PID-reuse window
+  minimal. The single rationale lives next to the supervisor as the
+  "POST-EXIT GROUP-KILL POLICY" so all three sites stay in agreement.
 
 `--strictPort` is intentionally retained: the proxy target is hardcoded to
 `:3100`, so the port is reliably freed rather than letting Vite drift to another
--- a/.changeset/dev-server-restart-race-502.md
+++ b/.changeset/dev-server-restart-race-502.md
@@ -16,10 +16,15 @@ Fixes:
 - **Process-group kill** — the frontend is spawned `detached` on POSIX (its own
   process group) and cleanup/restart now signal the entire group via
   `process.kill(-pid, …)`, reaping the Vite grandchild and freeing `:3100`.
-  Windows (no POSIX groups) falls back to a direct child kill.
+  Windows (no POSIX groups) reaps the tree with `taskkill /T /F /PID <pid>`,
+  which walks the child tree by PID so the Vite grandchild is killed too; it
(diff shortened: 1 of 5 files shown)


osama-rizk:
## Code review — 4 findings

Solid fix; the gaps are at the edges, not the core mechanism.

1. **`dev-server.ts:736` (Windows leak + inline duplication)** — the `process.on('exit')` safety net early-returns on `!usePosixProcessGroups`, so on Windows any path that bypasses `cleanup` (uncaught exception, abrupt exit) orphans the Vite tree on `:3100`, reproducing the exact 502 wedge; it also hand-rolls a third copy of the group-kill that diverges from `killFrontendTree`, which already reaps Windows via `taskkill` (and `spawnSync` is legal in a sync `exit` handler) — just call `killFrontendTree(child, 'SIGKILL')`.
2. **`sandbox.ts:181` (sibling entrypoint orphans Vite)** — `cleanup()` does a synchronous `devServer.kill(); process.exit(0)` and never awaits the dev server's new ~2s `terminateFrontend` drain, so the detached Vite (a great-grandchild here) is never reaped and the next run hits `--strictPort`/502; the PR hardened one entrypoint but not its sibling in the same package.
3. **`dev-server.ts:640-647` (post-exit branch skips the port-free wait)** — when the shell already exited, `terminateFrontend` fires SIGKILL and returns synchronously with no exit-wait/port poll, bypassing the "wait for the port to free before exiting" invariant in exactly the orphan case the PR targets; if the kernel hasn't released the socket before `tsx watch` relaunches, the fresh `--strictPort` Vite gets `EADDRINUSE` (low probability, but the one branch that drops the guarantee).
4. **`dev-server.ts:565` (budget reset on a liveness-only probe)** — `announceFrontendReady` resets `frontendRestarts` whenever `waitForPort` sees *anything* on `:3100`, so a foreign listener (a leftover Vite, or another dev server) makes every `--strictPort`-failing respawn count as success, neutralizing the `maxRestarts` cap and hot-looping forever instead of giving up.

Findings 1 and 2 are the actionable ones; 3 and 4 are real edge-case gaps worth a guard.

soberm:
Thanks for the careful review @osama-rizk — all four findings were valid. Fixed in `00f02c4` (pushed to `fix/dev-server-restart-race-502`).

Common theme you flagged ("don't hand-roll divergent copies"): I extracted the POSIX group-kill, the Windows `taskkill` tree-kill, and a bounded `SIGTERM→SIGKILL` teardown into one shared module, **`packages/core/src/scripts/process-tree.ts`**, now used by every entrypoint (dev server, respawn handler, `exit` net, and sandbox). So the four fixes below all route through the same primitives instead of re-implementing them.

**Finding 1 — `dev-server.ts:736` exit-net Windows leak + third copy** ✅
The `process.on('exit')` net no longer early-returns on `!usePosixProcessGroups` and no longer hand-rolls its own group kill. It now just calls `killFrontendTree(child, 'SIGKILL')` (`dev-server.ts:633`). `killFrontendTree` reaps via `process.kill(-pid)` on POSIX and `taskkill /T /F /PID` on Windows, so the Vite tree is reaped on Windows too, and there is no longer a divergent third copy.

**Finding 2 — `sandbox.ts:181` sibling entrypoint orphans Vite** ✅
`cleanup()` is now `async` and `await`s a bounded group teardown before exiting, replacing the synchronous `devServer.kill(); process.exit(0)`. Concretely: the dev server is now spawned `detached` (its own process group), and cleanup calls `await terminateProcessTree(devServer, 6000)` — which group-`SIGTERM`s the dev server so its **own** shutdown handler runs the `terminateFrontend` drain and reaps the detached Vite great-grandchild, then escalates to a tree `SIGKILL` and exits regardless if it hangs (6s budget > the dev server's ~4s drain). The handlers are registered idempotently (`SIGINT`/`SIGTERM`), so a second signal can't re-enter. Next `npm run sandbox` no longer races a survivor on `:3100`.

**Finding 3 — `dev-server.ts:640-647` post-exit branch skips the port-free wait** ✅
`terminateFrontend` now delegates the kill to `terminateProcessTree(child, 1500)` (which handles both the live and already-exited cases) and then **always** `await waitForPortFree(frontendPort)` (`dev-server.ts:446-450`). The port-free poll runs on **both** paths now, so the post-exit/orphan case — exactly the one this PR targets — no longer drops the "wait for the port to free before exiting" invariant, and a `tsx watch` relaunch can't race the kernel's socket teardown into `--strictPort` `EADDRINUSE`. (Budget: 1500ms SIGTERM grace + 500ms SIGKILL grace + ≤2000ms port poll ≈ 4s, under tsx-watch's ~5s force-kill window.)

**Finding 4 — `dev-server.ts:565` budget reset on a liveness-only probe** ✅
`announceFrontendReady` now takes the spawned child as its first argument and resets `frontendRestarts` only via `shouldCreditFrontendReady(child, frontendProcess)` (`dev-server.ts:368`), which credits the probe **only** when our own spawned child is still the live frontend (same identity, `exitCode`/`signalCode` still `null`). `waitForPort` seeing *something* on `:3100` no longer counts — so a foreign listener (leftover Vite, a second dev server) can't make every `--strictPort`-failing respawn look successful and neutralize the `maxRestarts` cap. A child that lost the `--strictPort` bind race is no longer `current`, so its failed attempt still counts toward the budget.

**Tests**: added unit coverage for `shouldCreditFrontendReady` (own-live-child vs. foreign/exited/killed/null), `waitForPortFree` (free / waits-then-frees / times-out), and `terminateProcessTree` (SIGTERM→SIGKILL escalation + post-exit reap, with injected kill/sleep so it's deterministic). The real-process integration test stays gated behind `RUN_SLOW_TESTS=1`.

Verification on Node 22:
- `npm run build:packages` — clean (exit 0).
- `npm test --workspace=@aws-blocks/core` (default) — **509 pass / 0 fail**.
- `RUN_SLOW_TESTS=1 npm test --workspace=@aws-blocks/core` — **510 pass / 0 fail** (real-process integration test runs).

Changeset updated with the round-2 behavior notes.

osama-rizk (dismissed):

soberm:
Round-3 review findings addressed in `71a4cc5`.

**1. [LOW] `sandbox.ts` — `cdkWatch` spawned without a process group + bare `cdkWatch.kill()`** ✅
`cdk watch` is now spawned in its own process group on POSIX (`detached: process.platform !== 'win32'`) and reaped via the shared `terminateProcessTree(cdkWatch, 2000)` instead of `cdkWatch.kill()`. A bare kill hit only the `npx` shell and could orphan the real `cdk watch` node (`npx → cdk → node`) — the exact shell-only-kill leak this PR removes for the dev server. Both child trees are now reaped **concurrently** (`Promise.all`) so the cdk-watch teardown doesn't serialize on top of the dev server's 6s drain; only the dev-server child owns the `:3100` port-free wait (its own SIGTERM handler runs `terminateFrontend`).

**2. [LOW] `process-tree.ts` — `terminateProcessTree` boolean overstates what it guarantees** ✅
Tightened the JSDoc to state the boolean reflects only the **direct child's** exit (`exitCode`/`signalCode`), **not** whole-group teardown or port release — on POSIX the SIGKILL goes to the whole group (`-pid`), but a surviving detached grandchild can outlive the shell child and still hold the port after this resolves `true`. Callers needing a freed port must follow with a bounded port-free wait (as `dev-server.ts` does via `waitForPortFree`); also noted that the dev-server child itself owns that drain. Doc-only — the boolean's semantics are unchanged, so existing callers/tests are untouched.

**3. [NIT] `process-tree.ts` — hardcoded 500ms SIGKILL grace vs injectable `graceMs`** ✅
Extracted into `export const KILL_GRACE_MS = 500`, with a comment explaining why the post-SIGKILL grace is shorter than and decoupled from the injectable SIGTERM `graceMs` (SIGKILL can't be caught/blocked, so we only need a brief beat to observe `exit`). Still injectable-for-tests via the existing `sleep` seam. Added a unit test asserting the SIGTERM grace uses `graceMs` and the post-SIGKILL grace uses `KILL_GRACE_MS`.

**4. [NIT] `dev-server.ts` — auto-respawn path skipped the port-free wait** ✅
The respawn timer now `await waitForPortFree(frontendPort)` before `spawnFrontend`, folded into the timer callback — parity with the graceful `terminateFrontend` path — so a slow socket teardown can't hand the relaunched `--strictPort` Vite an `EADDRINUSE` and burn a restart-budget slot. It re-checks `isShuttingDown` after the bounded wait (a shutdown arriving mid-wait cancels the relaunch), the budget is debited once at exit time (no double-count), and `waitForPortFree` is bounded (no deadlock).

**Build / test (Node 22):**
- `npm run build:packages` — clean (exit 0).
- `npm test --workspace=@aws-blocks/core` — **510/510 pass** (slow integration tests skipped by default).
- `RUN_SLOW_TESTS=1 npm test --workspace=@aws-blocks/core` — **512/512 pass**, including the new RUN_SLOW_TESTS integration test that reaps a detached non-shell `npx → node` tree (the cdk-watch shape) via `terminateProcessTree` post-exit.

Pushed to `fix/dev-server-restart-race-502` @ `71a4cc5`.

soberm:
Round-4 review feedback addressed in `ca45cbb` (build clean; `@aws-blocks/core` tests **516/516** default and **519/519** with `RUN_SLOW_TESTS=1`).

**#1 (MEDIUM) — post-exit group kill could hit a recycled group.** Took the *scope* option (preferred). `terminateProcessTree`'s post-exit branch now reaps the group only when it still has a live member, gated by a new `isProcessGroupAlive(pid)` probe: POSIX `kill(-pid, 0)` sends no signal, just tests group existence — `ESRCH` ⇒ drained → skip, `EPERM` ⇒ alive, Windows ⇒ always allow (taskkill is PID-tree-scoped, no `-pid` reuse hazard). When the group has fully drained there's also nothing of ours left to reap, so skipping is correct *and* eliminates the recycled-`-pid` hazard. Cross-referenced the dev-server `POST-EXIT GROUP-KILL POLICY` from the shared primitive, and added a note back in `dev-server.ts` explaining the two in-file kill sites (respawn reap + `exit` net) stay unconditional because they fire synchronously at the instant of observed exit, whereas the `terminateProcessTree` path can run outside that window and so additionally probes. No behavior change for real callers — the live-grandchild integration test still reaps and frees the port.

**#2 (MEDIUM) — nested-SIGTERM-handler claim unverified.** Added the test (preferred). New opt-in (`RUN_SLOW_TESTS`), POSIX-only integration test spawns a detached `shell → node` child whose node installs a SIGTERM handler *then* binds a port to signal readiness; the test sends a **group** SIGTERM (`process.kill(-pid, 'SIGTERM')`, exactly what `killFrontendTree` does on POSIX) and asserts the node handler **observed** it (writes a marker file). `sandbox.ts` now cross-references this test at the 6s-budget comment.

**#3 (LOW) — respawn timer `unref`.** Documented as intentional: the listening HTTP server owns process lifetime, so the backoff timer must not by itself keep the loop alive on shutdown; `cleanup` clears the timer and both the timer body and the awaited relaunch re-check `isShuttingDown`. Added an explicit "do not remove the unref" note.

**#4 (LOW) — `windowsTreeKill` ignored `status`.** Now treats only `status === 0` (reaped) or `128` (already gone) as handled and returns `false` for any other taskkill exit (e.g. `1` = access denied) so the caller falls back to `child.kill` instead of silently leaking the tree. JSDoc updated; added a unit test for the non-zero/non-128 case.

**#5 (NIT) — changeset bump.** Intentionally keeping this as `patch` for this revision.

New tests: `isProcessGroupAlive` unit suite, a post-exit "group drained → skip" case, a `windowsTreeKill` non-zero/non-128 status case, and the gated nested-SIGTERM integration test.

osama-rizk (approved):

chai3:
@osama-rizk Nice to meet you. Thank you for AWS-Blocks.
I believe the linked issue #92 (Development #92) is a mistake.

osama-rizk:
Apologies — issue #92 was linked to this PR by mistake. This PR actually relates to issue #78 (referenced in the description). The incorrect #92 link is being removed. Sorry for the confusion.

svidgen (approved):

pranavosu (approved):

PR #82, opened 2026-06-23
fix(bb-auth-oidc): surface signIn errors, idempotent redirect callback, broadcastAuthChange bridge

soberm:
## Summary

Three related developer-experience bugs in the `@aws-blocks/bb-auth-oidc` **browser** client made client-initiated OIDC sign-in fail silently in a React SPA. All three were confirmed against the source on `main` and fixed with minimal, well-scoped changes plus tests.

Refs bench cell `oidc-dsql-notes · react` (PR #31, run 27988095874, scored 1/5).

### (a) `signIn()` was fire-and-forget — errors were swallowed
`signIn()` returned `void` and ran `void this._signInPKCE(...)` with no `.catch`. When the cross-origin `authorize-params` fetch failed (`!paramsResp.ok` → `throw`), the rejection became an unobservable unhandled promise rejection. A caller writing `onClick={() => auth.signIn('google')}` had no way to know sign-in never started, and `await auth.signIn(...)` was impossible because the method returned `void`.

### (b) `handleRedirectCallback()` was not idempotent
The method read the single-use PKCE state from `sessionStorage` but only removed it **after** the `await fetch(exchange)` round-trip. Under React 18 StrictMode (effects run twice in dev), a double-click, or any re-render, a second invocation read the *same* pending state and replayed the single-use authorization `code` — which the IdP/exchange rejects — throwing on the second call and stranding the app. There was also no dedup of the in-flight exchange.

### (c) Auth state was never bridged to the shared `auth-common` bus
`index.browser.ts` had no import from `@aws-blocks/auth-common` and never called `broadcastAuthChange()`. A successful exchange called only the package-local `notify()` (its own `onAuthStateChange` subscribers). So `onAuthChange` listeners and the shared `<Authenticator>` / `<AuthenticatedContent>` components — which react to `broadcastAuthChange` — never updated on OIDC sign-in. The README had no OIDC + SPA example showing how to wire this.

### (a) `signIn()` surfaces errors
- Signature changed to return `Promise<void>`, so `await auth.signIn(...)` and `.catch(...)` both observe failures.
- A `.catch` is attached internally so fire-and-forget callers (`onClick={() => auth.signIn('google')}`) log to `console.error` instead of producing a silent unhandled rejection — backward compatible (statement-position calls still work).
- Public `OIDCClient` interface in `types.ts` updated to `Promise<void>`; the server-side stub in `auth-oidc.ts` is now `async`.

### (b) `handleRedirectCallback()` is idempotent
- The single-use PKCE state is read **and removed from `sessionStorage` synchronously, before the first `await`**, so a racing second client instance can't read it again (it returns `null` harmlessly instead of replaying the `code`).
- The in-flight exchange is cached in a per-instance `Map<code, Promise<User|null>>`. A StrictMode double-mount / double-click / re-render reuses the same promise and **both callers resolve to the same user**. The first successful flow always completes and renders.

### (c) Bridge to the shared auth bus + README
- After a successful exchange, the client now calls `broadcastAuthChange(user)` from `@aws-blocks/auth-common` in addition to the existing local `notify()`. `signOut()` broadcasts `null`. `onAuthChange` listeners and the shared auth UI components now update automatically (same window **and** across tabs).
- `README.md` gains a **React SPA** section: a dedicated callback-route component that calls `handleRedirectCallback()` (idempotent under StrictMode), `signIn().catch(...)` error handling, and an `onAuthChange` example that updates automatically — no manual wiring.

## Tests (`packages/bb-auth-oidc/src/index.browser.test.ts`)

- **signIn surfaces errors**: rejects when the `authorize-params` fetch fails; fire-and-forget callers get `console.error`; returns a `Promise`.
- **handleRedirectCallback idempotent**: concurrent double-invoke exchanges **once** and both callers get the same user; sequential re-invoke returns the same user without a second exchange; the second invocation never throws (StrictMode safety).
- **broadcastAuthChange bridge**: a successful exchange notifies real `onAuthChange` consumers; a same-window `blocks-auth-change` `CustomEvent` is dispatched carrying the exchanged user.

A `BroadcastChannelShim` (mirroring `auth-common/ui.test.ts`) keeps Node's real `BroadcastChannel` singleton from holding the event loop open; same-window delivery (which the tests assert) uses `window.dispatchEvent`.

## Verification (Node 22 + WebCrypto)

```
# bb-auth-oidc full suite
# tests 107
# pass 107
# fail 0          (clean exit, RC=0)

# auth-common (dependency sanity)
# tests 50
# pass 50
# fail 0

# full monorepo build (downstream consumers, incl. @aws-blocks/blocks umbrella)
BUILD_RC=0
```

## Changeset

Adds `.changeset/bb-auth-oidc-spa-fixes.md` — **patch** bump for `@aws-blocks/bb-auth-oidc`. `auth-common` was not modified (only imported), so no changeset is needed for it.

## Scope

Changes are confined to `@aws-blocks/bb-auth-oidc` (`index.browser.ts`, `types.ts`, `auth-oidc.ts`, `index.browser.test.ts`, `README.md`) plus the changeset. No unrelated refactors.

### DIFF
--- a/.changeset/bb-auth-oidc-spa-fixes.md
+++ b/.changeset/bb-auth-oidc-spa-fixes.md
@@ -1,0 +1,11 @@
+---
+"@aws-blocks/bb-auth-oidc": patch
+---
+
+fix(bb-auth-oidc): surface `signIn` errors, make the redirect callback idempotent, and bridge to the shared auth bus
+
+Three browser/SPA developer-experience fixes:
+
+- **`signIn()` no longer swallows errors.** It now returns `Promise<void>`, so `await auth.signIn(...)` and `.catch(...)` observe failures (e.g. an unreachable `authorize-params` endpoint). Fire-and-forget callers (`onClick={() => auth.signIn('google')}`) still work and now log the failure to `console.error` instead of producing a silent unhandled rejection.
+- **`handleRedirectCallback()` is idempotent.** It consumes the single-use PKCE state from `sessionStorage` synchronously *before* the exchange await, and caches the in-flight exchange per authorization `code`. A React StrictMode double-mount, double-click, or re-render reuses the same exchange and resolves to the same user instead of replaying the single-use `code` (which the IdP rejects). The first successful flow always completes and renders; a genuinely-already-consumed second call returns `null` harmlessly.
+- **Auth state is bridged to the shared bus.** A successful exchange and `signOut()` now call `broadcastAuthChange(...)` from `@aws-blocks/auth-common`, so `onAuthChange` listeners and the `<Authenticator>` / `<AuthenticatedContent>` components update automatically (same window and across tabs) — previously only the package-local `onAuthStateChange` subscribers were notified. The README gains a React SPA example wiring the callback and `onAuthChange`.
--- a/packages/bb-auth-oidc/README.md
+++ b/packages/bb-auth-oidc/README.md
@@ -63,6 +63,67 @@ auth.signIn('google', { redirectPath: '/auth-return' });
 
 `redirectPath` becomes the OAuth `redirect_uri`, so it must be a page your frontend serves **and** a redirect URI registered with the provider (the stub IdP accepts any HTTPS or localhost URL, so local/sandbox needs no registration).
 
+### React SPA: handle the callback and react to auth state
+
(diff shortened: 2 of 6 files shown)


soberm:
Superseded by a **3-PR stack** that splits this change into one PR per fix for easier review (per reviewer request):

- **#83** — (a) `signIn()` error surfacing — base `main`
- **#84** — (b) `handleRedirectCallback()` idempotency — base #83
- **#85** — (c) `broadcastAuthChange` bridge + SPA README — base #84

Review/merge **bottom-up: #83 → #84 → #85**. The cumulative tip (#85) is **byte-identical** (package source) to this PR. The single combined changeset here is replaced by one changeset entry per stacked PR — all patch bumps for `@aws-blocks/bb-auth-oidc`, which collapse to a single patch bump on merge.

Closing in favor of the stack.

Issue #79, opened 2026-06-23
AuthOIDC browser client: signIn() swallows failures, handleRedirectCallback() not idempotent, onAuthChange bridge undocumented


hfurkanbozkurt:
## Summary

Three related developer-experience gaps in the `@aws-blocks/bb-auth-oidc` browser client make OIDC sign-in fail silently in a React SPA:

1. `signIn()` is fire-and-forget and swallows all sign-in failures.
2. `handleRedirectCallback()` is not idempotent (breaks under React StrictMode / double-mount).
3. The bridge to `@aws-blocks/auth-common` `onAuthChange` is undocumented: `handleRedirectCallback()` does not drive `onAuthChange`, and no example shows the wiring.

Surfaced by the agent bench (PR #31, run 27988095874), cell `oidc-dsql-notes · react`, scored 1/5 -- the post-sign-in view never renders.

### `packages/bb-auth-oidc/src/index.browser.ts`

`signIn()` returns `void` and fires `_signInPKCE` with no `.catch` (L237-239):

```ts
signIn(provider: Provider, opts?: { state?: string; redirectPath?: string }): void {
  void this._signInPKCE(provider, opts?.state, opts?.redirectPath);
}
```

`_signInPKCE` does a cross-origin credentialed fetch, throwing on non-OK (L252-256):

```ts
const paramsResp = await fetch(`${baseUrl}${this.authorizeParamsBasePath}/${encodeURIComponent(String(provider))}`, {
  credentials: 'include',
});
if (!paramsResp.ok) {
  throw new Error(`AuthOIDC: failed to fetch authorize params for '${String(provider)}': ${paramsResp.status}`);
}
```

Navigates to the IdP only on success (L289): `window.location.href = authorizeUrl.toString();`

`handleRedirectCallback()` reads single-use PKCE state from sessionStorage (L308-309):

```ts
const raw = sessionStorage.getItem(_PENDING_STORAGE_KEY);
if (!raw) return null;
```

Removes it only AFTER the exchange await (L341) and throws on failure (L343-345); no idempotency guard:

```ts
sessionStorage.removeItem(_PENDING_STORAGE_KEY);

if (!resp.ok) {
  const err = await resp.json().catch(() => ({ error: 'Exchange failed' }));
  throw new Error(`AuthOIDC exchange failed: ${(err as any).error ?? resp.status}`);
```

On success it notifies only the OIDC client's own subscribers (L354): `notify(user, { state: pending.appState });`

[...]

PR #88, opened 2026-06-23
fix(bb-auth-oidc): make handleRedirectCallback() idempotent under double invocation

soberm:
## Problem

In the `@aws-blocks/bb-auth-oidc` browser client, `AuthOIDCClient.handleRedirectCallback()` was not idempotent. It read the single-use PKCE pending entry from `sessionStorage` and removed it **only after** awaiting the `/aws-blocks/auth/exchange` round-trip. A second concurrent invocation — most commonly **React StrictMode**, which mounts → unmounts → mounts and fires the callback effect twice synchronously — would either:

- replay the same single-use `code` (the second exchange fails server-side), or
- find the pending entry already consumed and resolve `null`.

In the double-mount race the second call's result is the one the component keeps, so a successful sign-in could leave the app stranded on a signed-out screen.

**Issue #, if available:** #79

## Changes

- Added a module-level in-flight guard (`_callbackInflight`) keyed by the PKCE `code`. While an exchange is in flight, a concurrent/duplicate `handleRedirectCallback()` for the same code returns the **first call's promise** instead of starting a second exchange — both callers resolve to the same user and subscribers fire exactly once.
- Extracted the exchange into a private `_exchangeCallback()` that:
  - consumes the pending `sessionStorage` entry **up front** (before the network round-trip), so a late duplicate that somehow misses the guard still can't replay it, and
  - releases the guard in a `finally` once the exchange settles, so a genuinely new sign-in flow on the same page is never blocked.
- All existing behavior is preserved: RFC 9207 `iss` forwarding, the state-mismatch throw, exchange-error handling, and `{ user }` / bearer-mode unwrapping.

Added a changeset (`patch` for `@aws-blocks/bb-auth-oidc`).

## Validation

Added a new `handleRedirectCallback — idempotency under double invocation` suite in `src/index.browser.test.ts`:

- **concurrent double invocation** — two `handleRedirectCallback()` calls fired via `Promise.all` (exchange stubbed to settle on a later tick): both resolve to the same user, the exchange POST is counted **exactly once**, subscribers are notified **exactly once**, and the pending entry is consumed. (Without the fix the exchange runs twice — the assertion fails.)
- **sequential double invocation** — two calls captured before awaiting either: still one exchange, both resolve to the user.
- **guard release** — after a settled flow, a brand-new flow (new `code`/`state` + freshly stored pending blob) runs its own exchange, proving the guard is released and doesn't block subsequent sign-ins.

Build + full package test suite pass (Node 22):

```
npm run build -w packages/bb-auth-oidc   # tsc --build, clean
npm run test  -w packages/bb-auth-oidc   # tests 103 | pass 103 | fail 0
```

(100 pre-existing tests + 3 new.)

## Checklist

- [x] PR description included

---

### DIFF
--- a/.changeset/issue-79-idempotent-callback.md
+++ b/.changeset/issue-79-idempotent-callback.md
@@ -1,0 +1,21 @@
+---
+"@aws-blocks/bb-auth-oidc": patch
+---
+
+fix(bb-auth-oidc): make `handleRedirectCallback()` idempotent under double invocation
+
+`handleRedirectCallback()` consumed the single-use PKCE pending entry from
+`sessionStorage` and only removed it **after** the `/aws-blocks/auth/exchange`
+round-trip. A second concurrent invocation — most commonly React StrictMode's
+mount → unmount → mount, which fires the callback effect twice synchronously —
+either replayed the already-consumed code (failing the second exchange) or
+found the pending entry gone and resolved `null`, stranding the app on a
+signed-out screen despite a successful sign-in.
+
+The callback now guards on an in-flight promise keyed by the PKCE `code`:
+concurrent/duplicate invocations for the same code share the first call's
+promise instead of starting a second exchange, so both callers resolve to the
+same user and subscribers are notified exactly once. The pending entry is also
+consumed up front (before the network round-trip) so a late duplicate can't
+replay it, and the guard is released once the exchange settles so a genuinely
+new sign-in flow on the same page is never blocked.
--- a/packages/bb-auth-oidc/DESIGN.md
+++ b/packages/bb-auth-oidc/DESIGN.md
@@ -264,3 +264,34 @@ encode to ~43 characters, comfortably exceeding the floor. The contract is
 stated once here: `csrf` MUST be ≥32 characters; SDKs SHOULD generate 32
 random bytes (≈43 base64url chars). Source files defer to this decision rather
 than restating the threshold in bytes.
+
+### D12 — `handleRedirectCallback` shares one in-flight exchange per `(exchangePath, code)`
+
+**Status:** Accepted.
+
+**Rationale.** A PKCE `code` is single-use and must be exchanged exactly once. A
+double invocation — React StrictMode's mount → unmount → mount fires the callback
+effect twice synchronously — would otherwise race to exchange the same code twice;
+the loser finds the pending entry already consumed and resolves `null`, stranding
+the app on a signed-out screen despite a successful sign-in.
+
+`_callbackInflight` (in `index.browser.ts`) is a **module-scoped** variable, so the
+guard is shared across every `AuthOIDCClient` instance in the tab — it is not a
+per-instance field. It is keyed on `(exchangePath, code)`:
+
+- **`code`** — distinct sign-in flows carry distinct single-use codes, so they get
+  distinct exchanges. Two instances in one app (the D4 admin-auth + customer-auth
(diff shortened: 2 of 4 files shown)


pranavosu (dismissed):

osama-rizk:
## Code review — 3 findings

Core fix is sound: I verified the sync prefix sets `_callbackInflight` before any `await` (so the StrictMode double-mount race is genuinely closed), `notify`/`lastUser` fire exactly once, and the code-keyed `finally` clear doesn't clobber a newer flow. Several candidates were refuted and dropped — up-front consumption is **not** a retry regression (the old code also removed the entry before throwing on both the state-mismatch and `!resp.ok` paths), storing the shared promise adds no unhandled-rejection hazard (same object the caller awaits), and the non-async sync-throw concern is moot (`new URL(window.location.href)` can't throw in a browser, and both real callers `await` inside try/catch). The two defenses (in-flight guard + up-front consumption) are genuinely complementary, not redundant — the fix is right-altitude.

1. **`index.browser.test.ts` (test gap) — no test covers the error path under double invocation.** The whole point of the guard is that both concurrent callers share one promise and the `finally` releases the guard, but only the success path is tested. A future refactor that mishandled the shared rejection or failed to release the guard on failure (blocking same-page retry) would keep the suite green. Add a test where `fetch` returns `{ ok: false }` (or a state mismatch) under `Promise.all`, asserting both callers reject identically and a subsequent fresh-code flow still runs its own exchange.

2. **`index.browser.ts:158` (module-level guard) — `_callbackInflight` is a module singleton keyed only by `code`, shared across all client instances.** *(low severity)* Two `AuthOIDCClient` instances with different `exchangePath`/config processing the same `code` would have instance #2 return instance #1's exchange (POSTed to #1's endpoint). Not realistically triggerable given the one-`getClient()`-per-page hydration design, and it follows the existing `lastUser`/`subscribers` pattern — but consider keying on `(this.exchangePath, code)`. Separately, the doc comment overstates the guarantee: the module variable doesn't survive a real page reload, where coordination falls back to up-front `sessionStorage` removal (→ `null` for a late caller); scope the comment to same-tab in-process concurrency.

3. **`index.browser.ts:251` (cleanup) — `_exchangeCallback(url, …)` takes the whole `URL` but uses it only for `url.searchParams.get('iss')`.** The caller already has `searchParams` in scope; passing `iss: string | undefined` directly makes the helper's real URL dependency explicit.

soberm:
Thanks for the review. Addressed all three findings in 70375f3:
1. (Guard keying) `_callbackInflight` is now keyed on `(exchangePath, code)` instead of just `code`, with the check applied at all three sites (in-flight match, assignment, and the finally release), so two clients with distinct endpoints no longer share an exchange. Reworded the doc comment to scope the guarantee to same-tab in-process concurrency and note the module var doesn't survive a page reload (a late caller after reload falls back to the up-front sessionStorage removal returning null).
2. (Signature) `_exchangeCallback` now takes `iss: string | undefined` directly; `iss` is read from searchParams in `handleRedirectCallback` (RFC 9207 comment moved there). The full URL is no longer passed.
3. (Test gap) Added an error-path-under-concurrent-double-invocation test: both callers reject with the identical Error instance, exactly one exchange POST occurs, and a subsequent fresh-code flow still runs its own exchange — proving the guard releases on failure.

Build clean, 104/104 tests pass.

osama-rizk (approved):

pranavosu (approved):

PR #89, opened 2026-06-23
fix(bb-auth-oidc): wire broadcastAuthChange on successful OIDC callback

soberm:
## Problem

The browser OIDC client's `handleRedirectCallback()` completes the client-PKCE exchange and, on success, notifies **only** this client instance's own `onAuthStateChange` listeners. Components that subscribe to auth state through `@aws-blocks/auth-common` — `onAuthChange(...)` and `<AuthenticatedContent>` — are never told about the sign-in. As a result, a React SPA that signs in via the client-initiated redirect flow does not re-render after the callback completes (only the server-initiated sign-in path, which already broadcasts, updated them). This is the third of the three independent bugs in #79.

**Issue #, if available:** #79

## Changes

- `packages/bb-auth-oidc/src/index.browser.ts`: on a successful exchange, `handleRedirectCallback()` now also calls `broadcastAuthChange(user)` from `@aws-blocks/auth-common/ui` (already a dependency — no new dep). This fires the same-window `blocks-auth-change` `CustomEvent` and posts to the cross-tab `BroadcastChannel`, so every `onAuthChange` consumer and `<AuthenticatedContent>` re-render on client-PKCE sign-in, in this tab and others. The existing local `notify(...)` call is unchanged.
- `packages/bb-auth-oidc/README.md`: new "Re-rendering your UI on sign-in (React SPA)" section documenting the `onAuthChange` / `broadcastAuthChange` wiring (and that the client bridges into it automatically on a successful callback), with an OIDC + React SPA example.
- `.changeset/issue-79-onauthchange-bridge.md`: patch changeset for `@aws-blocks/bb-auth-oidc`.

`packages/auth-common/src/ui.ts` is intentionally **not** modified.

## Validation

- `npm run build -w packages/auth-common` and `npm run build -w packages/bb-auth-oidc` — both clean.
- `npm run test -w packages/bb-auth-oidc` — **103/103 pass** (100 baseline + 3 new), process exits cleanly.
- New tests in `packages/bb-auth-oidc/src/index.browser.test.ts` (suite "@aws-blocks/auth-common bridge"):
  - dispatches a same-window `blocks-auth-change` event carrying the signed-in user, so on-page `onAuthChange` consumers re-render;
  - posts the user across tabs via `BroadcastChannel` (exactly one post);
  - does **not** broadcast when the callback fails (state mismatch) — no event, no cross-tab post.

  The test harness backs `window` with a real `EventTarget` and stubs `BroadcastChannel` (a real channel is a ref'd handle that would otherwise keep `node --test` alive).
- `npx biome lint` on the changed files — no new diagnostics (only pre-existing warnings in untouched code).

## Checklist

- [x] PR description included

---


pranavosu (dismissed):

osama-rizk:
**Finding 1 (most severe) — `packages/bb-auth-oidc/src/index.browser.ts:363`: sign-out is not bridged; the fix is half-applied.**

`signOut()` calls `notify(null, null)` + `window.location.reload()` but never `broadcastAuthChange(null)`. The reload repaints only the current tab; `BroadcastChannel.postMessage` doesn't fire in the originating tab, so other open tabs' `onAuthChange` / `<AuthenticatedContent>` consumers keep rendering the signed-in user (username, protected content) after sign-out — the exact cross-tab staleness this PR fixes for sign-in, left in place for sign-out.

The session is terminated server-side so it's not an auth bypass, but it's a real stale-UI issue and squarely within the changeset's stated goal ("every auth-common consumer and other open tabs re-render" on auth change). One-line fix: `broadcastAuthChange(null)` before the reload.

osama-rizk:
**Finding 2 — `packages/bb-auth-oidc/src/index.browser.ts:354`: the bridge broadcasts without priming the shared cache (transient flash).**

All four state-mutating sites in `auth-common/ui.ts` pair `updateState(api, …)` with `broadcastAuthChange(…)` (lines 497-498, 609-610, 795-797, 1020-1021); the bridge calls `broadcastAuthChange` alone, so `cache.state` stays `null`. A component that mounts and subscribes via `onAuthChange` *after* the callback fired (e.g. router navigates post-callback) reads the cold cache, emits `null` synchronously (one-frame signed-out flash), then self-corrects via a `getAuthState()` round-trip.

Impact is a transient flicker + an extra network call, never a stuck state. Note: `updateState` is private and the OIDC client holds no `AuthStateApi` handle, so this is a structural limitation of the package boundary, not a dropped line — a real fix would need auth-common to expose a cache-priming primitive (e.g. `primeAuthState(api, user)`). Worth a code comment acknowledging the divergence even if deferred.

osama-rizk:
**Finding 3 (PLAUSIBLE) — `packages/auth-common/src/ui.ts:538`: a same-tab `<Authenticator>` won't react to the OIDC bridge.**

`Authenticator` subscribes only via `getChannel().addEventListener('message', …)`, which does not fire in the originating tab — unlike `onAuthChange`, which also listens on the same-window `AUTH_LOCAL_EVENT`. A SPA rendering both `<Authenticator>` and the OIDC `AuthCallback` in the same tab: `onAuthChange`-based components re-render, but the same-tab `Authenticator` does not.

Pre-existing in `auth-common` (untouched by this PR) but newly reachable via the bridge — worth a follow-up note rather than a blocker.

---

**Lower-priority (non-blocking):**
- `index.browser.ts:354` — the `user as unknown as AuthUser` double-cast launders the generic `User` to `AuthUser` with no compile-time guard (benign today since `User` is a superset of `{userId, username}`).
- The `BroadcastChannel` test stub relies on `auth-common`'s un-resettable module-level `channel` singleton, making the cross-tab assertions dependent on describe-block ordering.

Overall the core change is correct and minimal; the standout actionable item is Finding 1 (one-line symmetric fix).

soberm:
Thanks for the thorough review. Addressed everything in 3940384:

1. (signOut not bridged — main finding) `signOut()` now calls `broadcastAuthChange(null)` before `window.location.reload()`, so other tabs/`onAuthChange` consumers see the sign-out — matching what the README already promised. Added 2 tests (cross-tab `null` broadcast; same-window `auth-change(null)` before reload).
2. (one-frame `null` flash for components that subscribe after the callback) The clean fix needs a new auth-common primitive to prime the shared cache, which is out of scope for this package-local PR. Added a code comment at the bridge site documenting the divergence and the self-heal via `getAuthState()`.
3. (same-tab `<Authenticator>` only listens cross-tab) Pre-existing auth-common behavior — fixing it would be cross-package scope creep. Added a code comment documenting the limitation and steering to `onAuthChange`/`<AuthenticatedContent>`.
4. (lower-priority) Documented why the `user as unknown as AuthUser` cast is safe (`User` ⊇ `{userId, username}`), and documented the BroadcastChannel test-stub singleton coupling + the shared-array invariant.

Build clean, 105/105 tests pass (auth-common 50/50 unchanged).

osama-rizk (dismissed):

osama-rizk (dismissed):

pranavosu (dismissed):

osama-rizk (approved):

pranavosu (approved):

PR #91, opened 2026-06-23
fix: remove scan() fallback in demo listTodos to prevent multi-tenant…

pjkroker:
## Problem

**Issue #, if available:**

## Checklist

- [ ] PR description included

---


packages/create-blocks-app/templates/demo/aws-blocks/index.ts
```diff
@@ -117,17 +117,18 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
...
+      createdAt: 'byCreatedAt'
     } as const;
-    
-    const iterator = sortBy 
-      ? todos.query(indexMap[sortBy], { userId: { equals: user.username } })
-      : todos.scan();
-    
+
+    const iterator = todos.query({
+      index: sortBy ? indexMap[sortBy] : 'byCreatedAt',
```
soberm, line 123: When sortBy is omitted this routes through the byCreatedAt GSI rather than the primary key. That's a reasonable default for ordering, but GSI reads are eventually consistent, so a todo just written in createTodo() can occasionally be missing from the very next listTodos() call. Querying the primary key ({ partitionKey: 'userId', sortKey: 'todoId' }) instead would be strongly consistent — and since todoId is a ULID it'd still come back in roughly chronological order. The tradeoff is you'd lose createdAt as a sort key (no range queries later) and diverge from the explicit sortBy: 'createdAt' path. Not a defect, just worth a one-line comment noting the default is eventually consistent so a freshly created todo may lag a list call.

packages/create-blocks-app/templates/demo/aws-blocks/index.ts
```diff
@@ -117,17 +117,18 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
...
+      where: { userId: { equals: user.username } }
+    });
+
     return await Array.fromAsync(iterator);
   },
```
soberm, line 129: Array.fromAsync pulls the entire result set into memory here — if the iterator handles DynamoDB's 1MB pagination under the hood, a user with thousands of todos gets all of them back in a single response. Fine for a demo, but this template tends to get copied into real apps, so it might be worth an optional limit param or just a `// no pagination: demo only` comment to make it a deliberate choice.

soberm (commented):
packages/create-blocks-app/templates/demo/aws-blocks/index.ts (createTodo, ~L104/L112):

The ULID prefix on line 104 and createdAt here come from two separate Date.now() calls. They'll almost always land in the same millisecond, but if the clock ticks between them the timestamp baked into todoId won't match createdAt. Easy to keep them in sync by capturing it once:

```ts
const now = Date.now();
const ulid = now.toString(36) + crypto.randomBytes(8).toString('hex');
const todo = { userId: user.username, todoId: ulid, title, completed: false, priority, createdAt: now };
```

packages/create-blocks-app/templates/demo/aws-blocks/index.ts
pranavosu: Good point. For the demo we deliberately keep the `byCreatedAt` GSI as the default so listings come back in stable chronological order, and we don't want to give up `createdAt` as a range key by switching to the primary key. The eventual-consistency lag is acceptable for a demo, so we'll add a one-line comment on the default path noting that reads are eventually consistent and a freshly created todo may not appear in the immediately following `listTodos()` call.

packages/create-blocks-app/templates/demo/aws-blocks/index.ts
pranavosu: Agreed this shouldn't read as production-ready. The intent is to keep the example minimal, and `query()` already accepts a `limit` param so anyone copying this into a real app can opt into pagination. We'll add a `// demo only: loads all todos into memory, no pagination` comment to make that a deliberate choice rather than an oversight.

pranavosu (approved):

soberm (approved):

PR #104, opened 2026-06-24
feat(bb-knowledge-base): ingestion sync API + S3 Vectors teardown

hfurkanbozkurt:
Add an ingestion-**sync** API to the `KnowledgeBase` block and fix S3 Vectors teardown so a destroy/sandbox teardown no longer leaves resources behind.

> **This revision** renames the API to Bedrock's own terminology (`isSynced()` / `waitUntilSynced()`) and addresses review feedback — see **Review feedback** at the bottom.

## What changed

**Sync API** — Adds `isSynced()` and `waitUntilSynced({ timeoutMs?, pollIntervalMs?, maxConsecutiveTransientErrors?, signal? })` to the `KnowledgeBase` block, mirroring Bedrock's console "Sync" / "sync with your latest data" wording. Bedrock ingestion runs asynchronously after deploy, so during the initial pre-sync window `retrieve()` returns an empty array even for queries that would later match — making "empty" ambiguous between "not yet synced with your latest data" and "synced, no match". `isSynced() === false` means the KB is not yet synced with your latest data; it is **not** unavailable — the KB stays queryable and serves the prior snapshot throughout, and `retrieve()` is always callable (it returns empty only during the initial pre-sync window). In production the new methods poll the Bedrock ingestion job status (`ListIngestionJobs` / `GetIngestionJob`) so callers can tell when async ingestion has actually finished; the local mock is always-synced (there is no async ingestion window in local dev). Two new typed errors are added — `KnowledgeBaseErrors.IngestionFailed` (carries the reported `failureReasons`) and `KnowledgeBaseErrors.Timeout` (raised by `waitUntilSynced()` when it exceeds its timeout). Purely additive — `retrieve()` and all existing signatures are unchanged. `waitUntilSynced()` defaults to `timeoutMs` 300000 (5 min) and `pollIntervalMs` 5000 (5 s). It also accepts an optional `signal?: AbortSignal` for cooperative cancellation — when the signal fires (e.g. `AbortSignal.timeout(...)` or a caller-owned `AbortController`) the wait rejects promptly with the signal's abort reason, and the signal is checked both before each poll and during the inter-poll delay. The poll interval carries light random jitter (±20%) so that many knowledge bases polling at once don't fall into lockstep; jitter only varies the delay between polls (never the number of polls), and each delay is still clamped to the remaining deadline.

**Transient-error tolerance** — Because `waitUntilSynced()` exists for the noisy post-deploy window, its poll loop absorbs a bounded run of *transient* control-plane errors instead of aborting on the first blip. Two situations count as transient: a throttling / transient network failure surfaced as `RetrievalFailedException`, and a *not-yet-visible* knowledge base — immediately after deploy the control plane can briefly return `ResourceNotFoundException` because the KB or its data source hasn't propagated yet. That `ResourceNotFoundException` maps to `KnowledgeBaseErrors.NotReady`, and `waitUntilSynced()` distinguishes it from a genuine configuration error by inspecting the mapped error's underlying cause: a `NotReady` whose cause is a control-plane `ResourceNotFoundException` is treated as transient and ridden out, whereas a missing-`KB_ID` config error (also surfaced as `NotReady`, but with no underlying cause) stays terminal and short-circuits immediately. `maxConsecutiveTransientErrors` (optional, default 3, minimum 0) bounds how many *consecutive* transient failures are tolerated; any clean poll resets the counter. When the tolerance is exhausted the wait logs a `warn` ("gave up after N consecutive transient control-plane errors — tolerance exhausted") before rethrowing, so that case is distinguishable in logs from a Timeout on a healthy-but-still-ingesting KB. Other terminal conditions still short-circuit immediately — a `FAILED` ingestion job (`KnowledgeBaseErrors.IngestionFailed`) and validation errors are never retried. Set the knob to `0` to fail fast.

**S3 Vectors teardown fix** — The S3 Vectors store (the vector bucket + index) now honors the same removal policy as the data bucket: `RemovalPolicy.DESTROY` → `DeletionPolicy: Delete`, otherwise → `Retain`. Previously the data bucket was force-deleted and auto-emptied on a `destroy`/sandbox teardown, but the vector bucket and index relied solely on their default CloudFormation deletion policy and were left behind. The data bucket keeps `autoDeleteObjects` so CloudFormation can empty it before deletion. A new CDK template test covers the destroy / retain / sandbox cases.

**Synth-time guards** — The CDK `KnowledgeBase` construct now stubs its runtime methods (`retrieve` / `isSynced` / `waitUntilSynced`) with the shared synth-guard helper, mirroring the canonical pattern the other blocks use (e.g. `bb-kv-store`). Under `--conditions=cdk` a `KnowledgeBase` resolves to the infrastructure construct, which has no data plane; an accidental synth-time call (e.g. invoking a data/sync method at the top level of a backend module, which executes during CDK synth) now throws an actionable "cannot be called during CDK synth" message instead of a cryptic `TypeError: not a function`. Additive — it only changes behavior on misuse at synth time and does not touch the runtime build.

**Docs** — README documents the sync API: an `isSynced()` / `waitUntilSynced(options?)` row in the method table, `IngestionFailed` / `Timeout` rows in the error-handling table, and a short Sync usage example (including the `signal` option and the transient-vs-terminal distinction). DESIGN records the rationale in its consolidated "Design Decisions" section — why the block uses a raw `s3.Bucket` for the data source (the Bedrock service principal must read it, which needs `bucketArn` / `grantRead` / `enforceSSL` and a generated bucket name), why the S3 Vectors resources mirror the data bucket's removal policy on teardown, and the sync contract (fail-fast on terminal errors, tolerate a bounded run of transient ones). DESIGN also notes the teardown caveat — a stack-level force-delete aspect applied without also passing `removalPolicy: 'destroy'` to the block cannot auto-empty the bucket — and clarifies that sync covers both source types: both a local-folder source and an imported `s3://` source register a BB-managed data source (for an `s3://` source the construct simply skips the document-upload step), so `isSynced()` / `waitUntilSynced()` track ingestion in either case. The shortcut where `isSynced()` returns `true` immediately applies only to deployments that predate this API, which have no data source id injected and therefore nothing to track. DESIGN also records that after an ingestion job reaches `COMPLETE`, non-Aurora vector stores (this BB uses S3 Vectors) can take a few minutes for the newly embedded vectors to become queryable — so `isSynced() === true` means the ingestion job completed, with a possible short embedding-propagation lag.

**Comprehensive test app** — The knowledge-base end-to-end test now gates on `waitUntilSynced()` before asserting, and prints a friendly "still syncing" line instead of error noise while ingestion is still in progress.

## Tests

- The `bb-knowledge-base` unit/integration suite passes (129 tests on Node 22), including the sync state-machine tests (driven by a mocked Bedrock client) and the CDK template tests covering destroy / retain / sandbox teardown, the ingestion-status IAM grant (`bedrock:GetIngestionJob` / `bedrock:ListIngestionJobs`, scoped to the KB ARN exactly like the existing `bedrock:Retrieve` grant), and the `DATA_SOURCE_ID` config registration (wired to the Bedrock data source and surfaced in the synthesized config blob) — now covered for both the local-folder source and the imported `s3://` source — and the synth-time guards on the runtime methods. The transient-error-tolerance tests cover both transient kinds: transient-then-synced resolves, repeated transient errors throw past the limit, the counter resets after a clean poll, a not-yet-visible KB (`ResourceNotFoundException` → `NotReady` with cause) is ridden out and then resolves once ingestion is `COMPLETE`, repeated not-yet-visible errors throw past the limit, and a missing-`KB_ID` `NotReady` short-circuits with no polling. Cancellation tests confirm an already-aborted signal rejects before any poll and an abort fired during the inter-poll delay rejects promptly with the supplied reason.
- The conditional-exports parity suite passes.
- Validated end-to-end on real AWS: deploy → `isSynced()` flips from `false` to `true` once ingestion completes → `retrieve()` returns real documents → `destroy` leaves zero orphaned buckets (both the data bucket and the S3 Vectors bucket are gone).

## Notes

- Additive / non-breaking — no existing error constants changed; `maxConsecutiveTransientErrors` and `signal` are optional, defaulted fields on `WaitUntilSyncedOptions` (`signal?: AbortSignal`). The rename from the earlier `isReady()` / `waitUntilReady()` / `WaitUntilReadyOptions` naming is pre-release and has no released consumers. The CDK runtime-method synth guards throw only on misuse at synth time and leave the runtime build untouched.
- Changesets: `minor` for the sync API, `patch` for the teardown fix.
- Adds a dependency on `@aws-sdk/client-bedrock-agent` (used for the ingestion-job status calls).

## Review feedback (this revision)

- **Naming** — Adopted `isSynced()` / `waitUntilSynced()` / `WaitUntilSyncedOptions`, matching Bedrock's console "Sync" / "sync with your latest data" wording (freshness, not availability). Deferred a separate availability latch (`isPopulated`) as YAGNI — `retrieve()` is always callable, so one freshness signal covers the real need; a second method can be added additively later.
- **Browser stub docs** — DESIGN (D-KB-8) now notes the browser stub throws `BrowserNotSupportedException` for `isSynced()` / `waitUntilSynced()` too, same contract as `retrieve()`.
- **Give-up log** — `waitUntilSynced()` now logs a `warn` when it exhausts its transient tolerance, before rethrowing, so it is distinguishable from a Timeout on a healthy KB.
- **`timeoutMs: 0` doc** — Documented that no inter-poll sleep occurs when the budget is already spent — one-shot poll, then Timeout.
- **Log level** — Downgraded the transient control-plane error-mapping log on the poll path (`fetchLatestIngestionJob`) from `error` to `debug`, so the warm-up window no longer emits spurious ERROR entries; the user-facing `error` in `retrieve()` is untouched.
- **`s3://` CDK test** — Added CDK synth tests for the imported `s3://` source path (IAM grants + `DATA_SOURCE_ID` wiring), mirroring the folder-source tests.


packages/bb-knowledge-base/src/index.browser.ts
```diff
@@ -40,4 +40,12 @@ export class KnowledgeBase {
 	async retrieve(_query: string, _options?: RetrieveOptions): Promise<RetrieveResult[]> {
 		throw browserError();
 	}
+
+	async isReady(): Promise<boolean> {
+		throw browserError();
+	}
```
soberm, line 46: The `Mock vs AWS Behavior Differences` table in DESIGN.md covers the mock's readiness behavior (always ready), and D-KB-8 explains why the browser stub throws at construction — but neither mentions that `isReady()` and `waitUntilReady()` also throw `BrowserNotSupportedException` here, the same contract as `retrieve()`. A one-liner in DESIGN.md would close the gap and document all three layers' readiness behavior in one place.

packages/bb-knowledge-base/src/index.aws.ts
```diff
@@ -227,6 +333,260 @@ export class KnowledgeBase extends Scope {
...
+			} catch (err) {
+				// Terminal errors (FAILED job, missing-KB config, validation) short-circuit.
+				// isTransientControlPlaneError() only returns true after its own
+				// `instanceof Error` guard, so pairing it with one here narrows `err` to
+				// `Error` for the rest of the catch — no per-use `as Error` casts needed.
+				if (!(err instanceof Error) || !isTransientControlPlaneError(err)) throw err;
+				// Transient control-plane blip: absorb a bounded run, then give up.
+				consecutiveTransientErrors += 1;
+				lastTransient = err;
+				if (consecutiveTransientErrors > maxConsecutiveTransientErrors) throw err;
```
soberm: nit: the `warn` below only fires on the tolerated-and-retried path — it's not reached when we give up here. A quick `this.log.warn` (or `error`) immediately before this `throw`, noting that transient tolerance was exhausted, would make "gave up after N consecutive errors" distinguishable from "timed out with a healthy-looking KB" in logs.

packages/bb-knowledge-base/src/types.ts
```diff
@@ -153,3 +153,64 @@ export interface RetrieveResult {
...
+ * ```
+ */
+export interface WaitUntilReadyOptions {
+	/**
+	 * Maximum time to wait for ingestion to complete, in milliseconds. Default: 300000 (5 minutes).
+	 *
+	 * `timeoutMs: 0` is a one-shot check, not a no-poll fast-fail: the deadline is
+	 * evaluated *after* the first readiness poll, so exactly one `isReady()` poll
+	 * always runs (and can resolve the wait) before a `KnowledgeBaseTimeoutException`
+	 * is thrown. Clamped to a minimum of 0.
```
soberm: nit: the one-shot semantics are well explained, but there's one more subtlety worth noting: no inter-poll sleep occurs in this case either, since the sleep is `Math.min(jitter, deadline - now)` and the remaining-budget clamp drives that to 0 when `timeoutMs: 0`. Optional half-sentence, e.g. "No inter-poll sleep occurs because the remaining-budget clamp is zero."

packages/bb-knowledge-base/src/index.aws.ts
```diff
@@ -227,6 +333,260 @@ export class KnowledgeBase extends Scope {
...
+
+		if (job.status === 'COMPLETE') return true;
+		if (job.status === 'FAILED') {
+			const reasons = await this.fetchFailureReasons(knowledgeBaseId, dataSourceId, job.ingestionJobId);
+			throw blocksError(
+				KnowledgeBaseErrors.IngestionFailed,
+				`Knowledge base ingestion failed.${reasons.length ? ` Reasons: ${reasons.join('; ')}` : ''}`,
+			);
+		}
+		// STARTING | IN_PROGRESS | STOPPING | STOPPED → not ready.
```
soberm: Naming thought worth considering before this API surface locks: once the first ingestion completes, `retrieve()` continues to serve the prior snapshot throughout any subsequent re-ingestion — Bedrock doesn't go dark while re-indexing. So `false` here during a re-sync means "latest content not indexed yet", not "KB unavailable". Those are different things, but the name `isReady()` reads like the latter.

In practice, a caller gating every `retrieve()` call on `isReady()` will back off unnecessarily on every document update cycle, even though the previous snapshot is still fully queryable.

Two directions worth considering:
- Rename to signal the actual contract — `isIngestionComplete()`, `isUpToDate()`, `isSynced()` — and reserve "ready" for availability.
- Or keep `isReady()` as a latch-once availability check (stays `true` after first `COMPLETE`) and add a separate freshness signal (`isSynced()`, `ingestionStatus()`) for callers who actually need to know about the latest content.

packages/bb-knowledge-base/src/index.aws.ts
```diff
@@ -227,6 +333,260 @@ export class KnowledgeBase extends Scope {
...
+					knowledgeBaseId,
+					dataSourceId,
+					sortBy: { attribute: 'STARTED_AT', order: 'DESCENDING' },
+					maxResults: 1,
+				}),
+			);
+			return response.ingestionJobSummaries?.[0];
+		} catch (err) {
+			const mapped = mapSdkError(err);
+			this.log.error(mapped.message);
```
soberm: nit: this `error` log fires for transient control-plane errors — `ResourceNotFoundException` → `NotReady`, SDK errors → `RetrievalFailed` — that `waitUntilReady` is specifically designed to absorb and retry. During the normal post-deploy warm-up window, every tolerated blip therefore logs at `error` level here and *also* at `warn` via the retry path, producing spurious ERROR entries in CloudWatch during expected behavior. If this is mainly useful for direct `isReady()` callers, `warn` or `debug` would fit better — or omit it entirely and let the caller own the signal (`waitUntilReady` already does with its own warn at the retry site).

packages/bb-knowledge-base/src/index.cdk.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  const { kb } = buildStack();
+  const construct = kb as unknown as Record<string, (...args: unknown[]) => unknown>;
+  for (const method of ['retrieve', 'isReady', 'waitUntilReady']) {
+    assert.throws(
+      () => construct[method]('x'),
+      /cannot be called during CDK synth/,
+      `${method}() should throw the actionable synth-time error`,
+    );
+  }
+});
```
soberm, line 281: suggestion: no CDK test exercises the `s3://` source path, which has meaningfully different behavior — it imports an existing bucket instead of creating one, skips `BucketDeployment`, optionally sets `inclusionPrefixes`, but still registers `DATA_SOURCE_ID` and emits the same IAM grants. Worth a synth test pinning that the handler still receives `bedrock:Retrieve` + `bedrock:GetIngestionJob`/`bedrock:ListIngestionJobs` and that `*_DATA_SOURCE_ID` is wired to the data source ID when the source is an S3 URI — mirrors what the folder-source tests above already cover.

packages/bb-knowledge-base/src/index.browser.ts
hfurkanbozkurt: Done — added a line to D-KB-8 in DESIGN.md noting that `retrieve()`, `isSynced()`, and `waitUntilSynced()` all throw `BrowserNotSupportedException` in the browser stub (none silently no-op or return a fake "synced"), and tied it to the other layers so all three sync behaviors are documented in one place: mock is always synced, AWS polls the real ingestion job, browser throws for the data and sync methods alike. (Note: the API is now `isSynced()` / `waitUntilSynced()` — see the naming thread.)

packages/bb-knowledge-base/src/index.aws.ts
hfurkanbozkurt: Done — added a `this.log.warn` immediately before the transient-exhausted `throw` in `waitUntilSynced()`. It logs "giving up after N consecutive transient control-plane error(s) — tolerance (M) exhausted", so that case is greppable in CloudWatch and clearly distinct from a Timeout on a healthy-but-still-ingesting KB. The existing tolerated-and-retried `warn` is unchanged.

packages/bb-knowledge-base/src/types.ts
hfurkanbozkurt: Done — extended the `timeoutMs` doc: with `timeoutMs: 0`, no inter-poll sleep occurs either, since the sleep is `min(jitter, deadline − now)` and the remaining-budget clamp drives it to 0 when the budget is already spent. So it's a single one-shot poll, then Timeout.

packages/bb-knowledge-base/src/index.aws.ts
hfurkanbozkurt: Adopted your direction #1: `isReady()` → `isSynced()`, `waitUntilReady()` → `waitUntilSynced()` (and `WaitUntilReadyOptions` → `WaitUntilSyncedOptions`). This matches Bedrock's own console "Sync" / "sync with your latest data" wording and reads as freshness rather than availability. Docs/JSDoc now frame `isSynced() === false` as "not yet synced with your latest data" — the KB is still queryable and serves the prior snapshot, never "unavailable"; `retrieve()` is always callable and returns empty only during the initial pre-sync window.

I deferred a separate availability latch (`isPopulated`) as YAGNI for now: since `retrieve()` is always callable, one freshness signal covers the real need, and a second method can be added additively later if a concrete case emerges. Pure rename, non-breaking (pre-release).

packages/bb-knowledge-base/src/index.aws.ts
hfurkanbozkurt: Done — downgraded the control-plane error-mapping log on the poll path (`fetchLatestIngestionJob`) from `error` to `debug`. Those are exactly the transient blips `waitUntilSynced()` is designed to absorb during the warm-up window, so they no longer emit spurious ERROR entries in CloudWatch. The operator signal now comes solely from `waitUntilSynced()`'s own warns (tolerated / gave-up); the genuinely user-facing `error` in `retrieve()` is untouched.

packages/bb-knowledge-base/src/index.cdk.test.ts
hfurkanbozkurt: Done — added two CDK synth tests for the imported `s3://` source path. They assert the handler still gets `bedrock:Retrieve` + `bedrock:GetIngestionJob` / `bedrock:ListIngestionJobs` scoped to the same KB ARN, and that `*_DATA_SOURCE_ID` is wired to the data source id (`Fn::GetAtt … DataSourceId`) — mirroring the folder-source tests. One also asserts the imported-bucket branch is genuinely taken (no documents `BucketDeployment`). Suite is now 129 tests, all green.

soberm (approved):

ahmedhamouda78 (approved):

PR #107, opened 2026-06-25
feat(bb-data): configurable TLS for external database connections

vishaalmehrishi:
## What

Adds an `ssl` option to the connection-string form of `fromExisting(...)` and makes server-certificate verification the default for external PostgreSQL connections (Supabase, Neon, RDS). `db pull` captures the provider's CA so generated apps verify by default — including in the deployed Lambda — with no runtime configuration.

Previously the external-database paths hardcoded `ssl: { rejectUnauthorized: false }`, so there was no supported way to configure TLS verification for these connections.

## Approach

The connection's TLS policy is now configurable and verifies by default, with the CA delivered so verification actually works at every layer:

1. **Type** — `ExternalDatabaseRef`'s connection-string variant gains `ssl?: { rejectUnauthorized?: boolean; ca?: string }`.
2. **Runtime** (`index.aws.ts`) — passes the caller's `ssl` through; `PgClientEngine` already defaults to `{ rejectUnauthorized: true }`, so omitting `ssl` verifies.
3. **Local dev** (`index.mock.ts`) — honors `connection.ssl`; keeps an unverified default only for self-signed local databases.
4. **`db pull`** — prompts for the provider CA and writes it to a generated, committed **`database.ca.ts`** (a public, non-secret certificate — public key + issuer metadata, no private key). The generated `supabase.ts` pins it via a visible `resolveDbSsl()`. Because it's a bundled JS import (not a runtime file/env read), verification works in the deployed Lambda with **no env/SSM/CDK plumbing**. `DATABASE_CA_CERT` (inline PEM or path) overrides it; with no CA the wiring falls back to a visible, editable `rejectUnauthorized: false`.
5. **Operational paths** (`introspect`, migrate CLI, external migrations) — resolve `ssl` via a shared `externalDbSsl()` helper and strip `sslmode` from the URL so a pinned CA takes effect (node `pg` ignores a programmatic `ssl.ca` when `sslmode` is present).

Why a committed cert (not an env var): `.env.*` is only loaded on the deploy host, never injected into the Lambda — so an env-only CA would leave the deployed runtime unverified. The CA isn't secret, so committing + bundling it is the simplest mechanism that verifies in production. It's regenerated on each `db pull` (rotation) and preserved when a re-pull doesn't re-supply it (no silent downgrade).

## What a customer sees

`npx bb-data pull` now asks for the CA and explains what it does:

```
TLS verification (recommended): download your CA certificate from the Supabase
dashboard → Database Settings → SSL Configuration (prod-ca-2021.crt).
  Path to CA certificate [Enter to skip]: ./prod-ca-2021.crt
  ✓ CA captured. It will be written to aws-blocks/database.ca.ts (a public cert, committed with
    your app). The generated connection (resolveDbSsl in supabase.ts) pins it, so
    your database's TLS certificate is verified — both with `npm run dev` and in the
    deployed function (the file is bundled). Re-run `bb-data pull` to refresh a
    rotated CA. Details: MIGRATION_GUIDE.md → "Securing the connection (TLS)".
```

Generated `aws-blocks/supabase.ts` (the wiring — TLS policy is visible and editable):

```ts
import { DATABASE_CA_CERT } from './database.ca.js';

function resolveDbSsl(): { ca?: string; rejectUnauthorized: boolean } {
  // DATABASE_CA_CERT (inline PEM or a file path) overrides the committed cert.
  const envCa = process.env.DATABASE_CA_CERT;
  const ca = envCa
    ? (envCa.includes('-----BEGIN') ? envCa : readFileSync(envCa, 'utf8'))
    : (DATABASE_CA_CERT || undefined);
  if (ca) {
    console.log('[bb-data] DB TLS: verifying the server certificate against the pinned CA (verify-full equivalent).');
    return { ca, rejectUnauthorized: true };
  }
  console.warn('[bb-data] DB TLS: server certificate NOT verified — encrypted only (no CA). ...');
  return { rejectUnauthorized: false };
}

export const db = new Database(scope, 'db', {
  connection: fromExisting({ connectionString: { get: resolveConnString }, ssl: resolveDbSsl() }),
  rlsPolicy: 'enforce',
  schema: tableMeta,
});
```

Generated `aws-blocks/database.ca.ts` (regenerated each pull; the customer's actual project CA):

```ts
// Generated by `db pull` — safe to regenerate.
export const DATABASE_CA_CERT = `-----BEGIN CERTIFICATE-----
MII... (your project's CA)
-----END CERTIFICATE-----`;
```

How the customer uses it is unchanged — spread `supabaseCrud` into an `ApiNamespace`:

```ts
import { supabaseCrud } from './supabase.js';
export const api = new ApiNamespace(scope, 'api', (context) => ({
  ...supabaseCrud(context, auth),
}));
```

On first connect the runtime logs the resolved mode, e.g. `[bb-data] DB TLS: verifying the server certificate against the pinned CA (verify-full equivalent).`

## Behavior change

For hand-written `fromExisting({ connectionString })` with no `ssl`, the connection now verifies the server certificate (it previously did not). Private-CA providers (e.g. Supabase) require pinning the CA via `ssl: { ca }` (the certificate contents). Pass `ssl: { rejectUnauthorized: false }` to keep the old behavior explicitly. `db pull`-generated apps are unaffected in default connectivity. Released as a `minor` bump (see the changeset's upgrade note).

## Testing

**Automated**
- `bb-data` unit suite green, including: the engine verifies by default and honors an explicit `ssl` option; the SSL resolver and CA-file handling; `sslmode` is stripped so a pinned CA is honored; the deployed-runtime path forwards the caller's `ssl`; and a test that transpiles and runs the generated connection wiring to exercise its fail-closed and verify branches.
- The generated-app type-check passes against the regenerated snapshot.
- The live database end-to-end CI job now connects with the provider CA pinned, so it exercises full certificate verification (it previously connected unverified). Lint runs in CI.

**Manual, end-to-end against a real managed Postgres provider**
- Pulled the schema from a real database: the flow prompts for the provider's CA certificate and the connection then verifies the server certificate end-to-end (encryption plus identity, including hostname).
- Negative test: pointing the connection at an incorrect CA causes it to fail certificate verification and refuse to connect, rather than silently falling back to an unverified connection — confirming it fails closed.
- Ran a generated application locally and performed create/read/delete against the database; the server certificate is verified against the captured CA on each connection.
- Deployed the same application to both a sandbox and a production environment; both connect and operate with the server certificate verified in the deployed function — the captured CA travels with the app, so no extra runtime configuration is required.

## Notes

- `database.ca.ts` holds a certificate only (no private key); the connection string and other secrets continue to flow through SSM, not the repo. The MIGRATION_GUIDE notes keeping it committed and that a separate production project may present its own CA.
- Operational `db pull` introspection resolves `ssl` from the captured CA / environment (it runs on the operator host, not the Lambda).


packages/bb-data/src/db-pull/templates.ts
```diff
@@ -75,11 +77,54 @@ export const WIRING_RESOLVE_CONN_FN = `async function resolveConnString(): Promi
...
+    // Pin the provider CA → fully verified (sslmode=verify-full equivalent).
+    console.log('[bb-data] DB TLS: verifying the server certificate against the pinned CA (verify-full equivalent).');
+    return { ca, rejectUnauthorized: true };
+  }
+  // ⚠️ No CA available: the connection is encrypted but the server certificate is
+  // NOT verified (no man-in-the-middle protection). Download your Supabase CA
+  // ('prod-ca-2021.crt' from Database Settings → SSL Configuration) and re-run
+  // \`npx bb-data pull\` to capture it (or set DATABASE_CA_CERT). See MIGRATION_GUIDE.md.
+  console.warn('[bb-data] DB TLS: server certificate NOT verified — encrypted only (no CA). Run bb-data pull to capture your CA, or set DATABASE_CA_CERT. See MIGRATION_GUIDE.md.');
+  return { rejectUnauthorized: false };
```
soberm, line 145: When no CA is configured this falls back to rejectUnauthorized: false and deploys with no certificate verification. The console.warn only surfaces in logs after deploy, and the committed wiring file shows an explicit rejectUnauthorized: false that reads as intentional. In a Lambda context this should throw instead so a missing CA fails loudly. Suggest gating on process.env.AWS_LAMBDA_FUNCTION_NAME and keeping the insecure fallback for local dev only.

packages/bb-data/src/external-ssl.ts
```diff
@@ -0,0 +1,38 @@
...
+export function externalDbSsl(): { ca?: string; rejectUnauthorized: boolean } {
+  const ca = process.env.DATABASE_CA_CERT;
+  if (ca && ca.trim() !== '') {
+    const pem = ca.includes('-----BEGIN') ? ca : readFileSync(ca, 'utf8');
+    return { ca: pem, rejectUnauthorized: true };
+  }
+  // No CA available — encrypted but unauthenticated. Acceptable for these
+  // ephemeral, operator-driven connections to a database the operator owns;
+  // pin DATABASE_CA_CERT to verify the server identity.
+  return { rejectUnauthorized: false };
```
soberm, line 118: This returns rejectUnauthorized: false whenever DATABASE_CA_CERT is unset, so CI/CD migrations and first-pull introspection run unverified. Migrations apply DDL, which is a higher-privilege path than the Lambda. Consider reading the committed database.ca.ts as a fallback CA source, or failing when no CA is available in a non-interactive run.

packages/bb-data/src/db-pull/templates.ts
```diff
@@ -75,11 +77,54 @@ export const WIRING_RESOLVE_CONN_FN = `async function resolveConnString(): Promi
...
+ * `DATABASE_CA_CERT` (inline PEM or a file path) overrides the committed cert.
+ * If neither is available, this falls back to an ENCRYPTED-BUT-UNVERIFIED
+ * connection (rejectUnauthorized: false): safe against passive eavesdropping but
+ * NOT an active man-in-the-middle. Provide the CA for production.
+ */
+export const WIRING_RESOLVE_SSL_FN = `function resolveDbSsl(): { ca?: string; rejectUnauthorized: boolean } {
+  // DATABASE_CA_CERT (inline PEM or a file path) overrides the committed cert.
+  const envCa = process.env.DATABASE_CA_CERT;
+  const ca = envCa
+    ? (envCa.includes('-----BEGIN') ? envCa : readFileSync(envCa, 'utf8'))
```
soberm: readFileSync runs at module load (cold start). If DATABASE_CA_CERT points to a missing file this throws ENOENT with no TLS context, so the operator sees an unrelated startup crash. Wrap it and throw a clear error, and validate the content contains BEGIN CERTIFICATE.

packages/bb-data/src/external-ssl.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Note: for the CA to take effect, any `sslmode` must be stripped from the URL —
+ * node `pg` verifies against the system trust store and ignores a programmatic
+ * `ssl.ca` when `sslmode` is present in the connection string. The callers in
+ * this package already normalize the URL before connecting.
+ */
+export function externalDbSsl(): { ca?: string; rejectUnauthorized: boolean } {
+  const ca = process.env.DATABASE_CA_CERT;
+  if (ca && ca.trim() !== '') {
+    const pem = ca.includes('-----BEGIN') ? ca : readFileSync(ca, 'utf8');
```
soberm: If the file is empty or corrupt, pem is empty but the function still returns rejectUnauthorized: true. Node may fall back to the system trust store or throw an opaque error. Validate the PEM is non-empty and contains BEGIN CERTIFICATE before returning.

packages/bb-data/src/index.mock.ts
```diff
@@ -40,20 +40,24 @@ export class Database extends Scope {
...
     if (options?.connection && isConnectionString(options.connection)) {
-      // External database via connection string — connect directly
+      // External database via connection string — connect directly.
+      // Local dev defaults to NOT verifying the certificate (self-signed local
+      // databases are common); a caller-supplied `ssl` (e.g. the `db pull`-generated
+      // wiring pinning the provider CA) overrides this.
       const engine = new PgClientEngine({
         connectionString: options.connection.connectionString,
-        ssl: { rejectUnauthorized: false },
+        ssl: options.connection.ssl ?? { rejectUnauthorized: false },
```
soberm: The mock defaults to rejectUnauthorized: false while the AWS path defaults to true. Hand-written fromExisting with no ssl passes locally but fails on deploy against a private CA, and the failure only shows up after deploying. Align the defaults or warn when ssl is omitted.

packages/bb-data/src/external-ssl.ts
```diff
@@ -0,0 +1,38 @@
...
+    return { ca: pem, rejectUnauthorized: true };
```
soberm: No minVersion is set. Node 18+ defaults to TLS 1.2 but that is not guaranteed across all targets. Add minVersion: 'TLSv1.2'.

packages/bb-data/src/types.ts
```diff
@@ -44,4 +44,19 @@ export interface DatabaseOptions {
...
+       *
+       * Managed Postgres providers (Supabase, Neon, RDS) present certificates
+       * signed by a provider-specific CA that is not in Node's built-in trust
+       * store, so verification requires pinning that CA via `ca`. For Supabase,
+       * download `prod-ca-2021.crt` from Database Settings → SSL Configuration.
+       *
+       * Set `rejectUnauthorized: false` only if you accept the man-in-the-middle
+       * risk (encrypted but unauthenticated); prefer pinning the CA instead.
+       */
+      ssl?: { rejectUnauthorized?: boolean; ca?: string };
```
soberm: This type allows { ca, rejectUnauthorized: false }, where pg ignores the CA and verification is off, so it looks secure but is not. A discriminated union ({ rejectUnauthorized: true; ca: string } | { rejectUnauthorized: false }) would make that combination a compile error.

packages/bb-data/src/db-pull/pull.ts
```diff
@@ -501,7 +516,41 @@ async function dbPullDevInteractive(outputDir: string): Promise<void> {
   }
   console.log('  Detected provider: Supabase\n');
 
-  await dbPull({ connectionString, outputDir, projectRef });
+  // Optional: capture the project's CA certificate so the generated connection
+  // verifies the server's TLS identity (otherwise it connects encrypted but
+  // unverified). The CA is a public, non-secret cert; it is committed to
+  // database.ca.ts and bundled into the deployed function.
+  const caFilePath = path.join(outputDir, 'database.ca.ts');
+  const hasExistingCa = fs.existsSync(caFilePath) && fs.readFileSync(caFilePath, 'utf-8').includes('-----BEGIN');
```
soberm: includes('-----BEGIN') also matches a CSR or a private key. Use '-----BEGIN CERTIFICATE-----' so only an actual certificate counts as an existing CA.

soberm (commented):
What happens with an app that used a prior version? Will it keep ssl off?

packages/bb-data/src/db-pull/templates.ts
vishaalmehrishi: Fixed. The generated `resolveDbSsl()` now **fails closed in the deployed function**: when `process.env.AWS_LAMBDA_FUNCTION_NAME` is set and no CA is available (neither the committed `database.ca.ts` nor `DATABASE_CA_CERT`), it throws instead of returning `rejectUnauthorized: false`. The encrypted-but-unverified fallback (and its `console.warn`) is now local-dev only — so a missing CA in production fails loudly at startup rather than silently deploying unverified.

packages/bb-data/src/external-ssl.ts
vishaalmehrishi: Fixed. `externalDbSsl()` now **fails closed in non-interactive runs**: if `process.env.CI` is set and no CA is configured, it throws rather than connecting unverified — so CI/CD migrations (which apply DDL) won't run against an unverified server. First-pull `db pull` introspection is the one deliberate exception: it runs *before* a CA has been captured, so it opts in explicitly via `externalDbSsl({ allowUnverifiedInCi: true })`. Interactive operator runs keep the warned fallback.

packages/bb-data/src/db-pull/templates.ts
vishaalmehrishi: Fixed. The generated wiring now reads the CA through a `readCaFile()` helper that (a) wraps `readFileSync` and rethrows with a clear TLS message including the path, instead of a bare `ENOENT` startup crash, and (b) validates the contents contain `-----BEGIN CERTIFICATE-----`. A misconfigured `DATABASE_CA_CERT` now produces an actionable error at startup.

packages/bb-data/src/external-ssl.ts
vishaalmehrishi: Fixed. `externalDbSsl()` now validates that the resolved PEM is non-empty and contains `-----BEGIN CERTIFICATE-----` before returning `rejectUnauthorized: true`; otherwise it throws. Same validation as the generated path, so an empty or corrupt CA can no longer silently fall through to the system trust store.

packages/bb-data/src/index.mock.ts
vishaalmehrishi: The asymmetry is intentional — local dev often points at a self-signed Postgres, so the mock defaults to not verifying, while the deployed runtime verifies. But you're right that the gap only shows up at deploy time. Fixed by **warning when `ssl` is omitted** on a hand-written external connection in the mock, so the dev/prod parity gap surfaces during `npm run dev` instead of after deploy. Pinning `ssl: { ca }` (or setting `ssl: { rejectUnauthorized: false }` explicitly) silences the warning.

packages/bb-data/src/external-ssl.ts
vishaalmehrishi: Done. Rather than scatter it across call sites, I enforced the floor **centrally in `PgClientEngine`** (`ssl: { minVersion: 'TLSv1.2', ...callerSsl }`), so every path gets TLS 1.2+ — deployed runtime, the operational helper, and the generated wiring — including the unverified opt-out path (TLS version is orthogonal to certificate verification). A caller-supplied `minVersion` still wins.

packages/bb-data/src/types.ts
vishaalmehrishi: Great suggestion — done. `ssl` is now a discriminated union (exported as `ExternalSslOptions`), so `{ ca, rejectUnauthorized: false }` is a compile error.

One refinement on the exact shape: I kept `ca` optional on the verify branch —

```ts
type ExternalSslOptions =
  | { rejectUnauthorized?: true; ca?: string }
  | { rejectUnauthorized: false };
```

— so verifying against Node's built-in trust store with no CA (valid for publicly-trusted certs, e.g. some RDS configs) still type-checks, while the misleading `{ ca, rejectUnauthorized: false }` combo is still rejected.

packages/bb-data/src/db-pull/pull.ts
vishaalmehrishi: Fixed. Switched to `-----BEGIN CERTIFICATE-----` here and everywhere else we sniff for a cert — the `externalDbSsl()` inline-PEM-vs-path check, the generated `resolveDbSsl()`, and this `hasExistingCa` probe — so a CSR or a private key isn't mistaken for a CA.

vishaalmehrishi:
> What happens with an app that used a prior version? Will it keep ssl off?

Two cases:

- **`db pull`-generated apps:** the verified wiring (`resolveDbSsl()` + the committed `database.ca.ts`) is only emitted on a **re-pull**. An app generated before this change keeps its existing `supabase.ts` (no `ssl`) until the next `npx bb-data pull` — so yes, it stays ssl-off until re-pulled. A re-pull regenerates the wiring and prompts for the CA.
- **Hand-written `fromExisting({ connectionString })` with no `ssl`:** behavior changes to **verify by default**. A private-CA provider (e.g. Supabase) will now fail until the CA is pinned via `ssl: { ca }` — or pass `ssl: { rejectUnauthorized: false }` to keep the old behavior explicitly. This is captured as an upgrade note in the changeset (`minor` bump).

Happy to add the "re-pull to pick up verification" note to the generated migration guide so existing-app users aren't surprised.

packages/bb-data/src/index.aws.ts
```diff
@@ -66,7 +66,13 @@ export class Database extends Scope {
...
         ? conn.connectionString
         : await conn.connectionString.get();
-      return new RLSEnabledDatabase(new PgClientEngine({ connectionString: connStr, ssl: { rejectUnauthorized: false } }));
+      // Verify the server's TLS certificate by default (PgClientEngine defaults to
+      // rejectUnauthorized: true when ssl is undefined). The `db pull`-generated
+      // wiring supplies an ssl config that pins the provider CA; callers using
+      // fromExisting() directly can pass `ssl` to pin a CA or opt out. Previously
+      // this hardcoded rejectUnauthorized:false, leaving the deployed Lambda's
+      // connection to external DBs (Supabase/Neon/etc.) unauthenticated (MITM-exposed).
+      return new RLSEnabledDatabase(new PgClientEngine({ connectionString: connStr, ssl: conn.ssl }));
```
soberm, line 75: This is the line that restores verify-by-default in the deployed runtime: conn.ssl being undefined falls through to PgClientEngine's default. The engine default is covered in pg-client-engine.test.ts, but there is no aws-layer test asserting that _initBase forwards conn.ssl and that an omitted ssl yields a verifying pool here. This is exactly where a future refactor could regress the deployed default back to unverified. Worth a direct test that builds a Database with a string connection and inspects the resulting pool's ssl: rejectUnauthorized true when no ssl is passed, and that an explicit ssl is forwarded.

packages/bb-data/src/db-pull/introspect.ts
```diff
@@ -65,10 +67,20 @@ export async function introspect(connectionString: string): Promise<Introspectio
...
+  // the system store — so verification needs that CA pinned. When the caller
+  // provides one (e.g. the CA captured by `db pull`), verify against it;
+  // otherwise externalDbSsl() applies DATABASE_CA_CERT or an unverified fallback
+  // for this ephemeral, operator-driven introspection. First-pull introspection
+  // runs *before* a CA has been captured, so it explicitly tolerates the
+  // unverified fallback even in CI (unlike the migration/DDL paths).
+  parsed.searchParams.delete('sslmode');
+  const connStr = parsed.toString();
+  const ssl = caCert ? { ca: caCert, rejectUnauthorized: true } : externalDbSsl({ allowUnverifiedInCi: true });
+  const pool = new pg.Pool({ connectionString: connStr, ssl });
```
soberm: This pool is created with a raw pg.Pool instead of going through PgClientEngine, so it misses the minVersion: 'TLSv1.2' floor the engine sets. cli.ts, baseline.ts, and external-migrations.ts all route through PgClientEngine and get the floor; introspection is the only outlier. Separately, the caCert branch on the line above treats caCert as inline PEM only, while externalDbSsl() accepts inline PEM or a file path. Either add minVersion here, or run introspection through PgClientEngine so the floor and the PEM/path handling are shared.

packages/bb-data/src/engines/pg-client-engine.ts
```diff
@@ -12,8 +12,8 @@ import { DatabaseErrors } from '../errors.js';
 export interface PgClientEngineConfig {
   /** PostgreSQL connection URI (e.g. postgresql://user:pass@host:5432/db). */
   connectionString: string;
-  /** SSL configuration. Defaults to `{ rejectUnauthorized: true }`. */
-  ssl?: { rejectUnauthorized?: boolean; ca?: string };
+  /** SSL configuration. Defaults to `{ rejectUnauthorized: true }`. A TLS 1.2 floor (`minVersion: 'TLSv1.2'`) is applied unless overridden here. */
+  ssl?: { rejectUnauthorized?: boolean; ca?: string; minVersion?: 'TLSv1.2' | 'TLSv1.3' };
```
soberm: This ssl type is loose ({ rejectUnauthorized?: boolean; ca?: string }), so it still permits { ca, rejectUnauthorized: false }, the combination the ExternalSslOptions union in types.ts was designed to make a compile error. fromExisting keeps that boundary strict, but PgClientEngine and PgClientEngineConfig are exported, so a caller using the engine directly bypasses the guarantee. Consider reusing ExternalSslOptions (intersected with the optional minVersion) here, or add a short comment noting the broader shape is intentional for internal callers.

packages/bb-data/src/index.aws.ts
vishaalmehrishi: Added a direct aws-layer test for exactly this seam: [`index.aws.test.ts`](https://github.com/aws-devtools-labs/aws-blocks/blob/226e4be7c7b224a9cb6e666219b70f4b47635a75/packages/bb-data/src/index.aws.test.ts#L35). It builds a `Database` with a string connection, triggers `_initBase()` via `getEngine()`, and inspects the resulting `pg.Pool`'s `ssl`:

- **ssl omitted → `{ minVersion: 'TLSv1.2', rejectUnauthorized: true }`** — the verify-by-default guarantee this PR restores for the deployed runtime
- explicit CA-pin → forwarded as-is
- explicit `{ rejectUnauthorized: false }` → forwarded (opt-out honored)
- async (`{ get }`) connection string → also verifies by default

So a future refactor that drops the `conn.ssl` forwarding or flips the default now fails a test. (`rejectUnauthorized: true` is Node's own TLS default — we simply stopped overriding it to `false`; see the [Node TLS docs](https://nodejs.org/api/tls.html#tlsconnectoptions-callback).)

packages/bb-data/src/db-pull/introspect.ts
vishaalmehrishi: Both points fixed by routing introspection through `PgClientEngine` instead of a raw `pg.Pool`: [introspect.ts#L87](https://github.com/aws-devtools-labs/aws-blocks/blob/226e4be7c7b224a9cb6e666219b70f4b47635a75/packages/bb-data/src/db-pull/introspect.ts#L87). It now inherits the engine's `minVersion: 'TLSv1.2'` floor — the same posture as the migrate CLI, baseline, and migration runner (introspect was the only raw-pool outlier) — plus the connect-time TLS confirmation.

For the inline-PEM-vs-path inconsistency: the caller CA is now normalized via a shared [`resolveCaPem()`](https://github.com/aws-devtools-labs/aws-blocks/blob/226e4be7c7b224a9cb6e666219b70f4b47635a75/packages/bb-data/src/external-ssl.ts#L88) (inline PEM **or** file path) — the same helper [`externalDbSsl()`](https://github.com/aws-devtools-labs/aws-blocks/blob/226e4be7c7b224a9cb6e666219b70f4b47635a75/packages/bb-data/src/external-ssl.ts#L98) uses, so the two can't drift. (node-postgres passes `ssl` straight through to `tls.connect` — [node-postgres SSL docs](https://node-postgres.com/features/ssl).)

packages/bb-data/src/engines/pg-client-engine.ts
vishaalmehrishi: Done — `PgClientEngineConfig.ssl` now reuses the [`ExternalSslOptions`](https://github.com/aws-devtools-labs/aws-blocks/blob/226e4be7c7b224a9cb6e666219b70f4b47635a75/packages/bb-data/src/types.ts#L79) discriminated union, intersected with the optional `minVersion`: [pg-client-engine.ts#L27](https://github.com/aws-devtools-labs/aws-blocks/blob/226e4be7c7b224a9cb6e666219b70f4b47635a75/packages/bb-data/src/engines/pg-client-engine.ts#L27). So `{ ca, rejectUnauthorized: false }` is now a compile error for direct engine callers too, not just `fromExisting` — closing the bypass you flagged. node-pg ignores `ca` when `rejectUnauthorized: false`, so that combo *looks* secure but isn't; the union makes it unrepresentable.

vishaalmehrishi:
> What happens with an app that used a prior version? Will it keep ssl off?

Good question — it's a deliberate behavior change, and it splits by how the app connects:

- **`db pull`-generated apps:** unaffected in default connectivity. The generated `supabase.ts` wiring (`resolveDbSsl()`) is regenerated on each pull and supplies an explicit `ssl` config (pins the captured CA, or a visible commented opt-out), so upgrading the library doesn't silently change their behavior.
- **Hand-written `fromExisting({ connectionString })` with no `ssl`:** this **now verifies** the server certificate (it previously hardcoded `rejectUnauthorized: false`). For a private-CA provider like Supabase that means the connection will *fail* until they pin the CA via `ssl: { ca }` — which is the intent: the old default silently accepted any certificate (MITM-exposed). Passing `ssl: { rejectUnauthorized: false }` keeps the old behavior explicitly.

Shipped as a **minor** with an upgrade/remediation note in the changeset so the break is discoverable. So no — it does not keep ssl off by default anymore; that default was the vulnerability.

soberm (approved):

svidgen (approved):

PR #114, opened 2026-06-26
fix: scope external-database connection-string parameter per app

vishaalmehrishi:
# Scope the external-database connection-string parameter per app

## Problem

When a Blocks app connects to an external database (for example Supabase), the
deploy process stores the database connection string in an encrypted SSM
SecureString parameter so the deployed Lambda can read it at runtime.

Today that parameter is named only after the stage:

```
/blocks/production/db-connection-string
```

The name contains nothing that identifies the app. So if two different Blocks
apps are deployed to the same AWS account, region, and stage, they both compute
the **same** parameter name and overwrite each other's connection string. The
second deploy wins, and the first app's Lambda then silently connects to the
wrong database with no error.

## Root cause

The parameter name had no app identity in it. That constraint no longer holds —
PR #51 gives every app a stable, committed stack name (`.blocks/config.json`)
known *before* template generation.

> A first attempt at this fix (#39) addressed the collision with a
> deploy-then-write approach — naming the parameter `/${fullId}` at synth,
> emitting it as a `CfnOutput`, and writing the value *after* `cdk deploy` by
> reading the name back from stack outputs. That was necessary only because the
> final name was assumed unknowable before synth. Once #51 made the stack name
> committed and pre-computable, that machinery became unnecessary; this PR
> supersedes #39 with the simpler approach below.

## Fix

The parameter is now named after the app's deployment (stack):

```
/<stack-name>-db-url        e.g.  /my-app-k7x2mf-prod-db-url
```

Built from `getStackName({ sandbox, projectRoot? })` — a single options-object
function that is the sole source of truth for an app's CloudFormation stack name:

- production: `<stackId>-prod`
- sandbox:    `<stackId>-<sandboxId>`

CDK templates call `getStackName`; `dbConnectionParameterName(stackName)` is a
pure string transform (`/${stackName}-db-url`) that takes the result.

### Why written name == read name

1. `ensureSecrets` **writes** using `dbConnectionParameterName(getStackName({ sandbox, projectRoot }))`
2. Generated wiring at synth **records** the name with the same two calls and same inputs
3. Lambda **reads** the name stamped at synth (never recomputes)

Both callers visibly call `getStackName` then `dbConnectionParameterName` — the
agreement is obvious at each call site.

### Pure name derivation + explicit sandbox-id creation

`getStackName` is now **read-only**: it uses `readSandboxId` (throws if the
sandbox-id file is absent) instead of `getSandboxId` (which creates it). This
removes a hidden filesystem side effect — importing the generated wiring during
`npm run dev` or tests no longer mutates the working tree.

Because the old create-if-missing behavior lived *inside* `getStackName`, the
sandbox orchestrator (`sandbox.ts`) now **explicitly** materializes the sandbox
id via `getSandboxId(process.cwd())` before any name derivation runs. This is the
one place creation belongs; everything downstream (`ensureSecrets`, synth) is
read-only.

The generated wiring only computes the parameter name at synth/deploy (when
`BLOCKS_STAGE` is set). At local dev, neither `BLOCKS_SSM_PARAM_DB_URL` nor
`BLOCKS_STAGE` is set, so it uses a placeholder — the mock reads from
`.env.local` and the name is unused. This keeps module import side-effect-free.

## Compatibility

- **Breaking signatures:** `getStackName({ sandbox, projectRoot? })` (was
  positional args); `dbConnectionParameterName(stackName)` (was
  `(stage: string)`). Pre-1.0 preview, no external consumers on old forms. Apps
  with stale generated `supabase.ts` must run `npx bb-data pull`.
- **Prerequisite (PR #51):** `.blocks/config.json` must have a `stackId`. Pre-#51
  apps must set it manually (see D-012 in `docs/DECISIONS.md`) or deploys will
  throw.
- **Old parameter orphaned:** `/blocks/{stage}/db-connection-string` is left
  behind unused. Can be deleted manually.

## Tradeoffs

- **Write before deploy:** Connection string persists to SSM even if deploy
  fails. Harmless — idempotent, overwritten next deploy.
- **`process.cwd()` vs `projectRoot`:** Generated wiring omits `projectRoot`
  (defaults to cwd) while the deploy orchestrators pass it explicitly. They agree
  because orchestrators set `cwd === projectRoot`. TODO left for future PR to
  thread context explicitly.

## Testing

- `getStackName`: production, sandbox, missing config, purity (does NOT write)
- `readSandboxId`: returns existing, throws without creating
- `dbConnectionParameterName`: pure formatting, distinct names
- Cross-site invariant: write name == read name for same `(projectRoot, sandbox)`
- `npm run build`, `npm run check:api` pass for `@aws-blocks/core`, `@aws-blocks/bb-data`

## Manual testing

Verified end-to-end with a scaffolded sample app backed by a real Supabase DB:

- **Local dev** (`npm run dev`) — reads the connection string from `.env.local`
  via the mock; importing the generated wiring no longer mutates the working tree
  (no `.blocks-sandbox/sandbox-id.txt` created on import).
- **Sandbox** (`npm run sandbox`) — orchestrator creates the sandbox id,
  `ensureSecrets` writes the stack-scoped parameter, and the deployed Lambda reads
  it back at runtime (create/list todos and notes succeed through the deployed API).
- **Production** (`npm run deploy` with a `.env.production` connection string) —
  same path with the `-prod` stack name; deployed runtime resolves the parameter.

The two stages produced distinct, app-scoped parameter names as intended:

```
db-pull-test-1e43e5-vishaalm-e94612-db-url   (sandbox)
db-pull-test-1e43e5-prod-db-url              (prod)
```

## Files changed

- `packages/core/src/scripts/stack-id.ts` — `getStackName` pure, new `readSandboxId`
- `packages/core/src/scripts/sandbox.ts` — explicit `getSandboxId` before name derivation
- `packages/core/src/db-naming.ts` — `dbConnectionParameterName(stackName)` pure transform
- `packages/core/src/scripts/ensure-secrets.ts` — calls `getStackName` + `dbConnectionParameterName`
- `packages/core/src/scripts/external-migrations-step.ts` — same pattern
- `packages/core/src/scripts/index.ts` — exports `readSandboxId`
- `packages/bb-data/src/db-pull/generate.ts`, `templates.ts` — generated wiring
  computes the name only at synth, placeholder at dev
- `test-apps/db-pull-typecheck/generated/index.ts` — regenerated fixture
- `docs/DECISIONS.md` — Compatibility section on D-013
- Tests: `db-naming.test.ts`, `stack-id.test.ts`, `db-pull.test.ts`
- CDK templates (7) — use `getStackName`
- Changeset, decision record


packages/core/src/scripts/stack-id.ts
```diff
@@ -34,10 +34,27 @@ export function getStackId(projectRoot?: string): string {
...
+ * Use this in pure name-derivation paths that must not mutate the filesystem.
+ */
+export function readSandboxId(projectRoot?: string): string {
+  const root = projectRoot || process.cwd();
+  const filePath = join(root, '.blocks-sandbox', 'sandbox-id.txt');
+  if (existsSync(filePath)) return readFileSync(filePath, 'utf-8').trim();
+  throw new Error(
+    `Sandbox id not found at ${filePath} — run \`npm run sandbox\` to initialize it.`,
+  );
+}
```
svidgen: What's the rationale for not just using `getSandboxId()`? Even though a sandbox may not have been created when `generate()` pulls it, it's not harmful to generate an identifier. If you call `getStackName()` when sandbox hasn't been run yet, this will throw an exception. Is that what we actually want?

This might be answered as I read on, but what happens to the DX after the customer clobbers or changes the their `sandbox-id.txt` file? When and where exactly does this blow up?

packages/core/src/scripts/stack-id.ts
```diff
@@ -34,10 +34,27 @@ export function getStackId(projectRoot?: string): string {
...
+
 /**
  * Get or create a per-machine sandbox identifier.
  * Stored in `.blocks-sandbox/sandbox-id.txt` (gitignored).
  * Format: `<username(8)>-<random(6)>` — identifies the developer's sandbox.
+ *
+ * NOTE: This function WRITES the file if absent (side effect). Use only in the
+ * sandbox orchestrator. For pure name derivation, use {@link readSandboxId}.
```
svidgen: As noted above, I'm not sure this is a critical distinction.

packages/core/src/db-naming.ts
```diff
@@ -35,7 +41,13 @@ export function extractDbRef(connectionString: string): string {
...
-  return `/blocks/${stage}/db-connection-string`;
+/**
+ * SSM SecureString parameter name for a deployment's external database
+ * connection string.
+ *
+ * Pure string transform: `/<stackName>-db-url`. The caller is responsible for
+ * computing the stack name via `getStackName({ sandbox, projectRoot })`.
+ */
+export function dbConnectionParameterName(stackName: string): string {
+  return `/${stackName}-db-url`;
```
svidgen, line 52: Should this be scoped to the specific `Database` it connects to? A customer may have more than one. (Do we support having more than one otherwise?)

svidgen (commented):
It escaped me in the past, but it looks like we support only a single (external?) SQL database. We need at least start taking steps to remedy this by having associated secrets and resource identifiers scoped to the `Database` instance if possible; not just named after the stack.

Is that possible?

packages/core/src/scripts/stack-id.ts
vishaalmehrishi: Yeah, you're right. Switched it back to `getSandboxId` (create-on-missing) and removed `readSandboxId` — pushed in `fe80e81`.

I'd originally gone read-only because I thought creating the id here could mint one that disagrees with what the real sandbox used. That was wrong, and your instinct was right: the id just lives in `.blocks-sandbox/sandbox-id.txt`, so whoever gets there first writes it and everyone after reads the same value ([getSandboxId](https://github.com/aws-devtools-labs/aws-blocks/blob/fe80e81/packages/core/src/scripts/stack-id.ts#L47)). The throw only fired when the file didn't exist yet — and in that case there's nothing to disagree with. So creating it is perfectly safe, and get-or-create is the more natural pattern here anyway.

Once that reasoning fell away, all that was left for the throw was a bit of tidiness plus a fail-fast error, which isn't worth making people trip over an exception. So `getStackName` now lazily creates the id, and I dropped the explicit pre-creation step I'd added to the sandbox orchestrator.

On your two questions: it only ever threw if you derived a sandbox name *outside* `npm run sandbox` — a bare `cdk synth`, CI, a stray import — and the normal flow always created the id first, so you'd never hit it. And if someone deletes or edits `sandbox-id.txt`, nothing dramatic: the stack name changes, so the next `npm run sandbox` just spins up a new sandbox stack and leaves the old DB param orphaned. Sandboxes are per-dev and disposable (the file's gitignored), so that's fine.

(Prod was never in the picture either way — it's `<stackId>-prod`, no sandbox id.)

packages/core/src/scripts/stack-id.ts
vishaalmehrishi: Agreed (see comment above)

vishaalmehrishi:
> It escaped me in the past, but it looks like we support only a single (external?) SQL database. We need at least start taking steps to remedy this by having associated secrets and resource identifiers scoped to the `Database` instance if possible; not just named after the stack.
> 
> Is that possible?

The multi-external DB support story needs a bit more than naming, and it should be done as one piece. I will create a feature task for multi-DB support and we can take it from there.

---

Update: Here's the feature request issue: https://github.com/aws-devtools-labs/aws-blocks/issues/123

packages/core/src/db-naming.ts
vishaalmehrishi: We do not support multiple external DB connections at the moment. Given we only support one external DB provider (Supabase) at preview and it's unlikely that a customer would want to connect multiple Supabase DBs in a single application, I figured we could defer this support to post-preview launch (i.e. now).

Multi-DB support needs more than just naming changes, so I will leave it out of this PR. I will create a feature task for multi-DB support and we can take it from there.

vishaalmehrishi: Here's the feature request issue: https://github.com/aws-devtools-labs/aws-blocks/issues/123

docs/DECISIONS.md
```diff
@@ -430,3 +430,97 @@ Both helpers (`getStackId`, `getSandboxId`) are exported from `@aws-blocks/block
...
+inside the construct tree.
+
+### Decision
+
+The parameter name is stack-scoped: `/<stackName>-db-url`, where `stackName` comes
+from a single new `getStackName(projectRoot, { sandbox })` helper. That helper is
+also the one place the CDK templates compute the stack name (the `-prod` / sandbox
+suffix assembly that D-012 left duplicated inline across templates is removed).
+
+`dbConnectionParameterName(projectRoot, { sandbox })` derives the parameter name
```
bobbor: `getStackName` and `dbConnectionParameterName` have stale implementations in this Decision.

.changeset/external-db-param-stack-scoped.md
```diff
@@ -0,0 +1,24 @@
...
+
+The external-database connection string was stored in an SSM parameter named only
+by stage (`/blocks/{stage}/db-connection-string`), so two Blocks apps deployed to
+the same AWS account + region + stage computed the same name and silently
+overwrote each other's credentials.
+
+The parameter name is now stack-scoped (`/<stackName>-db-url`), derived from a
+single new `getStackName(projectRoot, { sandbox })` helper that is also the one
+place the CDK templates compute the stack name (replacing logic duplicated across
+templates). The same `dbConnectionParameterName(projectRoot, { sandbox })` is used
```
bobbor: both function calls have stale implementation signatures.

bobbor (changes requested):
2 smaller (documentation) things for that require addressing

docs/DECISIONS.md
vishaalmehrishi: Fixed in 111bd1f. The Decision section now uses the current signatures — `getStackName({ sandbox, projectRoot })` and `dbConnectionParameterName(stackName)` (the caller computes the stack name via `getStackName`) — matching the code after fe80e81 and this decision's own Compatibility section.

Note: this decision was renumbered **D-013 → D-014** in the merge, since #107 landed a `D-013` (TLS) on `main` first.

.changeset/external-db-param-stack-scoped.md
vishaalmehrishi: Fixed in 111bd1f. The changeset now reads `getStackName({ sandbox, projectRoot })` and `dbConnectionParameterName(stackName)` (fed the stack name from `getStackName`).

packages/core/src/db-naming.ts
vishaalmehrishi: Confirming this is unchanged by the #107 merge: multi-`Database` scoping stays deferred to post-preview, tracked in #123. The current single-external-DB naming is intact.

vishaalmehrishi:
Merged latest `main` into this branch — merge commit 111bd1f. `main` now includes #107 (external-DB TLS verify-by-default), which overlapped this PR.

**Conflict resolution**
- `docs/DECISIONS.md`: #107 and this PR both added a `## D-013`. Kept #107's TLS decision as **D-013** and renumbered this PR's stack-scoped-param decision to **D-014**.
- `test-apps/db-pull-typecheck/generated/index.ts`: regenerated from the merged generators, so the snapshot carries **both** #107's TLS wiring (`resolveDbSsl`/`DATABASE_CA_CERT`) and this PR's stack-scoped param wiring.
- Other overlapping files (`db-pull.test.ts`, `db-pull/generate.ts`, `db-pull/templates.ts`, `sandbox.ts`) auto-merged.

**Feedback (@bobbor)**: fixed the stale `getStackName`/`dbConnectionParameterName` signatures in the decision doc and changeset.

**Verified on the merged tree**: `bb-data` 359 pass / 0 fail, `core` 544 pass / 0 fail, `db-pull-typecheck` typecheck clean.

bobbor (approved):

svidgen (approved):

PR #115, opened 2026-06-26
feat(hosting,pipeline): secret() / config() values for self-hosted deployments

osama-rizk:
## What & why

`Hosting` and `Pipeline` are framework-neutral leaf packages with **no Blocks backend to lean on**, so a standalone Next/Nuxt/Astro app, an Amplify Gen2 self-hosting app, the pipeline's own credentials, and a Blocks app's hosting-only values all need a first-class way to keep secrets/config out of source. This PR adds that.

After a design review, we landed on **two intent functions** (I1 “Approach B”): the developer states *what kind of value* it is by **which function they call**, and the framework owns the store.

| Function | Store | Read at runtime |
|---|---|---|
| **`secret('KEY')`** | AWS **Secrets Manager** (sensitive) | `getSecret('KEY')` |
| **`config('KEY')`** | AWS **SSM Parameter Store** (non-sensitive, free tier) | `getConfig('KEY')` |

The developer never picks a store — it’s derived from the function (`storeForKind`), so the CLI write, the IAM grant, the synth-time fetch, and the runtime read can’t drift.

---

### Declare (infra) — two intent functions
```ts
import { Hosting, secret, config } from '@aws-blocks/blocks/cdk'; // or the leaf @aws-blocks/hosting

await Hosting.create(stack, 'Web', {
  root: '.', framework: 'nextjs', api: blocksStack,
  environment: {
    STRIPE_KEY:    secret('STRIPE_KEY'),     // → Secrets Manager
    FEATURE_FLAGS: config('FEATURE_FLAGS'),  // → SSM Parameter Store
  },
  domain: { domainName: config('DOMAIN_PROD') },      // resolved to a literal at synth
  // separate per-kind namespace/cache config (optional):
  secretStore: { prefix: '/myapp/secrets', cacheTtlSeconds: 0 },
  configStore: { prefix: '/myapp/config',  cacheTtlSeconds: 30 },
});
```
- The marker is **inert** (`{ key, kind }`) — safe to commit; only the store *locator* (never the value) is injected, and the compute role is granted least-privilege read (`secretsmanager:GetSecretValue` / `ssm:GetParameter`, scoped to the exact ARN) + `kms:Decrypt`.
- A `domain` marker resolves at **synth time** (SDK read → inlined literal) and requires the async `Hosting.create(...)`.

### Read (runtime) — one getter per store
```ts
import { getSecret, getConfig } from '@aws-blocks/hosting'; // CDK-free subpath: '@aws-blocks/hosting/secret'
const key   = await getSecret('STRIPE_KEY');    // Secrets Manager
const flags = await getConfig('FEATURE_FLAGS');  // SSM
```
- Reads `process.env.KEY` first, so **local dev needs no AWS** (`.env`). Otherwise fetches + decrypts, cached per cold start (or per `cacheTtlSeconds` for rotation-without-redeploy).
- Each getter reads its own env prefix (`HOSTING_SECRET_PARAM_*` vs `HOSTING_CONFIG_PARAM_*`), so the store is unambiguous.

### Bring-your-own (I4)
`environment` (hosting) and `buildSecrets` (pipeline) also accept an **existing CDK handle** — managed *provisions*, BYO *references*:
```ts
import { Secret } from 'aws-cdk-lib/aws-secretsmanager';
const stripe = Secret.fromSecretNameV2(this, 'Stripe', 'prod/stripe'); // already exists
environment: { STRIPE_KEY: stripe };  // granted + wired; getSecret('STRIPE_KEY') resolves it
```
`ISecret` → the secret path, `IParameter` → the config path (the handle *is* the store).

### Pipeline
```ts
await Pipeline.create(stack, 'Pipeline', {
  source: { repo: 'org/app', connectionArn: config('CONNECTION_ARN') }, // synth-time literal
  buildSecrets: { NPM_TOKEN: secret('NPM_TOKEN') },                     // CodeBuild SECRETS_MANAGER env, fetched per build
  secretStore: { prefix: '/myapp/secrets' },
  configStore: { prefix: '/myapp/config' },
});
```
`connectionArn` accepts `secret()`/`config()`; `buildSecrets` is Secrets-Manager-only (`secret()` or a BYO `ISecret`) — build-time credentials are secrets.

### CLI (I2) — one command family per kind
```bash
# Blocks apps (pinned to /blocks/secrets and /blocks/config):
npm run secret -- set STRIPE_KEY            # → Secrets Manager  (hidden prompt; or --value-stdin)
npm run config -- set FEATURE_FLAGS '{"x":1}'  # → SSM
npm run secret -- list                      # names only, never values
npm run config -- remove FEATURE_FLAGS

# Standalone / pipeline apps: the shipped bins
npx hosting-secret set STRIPE_KEY --prefix /myapp/secrets
npx hosting-config set FEATURE_FLAGS --prefix /myapp/config
```
`set` never puts the value in `argv`/shell history (hidden prompt or `--value-stdin`); `list` prints names only.

---

## Verification

- **Unit tests green:** `@aws-blocks/hosting` 960, `@aws-blocks/core` 688, `@aws-blocks/pipeline` 89 (build + lint clean).
- **Deployed to a real account (us-east-1) and verified end-to-end.** An SSR probe route resolved both at runtime, each from its own store:
  - `secret('DEMO_SECRET')` → `getSecret()` ← **Secrets Manager** ✅
  - `config('FEATURE_FLAGS')` → `getConfig()` ← **SSM** ✅
  - Rotated a value via the CLI and confirmed the new value flowed through.
- **Pipeline synth-verified:** `connectionArn` (config) resolves to the literal ARN; `buildSecrets` (secret) emit as `SECRETS_MANAGER` CodeBuild env vars at the configured namespace.

---

## Notes / decisions

- **Runtime getters live in `@aws-blocks/hosting`** (not re-exported from `@aws-blocks/core`/`blocks`) to avoid clashing with core's existing backend `getConfig`. The declare functions `secret()`/`config()` are re-exported from `@aws-blocks/blocks/cdk` for infra.
- **Separate per-kind config props** `secretStore` / `configStore` (`{ prefix, stage, cacheTtlSeconds }`) replace the old single `secretsConfig` knob — no store selection, since the kind implies it.
- The `resolveAt:'deploy'` plaintext escape hatch was **dropped** (secure by default; a domain still resolves at synth inherently).
- Fixes a latent pipeline bug: stage-stack validation uses `cdk.Stack.isStack()` instead of `instanceof cdk.Stack`, so a consumer resolving a different `aws-cdk-lib` copy (monorepo / linked installs) isn't misdetected as "no stacks in stage".


packages/pipeline/src/pipeline-construct.ts
```diff
@@ -135,217 +125,296 @@ const DEFAULT_SYNTH_PARTIAL_BUILD_SPEC = codebuild.BuildSpec.fromObject({
...
+ * Resolve a `secret()` marker on `source.connectionArn` to a plaintext ARN at
+ * synth time, returning props with the resolved literal. A plain-string ARN is
+ * returned unchanged. Uses the shared hosting resolver (SSM/Secrets Manager),
+ * so the pipeline needs no store code of its own.
+ */
+async function resolveSourceSecrets<TConfig>(props: PipelineProps<TConfig>): Promise<PipelineProps<TConfig>> {
+	const arn = props.source.connectionArn;
+	if (!isSecret(arn)) return props;
+
+	const resolved = await resolveSecretsAtSynth([arn.key]);
```
soberm: [major] `resolveSourceSecrets` calls `resolveSecretsAtSynth([arn.key])` with no options, so `connectionArn` always resolves using the defaults: prefix `/hosting/secrets`, store `secrets-manager`. But `buildSecretEnvVars` (lines 307-308) correctly threads `props.secrets?.prefix` and `props.secrets?.store` through to the locator. If a caller sets `secrets: { store: 'ssm', prefix: '/myapp/secrets' }` and uses both a `secret()` `connectionArn` and `buildSecrets`, the two resolve from different stores and prefixes. The `connectionArn` will silently look in Secrets Manager under `/hosting/secrets/CONNECTION_ARN` no matter what they configured, and the value they set via the CLI under their chosen store/prefix is never found. `types.ts` documents the `secrets` prop as governing both `buildSecrets` and a `secret()` `connectionArn`, so this contradicts the stated contract. The test gap: the `connectionArn` test uses `_setSynthSecretFetcher(async () => MOCK_CONNECTION_ARN)`, a global override that bypasses the locator entirely, so the store/prefix path is never exercised. Fix: thread `props.secrets` the same way `buildSecretEnvVars` does -- `resolveSecretsAtSynth([arn.key], { prefix: props.secrets?.prefix, store: props.secrets?.store })` -- and add a test that asserts the locator uses the configured store and prefix.

packages/hosting/src/secret-resolve.ts
```diff
@@ -0,0 +1,291 @@
...
+export type EnvValue = string | SecretValue;
+
+/** A custom-domain name: a literal, a secret, or a mix in an array. */
+export type DomainNameInput = string | SecretValue | Array<string | SecretValue>;
+
+/** Options shared by the resolve/wire helpers. */
+export interface SecretResolveOptions {
+	/** SSM path prefix (no trailing slash). Default {@link DEFAULT_SECRET_PARAMETER_PREFIX}. */
+	prefix?: string;
+	/** Backing store. Default `'ssm'`. */
```
soberm: [minor] The JSDoc says `Default 'ssm'`, but `DEFAULT_SECRET_STORE` is `'secrets-manager'` (defined in `secret.ts` line 93) and every consumer in this file resolves `options.store ?? DEFAULT_SECRET_STORE`. The documented default is the opposite of the real one. Should be: `/** Backing store. Default {@link DEFAULT_SECRET_STORE} ('secrets-manager'). */`

test-apps/pipeline/aws-blocks/pipeline-secret.cdk.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// E2E variant: connectionArn comes from a secret() marker, resolved at synth
+// time from SSM (/hosting/secrets/CONNECTION_ARN) via Pipeline.create().
```
soberm: [minor] The comment says the `connectionArn` is "resolved at synth time from SSM (`/hosting/secrets/CONNECTION_ARN`)". With `DEFAULT_SECRET_STORE = 'secrets-manager'` and `resolveSourceSecrets` passing no store option, it actually resolves from Secrets Manager under the name `hosting/secrets/CONNECTION_ARN` (no leading slash). Both the store and the path form are wrong.

packages/core/src/secret-naming.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Blocks-owned SSM namespace for hosting secrets.
+ *
+ * The secret *mechanism* (marker, runtime resolver, path/env naming) lives in
+ * the framework-neutral `@aws-blocks/hosting` package, which defaults to a
+ * neutral `/aws-hosting/secrets` prefix so a non-Blocks consumer (a plain
```
soberm: [nit] Module JSDoc describes the default prefix as `/aws-hosting/secrets`, but `DEFAULT_SECRET_PARAMETER_PREFIX` is `/hosting/secrets`. The PR description even notes that `/aws-hosting/secrets` was a bug (SSM rejects paths starting with `aws`) that was caught and fixed, so this comment now names the exact value that was corrected. Suggested replacement: `neutral \`/hosting/secrets\` prefix so a non-Blocks consumer (a plain`

packages/core/src/hosting-secrets.ts
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Resolution glue between the dependency-free `secret()` marker (defined in
+ * `@aws-blocks/hosting`) and the `Hosting` CDK block. The marker is inert; this
+ * module is where it actually gets wired into infrastructure.
+ *
+ * Two resolution strategies, chosen by where the marker appears:
+ *
+ *   • `compute.environment` (runtime, the secure default) — we inject the SSM
+ *     parameter NAME (never the value) as `BLOCKS_SECRET_PARAM_<KEY>` and grant
```
soberm: [nit] The module JSDoc says the parameter name is injected as `BLOCKS_SECRET_PARAM_<KEY>`, but `wireRuntimeSecret` (line 149) calls `secretEnvVarName(key)`, which returns `HOSTING_SECRET_PARAM_<KEY>`. Doc-only issue; the code is correct. Update to `HOSTING_SECRET_PARAM_<KEY>`.

packages/pipeline/src/pipeline-construct.ts
osama-rizk: Fixed in 822a568. `resolveSourceSecrets` now threads `{ prefix: props.secrets?.prefix, store: props.secrets?.store }` into `resolveSecretsAtSynth`, so `connectionArn` and `buildSecrets` resolve from the same store/prefix. Added a test that captures the locator (not the value-bypass override) and asserts `ssm` + `/myapp/secrets` → `/myapp/secrets/CONNECTION_ARN`. Verified live too: with `secrets:{prefix:'/stress-next-app/secrets'}`, the deployed source action's ConnectionArn came from that namespace.

packages/hosting/src/secret-resolve.ts
osama-rizk: Fixed in 822a568 → `Default {@link DEFAULT_SECRET_STORE} ('secrets-manager')`.

test-apps/pipeline/aws-blocks/pipeline-secret.cdk.ts
osama-rizk: Fixed in 822a568 — comment now says resolved from Secrets Manager under `hosting/secrets/CONNECTION_ARN` (slash-free).

packages/core/src/secret-naming.ts
osama-rizk: Fixed in 822a568 → `/hosting/secrets`.

packages/core/src/hosting-secrets.ts
osama-rizk: Fixed in 822a568 → `HOSTING_SECRET_PARAM_<KEY>`.

packages/core/src/scripts/secret.ts
```diff
@@ -0,0 +1,142 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * `blocks secret` — manage hosting/pipeline secrets in SSM Parameter Store.
```
soberm: nit: Should this be SecretsManager and not SSM Parameter Store.

packages/core/src/scripts/secret.ts
```diff
@@ -0,0 +1,142 @@
...
+	}
+}
+
+/** Set (create or overwrite) a secret value. */
+export async function setSecret(key: string, value: string): Promise<void> {
+	assertValidKey(key);
+	if (value === undefined || value === null) {
+		throw new Error(`No value provided for secret '${key}'.`);
+	}
+	const { SSMClient, PutParameterCommand } = await import('@aws-sdk/client-ssm');
```
soberm: The PR description mentions Secrets Manager by default and SSM is opt-in. I only see SSM here.

packages/core/src/scripts/secret.ts
osama-rizk: Intentional — this is the **Blocks**-namespaced CLI, which is deliberately distinct from the framework-neutral `@aws-blocks/hosting` CLI that standalone-hosting and pipeline apps use. Only the neutral leaf defaults to Secrets Manager (per the AWS credentials guidance); this Blocks path stays pinned to SSM SecureString on the pre-existing `/blocks/secrets` namespace (alongside `/blocks/{stage}/db-connection-string`), so a Blocks app's secret store doesn't silently change shape. A Blocks app that wants Secrets Manager / per-stage / store control uses the leaf CLI directly. Clarified the docstring to say this explicitly in 5380cfa (also noting pipeline uses the leaf, not this CLI).

packages/core/src/scripts/secret.ts
osama-rizk: The "Secrets Manager by default, SSM opt-in" contract is the **leaf** (`@aws-blocks/hosting`) CLI's — that's what standalone-hosting and pipeline apps use, and its default flipped to SM in an earlier commit. This file is the separate Blocks-namespaced CLI, intentionally SSM-only on `/blocks/secrets` (unchanged shape for existing Blocks apps). So no missing SM path here — it's a different, deliberate consumer. Docstring updated in 5380cfa to make the split explicit.

packages/pipeline/src/pipeline.test.ts
```diff
@@ -26,1792 +28,2225 @@ const MOCK_CONNECTION_ARN = 'arn:aws:codeconnections:us-east-1:123456789012:conn
...
+				stages: [{ name: 'beta' }, { name: 'prod', requireApproval: true }],
+			},
+		],
+		stageFactory: minimalStageFactory,
+		...overrides,
+	};
 }
 
 describe('Pipeline', () => {
-  describe('basic pipeline creation', () => {
```
soberm, line 53: question: why do we have so many additions and deletions in this file?

packages/pipeline/src/pipeline-construct.ts
```diff
@@ -1,27 +1,18 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { pathToFileURL } from 'node:url';
```
soberm, line 4: question: why did this file change so much?

packages/pipeline/src/types.ts
```diff
@@ -13,52 +14,58 @@ import type { IFileSetProducer } from 'aws-cdk-lib/pipelines';
  * connection is created once via the AWS Console.
  */
 export interface PipelineSourceConfig {
```
soberm, line 18: same here. it seems that almost the whole file was re-written.

osama-rizk: Good catch — that was noise, now removed in 3cf3c2a. The pipeline package is space-indented on `main`, but biome's format-on-save reindented it to tabs (the repo's format standard) when I edited it, so the whole file showed as changed. CI only runs `biome lint` (not the formatter), so none of that reflow was required. I reverted this file to `main`'s formatting and re-applied only the real change — the diff is now **+45/-1**: the one changed line is `connectionArn: string` → `string | SecretValue`; everything else is the two new optional props (`buildSecrets`, `secrets`) + their JSDoc.

packages/pipeline/src/pipeline-construct.ts
osama-rizk: Same root cause as `types.ts` — format-on-save reindented spaces→tabs and re-sorted/-wrapped pre-existing code, none of which CI requires (it lints, doesn't format). Reverted to `main`'s formatting and re-applied only the feature in 3cf3c2a. The diff is now **+92/-5**: `resolveSourceSecrets()` (synth-time connectionArn resolution, threading the configured store/prefix — the earlier `[major]` fix), `buildSecretEnvVars()` (build-time CodeBuild env vars), the sync-path secret rejection in `validateProps`, and the env-var injection into the synth CodeBuild defaults. The 5 deletions are exactly the lines those changes replace.

packages/pipeline/src/pipeline.test.ts
osama-rizk: Same cause, now fixed in 3cf3c2a — the additions/deletions were format-on-save (import re-sort + re-wrapping existing `assert.throws` blocks), not logic. Reverted to `main`'s formatting and re-applied only the new tests, so the diff is now **purely additive (+138/-0)**: two new imports (`_setSynthSecretFetcher`, `secret`) and two new describe blocks — `connectionArn secret()` (3 tests: sync rejection, synth-time resolution, and the store/prefix locator assertion for the `[major]` fix) and `buildSecrets` (4 tests: SM default, SSM opt-in, non-marker rejection, empty case). 89 tests pass.

soberm (dismissed):

packages/hosting/src/secret-resolve.ts
```diff
@@ -0,0 +1,291 @@
...
+	// if the stage-specific value is not found.
+	fn.addEnvironment(secretEnvVarName(key), primary);
+	if (store !== 'ssm') fn.addEnvironment(`${secretEnvVarName(key)}_STORE`, store);
+	if (fallback) fn.addEnvironment(secretFallbackEnvVarName(key), fallback);
+
+	// Grant read+decrypt on every locator the runtime might read (primary +
+	// fallback), deduped.
+	const locators = [...new Set([primary, ...(fallback ? [fallback] : [])])];
+	for (const locator of locators) {
+		grantSecretRead(fn, locator, store);
```
bobbor: **🔴 Blocking: cross-stage IAM over-grant on the shared fallback**

This loop grants the compute role read on **both** the stage locator and the shared fallback `<prefix>/<key>`, unconditionally. The two-try fallback is only enforced in application logic; the IAM permission is standing. That means every ephemeral PR/preview deploy holds IAM read access to the shared fallback secret at all times, and a compromised handler in a preview stage can call `GetSecretValue` on the fallback ARN directly, bypassing the two-try logic.

The docs example (`sk_test_shared` as fallback) is safe, but nothing stops or warns the inverse (someone putting a prod value in the shared slot and letting previews "inherit" it).

Suggestions: make the fallback grant opt-in per stage, or document loudly that the shared fallback must never hold production values (safe dev defaults only).

packages/hosting/src/secret-resolve.ts
```diff
@@ -0,0 +1,291 @@
...
+	}
+}
+
+/** Grant a Lambda role least-priv read+decrypt on one secret locator. */
+function grantSecretRead(fn: cdk.aws_lambda.Function, locator: string, store: SecretStore): void {
```
bobbor: **🔴 Blocking: duplicate `kms:Decrypt` statements per secret**

`grantSecretRead` is called once per locator, and each call appends its own `kms:Decrypt` statement (`Resource: '*'` + `kms:ViaService` condition). CDK doesn't deduplicate `addToRolePolicy`, so N secrets with stage fallbacks means up to 2N identical `kms:Decrypt` statements. Beyond the noise, this bloats the inline policy toward the 10KB limit for apps with several secrets.

Suggestion: hoist the KMS grant out of the per-locator path and add it once per store type after the read grants.

(The `Resource: '*'` + `ViaService` pattern itself is fine 👍 that's the documented approach for service-default keys and matches the existing `AppSetting` code.)

packages/hosting/src/secret-cli.ts
```diff
@@ -0,0 +1,240 @@
...
+	// Pull an optional `--stage <name>` (or `--stage=<name>`) out of argv; a
+	// CLI-supplied stage overrides any preset on `opts`. Everything else is
+	// positional, so set/list/remove parsing below stays unchanged.
+	const { stage, positional } = extractStageFlag(argv);
+	const effectiveOpts: SecretCliOptions = stage !== undefined ? { ...opts, stage } : opts;
+	const [subcommand, ...rest] = positional;
+	switch (subcommand) {
+		case 'set': {
+			const [key, ...valueParts] = rest;
+			const value = valueParts.join(' ');
```
bobbor: **🟠 Secret values land in argv and shell history**

`secret set STRIPE_KEY sk_live_…` puts the value into `process.argv`, which makes it visible in `ps` output, `/proc/*/cmdline`, and shell history files. For a tool whose whole pitch is "secrets never end up in git", this is a bit ironic 😅 and it's exactly the kind of footgun the target audience (external devs) will hit.

Suggestion: support `--value-stdin` or an interactive prompt, or at minimum document the exposure prominently in the CLI help.

packages/hosting/src/secret.ts
```diff
@@ -0,0 +1,258 @@
...
+ * when a consumer does not supply its own. A branded consumer overrides this by
+ * passing an explicit `prefix` to {@link secretParameterName} (e.g. Blocks
+ * passes `/blocks/secrets`), so this package never hardcodes a framework's
+ * namespace into the shared upstream.
+ *
+ * NOTE: must NOT begin with `aws` or `ssm` — SSM Parameter Store reserves
+ * those prefixes and rejects create/read with "No access to reserved parameter
+ * name". (An earlier `/aws-hosting/...` default hit exactly that.)
+ */
+export const DEFAULT_SECRET_PARAMETER_PREFIX = '/hosting/secrets';
```
bobbor, line 172: **🟠 Default prefix is account-global**

Two standalone apps in the same account using this default share one namespace: same key name means literally the same secret, and `secret list` enumerates entries across apps. That's a realistic scenario for the target audience (one personal account, several apps).

Suggestion: derive the default from the app/stack name, or document clearly that multi-app accounts must set a unique `prefix`.

packages/hosting/src/secret-resolve.ts
```diff
@@ -0,0 +1,291 @@
...
+	exposeSecrets: SecretValue[];
+} {
+	const plain: Record<string, string> = {};
+	const runtimeSecrets: SecretValue[] = [];
+	const exposeSecrets: SecretValue[] = [];
+
+	for (const [key, value] of Object.entries(environment ?? {})) {
+		if (isSecret(value)) {
+			if (value.key !== key) {
+				throw new Error(
```
bobbor: **🟠 Convention: plain `Error` vs `HostingError`**

`hosting_construct.ts` uses structured `HostingError` (`{code, message, resolution}`) exclusively, 14 instances and zero plain throws, but the throws in this file (`partitionEnvironment`, `resolveSecretsAtSynth`, `resolveDomainNames`) are plain `new Error(...)`. The `exposeSecrets` path in the construct got it right with `HostingError('UnresolvedSecretError', …)`. Consumers lose the error-code contract for these paths.

packages/pipeline/package.json
```diff
@@ -24,6 +24,9 @@
     "test": "node --test dist/*.test.js",
     "dev": "tsc --build --watch"
   },
+  "dependencies": {
+    "@aws-blocks/hosting": "*"
```
bobbor: **🟠 Bare `"*"` dependency on a published package**

Pipeline previously had no runtime deps. In the workspace this resolves locally, but in the **published** package `"*"` accepts any hosting version whatsoever, which will break consumers on the first incompatible hosting release. Suggest pinning to a proper semver range.

packages/hosting/src/secret-cli.ts
```diff
@@ -0,0 +1,240 @@
...
+	if (store === 'secrets-manager') {
+		// SM names are the slash-free locator form (`hosting/secrets/<KEY>`).
+		const smPrefix = prefix.replace(/^\//, '');
+		const { SecretsManagerClient, ListSecretsCommand } = await import('@aws-sdk/client-secrets-manager');
+		const client = new SecretsManagerClient({});
+		const keys: string[] = [];
+		let nextToken: string | undefined;
+		do {
+			const result = await client.send(
+				new ListSecretsCommand({
```
bobbor: **🟠 SM `list` leaks stage-scoped entries into the shared listing**

The SSM path uses `Recursive: false` to keep the listing flat, but this Secrets Manager name filter is a plain prefix match with no depth control, so stage-scoped secrets show up in the shared-namespace listing as `prod/KEY` etc.

Suggestion: after slicing the prefix, filter out names containing `/` (or group them by stage, which would actually be a nice UX).

packages/core/src/hosting-secrets.ts
```diff
@@ -0,0 +1,193 @@
...
+		return value;
+	});
+	return Array.isArray(domainName) ? arr : arr[0];
+}
+
+/**
+ * Inject the SSM parameter NAME (not the value) for a runtime secret and grant
+ * the compute role read+decrypt access scoped to that one parameter.
+ */
+export function wireRuntimeSecret(fn: cdk.aws_lambda.Function, key: string): void {
```
bobbor: **🟠 SSM-only duplicate of the hosting resolve engine**

This is a reimplementation of `hosting/src/secret-resolve.ts`'s engine, hardcoded to SSM (no `_STORE` hint, no stage/fallback, SSM-only IAM). That's consistent today because core pins the Blocks SSM namespace, but there's no comment or assertion saying "SSM-only by design", so the two engines can silently diverge (e.g. if someone later threads a store option through core, they'd get SSM grants with a Secrets Manager runtime read → AccessDenied).

Suggestion: a one-line comment/assertion making the SSM-only contract explicit, or share the engine.

packages/hosting/src/secret-runtime.ts
```diff
@@ -0,0 +1,140 @@
...
+import { secretEnvVarName, secretFallbackEnvVarName } from './secret.js';
+
+/** True if a store error means "the secret/parameter does not exist". */
+function isNotFoundError(error: unknown): boolean {
+	const name = (error as { name?: string })?.name;
+	return name === 'ParameterNotFound' || name === 'ResourceNotFoundException';
+}
+
+/** Cache of resolved secret values, keyed by logical secret name. */
+const cache = new Map<string, string>();
```
bobbor: **🟡 No TTL: rotation only lands on cold start**

The docs say rotated values are picked up "on the next cold start", which is accurate, but with provisioned concurrency or steady traffic a warm sandbox can serve a stale value for hours. Not blocking, but a configurable TTL (even a generous default like 5-15 min) or a prominent note in the rotation docs would save someone a confusing incident.

packages/hosting/package.json
```diff
@@ -26,6 +26,10 @@
     "./error": {
       "types": "./dist/hosting_error.d.ts",
       "default": "./dist/hosting_error.js"
+    },
+    "./secret": {
```
bobbor: **🟡 Missing subpath export test**

`sub_path_exports.test.ts` covers `./constructs`, `./adapters`, and `./error` but wasn't updated for this new `./secret` entry. Cheap test, catches broken published artifacts.

bobbor (dismissed):
Hey @osama-rizk, really impressive work here 🙌 The PR description alone deserves a shout-out, and the design has a lot of genuinely great calls: the shared `secretStoreLocator()` helper (love that it was born from a live-caught bug), the branded-symbol marker, the coalescing cache in `getSecret()`, the fail-fast on unresolved `exposeAsEnv`, and the CodeBuild-native `buildSecrets` wiring. Test coverage actually asserts the security-relevant behavior too, not just the happy paths 👏

That said, I dug in pretty deep and found a couple of things I'd consider blocking, plus some smaller ones. Details are in the inline comments; TL;DR:

**🔴 Blocking**
1. Cross-stage IAM over-grant: preview/PR stages get standing IAM read on the shared fallback secret (see `secret-resolve.ts` comment)
2. Duplicate `kms:Decrypt` statements per secret: policy bloat that can approach the 10KB inline policy limit

**🟠 Worth fixing in this PR**
3. Secret values land in argv / shell history via `secret set KEY value`
4. Default prefix `/hosting/secrets` is account-global, so two apps in one account collide
5. Plain `Error` throws in `secret-resolve.ts` while the rest of hosting uses structured `HostingError`
6. Pipeline's new dep on `"@aws-blocks/hosting": "*"` resolves to any version once published
7. Secrets Manager `list` leaks stage-scoped entries into the shared-namespace listing
8. core's `hosting-secrets.ts` duplicates the resolve engine as an SSM-only variant with no guard against divergence

**🟡 Minor / fast-follow**
- No cache TTL in `getSecret()`: rotation only lands on cold start, which can be hours with provisioned concurrency
- `sub_path_exports.test.ts` wasn't updated for the new `./secret` entry
- No README/docs updates despite the new public API and CLI workflow
- Template `secret.ts` files use 2-space indent while `biome.json` mandates tabs; also the `hosting/src/index.ts` reformat of existing exports inflates the diff a bit
- CLI region mismatch: you already called it out as a follow-up, a `--region` flag would be cheap insurance
- Worth confirming `API.md` came from api-extractor and not a hand edit

One thing I checked so nobody else trips on it: `requestHandler: { requestTimeout: 5000 }` is totally fine 👍 the plain-object short form is officially supported since SDK v3.521.0, so no change needed there.

Overall: really solid feature, and the two blocking items are fixable without touching the architecture. Happy to re-review quickly once they're addressed 🚀

packages/hosting/src/secret-resolve.ts
osama-rizk: Fixed in b7458c4. Hoisted the `kms:Decrypt` grant out of the per-locator path into `grantKmsDecrypt(fn, store)`, which emits it **once per (compute function, store)** via a WeakMap dedupe — the statement is identical for every secret in the same store (`Resource:*` + `kms:ViaService`), so per-locator copies were pure duplicates. Synth-verified: two stage-scoped secrets now produce **1** Decrypt statement (was 4). Added unit tests asserting the count for the stage-fallback case and the multi-secret case. (Kept the `Resource:*` + ViaService pattern as-is per your 👍.)

packages/hosting/src/secret-resolve.ts
osama-rizk: Addressed in b7458c4 (documented the trust model, per your second suggestion). The read grant on the shared ARN is required for the two-try fallback to function, so removing it would break the feature — instead I documented the boundary loudly on `wireRuntimeSecret` and in the changeset: the shared-fallback grant is **standing IAM, not enforced by the two-try logic**, so the shared slot must hold a **safe default for all stages** (e.g. a sandbox credential), never a production secret — give prod its own stage-scoped value (`secret set KEY <val> --stage prod`) so it's unreachable from a preview role. Happy to add an opt-in `fallbackGrant: false` toggle in a follow-up if you'd prefer the mechanism over the docs.

packages/hosting/src/secret-resolve.ts
osama-rizk: Fixed in b7458c4. `partitionEnvironment`, `resolveSecretsAtSynth` (both not-found paths), and `resolveDomainNames` now throw structured `HostingError` (`SecretKeyMismatchError`, `UnresolvedSecretError`) with `message`/`resolution`, matching the construct. Kept the existing message text so the error-code contract is added without breaking the message-substring assertions.

packages/pipeline/package.json
osama-rizk: Fixed in b7458c4 → `"@aws-blocks/hosting": "^0.1.6"` (matches the `^0.1.x` convention the bb-* packages use for their `@aws-blocks/core` dep). Lockfile updated (single-line change); workspaces still resolve locally.

packages/hosting/src/secret-cli.ts
osama-rizk: Fixed in b7458c4. When listing the shared namespace (no `--stage`), entries with a remaining `/` after slicing the prefix are stage-scoped (`<prefix>/<stage>/<key>`) and are now filtered out — mirroring the SSM path's `Recursive:false`. A stage listing already has the stage in the prefix, so its own keys still show.

packages/core/src/hosting-secrets.ts
osama-rizk: Documented in b7458c4 — added a prominent "⚠️ SSM-only by design" note to the `core/hosting-secrets.ts` module doc: no `store` option / no `_STORE` hint / SSM-only IAM, correct *because* core pins the `/blocks/secrets` SSM namespace and emits no hint, with an explicit warning that threading a store option here without also switching the grant + runtime read yields `AccessDenied` — and to adopt the shared `@aws-blocks/hosting` engine rather than parameterize this one if a second store is ever needed. (The two-solution split is intentional — confirmed with the maintainer.)

packages/hosting/package.json
osama-rizk: Fixed in b7458c4 — added `resolves ./secret` to `sub_path_exports.test.ts`. Passes.

packages/hosting/src/secret-cli.ts
osama-rizk: Fixed in cce5070. `secret set` now takes `--value-stdin` (`cat key.txt | … set KEY --value-stdin`) or, with no value on the command line, prompts interactively with the input **hidden** (no echo, muted readline). A positional value still works for trusted CI but is documented as the exposed path, and passing a value both ways is rejected. Documented the argv/history exposure prominently in the CLI module doc. Tests cover the precedence + the both-ways rejection + the non-TTY hint.

packages/hosting/src/secret-runtime.ts
osama-rizk: Fixed in cce5070. Added an optional `secrets.cacheTtlSeconds` that injects `HOSTING_SECRET_CACHE_TTL`; the runtime resolver then re-fetches after the TTL, so a warm compute (provisioned concurrency / steady traffic) picks up a rotated value without a cold start. Default is unchanged — cache for the process lifetime — so it's fully opt-in. (env-var-sourced values from local dev / exposeAsEnv are always cached forever since they can't rotate under a running process.) Synth-verified the env var is injected only when configured; unit tests cover TTL expiry/refetch and the no-TTL default. Also called it out in the rotation section of the changelog.

packages/hosting/src/secret.ts
osama-rizk: Good point on the multi-app collision. I looked at deriving the default from the stack name but it introduces a worse footgun: the `secret` CLI writes **out of band and has no Stack/CDK context**, so it can't see a stack-derived prefix. If the construct silently auto-derived `/hosting/<stackName>/secrets` while the CLI kept writing the neutral default, every `secret set` would land in the wrong place and the deploy would report "secret not set" — and it'd also be a breaking change to the current default locator. The safe version (a shared `hostingSecretPrefix(appName)` helper the template wires into **both** the construct `secrets:{prefix}` and the CLI wrapper so they can't drift) is a template/DX change rather than a leaf-behavior change. Deferring it to a follow-up; for now the neutral default stays and I'll make sure the docs say multi-app accounts must set a unique `prefix`. Does that split sound right to you?

packages/core/src/secret-naming.ts
```diff
@@ -0,0 +1,44 @@
...
+/** Blocks prefix for `secret()` values (Secrets Manager). */
+export const BLOCKS_SECRET_PARAMETER_PREFIX = '/blocks/secrets';
+
+/** Blocks prefix for `config()` values (SSM Parameter Store). */
+export const BLOCKS_CONFIG_PARAMETER_PREFIX = '/blocks/config';
+
+/** Blocks-namespaced name for a secret key. @example '/blocks/secrets/STRIPE_KEY' */
+export function blocksSecretParameterName(key: string): string {
+	return parameterName(key, BLOCKS_SECRET_PARAMETER_PREFIX);
+}
```
soberm, line 76: [blocker] This returns `/blocks/secrets/STRIPE_KEY` (leading slash, via `parameterName()`), but the actual Secrets Manager name used everywhere else (CLI writes, IAM grants, runtime locator) goes through `secretStoreLocator()`, which strips the leading slash for the `secrets-manager` store (`blocks/secrets/STRIPE_KEY`). This is exported public API (re-exported from `index.cdk.ts`) named like it should be "the secret's name", but it doesn't match what's actually stored. Anyone building an ARN or cross-checking against `secretStoreLocator` off this will get the wrong path. Worth routing this through `secretStoreLocator(key, { prefix, store: 'secrets-manager' })` instead, or renaming/documenting it as something other than the real Secrets Manager name.

packages/hosting/src/secret-resolve.ts
```diff
@@ -0,0 +1,328 @@
...
+ *
+ * **Trust boundary (stage fallback).** When a `stage` is configured the runtime
+ * tries `<prefix>/<stage>/<key>` first and falls back to the shared
+ * `<prefix>/<key>`. The IAM grant here is STATIC, so the compute role gets
+ * standing read on BOTH locators — not just the one a given request resolves. A
+ * stage's compute can therefore always read the shared value (that is what makes
+ * the fallback work), so treat the shared entry as readable by every stage that
+ * shares this prefix and put stage-private values under the stage locator only.
+ */
+export function wireManagedValue(fn: cdk.aws_lambda.Function, marker: ManagedValue, cfg: StoreConfig = {}): void {
```
soberm, line 323: [blocker] The doc above explains that with a `stage` configured, the IAM grant here is static and gives standing read on BOTH the stage locator and the shared `<prefix>/<key>` fallback — so any one stage's compute can always read the shared value. That's a reasonable tradeoff, but it means "stage" isn't a security boundary here, which is easy to miss if a team assumes per-stage `Hosting` constructs sharing a prefix are isolated from each other. Given that's paired with `ForceDeleteWithoutRecovery: true` on removal (no undo), I'd push this out of a JSDoc comment and into the README/HostingProps docs for `secretStore`/`configStore`/`stage` so it's visible where someone would actually configure it.

packages/hosting/src/secret-cli.ts
```diff
@@ -0,0 +1,313 @@
...
+	const store: SecretStore = storeForKind(kind);
+	const prefix = opts.prefix ?? defaultPrefixForKind(kind);
+	const name = secretStoreLocator(key, { prefix, store, stage: opts.stage });
+	const clientConfig = opts.region ? { region: opts.region } : {};
+
+	if (store === 'secrets-manager') {
+		const { SecretsManagerClient, DeleteSecretCommand } = await import('@aws-sdk/client-secrets-manager');
+		const client = new SecretsManagerClient(clientConfig);
+		try {
+			await client.send(new DeleteSecretCommand({ SecretId: name, ForceDeleteWithoutRecovery: true }));
```
soberm: [minor] `ForceDeleteWithoutRecovery: true` means `secret remove` permanently destroys the value immediately, no recovery window. A typo'd key in prod is unrecoverable. Might be safer to default to the normal recovery window and add a `--force` flag for the immediate-delete case.

packages/hosting/src/secret-runtime.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/** Shared resolver for both kinds. */
+async function resolveValue(kind: ValueKind, key: string): Promise<string> {
+	const cacheKey = `${kind}:${key}`;
+	const cached = cache.get(cacheKey);
+	if (cached !== undefined && Date.now() < cached.expiresAt) return cached.value;
+
+	// 1. Plaintext already in env (local dev).
+	const direct = process.env[key];
```
soberm, line 126: [minor] Both `getSecret('K')` and `getConfig('K')` check `process.env.K` (not kind-prefixed) first, so if you declare both `secret('K')` and `config('K')`, a single `.env` entry satisfies both getters identically in local dev. The "secret and config namespaces are independent" guarantee (tested below for the deployed-store path) doesn't hold for local dev. Probably fine as documented behavior, but worth a callout since it's the kind of thing that only bites when someone's debugging a local/prod mismatch.

packages/hosting/src/secret-cli.ts
```diff
@@ -0,0 +1,313 @@
...
+
+	if (store === 'secrets-manager') {
+		const smPrefix = scoped.replace(/^\//, '');
+		const { SecretsManagerClient, ListSecretsCommand } = await import('@aws-sdk/client-secrets-manager');
+		const client = new SecretsManagerClient(clientConfig);
+		const keys: string[] = [];
+		let nextToken: string | undefined;
+		do {
+			const result = await client.send(
+				new ListSecretsCommand({ Filters: [{ Key: 'name', Values: [`${smPrefix}/`] }], NextToken: nextToken }),
```
soberm, line 111: [nit] `ListSecretsCommand` prefix-filters then re-checks with `startsWith` — good defensive check, but this is an unbounded paginated `ListSecrets` call which is a relatively expensive/throttled API. Not urgent, just flagging for when `list` gets used against a large `/blocks/secrets` tree.

packages/pipeline/src/pipeline-construct.ts
```diff
@@ -265,6 +267,64 @@ export class Pipeline<TConfig = Record<string, unknown>> extends Construct {
...
+
+  const envVars: Record<string, codebuild.BuildEnvironmentVariable> = {};
+  for (const [name, marker] of Object.entries(buildSecrets)) {
+    if (isManagedValue(marker)) {
+      if (marker.kind !== 'secret') {
+        throw new Error(`Pipeline: buildSecrets['${name}'] must be a secret(...) — build credentials are secrets.`);
+      }
+      // secret marker → Secrets Manager locator (slash-free) at build time.
+      envVars[name] = { type, value: secretStoreLocator(marker.key, { prefix, store: 'secrets-manager' }) };
+    } else if (typeof marker === 'object' && marker !== null && 'secretName' in marker) {
```
soberm: [nit] The BYO branch here just checks `'secretName' in marker`, looser than the `isCdkSecret`/`isCdkParameter` duck-typing helpers in `secret-resolve.ts` (which also check `grantRead`). Might be worth reusing those helpers here for consistency, low risk given the types but easy win.

packages/hosting/src/secret-resolve.ts
```diff
@@ -0,0 +1,328 @@
...
+		fn.addEnvironment(cacheTtlEnvVarName(kind), String(Math.floor(cacheTtlSeconds)));
+	}
+}
+
+function grantStoreRead(fn: cdk.aws_lambda.Function, locator: string, store: SecretStore): void {
+	if (store === 'secrets-manager') {
+		const secretArn = cdk.Stack.of(fn).formatArn({
+			service: 'secretsmanager',
+			resource: 'secret',
+			resourceName: `${locator}-??????`,
```
soberm, line 372: [nit] The `-??????` suffix hack for the Secrets Manager ARN works today but is a bit fragile if AWS ever changes the random-suffix length/format. Not asking for a change, just noting it as a spot to watch.

packages/pipeline/src/pipeline-construct.ts
```diff
@@ -391,20 +459,25 @@ function buildCodePipeline<TConfig>(
 ): CodePipeline {
   const hasTriggerFilters = props.source.triggerFilters && props.source.triggerFilters.length > 0;
   if (branchConfig.triggerOnPush === true && hasTriggerFilters) {
-    throw new Error(`Pipeline branch '${branchConfig.branch}': triggerOnPush cannot be true when triggerFilters are set. CodePipeline uses filters instead of push triggers when filters are configured.`);
+    throw new Error(
+      `Pipeline branch '${branchConfig.branch}': triggerOnPush cannot be true when triggerFilters are set. CodePipeline uses filters instead of push triggers when filters are configured.`,
+    );
+  }
+  const triggerOnPush = hasTriggerFilters ? false : (branchConfig.triggerOnPush ?? props.source.triggerOnPush ?? true);
```
soberm, line 482: [nit] A good chunk of this file's diff is pure reformatting (ternaries collapsed to one line, error strings rewrapped, import sorting) unrelated to the secret/config feature — makes the actual functional changes (`resolveSourceSecrets`, `buildSecretEnvVars`, the `isManagedValue` check below) harder to spot in review. Might be worth keeping formatter-only changes in a separate commit next time.

soberm (changes requested):
This is a big, well-structured change (secret()/config() markers, synth-time + runtime resolution, CLI, BYO handle support, plus the isStack() cross-copy fix bundled in). Design and test coverage are solid overall. Left a couple of blockers around a naming/value mismatch in a public helper and the stage-fallback trust boundary needing more visibility, plus some smaller suggestions. Requesting changes mainly for the two blockers below — rest is take-it-or-leave-it.

packages/core/src/secret-naming.ts
osama-rizk: Good catch — that was a real public-API mismatch. `blocksSecretParameterName` / `blocksConfigParameterName` now delegate to `secretStoreLocator(key, { prefix, store })`, so they return exactly what the CLI writes, the IAM grant scopes to, and the runtime reads: `blocks/secrets/K` (slash-free, Secrets Manager) and `/blocks/config/K` (SSM). Test updated to assert equality against `secretStoreLocator` rather than the old `/blocks/secrets/K`. Fixed in bc55b267.

packages/hosting/src/secret-resolve.ts
osama-rizk: Agreed — moved it out of the JSDoc-only spot. It is now on `KindStoreOptions.stage` (so it surfaces on `HostingProps.secretStore`/`configStore` at the configuration site) and in the hosting README footguns: `stage` is not a security boundary — the grant is static over both the stage and shared locators, so keep prod-only secrets stage-scoped and put only a safe cross-stage default in the shared slot. Paired with the recoverable-delete change below. bc55b267.

packages/hosting/src/secret-cli.ts
osama-rizk: Done — `remove` on a secret now defaults to Secrets Manager's recovery window (restorable), and `--force` opts into immediate no-recovery delete. `config` removes stay immediate since SSM Parameter Store has no recovery window. bc55b267.

packages/hosting/src/secret-runtime.ts
osama-rizk: Documented as behavior in `secret-runtime.ts`: the env-first local path keys off the bare `process.env.K` (not kind-prefixed), so one `.env` entry satisfies both `getSecret('K')` and `getConfig('K')`; the per-kind independence only holds on the deployed store path. Added a note that reusing the same key for a secret and a config is best avoided. bc55b267.

packages/hosting/src/secret-cli.ts
osama-rizk: Agreed — `ListSecrets` is the paginated/throttled call; fine at today's `/blocks/secrets` sizes, and noted for if `list` is ever run against a large tree. Left as-is.

packages/pipeline/src/pipeline-construct.ts
osama-rizk: Done — exported `isCdkSecret`/`isCdkParameter` from the resolve engine; the pipeline BYO branch now uses `isCdkSecret(marker)` instead of the looser `'secretName' in marker`, and the now-unused `ISecret` import is dropped. bc55b267.

packages/hosting/src/secret-resolve.ts
osama-rizk: Noted — the `-??????` suffix is Secrets Manager's documented ARN wildcard pattern; keeping it, and watching in case AWS ever changes the suffix length/format.

packages/pipeline/src/pipeline-construct.ts
osama-rizk: Fair. I de-noised the bulk of the pipeline diff earlier (reverted to main's spacing so the real change shows), but some formatter churn remains; I'll keep formatter-only changes in a separate commit going forward.

soberm (dismissed):

pranavosu (dismissed):

soberm (approved):

PR #118, opened 2026-06-26
test(bb-realtime): confirm schema forward-compat for namespace evolution

svidgen:
## Summary

Investigates and closes the reported concern that additive namespace schema changes break already-connected clients.

**Finding:** The framework already handles this gracefully by design:
- Schema validation runs only on `publish()` (server-side) — there is no receive-side validation
- The raw deserialized payload is passed directly to user handlers — extra object fields are simply extra properties (standard JS behavior)
- Over WebSocket, handler errors are wrapped in try/catch — a handler that throws on unexpected fields does NOT kill the subscription

## What's in this PR

A new test file (`schema-compat.test.ts`) with 5 tests that document and verify this behavior:
1. Extra fields pass through without framework error
2. Handler destructuring only old fields still works fine
3. Various payload shapes (objects, strings, numbers) pass through intact
4. Over WebSocket: handler throwing on unexpected shape does not kill the subscription
5. Subsequent messages still delivered after a handler error

## What was tested

All 67 tests pass (62 existing + 5 new).

## Related

Closes R7 reports (3) about Realtime envelope schema evolution concerns.

### DIFF
--- a/.changeset/realtime-schema-forward-compat.md
+++ b/.changeset/realtime-schema-forward-compat.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add schema forward-compatibility tests for bb-realtime namespace evolution
--- a/.changeset/realtime-schema-forward-compat.md
+++ b/.changeset/realtime-schema-forward-compat.md
@@ -1,4 +1,5 @@
 ---
+"@aws-blocks/bb-realtime": patch
 ---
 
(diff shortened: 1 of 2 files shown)


svidgen:
Notably not yet convinced this test actually adds value; just looking to solidify expected subscription behavior when schemas change after a deployment.

Issue #98, opened 2026-06-23
PGlite .bb-data corruption on tsx-watch restart mid-init: next boot aborts and the dev-server port never binds


hfurkanbozkurt:
## Summary

When `tsx watch` SIGTERMs the dev server **while a `Database` block is still initializing**, the PGlite data directory under `$WORKSPACE/.bb-data/<fullId>` can be left half-written. On the next (re)boot, `new PGlite(dataDir)` aborts on the corrupt directory, the local-deploy phase rejects **before** the HTTP server ever calls `listen()`, and the dev-server port never binds — the app is permanently unreachable until `.bb-data` is manually deleted.

This is a sibling of #78: both are triggered by the same `tsx watch` restart, but #78 orphans the **frontend** port (502 / EADDRINUSE) whereas this one corrupts the **PGlite backend** state so the backend port never comes up at all.

## Framework files / behavior

- `packages/bb-data/src/engines/pglite-engine.ts`
  - The constructor does `mkdirSync(dataDir, { recursive: true })` → `cleanStaleLock(dataDir)` → `this.db = new PGlite(dataDir)`.
  - `cleanStaleLock()` only removes a stale `postmaster.pid`. It does **not** repair a half-written data directory. (The code comment already notes a leftover pid file makes PGlite "crash with `Aborted()`"; a partially-initialized data dir aborts the same way and is not handled.)
- `packages/bb-data/src/index.mock.ts` (L62): local `Database` uses `new PGliteEngine(`.bb-data/${this.fullId}`)`, so each DB persists a per-`fullId` directory across restarts.
- `packages/core/src/scripts/dev-server.ts`
  - `deployLocal(backend)` (L93) pushes every block's `initialize()` and `await Promise.all(initPromises)` (L101).
  - It is `await`ed during startup at **L176**, *before* `server.listen(port, '127.0.0.1', …)` at **L322**. So if any block's init throws (here: PGlite aborting on a corrupt dir), `listen()` is never reached and the port never binds.
  - `cleanup()` (L364) kills the frontend child on `SIGTERM`/`SIGINT` (L374–375) but does **not** ensure a block mid-`initialize()` reaches a consistent on-disk state before the process dies.

## Repro

1. Start the dev server on a template that uses a `Database` (e.g. `nextjs` / `react`) with an empty `.bb-data`.
[...]

PR #119, opened 2026-06-27
fix: recover incomplete PGlite data dirs

ikenyal:
## Problem

Issue #98 describes a local dev failure mode where `tsx watch` or another restart interrupts PGlite during first-boot initialization. That can leave the local `.bb-data/<fullId>` PostgreSQL directory partially written. On the next boot, `new PGlite(dataDir)` may abort before the dev server reaches `listen()`, so the app stays unreachable until the directory is manually moved or deleted.

## Changes

- Detect incomplete local PGlite data directories before opening PGlite.
- Treat a directory as initialized only after `PG_VERSION`, `base`, `global`, and `global/pg_control` exist.
- Recover partial PGlite leaf directories even when initialization stopped before `PG_VERSION` was written.
- Avoid quarantining parent directories that contain already-initialized PGlite child data directories.
- Move incomplete data directories to a `.corrupt-<timestamp>-<pid>-<suffix>` sidecar instead of deleting them.
- Keep empty directories and already-initialized data directories untouched.
- Add regression tests for partial init directories, including `PG_VERSION`-only, empty `base`/`global`, and pre-`PG_VERSION` leaf states.
- Add a patch changeset for `@aws-blocks/bb-data`.

This is intentionally narrower than deleting any non-empty directory: it only targets leaf directories with PGlite/PostgreSQL markers such as `PG_VERSION`, `base`, `global`, `pg_wal`, or `postgresql.conf`, and it skips parent directories that contain initialized child data directories.

This addresses the PGliteEngine recovery path from #98. It does not attempt to change the `tsx watch` / process-group shutdown behavior that can interrupt first-boot initialization in the first place.

## Validation

Manual before/after reproduction of the corrupt-state symptom:

```text
# Before/failure mode: opening a partial PGlite data dir directly
partial dir contents: PG_VERSION only
result: RuntimeError: unreachable

# After/fixed path: opening the same partial dir through PGliteEngine
[PGliteEngine] Moved incomplete PGlite data directory from .../partial to .../partial.corrupt-...; created a fresh directory. Delete matching .corrupt-* directories when they are no longer needed.
{"rows":[{"id":"ok"}],"hasBase":true,"hasGlobal":true}
```

Automated checks:

```text
npm run build -w packages/bb-data
npm run test -w packages/bb-data
./node_modules/.bin/changeset status --since=origin/main
./node_modules/.bin/tsx scripts/verify-changeset-coverage.ts
git diff --check
git diff --cached --check
```

## Checklist

- [x] PR description included

---

### DIFF
--- a/packages/bb-data/src/engines/pglite-engine.test.ts
+++ b/packages/bb-data/src/engines/pglite-engine.test.ts
@@ -194,3 +194,39 @@ test('constructor recovers an incomplete PGlite init directory', async () => {
   assert.strictEqual(corruptDirs.length, 1);
   assert.strictEqual(existsSync(join(TEST_DIR, corruptDirs[0], 'PG_VERSION')), true);
 });
+
+test('constructor recovers a PGlite init directory with empty required directories', async () => {
+  const partial = join(TEST_DIR, 'empty-required-dirs');
+  mkdirSync(join(partial, 'base'), { recursive: true });
+  mkdirSync(join(partial, 'global'), { recursive: true });
+  writeFileSync(join(partial, 'PG_VERSION'), '16\n');
+
+  engine = new PGliteEngine(partial);
+  await engine.execute('CREATE TABLE t (id TEXT PRIMARY KEY)');
+  await engine.execute("INSERT INTO t (id) VALUES ('ok')");
+
+  const rows = await engine.query<{ id: string }>('SELECT id FROM t');
+  assert.deepStrictEqual(rows, [{ id: 'ok' }]);
+  assert.strictEqual(existsSync(join(partial, 'global', 'pg_control')), true);
+
+  const corruptDirs = readdirSync(TEST_DIR).filter((entry) => entry.startsWith('empty-required-dirs.corrupt-'));
+  assert.strictEqual(corruptDirs.length, 1);
+  assert.strictEqual(existsSync(join(TEST_DIR, corruptDirs[0], 'global')), true);
+});
+
+test('constructor recovers a PGlite leaf directory before PG_VERSION is written', async () => {
+  const partial = join(TEST_DIR, 'pre-version-init');
+  mkdirSync(join(partial, 'base'), { recursive: true });
+
+  engine = new PGliteEngine(partial);
+  await engine.execute('CREATE TABLE t (id TEXT PRIMARY KEY)');
+  await engine.execute("INSERT INTO t (id) VALUES ('ok')");
+
+  const rows = await engine.query<{ id: string }>('SELECT id FROM t');
+  assert.deepStrictEqual(rows, [{ id: 'ok' }]);
+  assert.strictEqual(existsSync(join(partial, 'PG_VERSION')), true);
+
+  const corruptDirs = readdirSync(TEST_DIR).filter((entry) => entry.startsWith('pre-version-init.corrupt-'));
+  assert.strictEqual(corruptDirs.length, 1);
+  assert.strictEqual(existsSync(join(TEST_DIR, corruptDirs[0], 'base')), true);
+});
--- a/packages/bb-data/src/engines/pglite-engine.test.ts
+++ b/packages/bb-data/src/engines/pglite-engine.test.ts
@@ -5,7 +5,7 @@ import { test, afterEach } from 'node:test';
 import assert from 'node:assert';
 import { PGliteEngine } from './pglite-engine.js';
 import { DatabaseErrors } from '../errors.js';
-import { rmSync, existsSync } from 'node:fs';
+import { rmSync, existsSync, mkdirSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 
 const TEST_DIR = '.bb-data-test-' + process.pid;
@@ -174,4 +174,19 @@ test('constructor does not crash when parent directory is missing', async () => 
   await engine.execute('CREATE TABLE t (id TEXT PRIMARY KEY)');
   const rows = await engine.query('SELECT * FROM t');
   assert.deepStrictEqual(rows, []);
+});
+
+test('constructor recovers an incomplete PGlite init directory', async () => {
+  const partial = join(TEST_DIR, 'partial-init');
+  mkdirSync(partial, { recursive: true });
+  writeFileSync(join(partial, 'PG_VERSION'), '16\n');
+
+  engine = new PGliteEngine(partial);
+  await engine.execute('CREATE TABLE t (id TEXT PRIMARY KEY)');
+  await engine.execute("INSERT INTO t (id) VALUES ('ok')");
+
+  const rows = await engine.query<{ id: string }>('SELECT id FROM t');
+  assert.deepStrictEqual(rows, [{ id: 'ok' }]);
+  assert.strictEqual(existsSync(join(partial, 'base')), true);
+  assert.strictEqual(existsSync(join(partial, 'global')), true);
 });
--- a/packages/bb-data/src/engines/pglite-engine.test.ts
+++ b/packages/bb-data/src/engines/pglite-engine.test.ts
@@ -5,7 +5,7 @@ import { test, afterEach } from 'node:test';
 import assert from 'node:assert';
 import { PGliteEngine } from './pglite-engine.js';
 import { DatabaseErrors } from '../errors.js';
-import { rmSync, existsSync, mkdirSync, writeFileSync } from 'node:fs';
+import { rmSync, existsSync, mkdirSync, writeFileSync, readdirSync } from 'node:fs';
 import { join } from 'node:path';
 
 const TEST_DIR = '.bb-data-test-' + process.pid;
@@ -189,4 +189,8 @@ test('constructor recovers an incomplete PGlite init directory', async () => {
   assert.deepStrictEqual(rows, [{ id: 'ok' }]);
   assert.strictEqual(existsSync(join(partial, 'base')), true);
   assert.strictEqual(existsSync(join(partial, 'global')), true);
+
+  const corruptDirs = readdirSync(TEST_DIR).filter((entry) => entry.startsWith('partial-init.corrupt-'));
+  assert.strictEqual(corruptDirs.length, 1);
+  assert.strictEqual(existsSync(join(TEST_DIR, corruptDirs[0], 'PG_VERSION')), true);
 });
--- a/packages/bb-data/src/engines/pglite-engine.ts
+++ b/packages/bb-data/src/engines/pglite-engine.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { PGlite } from '@electric-sql/pglite';
-import { existsSync, mkdirSync, unlinkSync } from 'node:fs';
+import { existsSync, mkdirSync, readdirSync, rmSync, unlinkSync } from 'node:fs';
 import { join } from 'node:path';
 import type { DatabaseEngine, TransactionHandle } from '@aws-blocks/data-common';
 import { DatabaseErrors, wrapError } from '../errors.js';
@@ -12,6 +12,15 @@ const PG_UNIQUE_VIOLATION = '23505';
 
 /** PostgreSQL error code class for connection exceptions. */
 const PG_CONNECTION_EXCEPTION_CLASS = '08';
+const PGLITE_REQUIRED_DATA_DIR_ENTRIES = ['PG_VERSION', 'base', 'global'];
+const PGLITE_DATA_DIR_MARKERS = [
+  ...PGLITE_REQUIRED_DATA_DIR_ENTRIES,
+  'pg_wal',
+  'postgresql.conf',
+  'pg_hba.conf',
+  'pg_ident.conf',
+  'postmaster.pid',
+];
 
 /**
  * Translate a PGlite/PostgreSQL error to a standardized DatabaseErrors name.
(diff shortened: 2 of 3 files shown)


packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +62,22 @@ function cleanStaleLock(dataDir: string): void {
   }
 }
 
+function hasInitializedPgliteDataDir(dataDir: string): boolean {
+  return PGLITE_REQUIRED_DATA_DIR_ENTRIES.every((entry) => existsSync(join(dataDir, entry)));
+}
+
+function recoverIncompletePgliteDataDir(dataDir: string): void {
```
soberm, line 91: `recoverIncompletePgliteDataDir()` decides to delete the entire directory from a loose heuristic. A single matching filename triggering an unconditional recursive delete is broader than the stated intent of recovering an interrupted first boot. We could tighten the trigger so deletion only fires when the directory clearly looks like a real-but-incomplete PGlite init rather than on any single marker. For example, require at least `PG_VERSION` present (the unambiguous PGlite signature) before recreating, or require two or more markers; alternatively rename the incomplete dir to a .corrupt-<timestamp> sidecar instead of `rmSync` so the data is recoverable, then log its location.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +62,22 @@ function cleanStaleLock(dataDir: string): void {
...
+  const entries = readdirSync(dataDir);
+  if (entries.length === 0 || hasInitializedPgliteDataDir(dataDir)) return;
+
+  const looksLikePgliteInit = entries.some((entry) => PGLITE_DATA_DIR_MARKERS.includes(entry));
+  if (!looksLikePgliteInit) return;
+
+  rmSync(dataDir, { recursive: true, force: true });
+  mkdirSync(dataDir, { recursive: true });
+  console.warn(`[PGliteEngine] Recreated incomplete PGlite data directory ${dataDir}`);
```
soberm: The only signal that user-visible local state was just deleted is a single `console.warn`. The rest of the engine uses `console.log/console.debug` for benign events, so a destructive recreate is not distinguished in severity, and the message does not say what was removed.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +62,22 @@ function cleanStaleLock(dataDir: string): void {
   }
 }
 
+function hasInitializedPgliteDataDir(dataDir: string): boolean {
+  return PGLITE_REQUIRED_DATA_DIR_ENTRIES.every((entry) => existsSync(join(dataDir, entry)));
+}
+
+function recoverIncompletePgliteDataDir(dataDir: string): void {
+  const entries = readdirSync(dataDir);
```
soberm: Should we wrap the `readdirSync` in a try/catch that returns early on failure matching the defensive posture of `cleanStaleLock` above it?

ikenyal: Done. I wrapped readdirSync in a try/catch and return early on failure, so this recovery step stays best-effort like cleanStaleLock. If the directory is still unusable, the normal PGlite initialization path will surface that error.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Thanks, agreed. I tightened the recovery trigger to require PG_VERSION before treating a non-empty directory as an incomplete PGlite init, and changed the recovery path to move the existing directory to a dataDir.corrupt-<timestamp> sidecar instead of deleting it recursively. The regression test now also verifies that the partial PG_VERSION-only directory is retained in the sidecar while a fresh data dir is created.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Updated. The warning now includes both the original dataDir and the sidecar path, and the recovery now renames the incomplete directory instead of deleting it so the previous local state remains available for inspection.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +54,40 @@ function cleanStaleLock(dataDir: string): void {
   }
 }
 
+function hasInitializedPgliteDataDir(dataDir: string): boolean {
```
soberm, line 68: hasInitializedPgliteDataDir only checks that PG_VERSION, base, and global *exist*, not that they have content. PGlite unpacks the data dir as a tar via loadTar in a single depth-first pass, so the global/ directory inode gets created before the files inside it (pg_filenode.map, pg_authid, …) are written. If a boot is interrupted in that window this returns true, recovery skips the dir, and the next `new PGlite(dataDir)` falls over with an opaque Postgres/WASM error instead of self-healing — which is the exact failure mode this change is meant to prevent. Can we add one cheap content check, e.g. also require `global/pg_filenode.map`? It's the first file written into global/ and Postgres won't start without it.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +54,40 @@ function cleanStaleLock(dataDir: string): void {
...
+  try {
+    entries = readdirSync(dataDir);
+  } catch {
+    return;
+  }
+  if (entries.length === 0 || hasInitializedPgliteDataDir(dataDir)) return;
+
+  if (!entries.includes('PG_VERSION')) return;
+
+  const corruptDataDir = nextCorruptDataDir(dataDir);
```
soberm, line 103: These quarantined dirs pile up forever — every recovery renames to `.corrupt-<timestamp>` and nothing ever prunes them, so a repeated crash-on-boot loop keeps spawning full data-dir siblings. Could we at least log a one-liner telling the dev they can delete the `*.corrupt-*` dirs, or cap how many we keep? A note in DESIGN.md would do too.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +54,40 @@ function cleanStaleLock(dataDir: string): void {
...
+    suffix++;
+  }
+  return candidate;
+}
+
+function recoverIncompletePgliteDataDir(dataDir: string): void {
+  let entries: string[];
+  try {
+    entries = readdirSync(dataDir);
+  } catch {
```
soberm: This catch swallows everything, not just ENOENT. If the dir exists but we can't read it (EACCES/EPERM), we silently bail here and the real problem resurfaces later as a more confusing PGlite error. Worth narrowing to `err.code === 'ENOENT'` and letting other errors through.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +54,40 @@ function cleanStaleLock(dataDir: string): void {
...
 
+function hasInitializedPgliteDataDir(dataDir: string): boolean {
+  return PGLITE_REQUIRED_DATA_DIR_ENTRIES.every((entry) => existsSync(join(dataDir, entry)));
+}
+
+function nextCorruptDataDir(dataDir: string): string {
+  const timestamp = new Date().toISOString().replace(/[:.]/g, '-');
+  let candidate = `${dataDir}.corrupt-${timestamp}`;
+  let suffix = 1;
+  while (existsSync(candidate)) {
```
soberm: Minor: there's a check-then-act gap here — existsSync, then the caller renameSyncs into the candidate. Practically fine since this is single-process local dev, but the loop reads like it fully reserves the name when it doesn't. A short random/pid suffix would close the window (and let you drop the loop); otherwise a quick comment noting the single-process assumption would help.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +54,40 @@ function cleanStaleLock(dataDir: string): void {
...
+    return;
+  }
+  if (entries.length === 0 || hasInitializedPgliteDataDir(dataDir)) return;
+
+  if (!entries.includes('PG_VERSION')) return;
+
+  const corruptDataDir = nextCorruptDataDir(dataDir);
+  renameSync(dataDir, corruptDataDir);
+  mkdirSync(dataDir, { recursive: true });
+  console.warn(
```
soberm: This is a fully handled, self-healing event, so console.warn feels a bit loud — it'll surface as a warning on every clean recovery. console.log/info reads better here.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -12,6 +12,7 @@ const PG_UNIQUE_VIOLATION = '23505';
 
 /** PostgreSQL error code class for connection exceptions. */
 const PG_CONNECTION_EXCEPTION_CLASS = '08';
+const PGLITE_REQUIRED_DATA_DIR_ENTRIES = ['PG_VERSION', 'base', 'global'];
```
soberm: Could you add a short comment on why these three entries mean "initialised" and that the order matters (PG_VERSION is written first by loadTar)? The whole recovery heuristic leans on that assumption and it'd be easy for someone to break it by accident.

packages/bb-data/src/engines/pglite-engine.test.ts
```diff
@@ -175,3 +175,22 @@ test('constructor does not crash when parent directory is missing', async () =>
   const rows = await engine.query('SELECT * FROM t');
   assert.deepStrictEqual(rows, []);
 });
+
+test('constructor recovers an incomplete PGlite init directory', async () => {
+  const partial = join(TEST_DIR, 'partial-init');
+  mkdirSync(partial, { recursive: true });
+  writeFileSync(join(partial, 'PG_VERSION'), '16\n');
```
soberm, line 182: Nice-to-have: this covers the "PG_VERSION only" case but not "PG_VERSION + empty base/global", which is the false-positive state hasInitializedPgliteDataDir would let through today. Adding that case would lock the gap down.

packages/bb-data/src/engines/pglite-engine.ts
```diff
@@ -53,6 +54,40 @@ function cleanStaleLock(dataDir: string): void {
...
+function recoverIncompletePgliteDataDir(dataDir: string): void {
+  let entries: string[];
+  try {
+    entries = readdirSync(dataDir);
+  } catch {
+    return;
+  }
+  if (entries.length === 0 || hasInitializedPgliteDataDir(dataDir)) return;
+
+  if (!entries.includes('PG_VERSION')) return;
```
hfurkanbozkurt: This guard does double duty: it correctly protects a parent dir (the CLI's `new PGliteEngine('.bb-data')` root has no top-level PG_VERSION), but it also skips a corrupt leaf with NO PG_VERSION at all — the original #98 mode (a non-empty `.bb-data/<id>` lacking PG_VERSION aborts `new PGlite()`). Combined with the existence-vs-content gap already noted on this PR (global/ created before its files), the only recovered state is "PG_VERSION present, base/global missing" — a narrow slice of the interruption window.

Suggestion: detect a partial leaf affirmatively instead of gating on PG_VERSION existence — recover when the dir is non-empty, not complete *by content* (require PG_VERSION AND global/pg_control, not just entry existence), and is itself a data dir (has postmaster.opts/postmaster.pid/base/global), while never
touching a dir whose children are complete data dirs. That catches the no-PG_VERSION leaf and the empty-global/ window without re-introducing the parent-wipe risk this guard was added for.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Done. The sidecar path now includes the process id and a random suffix, so it no longer relies on an existsSync loop that appears to reserve the destination name.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Done. I tightened the initialized check to require global/pg_control in addition to PG_VERSION/base/global, so a directory with empty base/global no longer gets treated as complete. I used pg_control as the content sentinel since it is present in a successfully initialized PGlite/Postgres global directory and Postgres cannot start without it.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Updated the recovery log to explicitly say that matching .corrupt-* directories can be deleted when they are no longer needed. I kept automatic pruning out of this patch so recovery remains non-destructive and does not guess which quarantined data dir is safe to remove.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Done. The readdirSync catch now returns only for ENOENT and rethrows other filesystem errors, so EACCES/EPERM should surface directly instead of being hidden behind a later PGlite failure.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Done. This is now console.log instead of console.warn since the event is self-healing after the incomplete directory has been moved aside.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Added a short comment above the initialized-dir check. It calls out that PG_VERSION is written before the directory tree and that global/pg_control is used to confirm that global contains PostgreSQL state, so the heuristic is less likely to be simplified back to existence-only checks.

packages/bb-data/src/engines/pglite-engine.ts
ikenyal: Updated along these lines. Recovery no longer gates solely on PG_VERSION; it treats a non-empty directory with PGlite leaf markers as recoverable when it is not complete by content, while skipping directories that contain initialized child data dirs. I also moved stale postmaster.pid cleanup after recovery so a pre-PG_VERSION leaf can still be recognized, and added a regression test for the pre-PG_VERSION case.

packages/bb-data/src/engines/pglite-engine.test.ts
ikenyal: Added that regression case. The test now seeds PG_VERSION plus empty base/global directories and verifies that PGliteEngine quarantines the partial dir, creates a fresh usable data dir, and writes global/pg_control in the recovered directory.

soberm (approved):

hfurkanbozkurt (approved):

sarayev (approved):

PR #128, opened 2026-06-30
fix(bb-data): recover from interrupted-initdb PGlite .bb-data corruption (#98)

### LINKED ISSUE #98: PGlite .bb-data corruption on tsx-watch restart mid-init: next boot aborts and the dev-server port never binds

hfurkanbozkurt:
## Summary

Fixes #98 — PGlite `.bb-data` corruption on `tsx watch` restart mid-init, after which the next boot aborts and the dev-server port never binds.

When `tsx watch` SIGTERMs (then, after its grace period, SIGKILLs) the dev server while a local `Database` block is still running first-boot `initdb`, the PGlite data directory under `.bb-data/<fullId>` is left **non-empty but missing the `PG_VERSION` / `global/pg_control` markers** a complete data dir has. On the next boot `new PGlite(dir)` aborts on the corrupt directory; the dev server's `deployLocal()` phase rejects **before** `server.listen()` is ever reached (init is `await`ed at startup, before `listen`), so the port never binds and the app is unreachable until `.bb-data` is deleted by hand.

## Confirmed mechanism

Reproduced against `@electric-sql/pglite@0.2.x` (the pinned version):

- A **complete** PGDATA always contains `PG_VERSION` **and** `global/pg_control`.
- A directory that exists and is **non-empty but lacks those markers** (a half-written `initdb`) makes `new PGlite(dir)` abort: `ExitStatus - Program terminated with exit(1)`.
- Wiping that leaf dir and letting PGlite re-`initdb` recovers cleanly; a valid dir (even with only a stale `postmaster.pid`) is **not** flagged, so no data is lost.

Why the fix lives in `PGliteEngine` (`@aws-blocks/bb-data`), not the dev-server shutdown path (`@aws-blocks/core`): a clean-shutdown await can reduce the corruption window for SIGTERM, but `tsx watch` escalates to **SIGKILL** after its grace period (and crashes/power-loss exist), which no shutdown handler can intercept. Making the on-disk state **recoverable on next boot** is the durable fix and directly resolves the "port never binds" symptom regardless of how the previous process died. This matches the issue's "defense-in-depth in `PGliteEngine` … re-init the leaf dir instead of aborting" suggestion.

## Change

`PGliteEngine` constructor now calls `recoverIncompleteDataDir(dataDir)` **before** opening PGlite:

- **Before:** non-empty dir missing `PG_VERSION`/`global/pg_control` → `new PGlite()` aborts → `deployLocal()` rejects → `listen()` never called → port never binds.
- **After:** such a dir is detected and the leaf is re-initialized; PGlite `initdb` runs cleanly, init completes, `listen()` binds the port. Empty dirs and fully-initialized dirs are untouched.

## Tests

`packages/bb-data/src/engines/pglite-engine.test.ts`:
- `recovers from an incompletely-initialized data directory (interrupted initdb)` — a non-empty dir with no `PG_VERSION` (which aborts PGlite) is recovered instead of crashing.
- `preserves a fully-initialized data directory across restart` — a valid dir is never wiped; inserted data survives a reopen.

All 18 `pglite-engine` tests pass; `@aws-blocks/bb-data` builds/typechecks; biome lint clean on changed lines.

## Changeset

`@aws-blocks/bb-data`: patch (ref #98).

> Note: the issue/triage hypothesis pointed at `@aws-blocks/core`'s dev-server shutdown; the actual durable fix is in `@aws-blocks/bb-data`'s `PGliteEngine`, so the changeset bumps `bb-data`. No `core` changes were required.

### DIFF
--- a/.changeset/fix-pglite-bb-data-corruption-recovery.md
+++ b/.changeset/fix-pglite-bb-data-corruption-recovery.md
@@ -1,0 +1,9 @@
+---
+"@aws-blocks/bb-data": patch
+---
+
+fix(bb-data): recover from a PGlite `.bb-data` dir left half-written by an interrupted initdb (#98)
+
+When `tsx watch` SIGTERMs/SIGKILLs the dev server while a local `Database` block is still running first-boot `initdb`, the PGlite data directory under `.bb-data/<fullId>` could be left non-empty but missing the `PG_VERSION`/`global/pg_control` markers a complete data dir has. On the next boot `new PGlite(dir)` aborted on the corrupt directory, the dev server's local-deploy phase rejected before `server.listen()` was ever reached, and the port never bound — the app stayed unreachable until `.bb-data` was deleted by hand.
+
+`PGliteEngine` now detects an incompletely-initialized data directory before opening it and re-initializes the leaf directory instead of aborting, so a single interrupted boot is recoverable. A fully-initialized directory (even one with only a stale `postmaster.pid`) is left untouched, so real local data is preserved across restarts.
--- a/packages/bb-data/src/engines/pglite-engine.test.ts
+++ b/packages/bb-data/src/engines/pglite-engine.test.ts
@@ -5,7 +5,7 @@ import { test, afterEach } from 'node:test';
 import assert from 'node:assert';
 import { PGliteEngine } from './pglite-engine.js';
 import { DatabaseErrors } from '../errors.js';
-import { rmSync, existsSync } from 'node:fs';
+import { rmSync, existsSync, mkdirSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 
 const TEST_DIR = '.bb-data-test-' + process.pid;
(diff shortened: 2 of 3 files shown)


soberm:
Duplicate of https://github.com/aws-devtools-labs/aws-blocks/pull/119

PR #127, opened 2026-06-30
fix(create-blocks-app): serve react template from single-origin front door

hfurkanbozkurt:
## Problem

`packages/create-blocks-app/templates/react` was the **only SPA template without a single-origin dev front door** — the root cause behind the `oidc-dsql-notes` failure surfaced by the agent-bench.

- `aws-blocks/scripts/server.ts` ran the backend on `:3001` via `startDevServer({ backendPath, port: 3001 })` with **no `frontendCommand`**.
- `package.json` `dev` used `concurrently` to run Vite on `:3000` + backend on `:3001` as **two separate origins**.
- `vite.config.ts` had **no `/aws-blocks` proxy**.

Result: `/aws-blocks/*` — including the server-initiated OIDC redirect routes `/aws-blocks/auth/signin/*` — was **not reachable from the SPA origin (`:3000`)**, so any browser-navigation auth flow (OIDC) was broken locally. Every other SPA template (`default`/`bare`/`demo`/`auth-cognito`/`nextjs`) already exposes `:3000` as a single front door (backend + SPA same origin) via `startDevServer({ frontendCommand: ... })`.

## Fix

Matched the convention used by every other SPA template:

**`aws-blocks/scripts/server.ts`**
```diff
 startDevServer({
   backendPath: join(__dirname, '..', 'index.ts'),
-  port: 3001
+  frontendCommand: 'npx vite --port 3100 --strictPort',
+  frontendPort: 3100,
 });
```

**`package.json`** — single dev server, dropped `concurrently`/`dev:client`:
```diff
-    "dev:client": "vite",
-    "dev": "concurrently \"npm:dev:server\" \"npm:dev:client\"",
+    "dev": "tsx watch aws-blocks/scripts/server.ts",
...
-    "concurrently": "^8.2.0",
```

**`vite.config.ts`** — no longer pins the SPA to `:3000` (Vite now runs on `:3100` behind the front door):
```diff
-  server: {
-    port: 3000
-  },
```

Now `startDevServer` defaults the customer-facing front door to `:3000`, launches Vite on `:3100`, and proxies the SPA + `/aws-blocks/*` from the **same origin** — unblocking OIDC / browser-navigation auth in the react template.

Also fixed stale `README.md` references (`localhost:5173`, "concurrent").

## Changeset

Adds `.changeset/fix-react-template-single-origin.md` bumping `@aws-blocks/create-blocks-app` (patch).

## Verification

[...]

Issue #97, opened 2026-06-23
agent-bench: scope the builder agent step to a Bedrock-invoke-only role (drop s3:PutObject)


hfurkanbozkurt:
## Summary

The builder step ("2. Agent run") runs an LLM-driven shell agent while the live OIDC/AWS session is in scope. The assumed role grants BOTH Bedrock invoke AND `s3:PutObject` on `bench/*` — but only the separate "Persist to S3" step needs `s3:PutObject`. The agent step's blast radius is broader than it needs, even though it's bounded by the same-repo + member-only `ready-for-pr-benchmark` label gate (forks get no secrets).

## Where

- `.github/workflows/agent-bench.yml` — OIDC role assumption + the "2. Agent run" and "Persist to S3" steps.
- README "Security" / "Future hardening" already names this.

## Desired (future hardening)

Give the agent step a minimal Bedrock-invoke-only role and keep `s3:PutObject` only on the "Persist to S3" step, shrinking the blast radius within the trusted trigger.

## Note

Non-blocking; current posture acknowledged + documented. Tracking the least-privilege follow-up.

---
_Follow-up tracking issue for PR #31 (per-PR agent bench)._

hfurkanbozkurt:
Reopening to track as the concrete follow-up for the agent-step least-privilege hardening flagged again in PR #31 review (the 'should-fix' security item). Plan: split the OIDC role so the '2. Agent run' step assumes a Bedrock-InvokeModelWithResponseStream-only role, and only the separate 'Persist to S3' step holds s3:PutObject on bench/* — shrinking a compromised-agent's blast radius even within the same-repo-gated trigger. Currently bounded by the same-repo gate (fork PRs get no secrets) and documented in README Security; this issue tracks the deferred least-privilege split.

ikenyal:
I traced the current workflow to pin down the credential boundaries before changing it. One implementation detail is worth calling out:

- The bench job configures `AWS_ROLE_ARN` before the builder and keeps that session through the judge, per-cell analysis, and the S3 archive step. The Bedrock-only role therefore needs to cover the builder, judge, and per-cell analysis; credentials can then be reconfigured immediately before `Persist to S3` to an archive role scoped to the existing benchmark prefix.
- The summary job is a separate boundary: it reads and writes the S3 baseline, but also runs `bedrock-runtime converse` for the roll-up analysis. It has no shell-capable agent, so it could use a dedicated combined summary role, or switch roles around those operations.

[...]

Issue #81, opened 2026-06-23
AuthBasic setAuthState() discards the structured error name (e.name) — recommended client path can't branch on AuthBasicErrors


hfurkanbozkurt:
## Summary

`AuthBasic`'s **recommended client path** is `createApi()` → `setAuthState()` (state machine). When an action fails, `setAuthState()` swallows the thrown `ApiError` and returns an `AuthState` that carries **only `error: e.message`** — it drops the structured error **name** (`e.name`, e.g. `'InvalidCredentialsException'`). Combined with the fact that the `AuthState` interface has **no field for an error name/code**, the README's documented catch-by-name idiom (`isBlocksError(e, AuthBasicErrors.InvalidCredentials)`) is **unreachable** from the recommended client path. A client that wants to implement a common "try sign-in first, fall back to sign-up for a fresh user" flow cannot reliably branch on the error type — it is forced to string-match the human-facing `message`.

### `packages/bb-auth-basic/src/index.ts`

`signIn()` throws **with** the structured error name (L323 and L327):

```ts
323: 			throw new ApiError('Invalid username or password', 401, { name: AuthBasicErrors.InvalidCredentials });
...
327: 			throw new ApiError('Invalid username or password', 401, { name: AuthBasicErrors.InvalidCredentials });
```

The error constant resolves to a stable name (L104-110):

```ts
104: export const AuthBasicErrors = {
105: 	InvalidCredentials: 'InvalidCredentialsException',
106: 	UserAlreadyExists: 'UserAlreadyExistsException',
107: 	InvalidCode: 'InvalidCodeException',
108: 	SessionExpired: 'SessionExpiredException',
109: 	InvalidPassword: 'InvalidPasswordException',
110: } as const;
```

But the recommended client entry point `setAuthState()` (L438) swallows the throw in its catch (L479-483) and returns a **lossy** state at L482 that keeps only `e.message` and discards `e.name`:

```ts
438: 			setAuthState: async (input: AuthActionInput): Promise<AuthState> => {
...
479: 				} catch (e: any) {
480: 					const currentUser = await this.getUserFromCookie(context);
481: 					const base = currentUser ? signedInState(currentUser) : this.signedOutState();
482: 					return { ...base, error: e.message };
483: 				}
```

### `packages/auth-common/src/index.ts`

The shared `AuthState` interface (L152-181) exposes a free-text `error?: string` (L169) only — there is **no** field to carry the structured error name/code:

```ts
[...]

PR #132, opened 2026-07-01
feat(ci): per-PR agent benchmark (Strands + Bedrock)

### LINKED ISSUE #78: Dev server returns permanent HTTP 502 "Frontend server unavailable" after a tsx-watch restart orphans the --strictPort Vite frontend

hfurkanbozkurt:
## What this is

A **per-PR LLM-agent benchmark** that measures whether the framework — or a given PR — helps or hurts the *agentic* developer experience of building on AWS Blocks. It is **CI/tooling-only**: it adds nothing to `packages/` or `test-apps/` source. All changes live under `scripts/agent-bench/`, `tasks/`, `.github/workflows/`, plus the workspace wiring in `package.json` / `package-lock.json` and its own changeset.

## How it works

- **10 cells** (task × template) spanning the Building Blocks (auth, DSQL/SQL, KV, realtime, file storage, async jobs, observability, email, knowledge-base chat, etc.).
- A **builder agent** (Strands + Amazon Bedrock, Claude Sonnet) scaffolds a fresh app and implements each task's `PROMPT.md` from scratch.
- The generated app is **built** and its **Playwright** tests are run.
- A **judge agent** (Bedrock, Claude Opus) scores the produced source against a **uniform 5-dimension rubric**: `functional_completeness`, `selector_contract`, `persistence`, `code_quality`, `blocks_fidelity`.
- **Composite score = 60% objective test pass-rate + 40% gated judge score.**

## Design choices

- **N=1, observational — NOT a merge gate.** `BENCH_MIN_SCORE` is intentionally unset, so the bench never blocks a PR.
- **Green-regardless.** Bench failures never turn the check red; a skipped/absent results set is handled gracefully (no spurious ❌).
- **No PR comment noise.** Results render in the **Actions run step-summary**, not as a PR comment.
- **Triggers** on `pull_request` and on `push` to `main`.
- **Same-repo / fork-gated.** Forks receive no secrets, so the agent steps are gated to same-repo events.
- **`cancel-in-progress`** on new commits.
- **Results persisted to S3** keyed by commit, with a **PR-vs-baseline diff overview**.

## Latest results

**Mean 93.2 / 100, 10/10 cells pass** (run `28503266028`).

## Notes

- Recreated from #31 on a clean branch/commit off latest `main` (the original PR accumulated noisy iteration history + early automated comments). Identical validated content.
- While building this out, the bench surfaced and validated framework fixes: #78, #79, #81, #98, #127.

## Security note

The builder step runs an LLM shell agent under the job's OIDC session. This is bounded by the same-repo gate (forks get no credentials). Least-privilege hardening of the builder role is tracked in #97.


scripts/agent-bench/steps/lib/scoring.mjs
```diff
@@ -0,0 +1,219 @@
...
+	if (klass === 'agent_fail') return 'fail';
+	const stats = testStats(result);
+	if (stats.denom === 0) return verdict(0, 'unknown');
+	return verdict(testRate(stats), null);
+}
+
+/**
+ * Composite score 0..100: 60% objective pass-rate + 40% judge, with the judge
+ * term gated by min(1, 4*tr) so a 0-test cell floors to 0 and a judge failure
+ * (j=0) only drops the judge term, never the test-driven portion.
```
svidgen, line 157: **Ask: document the 25% gate threshold.**

The `4` in `min(1, 4*rate)` means the judge term reaches full weight when tr ≥ 0.25. This is a defensible design choice, but it's not obvious *why* 25% specifically (vs 10%, 50%, etc.). Could you add a brief comment here explaining the rationale? Something like:

```js
// The judge term is gated by min(1, 4*tr): it ramps linearly from 0 to full
// weight as the pass-rate goes from 0% to 25%, then stays at full weight.
// Threshold rationale: 25% represents [your reasoning here — e.g. "minimum
// evidence that the agent produced runnable code, below which a generous
// judge score would be misleading"].
```

Also: the README shows both `40*(judge/10)*min(1, 4*tr)` and `4*judge*min(1, 4*tr)` — they're algebraically identical but showing both invites confusion. Pick one canonical form.

scripts/agent-bench/steps/lib/scoring.mjs
```diff
@@ -0,0 +1,219 @@
...
+	return { klass: 'scored', reason: null };
+}
+
+/**
+ * Test counts for a cell. Missing / non-numeric fields read as 0.
+ * @param {{tests_passed?: number, tests_failed?: number}} result
+ * @returns {{passed: number, failed: number, denom: number}}
+ */
+export function testStats(result) {
+	const passed = num(result?.tests_passed);
```
svidgen, line 99: **Ask: clarify the `tests_total` omission.**

`testStats()` uses `passed + failed` as the denominator and ignores `tests_total` (which is passed through in the EVIDENCE payload and persisted in result.json). This means skipped/timed-out Playwright tests vanish from the score.

Is this intentional? If yes — skipped tests shouldn't penalize the agent — please add a comment here explaining that choice and note that `tests_total` in the evidence is audit-only. If no, this should use `tests_total` as the denominator (or at least `Math.max(passed + failed, tests_total)`).

scripts/agent-bench/steps/lib/overview.mjs
```diff
@@ -0,0 +1,161 @@
...
+ */
+export function meanComposite(cells) {
+	const scored = (cells ?? []).filter((c) => isScoredCell(c));
+	if (scored.length === 0) return null;
+	// cellComposite re-checks isScoredCell, but every `c` here already passed the
+	// filter above, so that guard is a defensive no-op on this path (never null).
+	const sum = scored.reduce((acc, c) => acc + cellComposite(c), 0);
+	return round1(sum / scored.length);
+}
+
```
svidgen, line 51: **Ask: widen the `deltaArrow` threshold to account for N=1 model variance.**

The current threshold of ±0.05 only guards against float rounding noise. With N=1 reps and LLM non-determinism (even at temperature=0), run-to-run variance of ±5 points or more is expected from model sampling alone — not from actual code changes.

Suggestion: widen the `=` band to something like ±5 (adjustable later once we have Athena data showing actual per-task stddev), and add an `≈` marker for deltas within that band. The numeric delta should still be shown, but annotated so reviewers know it may be noise:

```
| auth-notes | demo | 91.0 | 88.2 | ≈ -2.8 |   ← within expected variance
| file-gallery | bare | 95.0 | 82.1 | ▼ -12.9 |  ← likely real
```

Also: add a footnote to the rendered overview, something like:
> _N=1 per cell. Deltas within ±5 may reflect model variance, not a real change. Re-run the workflow for greater certainty._

This keeps wall-clock time unchanged while preventing misinterpretation of small deltas as regressions.

scripts/agent-bench/steps/2-agent-run.ts
```diff
@@ -0,0 +1,197 @@
...
+ *   TASK_PROMPT      path to PROMPT.md
+ *   OUTPUT           path to write the builder envelope JSON
+ *   BENCH_MODEL      Bedrock model ID (default: us.anthropic.claude-sonnet-4-6)
+ *
+ * Tools: a single `shell` running bash inside WORKSPACE. The runner is the
+ * sandbox; no extra isolation needed for our PR-CI threat model.
+ */
+import { spawnSync } from 'node:child_process';
+import { readFileSync, writeFileSync } from 'node:fs';
+import { Agent, BedrockModel, ModelStreamUpdateEvent, tool } from '@strands-agents/sdk';
```
svidgen: **Ask: document why this is intentionally a minimal single-tool agent.**

Future contributors may look at this and think it should be upgraded to a full coding agent (multi-file editing, LSP, search tools, etc.). The minimal design is actually load-bearing for the benchmark's purpose. Could you add a comment block here (or in the README) explaining:

1. **Stable measurement surface** — a full-featured agent brings its own capabilities that obscure whether the *framework* is agent-friendly
2. **Deterministic attribution** — regressions are attributable to framework changes, not agent version changes
3. **Token/cost accounting** — the custom hooks and SIGTERM envelope handling require control over the agent internals
4. **Reproducibility** — pinned model, temperature=0, MAX_TURNS cap — a standard agent's prompt engineering and retry logic changes across versions, introducing noise

Without this, someone will inevitably "helpfully" swap in a richer agent and break the benchmark's ability to isolate framework quality.

scripts/agent-bench/steps/summary.mjs
```diff
@@ -0,0 +1,368 @@
...
+
+// One row per cell, read from the single result.json the cell artifact holds.
+const cells = dirs.map((d) => {
+	const file = join(RESULTS_DIR, d, 'result.json');
+	try {
+		return JSON.parse(readFileSync(file, 'utf-8'));
+	} catch {
+		// No parseable result.json — surface the raw artifact suffix
+		// (<task>-<template>); can't split it reliably (both names have dashes).
+		return { task: d.replace('bench-result-', ''), template: '', error: 'unreadable' };
```
svidgen, line 70: **Ask: add Athena-ready NDJSON output for long-term trending.**

The current S3 output (`bench/runs/<sha>/results.json`) is great for per-PR baseline comparison, but it's not queryable at scale — no time-based partitioning, tiny individual objects, nested JSON. For trending composite scores over time and doing cost analysis, we'd like to also emit an Athena-friendly output.

Proposed: after writing the aggregate, append one NDJSON file (one line per scored cell) to a Hive-partitioned prefix:

```
s3://<bucket>/bench/athena/year=2026/month=07/<sha-short>.json
```

Flat schema per line: `sha, timestamp, event, pr_number, task, template, composite, test_rate, judge_score, verdict, klass, tokens_in, tokens_out, duration_sec`

This is ~15 lines of code, needs no new IAM grants (existing `bench/*` prefix covers it), and enables a `CREATE EXTERNAL TABLE` with partition projection — zero ongoing maintenance, instant SQL access to the full history.

svidgen (commented):
Thanks for the thorough work here — the architecture is solid and the scoring methodology is well-thought-out. A few asks below, mostly around clarity, long-term reportability, and statistical interpretation of the N=1 deltas.

svidgen (commented):
**Ask: remove all Blocks-specific guidance from the builder system prompt.**

The system prompt currently tells the agent to read `README.md` and then hints toward `node_modules/@aws-blocks/blocks/README.md`. This creates two problems:

1. **Drift risk.** The bench prompt will inevitably drift from how the scaffolded project actually steers agents. When it does, the prompt and the scaffold will *fight* each other — the agent gets conflicting signals, and you're no longer measuring the framework's real DX, you're measuring the bench's compensating instructions.

2. **Hidden regressions.** If the bench prompt tells the agent where to find docs, you'll never catch a regression in the scaffold's own `AGENTS.md` / `README.md` discovery path. The whole point of this bench is to measure whether an agent can build on Blocks successfully — if it can't discover what it needs from the scaffolded project alone, that's a real regression we *want* to detect.

The builder system prompt should be framework-agnostic. At most: "You are in a workspace. Read the root-level docs (AGENTS.md or README.md). The dev server is running on port X. Implement the task." Everything Blocks-specific — which packages exist, where to find API docs, how the building blocks work — should be discoverable from the scaffold's own files. If it isn't, that's a framework bug, not a bench bug.

This also makes the bench more honest as a competitive comparison tool (per Preamp): if we're giving our agent Blocks-specific hints that a real user's agent wouldn't get, the scores are inflated relative to how the framework actually performs in the wild.

svidgen:
To expand on my last comment: the dev server port (`/tmp/dev.port`) should also not be in the system prompt. If the scaffold's AGENTS.md doesn't tell the agent how to find the running dev server, that's a gap in the scaffold — not something the bench prompt should paper over. The system prompt should literally just be "you are in a workspace, read the root-level docs, implement the task." Full stop.

scripts/agent-bench/steps/3-build-and-test.sh
```diff
@@ -0,0 +1,288 @@
...
+  echo "::warning::playwright install failed; functional tests will not run"
+  exit 0
+fi
+if ! npx playwright install chromium > "${CELL_TMP}/pw-install.log" 2>&1; then
+  echo "::warning::playwright chromium download failed; functional tests will not run"
+  exit 0
+fi
+echo "playwright_installed=true" >> "$GITHUB_OUTPUT"
+
+mkdir -p bench-tests
```
soberm: [major] `mkdir -p bench-tests` creates the dir without clearing it first. An agent that writes Playwright spec files under `bench-tests/` during step 2 (the sandbox root is WORKSPACE, which includes this path once the agent creates it) would have those files picked up by `npx playwright test`, inflating `tests_passed` with agent-authored passing tests. The test-run outcome should reflect the objective spec only.

`rm -rf bench-tests && mkdir -p bench-tests` before the `cp` ensures the directory is clean each run.

.github/workflows/agent-bench.yml
```diff
@@ -0,0 +1,559 @@
...
+      # pr-agent-bench.yml: fork PRs get no secrets/OIDC and cannot trigger this
+      # with credentials (no external/fork triggering). Future hardening (not
+      # done): give the builder a minimal Bedrock-invoke-only role. See README
+      # "Security".
+      - name: 2. Agent run
+        id: agent
+        continue-on-error: true
+        if: steps.init.outcome == 'success'
+        timeout-minutes: 35
+        run: npx tsx scripts/agent-bench/steps/2-agent-run.ts
```
soberm, line 168: [major] The builder agent step runs with `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN` inherited from the OIDC session. The same-repo fork gate is the primary guard, but the OIDC role still carries `s3:PutObject` on `bench/*`, and the agent's bash tool can run arbitrary `aws` CLI commands. A sufficiently adversarial task `PROMPT.md` (including one contributed via a PR) could write to S3 or call other in-scope AWS APIs.

The SECURITY comment above this step flags this as "future hardening (not done)". Before the workflow goes live, the fix is to assume a second, narrow Bedrock-invoke-only role immediately before this step and overwrite the main session's `AWS_*` env vars so the full session is not in the agent's environment.

scripts/agent-bench/steps/1-init-bench-app.sh
```diff
@@ -0,0 +1,92 @@
...
+export NPM_CONFIG_FETCH_RETRY_MINTIMEOUT="${NPM_CONFIG_FETCH_RETRY_MINTIMEOUT:-20000}"
+export NPM_CONFIG_FETCH_RETRY_MAXTIMEOUT="${NPM_CONFIG_FETCH_RETRY_MAXTIMEOUT:-120000}"
+
+# `npm run build` is topology-aware and runs prebuild hooks in the right order.
+# `build:packages` runs alphabetically and trips over bb-data needing
+# bb-app-setting's generated version.ts.
+npm run build
+npm run publish:dry-run
+
+npx tsx scripts/publish/serve-local-registry.ts &
```
soberm, line 31: [minor] The registry process is started here and its PID saved to `/tmp/registry.pid`, but nothing ever kills it. There's no `trap` in this script and no corresponding kill step in the workflow, so the daemon outlives the scaffold phase and runs through steps 2-4 unnecessarily. For GitHub-hosted ephemeral runners this is harmless in practice, but it's easy to fix:

```bash
trap 'kill "$(cat /tmp/registry.pid)" 2>/dev/null || true' EXIT
```

Add this right after `echo $!` on the next line.

scripts/agent-bench/steps/4-judge.ts
```diff
@@ -0,0 +1,480 @@
...
+// don't read as imports. grep exits 1 (no error, no output) when nothing matches
+// — that clean "none found" is the enforce-able score-0 signal for the judge.
+function collectBlocksImports(workspace: string): string {
+	try {
+		// The --exclude-dir set MUST stay in sync with STAGE_EXCLUDE_DIRS above —
+		// the grep runs over the staged source-only copy, so blinding by absence
+		// and this scan must agree on what's off-limits. Globs are quoted so the
+		// shell can't expand them against the cwd before grep sees them.
+		const out = execSync(
+			'grep -rn "@aws-blocks/" . --include="*.ts" --include="*.tsx" --include="*.js" --include="*.jsx" --include="*.mjs" --include="*.cjs" --exclude-dir=node_modules --exclude-dir=.git --exclude-dir=dist --exclude-dir=.blocks-sandbox --exclude-dir=bench-tests | grep -Ew "import|from|require"',
```
soberm: [minor] The `--exclude-dir` flags here are a manual copy of `STAGE_EXCLUDE_DIRS` (line 54). The comment two lines above acknowledges they "MUST stay in sync", but there's no enforcement: adding a directory to the `Set` silently misses the grep, and adding it to the grep string alone silently misses the file staging. The fix is to derive the flags from the Set so there's only one list to update:

```typescript
const excludeArgs = [...STAGE_EXCLUDE_DIRS].map(d => `--exclude-dir=${d}`).join(' ');
// then interpolate excludeArgs into the grep invocation
```

.github/workflows/agent-bench.yml
```diff
@@ -0,0 +1,559 @@
+name: Agent Bench
+
+# Per-PR LLM-agent benchmark. Each cell scaffolds one template, hands the
+# scaffolded app to a builder agent, runs the build + Playwright spec, then a
+# judge agent grades. Runs entirely on the GitHub runner — no microVM, no S3
+# round-trip. Bedrock provides the model; the runner provides the sandbox.
+#
+# Required configuration (publish environment):
+#   AWS_ROLE_ARN  secret    OIDC role with bedrock:InvokeModelWithResponseStream
```
soberm, line 9: [question] `aws bedrock-runtime converse` (used in `analyze-cell.mjs` and `analyze.mjs`) maps to the IAM action `bedrock:Converse`, which is a separate permission from `bedrock:InvokeModelWithResponseStream`. Only the latter is listed here. If the OIDC role doesn't also grant `bedrock:Converse`, the analyze-cell step will get AccessDenied on every call. Because the step is `continue-on-error: true`, it won't turn the cell red, but per-cell analysis will silently degrade to `"analysis unavailable"` on every run. Is `bedrock:Converse` already on the role and just missing from this comment, or does the role policy need to be updated?

scripts/agent-bench/steps/3-build-and-test.sh
```diff
@@ -0,0 +1,288 @@
...
+# test data (usernames, file names, notes …). It carries the cell's TASK so two
+# cells sharing a run/attempt id still seed distinct data. Exported once here so
+# it stays identical across a spec's internal navigation; the specs' uniq()
+# helper adds a per-call counter + timestamp on top, so identifiers stay
+# collision-free.
+export RUN_ID="${GITHUB_RUN_ID:-local}-${GITHUB_RUN_ATTEMPT:-1}-${TASK:-x}-$(date +%s)"
+BLOCKS_URL="$APP_BASE_URL" APP_BASE_URL="$APP_BASE_URL" npx playwright test 2>&1 | tee "${CELL_TMP}/pw.log" || true
+
+if [ -f "$PW_RESULTS_JSON" ]; then
+  PW_RESULTS_JSON="$PW_RESULTS_JSON" node -e "
```
soberm: [nit] The dead-server handler on line 181 uses `node -e '...'` with single-quoted outer and double-quoted JS strings inside. Here the quoting is flipped: the outer shell string is double-quoted, which allows `$` characters and backticks to be shell-expanded. The JS body happens to be safe today, but the pattern is inconsistent and fragile. Using the same convention as line 181 (single-quoted outer, double-quoted inner) would be more consistent and less error-prone.

scripts/agent-bench/steps/lib/scoring.mjs
```diff
@@ -0,0 +1,301 @@
...
+ *   - it is gradeable (not a `harness_error`) AND actually produced test
+ *     results (denom>0).
+ * `harness_error` cells are always excluded.
+ * @param {object} result
+ * @returns {boolean}
+ */
+export function isScoredCell(result) {
+	const klass = result?.klass ?? classifyCell(result).klass;
+	if (klass === 'harness_error') return false;
+	if (klass === 'agent_fail') return true;
```
soberm, line 200: [question] `agent_fail` cells are included in the headline mean as composite 0 on the grounds that the agent is what's under test. That reasoning holds when the agent genuinely timed out or produced nothing, but the `failed_at='2-agent'` classification is also what a runner OOM or a network outage mid-step produces. That class of infrastructure fault is excluded when it surfaces at an earlier step (`harness_error`), but here it counts against the score. Is there a way to distinguish infra-caused agent timeouts from genuine quality failures, or is the asymmetry considered acceptable because such events are expected to be rare? Just checking whether the design is intentional.

.github/workflows/agent-bench.yml
svidgen: Ultimately this isn't a very high risk, since it should only run on our approval. But, seems trivial enough to safeguard in case the agent gets confused and starts googling for answers. 😅

.github/workflows/agent-bench.yml
```diff
@@ -0,0 +1,562 @@
...
+        if: always()
+        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6
+        with:
+          name: bench-result-${{ matrix.task }}-${{ matrix.template }}
+          path: /tmp/result.json
+          retention-days: 90
+          overwrite: true
+          # If npm ci fails before 0-init writes /tmp/result.json, there's no file
+          # to upload — ignore rather than fail the step (green-regardless).
+          if-no-files-found: ignore
```
Simone319, line 283: `green-regardless` is broken here. These three `if: always()` upload steps — **5. Upload result**, **6. Upload source**, **6b. Upload trace** — set `if-no-files-found: ignore` but not `continue-on-error: true`. `if-no-files-found: ignore` only covers *missing* files; a transient `upload-artifact` API/network error fails the step, and because the step is `if: always()`, that fails the whole job and reds the PR check — breaking the documented "a cell never turns the PR red" invariant.

The S3 persist step further down already sets `continue-on-error: true` — just mirror it on these three upload steps.

.github/workflows/pr-agent-bench.yml
```diff
@@ -0,0 +1,64 @@
...
+#   S3_BUCKET     variable  Registry bucket (existing).
+
+on:
+  # Runs unconditionally on same-repo PR open/update AND on every push to main
+  # (a main run records the merged commit's baseline for later PRs to diff
+  # against). There is NO label gate: the bench is green-regardless — a failed
+  # cell / agent-timeout lands in the run summary, not the check status — so it
+  # is safe to run on every commit. Fork PRs get no secrets/OIDC under
+  # `pull_request` and are additionally blocked by the same-repo `if` below.
+  pull_request:
```
Simone319, line 26: Consider a `paths-ignore` here so unrelated changes don't spin up 10 credentialed Sonnet + Opus cells. We do want docs changes to still trigger the bench, but a PR that only touches the benchmark's own test specs doesn't change framework behavior and doesn't need a full run.

`paths-ignore` skips a run only when *every* changed file matches the list, so this keeps docs (and everything else) triggering while skipping test-spec-only PRs:

```yaml
pull_request:
  types: [opened, synchronize, reopened]
  paths-ignore:
    - 'scripts/agent-bench/tasks/**/test.spec.ts'
```

(Adjust the glob to the actual test-spec path.) Worth documenting expected per-run cost alongside it.

Simone319:
Maybe not in scope for this PR, but we need some contributing guidance on when and how to check the benchmark result. Otherwise, since they are non-blocking, they can be ignored ultimately.

Simone319 (approved):

soberm (approved):

svidgen:
> Maybe not in scope for this PR, but we need some contributing guidance on when and how to check the benchmark result. Otherwise, since they are non-blocking, they can be ignored ultimately.

+1. Let's see what the spread of scores looks like for a week and then set thresholds (accuracy + token count) that we can *hopefully* improve over time.

PR #133, opened 2026-07-01
feat(telemetry): e2e test suite + create-blocks-app worker parity + cross-platform console

sarayev:
## What

Comprehensive end-to-end telemetry test suite that verifies real event delivery across all CLI commands, plus production fixes discovered during testing.

## Production Changes

- **create-blocks-app**: Port telemetry to detached subprocess worker (matches core), remove CI suppression
- **console**: Cross-platform opener (macOS/Linux/Windows) + read region from AWS_REGION
- **core**: Conditional stderr inherit when `NODE_DEBUG=blocks-telemetry` is set (enables delivery verification in tests)

## Test Suite (`test-apps/telemetry`)

- Invokes **real CLI scripts** (not mocks) with `--telemetry-file` for payload assertions
- Verifies **actual delivery** to the telemetry endpoint via `NODE_DEBUG` stderr (`sent (status=200)`)
- Covers: dev, sandbox, sandbox:destroy, deploy, destroy, console, create-blocks-app, vendorize
- Tests SUCCESS + FAIL paths for each command (cloud commands use real CDK deploys with valid creds)
- Disable mechanisms (env var, global config, project config)
- Network resilience, schema validation, privacy, identifier stability
- `blocks-telemetry` CLI consent commands

## CI

- New `Telemetry E2E` job with AWS credentials, 60-min timeout, stack cleanup
- Unique stack names per run (`BLOCKS_STACK_SUFFIX`)
- Pinned installation ID for deterministic assertions


packages/create-blocks-app/src/telemetry.ts
```diff
@@ -61,7 +62,6 @@ function isTelemetryEnabled(): boolean {
     if (config.telemetry?.enabled === true) return true;
   } catch { /* no config */ }
 
-  if (isCI()) return false;
```
svidgen, line 64: Can we distinguish between "human" and CI use in the dashboard?

svidgen (dismissed):

packages/create-blocks-app/src/telemetry.ts
sarayev: Yes, we can. Here, this check was preventing from collecting events in ci for the create-blocks-app command, which I am removing now.

packages/create-blocks-app/src/telemetry-send-worker.ts
```diff
@@ -0,0 +1,60 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
soberm, line 1: [suggestion] This worker is a near-identical copy of the one in `packages/core`. The "no project imports" constraint is real, but you now have two files to keep in sync — timeout value, debug output format, error messages. If the endpoint URL scheme or status-code handling ever changes, the fix needs to land in both places. At minimum a comment cross-referencing the core copy would make the intentional duplication visible; ideally these could share a single source artifact at build time.

packages/create-blocks-app/src/telemetry-send-worker.ts
```diff
@@ -0,0 +1,60 @@
...
+ * Reads JSON payload from stdin, POSTs it to the endpoint (argv[2]).
+ *
+ * Uses only Node built-ins — no project imports — so the compiled .js
+ * runs with bare `node` (no tsx needed).
+ */
+
+import { request as httpsRequest } from 'node:https';
+import { request as httpRequest } from 'node:http';
+
+const TIMEOUT_MS = 500;
```
soberm, line 21: [nit] Both workers use 500ms, but since they live in separate files there is nothing enforcing they stay in sync. A comment like `// matches TIMEOUT_MS in packages/core/src/telemetry/telemetry-send-worker.ts` would make the coupling obvious to whoever touches this next.

packages/create-blocks-app/src/telemetry.ts
```diff
@@ -61,7 +62,6 @@ function isTelemetryEnabled(): boolean {
...
   return true;
```
soberm, line 65: [suggestion] Removing the `isCI()` guard means `create-blocks-app` CI runs will now send real events to the production endpoint — not just the telemetry E2E suite (which seeds a pinned installation ID), but every other workflow that invokes it too. Is there a server-side filter on the `ci: true` field, or a dashboard exclusion, so this data doesn't skew metrics?

packages/create-blocks-app/src/telemetry.ts
```diff
@@ -175,29 +175,26 @@ function sendEvent(event: Record<string, unknown>): void {
...
+    const dir = dirname(fileURLToPath(import.meta.url));
+    const workerPath = join(dir, 'telemetry-send-worker.js');
+    const child = spawn(process.execPath, [workerPath, endpoint], {
+      detached: true,
+      stdio: ['pipe', 'ignore', process.env.NODE_DEBUG?.includes('blocks-telemetry') ? 'inherit' : 'ignore'],
+      // Clear NODE_OPTIONS so inherited flags don't interfere
+      env: { ...process.env, NODE_OPTIONS: '' },
+    });
+
+    child.stdin!.on('error', (err) => { debug('stdin write failed: %s', err.message); });
```
soberm: [nit] `child.stdin!` — the non-null assertion is technically correct (stdio `'pipe'` guarantees a writable stdin), but `child.stdin?.write(payload); child.stdin?.end()` is consistent with the rest of the defensive style in this file. Same pattern in core's `client.ts` if you want to align both.

packages/core/src/scripts/console.ts
```diff
@@ -10,6 +10,33 @@ export interface ConsoleOptions {
   outputsFile?: string;
 }
 
+function resolveRegion(): string {
+  const fromEnv = process.env.AWS_REGION || process.env.AWS_DEFAULT_REGION;
```
soberm, line 14: [nit] `process.env.AWS_REGION || process.env.AWS_DEFAULT_REGION` — matches SDK precedence so almost certainly right. Just worth noting that if someone has a stale `AWS_REGION` in their shell but is actively using `AWS_DEFAULT_REGION` to target a different region, they'll get a silently wrong console URL with no warning.

package.json
```diff
@@ -88,6 +89,7 @@
     "monte-carlo:cached": "tsx scripts/monte-carlo.ts --cached",
     "canary": "tsx scripts/canary/run-canary.ts",
     "test:e2e:supabase": "node --test packages/bb-data/dist/crud/crud-e2e.test.js packages/bb-data/dist/crud/migrations-e2e.test.js",
+    "test:telemetry": "tsx test-apps/telemetry/test/telemetry-e2e.test.ts",
```
soberm, line 92: [suggestion] This runs from the monorepo root. If anything in the test file resolves paths via `process.cwd()` it would silently pick up the wrong base. Worth confirming that `APP_ROOT` and every other path in the test is derived from `import.meta.url`/`__dirname` rather than cwd. (Given the number of path-related fixes in the commit history, just wanted it called out explicitly.)

test-apps/telemetry/README.md
```diff
@@ -0,0 +1,69 @@
...
+## What's tested
+
+- **`--telemetry-file` emission + attributes**: blocks version, template
+  name/version, os, ci, per-command name, and building-block counters
+  (official BB names + version, custom BB count, total blocks count).
+- **Identifier creation**: `installationId` and `projectId` are written when
+  they do not already exist, and emitted events carry them.
+- **Per-command success + failure**: `deploy`, `destroy`, `sandbox`,
+  `sandbox:destroy`, `console`, `create-blocks-app`, `vendorize`, and `dev`
+  emit correct SUCCESS and FAIL events (with error code/phase).
```
soberm: [minor] This bullet overclaims. `console` has a SUCCESS test only — there's no FAIL path. `create-blocks-app` and `vendorize` FAIL tests don't assert a FAIL event; they assert *no telemetry is emitted* and explicitly document known `process.exit`/pre-`trackCommand` limitations. Worth tightening to something like: "SUCCESS + FAIL events for `deploy`, `destroy`, `sandbox`, `sandbox:destroy`, and `dev`; `console` SUCCESS only; `create-blocks-app` and `vendorize` document the no-telemetry limitation for their failure paths."

test-apps/telemetry/test/telemetry-e2e.test.ts
```diff
@@ -0,0 +1,1171 @@
...
+async function waitForFile(filePath: string, timeoutMs = 5_000): Promise<boolean> {
+  const start = Date.now();
+  while (Date.now() - start < timeoutMs) {
+    if (existsSync(filePath)) return true;
+    await sleep(150);
+  }
+  return existsSync(filePath);
+}
+
+function readTelemetryFile(filePath: string): Record<string, any> {
```
soberm: [nit] Return type is `Record<string, any>`, so field accesses like `body.event.command` and `body.identifiers.projectId` are untyped throughout the suite. For a test whose explicit job is schema validation, a typed interface would catch field-name typos at compile time rather than silently returning `undefined`. Something minimal like:
```ts
interface TelemetryPayload {
  event: { command: string; state: string; error?: unknown };
  identifiers: { installationId: string; projectId: string };
}
```
would cover the assertions already in place.

test-apps/telemetry/test/telemetry-e2e.test.ts
```diff
@@ -0,0 +1,1171 @@
...
+  const content = readFileSync(filePath, 'utf-8');
+  const parsed = JSON.parse(content);
+  return Array.isArray(parsed) ? parsed[0] : parsed;
+}
+
+/** Assert that the event was delivered to the real endpoint. */
+/** Assert that the event was delivered to the real endpoint. */
```
soberm: [nit] Duplicate JSDoc — lines 209 and 210 are identical. Delete one.

soberm (commented):
Reviewed commit 88d9373. Production fixes look correct; main feedback is on the duplicate worker, the CI suppression removal, and a handful of test/doc cleanups. Nothing blocking.

packages/create-blocks-app/src/telemetry-send-worker.ts
sarayev: Intentional duplication: the worker runs as a standalone subprocess with zero project/package imports, so create-blocks-app can't import packages/core's copy. I've added reciprocal cross-reference comments at the top of both files (and on the shared timeout) so the coupling is visible and edits get mirrored. Collapsing them into a single build-time source artifact is the right long-term fix but needs a build-step change that's out of scope here — tracking it as a follow-up.

packages/create-blocks-app/src/telemetry-send-worker.ts
sarayev: Done — added `// Matches the timeout in packages/core/src/telemetry/telemetry-send-worker.ts — keep in sync.`

packages/create-blocks-app/src/telemetry.ts
sarayev: Confirmed this won't skew production metrics: telemetry events carry a `ci` flag, and the dashboards exclude CI data. So CI-originated events (from any workflow, not just the pinned-ID E2E suite) are filtered out at the dashboard layer.

packages/create-blocks-app/src/telemetry.ts
sarayev: Done — switched to `child.stdin?.write(payload); child.stdin?.end();` and aligned the identical pattern in core/src/telemetry/client.ts for consistency.

packages/core/src/scripts/console.ts
sarayev: Good catch — the `AWS_REGION || AWS_DEFAULT_REGION` precedence intentionally matches the SDK, so it's correct for the common case. Agreed the stale-`AWS_REGION` scenario would produce a silently wrong console URL; leaving the SDK-consistent behavior as-is for now, and can add a mismatch warning as a follow-up if it comes up.

package.json
sarayev: Verified — no change needed. All paths derive from `import.meta.url` (`__dirname` → `APP_ROOT`/`REPO_ROOT`) and `spawnCli` defaults `cwd` to `APP_ROOT`; there are no `process.cwd()` usages, so paths are stable regardless of invocation directory (including the monorepo root).

test-apps/telemetry/README.md
sarayev: Fixed — updated to: "SUCCESS + FAIL events for `deploy`, `destroy`, `sandbox`, `sandbox:destroy`, and `dev`; `console` SUCCESS only; `create-blocks-app` and `vendorize` document the no-telemetry limitation for their failure paths."

test-apps/telemetry/test/telemetry-e2e.test.ts
sarayev: Done — replaced `Record<string, any>` with a typed `TelemetryPayload` interface covering event/identifiers/environment/product/counters; `tsc --noEmit --strict` is clean, so field-name typos now fail at compile time.

test-apps/telemetry/test/telemetry-e2e.test.ts
sarayev: Done — removed the duplicated JSDoc line.

sarayev:
Thanks for the thorough pass! Addressed all the nits/minor items in 47ef470d. Left the send-worker de-duplication as a tracked follow-up (rationale in-thread), and confirmed the isCI() removal is safe on the metrics side (events carry a ci flag; dashboards exclude CI data). Replies inline per comment.

soberm (approved):

ahmedhamouda78 (commented):
8 comments below: 2 major, 3 minor, 3 nits. The two majors are the ones I'd resolve before merge; the rest are suggestions/nits.

test-apps/telemetry/test/telemetry-e2e.test.ts
```diff
@@ -0,0 +1,1183 @@
...
+
+function readTelemetryFile(filePath: string): TelemetryPayload {
+  const content = readFileSync(filePath, 'utf-8');
+  const parsed = JSON.parse(content);
+  return Array.isArray(parsed) ? parsed[0] : parsed;
+}
+
+/** Assert that the event was delivered to the real endpoint. */
+function assertDelivered(stderr: string, description = ''): void {
+  assert.match(stderr, SENT_REGEX, `Telemetry should be delivered to endpoint. ${description}\nstderr: ${stderr.slice(-500)}`);
```
ahmedhamouda78, line 224: [major] Delivery tests don't override `BLOCKS_TELEMETRY_ENDPOINT`, so they hit the production default (`https://blocks-telemetry.us-east-1.api.aws/metrics`, per `packages/core/src/telemetry/client.ts:16-20`) using the pinned installation ID `00000000-0000-0000-0000-000000000e2e` seeded in the workflow. Every source-changed same-repo PR posts real synthetic events to production telemetry. Confirm the backend excludes that ID from metrics, or route delivery verification to a test endpoint. I'd gate merge on this.

.github/workflows/pr-checks.yml
```diff
@@ -474,13 +474,62 @@ jobs:
...
+
+      - name: Telemetry E2E Tests
+        run: npm run test:telemetry
+
+      - name: Cleanup stacks
+        if: always()
+        run: |
+          cd test-apps/telemetry
+          npx tsx aws-blocks/scripts/sandbox-destroy.ts || true
+          npx tsx aws-blocks/scripts/destroy.ts || true
```
ahmedhamouda78, line 524: [major] Both destroys end in `|| true`, so a failed teardown is swallowed. Stacks are unique per run (`BLOCKS_STACK_SUFFIX: telemetry-<pr>-<attempt>` + a random sandbox id), so leaks are distinct and cumulative -> ongoing AWS cost. Fail the job (or emit a warning annotation) on cleanup error, and/or add a scheduled sweeper for stale `bb-telemetry-e2e-*` stacks (or tag `blocks:purpose=telemetry-e2e`).

packages/create-blocks-app/src/telemetry.ts
ahmedhamouda78, line 65: [minor] Removing `if (isCI()) return false;` here is a user-visible change: consumers running `create-blocks-app` in their own CI now emit telemetry that was previously suppressed. Intentional and consistent with core (`consent.ts` documents "CI environments do NOT suppress telemetry"), but please surface it in the user-facing consent docs and release notes, not just the changeset. (No dead code: `isCI()` is still used as the `ci:` event attribute.)

packages/create-blocks-app/src/telemetry-send-worker.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Reads JSON payload from stdin, POSTs it to the endpoint (argv[2]).
+ *
+ * Uses only Node built-ins — no project imports — so the compiled .js
+ * runs with bare `node` (no tsx needed).
+ */
+
+import { request as httpsRequest } from 'node:https';
+import { request as httpRequest } from 'node:http';
+
+// Matches the timeout in packages/core/src/telemetry/telemetry-send-worker.ts — keep in sync.
```
ahmedhamouda78, line 20: [minor] This worker is byte-identical to `packages/core/src/telemetry/telemetry-send-worker.ts` except this header/comment, so manual "keep in sync" will drift. Since they're already identical modulo the header, add a CI check asserting byte-equality (ignoring the header comment), or generate one from the other at build time so core is the single source of truth.

packages/create-blocks-app/src/telemetry-send-worker.ts
```diff
@@ -0,0 +1,66 @@
...
+const TIMEOUT_MS = 500;
+const debug = (process.env.NODE_DEBUG || '').includes('blocks-telemetry');
+
+const endpoint = process.argv[2];
+if (!endpoint) process.exit(1);
+
+let payload = '';
+process.stdin.setEncoding('utf-8');
+process.stdin.on('data', (chunk: string) => { payload += chunk; });
+process.stdin.on('end', () => {
```
ahmedhamouda78, line 30: [minor] The request is fired only inside the `end` handler, and `timeout: TIMEOUT_MS` is the HTTP request timeout, not a stdin watchdog -- if `end` never fires the process has no self-kill. Low risk (detached/`unref`'d, tiny payload, parent exit normally closes the pipe), but a top-level `setTimeout(() => process.exit(1), TIMEOUT_MS * 2).unref()` is cheap insurance.

packages/core/src/telemetry/client.ts
```diff
@@ -199,16 +199,16 @@ export function sendEvent(event: BlocksTelemetryEvent): void {
     const workerPath = path.join(dir, 'telemetry-send-worker.js');
     const child = spawn(process.execPath, [workerPath, endpoint], {
       detached: true,
-      stdio: ['pipe', 'ignore', 'ignore'],
+      stdio: ['pipe', 'ignore', process.env.NODE_DEBUG?.includes('blocks-telemetry') ? 'inherit' : 'ignore'],
```
ahmedhamouda78, line 202: [nit] `process.env.NODE_DEBUG?.includes('blocks-telemetry')` is a looser substring match than the `debuglog('blocks-telemetry')` token semantics used for the `debug` fn (e.g. `NODE_DEBUG=blocks-telemetry-foo` flips stdio to `inherit` but wouldn't activate `debuglog`). The same expression is duplicated in `packages/create-blocks-app/src/telemetry.ts`. Consider a tiny shared helper for the two non-worker sites (workers stay import-free).

packages/core/src/scripts/console.ts
```diff
@@ -10,6 +10,33 @@ export interface ConsoleOptions {
...
+  return 'us-east-1';
+}
+
+/** Launch the URL in the default browser. Best-effort: no opener (headless/CI) is not a failure. */
+function openInBrowser(url: string): void {
+  const opener =
+    process.platform === 'darwin' ? 'open' :
+    process.platform === 'win32' ? 'cmd' :
+    'xdg-open';
+  const args = process.platform === 'win32' ? ['/c', 'start', '', url] : [url];
```
ahmedhamouda78, line 31: [nit] `cmd /c start "" <url>` (the empty-title arg) is correct today, but if the CloudFormation URL template ever gains `&`-joined query params, `cmd`'s `start` will break on them. Add a comment noting the constraint, or switch to PowerShell `Start-Process`.

test-apps/telemetry/test/telemetry-e2e.test.ts
```diff
@@ -0,0 +1,1183 @@
...
+const CREATE_BLOCKS_APP_BIN = join(REPO_ROOT, 'packages', 'create-blocks-app', 'dist', 'index.js');
+// The blocks-telemetry consent CLI is the `blocks-telemetry` bin of @aws-blocks/core.
+// `npx blocks-telemetry` depends on the workspace bin symlink + exec bit surviving the
+// build step, which is fragile in CI; invoke the built entrypoint directly with `node`.
+const BLOCKS_TELEMETRY_CLI = join(REPO_ROOT, 'packages', 'core', 'dist', 'scripts', 'telemetry-cli.js');
+
+const PINNED_INSTALLATION_ID = '00000000-0000-0000-0000-000000000e2e';
+const PINNED_PROJECT_ID = '00000000-0000-0000-0000-0000000e2e57';
+const UUID_REGEX = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+const SENT_REGEX = /BLOCKS-TELEMETRY: sent \(status=200\)/;
```
ahmedhamouda78, line 47: [nit] These `status=200` assertions couple CI green to the live endpoint's uptime, risking flakiness on unrelated PRs. The infra to fix already exists -- the network-resilience test sets `BLOCKS_TELEMETRY_ENDPOINT` to a local address, so pointing the per-command matrix at a `127.0.0.1` sink is a small change; keep one real-endpoint smoke test.

test-apps/telemetry/test/telemetry-e2e.test.ts
sarayev: This is actually the existing repo-wide behavior, not something this suite introduces. Telemetry emits unconditionally in CI (consent.ts: "CI environments do NOT suppress telemetry" — only AWS_BLOCKS_DISABLE_TELEMETRY=1/project/global config disable it), no workflow overrides BLOCKS_TELEMETRY_ENDPOINT anywhere (grep .github/ → 0 hits), and six other e2e jobs (e2e-templates/sandbox/sandbox-vpc/production/supabase/hosting) already seed the same pinned ID …000000000e2e via ./.github/actions/seed-telemetry-id and run the real CLI against the prod endpoint. The backend also flags and excludes CI data (and this pinned ID) from metrics, so there's no metric pollution. So I'd treat changing this as a repo-wide follow-up rather than gating this PR. (The delivery-assertion flakiness angle from your :47 comment is the part genuinely worth tightening — tracking that as a follow-up.)

.github/workflows/pr-checks.yml
sarayev: The best-effort teardown here mirrors the existing convention: e2e-hosting's cleanup step uses the identical `if: always()` + `|| true` pattern (line 475), and the other e2e jobs rely on in-script cleanup + the separate cleanup-stacks.yml sweeper. Stacks are uniquely named per run, so a genuinely stronger story (fail-on-cleanup-error and/or a scheduled sweeper for bb-telemetry-e2e-*) is a reasonable improvement — but to stay consistent with the repo's current approach I'd do that as a follow-up (ideally across all e2e jobs, not just this one). Tracking it.

packages/create-blocks-app/src/telemetry.ts
sarayev: Our telemetry docs already clearly state that we collect telemetry and that collection is not disabled in CI environments — that's the documented, intended behavior. The previous isCI() suppression in create-blocks-app was inconsistent with both that documentation and the core package (consent.ts), so removing it is a bug fix that brings create-blocks-app in line with the documented behavior. It doesn't introduce a new user-visible policy, so no special customer-facing wording or release-note callout is needed beyond the changeset. (Confirmed no dead code — isCI() still feeds the ci: event attribute.)

packages/create-blocks-app/src/telemetry-send-worker.ts
sarayev: Good point — the two workers are identical modulo the header comment, so manual "keep in sync" will drift. Adding a CI byte-equality check (ignoring the header) or generating one from the other at build time (core as SSOT) is the right call. This is part of the send-worker de-duplication follow-up we're already tracking; I'll fold the drift guard into it.

packages/create-blocks-app/src/telemetry-send-worker.ts
sarayev: Fair — `timeout: TIMEOUT_MS` is the HTTP request timeout, not a stdin watchdog, so if `end` never fires there's no self-kill. Risk is low (detached/unref'd, tiny payload, parent exit closes the pipe), but a top-level `setTimeout(() => process.exit(1), TIMEOUT_MS * 2).unref()` is cheap insurance. Adding it to the follow-up.

packages/core/src/telemetry/client.ts
sarayev: Noted — NODE_DEBUG?.includes('blocks-telemetry') is looser than debuglog's token semantics (e.g. blocks-telemetry-foo would flip stdio but not activate debuglog), and it's duplicated in create-blocks-app's telemetry. A tiny shared helper for the two non-worker sites makes sense; rolling it into the follow-up.

packages/core/src/scripts/console.ts
sarayev: Good defensive catch — `cmd /c start "" <url>` is fine for today's console URLs, but would break if the URL template ever gains &-joined query params. I'll add a constraint comment (or switch to PowerShell Start-Process) as part of the follow-up.

test-apps/telemetry/test/telemetry-e2e.test.ts
sarayev: This is the one I most agree is worth tightening: unlike the other e2e jobs (which fire-and-forget telemetry), our suite asserts delivery status=200, which couples CI green to the live endpoint's uptime and can flake on unrelated PRs. The infra already exists — the network-resilience test points BLOCKS_TELEMETRY_ENDPOINT at a local sink, so I'll route the per-command delivery matrix at a 127.0.0.1 sink and keep a single real-endpoint smoke test. Tracking as a follow-up.

sarayev:
Thanks for the detailed review! Dispositions: the two majors (#1 prod endpoint, #2 `|| true` teardown) are pre-existing repo-wide conventions shared by the other e2e jobs (evidence in-thread), so I'd address those repo-wide as a follow-up rather than in this PR. #3 (CI suppression) is a bug fix aligning create-blocks-app with already-documented behavior — no customer-facing wording needed. The remaining items (#4 worker drift guard, #5 watchdog, #6 NODE_DEBUG helper, #7 Windows start, #8 local delivery sink) are bundled into a tracked follow-up; #8 (delivery-assertion flakiness) is the one I most want to land.

ahmedhamouda78 (approved):

PR #136, opened 2026-07-02
fix(core): robust dev server — startup port reclaim, singleton guard, :3000 EADDRINUSE handling

hfurkanbozkurt:
## Summary

Hardens the `@aws-blocks/core` dev server (`packages/core/src/scripts/dev-server.ts`) against the fragility surfaced by the agent bench (#132). PR #80 fixed the single‑supervisor self‑restart loop (detached process group + `killFrontendTree` + `waitForPortFree(:3100)` + bounded respawn). This PR fixes the **four problems that remained after #80**, all of which cause a contended dev server to either silently never serve or fight itself into a restart loop.

All of #80's single‑supervisor hardening and #125's `0.0.0.0` bind are preserved.

### 1. No singleton guard → two supervisors fight
Templates run `dev` = `tsx watch aws-blocks/scripts/server.ts`. A second `npm run dev` spawned a **second supervisor**: the backend hit `:3000` EADDRINUSE, its `--strictPort` Vite hit `:3100` EADDRINUSE, and the two supervisors fought (killing one Vite → the survivor respawned it) → restart loop.

**Fix — pidfile singleton guard.** On startup the server writes `.blocks-sandbox/dev-server.<port>.pid` containing `{ pid, ppid, port, startedAt }`. Before writing, it reads any existing pidfile and runs a pure `evaluateSingleton(record, currentPpid)` decision:
- `record.ppid === process.ppid` → this is a **tsx‑watch hot‑reload relaunch of our own script** → **proceed** (hot reload is never blocked).
- else if `isPidAlive(record.pid)` → a real competing supervisor → **exit(0)** with a clear message: `dev server already running on :PORT (pid N)`.
- else (stale pidfile, dead pid) → **proceed**.

The pidfile is removed in both `cleanup()` and the `exit` handler, and only if it still points at our own pid (so a relaunch never deletes the successor's file).

### 2. `:3000` front door had no EADDRINUSE robustness
The backend's only handling of a contended `:3000` was emitting a telemetry event — no console error, no reclaim/retry, no non‑zero exit → a contended front door **silently never served**.

**Fix — same treatment `:3100` already gets.** `server.listen` is now wrapped in a bounded retry loop (`DEFAULT_PORT_BIND_RETRY_POLICY` = 3 attempts). On EADDRINUSE it emits a **real `console.error`**, calls `reclaimPort(:3000)`, waits, and retries (`evaluatePortBindRetry` is the pure decision). The telemetry event is still emitted. On unrecoverable failure it `console.error`s a clear message and **exits non‑zero (1)**.

### 3. No startup reclaim → orphaned listeners crash a fresh start
A fresh `server.ts` never freed a pre‑existing `:3000`/`:3100` listener; it relied on the previous process's `cleanup()` finishing within tsx‑watch's ~5s. If that process was SIGKILL'd/crashed (or was a different invocation), the orphan persisted and `spawnFrontend` (`--strictPort`) collided.

**Fix — startup reclaim.** Before binding `:3000` and before `spawnFrontend` on `:3100`, the server calls `reclaimPort(port)`: probe the port (`isPortOpen`), and if held, kill the listener's process group and `waitForPortFree`. This **reuses the existing mechanisms** — `waitForPortFree` and the process‑group kill primitives already in the codebase (`killSignal`/`windowsTreeKill` via the new `findListenerPids`/`killListenerTree` in `process-tree.ts`, which mirror the `lsof -ti:<port>` approach already used by `cleanup.ts`). No new kill mechanism was invented; it mirrors the respawn path.

### 4. `--strictPort` turned contention into a hard crash
**Decision: keep `--strictPort`, rely on startup reclaim.** Startup reclaim now guarantees `:3100` is free *before* `spawnFrontend` runs, so `--strictPort` is safe and preserves the single‑front‑door model (backend proxy targets a fixed `:3100`). Dropping `--strictPort` would let Vite auto‑pick a port, which would require the backend to dynamically **learn** Vite's chosen port and rewrite its proxy target — an invasive change to the proxy logic for little benefit now that reclaim exists. If reclaim ever fails (e.g. the port is held by a non‑killable process), the failure is now surfaced with a clear console error + non‑zero exit (problem #2) instead of a silent crash.

## Tests

New file `packages/core/src/scripts/dev-server-reclaim.test.ts` — **29 tests, no mocks**, using real TCP sockets and injectable dependency seams:
- `isPortOpen` against a real free/bound port.
- `reclaimPort`: port already free; real stale listener reclaimed; port stays open after kill → `reclaimed: false`.
- `evaluatePortBindRetry` honoring `DEFAULT_PORT_BIND_RETRY_POLICY`.
- `parsePidRecord` (valid / invalid JSON / missing fields → `null`, never throws).
- `evaluateSingleton`: dead pid → proceed; live pid with matching ppid → proceed (hot reload); live pid different ppid → exit.

Also smoke‑tested the real server end‑to‑end: starting on a port held by a SIGKILL‑style orphan logs `♻️ Reclaimed port …` and serves; a second start exits 0 with `dev server already running on :PORT` while the first keeps serving.

## Verification
- `npm run build -w packages/core` — clean.
- `npm run test -w packages/core` — **573 pass, 0 fail** (incl. the 29 new tests; full dev‑server suites incl. `RUN_SLOW_TESTS=1`).
- `npm run lint` (biome gate) — exit 0 on touched files.
- Changeset added: `.changeset/dev-server-robustness.md` bumping `@aws-blocks/core` (patch).

## Follow‑up (not in this PR)
Problem #5 — tsx‑watch performing a full restart on **every** edit — is out of scope here (non‑trivial; would need `tsx watch` scoping or an in‑process reload). Tracking as a follow‑up.

Related: #78 (closed), #80, #125. Surfaced by the agent bench (#132).

### DIFF
--- a/packages/core/src/scripts/dev-server-reclaim.test.ts
+++ b/packages/core/src/scripts/dev-server-reclaim.test.ts
@@ -5,6 +5,7 @@ import assert from 'node:assert';
 import { createServer } from 'node:net';
 import {
   reclaimPort,
+  reclaimMessage,
   evaluatePortBindRetry,
   DEFAULT_PORT_BIND_RETRY_POLICY,
   evaluateSingleton,
@@ -104,6 +105,26 @@ describe('reclaimPort — frees a stale/orphaned port before startup', () => {
     assert.strictEqual(result.reclaimed, true);
   });
 
+  it('SIGKILLs the CURRENT owner, not the stale one, when the port changed hands during the wait', async () => {
+    // The original listener (111) is SIGTERM'd but exits, and a DIFFERENT process
+    // (222) grabs the port before the SIGKILL pass. The escalation must re-list
+    // and target the new owner (222) — never the stale pid (111) it already
+    // signalled — otherwise it SIGKILLs a dead pid and leaves the real owner.
+    const kills: Array<[number, NodeJS.Signals]> = [];
+    let call = 0;
+    const result = await reclaimPort(3000, {
+      // open, still open after SIGTERM (new owner now holds it), free after SIGKILL
+      probe: scriptedProbe([true, true, false]),
+      listPids: () => (call++ === 0 ? [111] : [222]),
+      killTree: (pid, sig) => kills.push([pid, sig]),
+      waitFree: async () => {},
+    });
+    assert.deepStrictEqual(kills, [[111, 'SIGTERM'], [222, 'SIGKILL']]);
+    assert.strictEqual(result.reclaimed, true);
+    // result.pids reflects the initial discovery (what we SIGTERM'd).
+    assert.deepStrictEqual(result.pids, [111]);
+  });
+
   it('reports the port free once its real listener is reclaimed (real sockets, injected kill)', async () => {
     const port = await getFreePort();
     const srv = createServer((c) => c.destroy());
@@ -123,6 +144,25 @@ describe('reclaimPort — frees a stale/orphaned port before startup', () => {
     assert.strictEqual(result.wasOpen, true);
     assert.strictEqual(result.reclaimed, true);
     assert.strictEqual(await isPortOpen(port, '127.0.0.1'), false, 'port should be free after reclaim');
+  });
+});
+
+// ── reclaimMessage — accurate startup reclaim-outcome messaging ──────────────
+describe('reclaimMessage — three-way reclaim outcome message', () => {
+  it('reports success when the port was reclaimed', () => {
+    const msg = reclaimMessage(3000, { wasOpen: true, reclaimed: true, pids: [] }, 'a stale/orphaned listener');
+    assert.match(msg, /Reclaimed port 3000 from a stale\/orphaned listener/);
+  });
+
+  it('names the holding pid(s) when reclaim failed but an owner is known', () => {
+    const msg = reclaimMessage(3100, { wasOpen: true, reclaimed: false, pids: [4242, 4243] }, 'a stale/orphaned dev server');
+    assert.match(msg, /Port 3100 held by pid\(s\) \[4242, 4243\]/);
+    assert.match(msg, /stop that process and retry/);
+  });
+
+  it('falls back to the generic message when reclaim failed with no owner PID', () => {
+    const msg = reclaimMessage(3000, { wasOpen: true, reclaimed: false, pids: [] }, 'a stale/orphaned listener');
+    assert.match(msg, /no owner PID found/);
   });
 });
 
--- a/packages/core/src/scripts/dev-server-reclaim.test.ts
+++ b/packages/core/src/scripts/dev-server-reclaim.test.ts
@@ -1,0 +1,290 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+import { createServer } from 'node:net';
+import {
+  reclaimPort,
+  evaluatePortBindRetry,
+  DEFAULT_PORT_BIND_RETRY_POLICY,
+  evaluateSingleton,
+  parsePidRecord,
+  isPidAlive,
+  isPortOpen,
+  type DevServerPidRecord,
+} from './dev-server.js';
+import { findListenerPids, killListenerTree } from './process-tree.js';
+
+function getFreePort(): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const srv = createServer();
+    srv.on('error', reject);
+    srv.listen(0, '127.0.0.1', () => {
+      const addr = srv.address();
+      const port = typeof addr === 'object' && addr ? addr.port : 0;
+      srv.close(() => resolve(port));
+    });
+  });
+}
+
+// ── reclaimPort — startup / EADDRINUSE port reclaim policy ──────────────────
+// Fully injected: no real processes are spawned. `probe` is scripted to model
+// the port's open/closed state across the reclaim sequence, so we assert exactly
+// which signals are sent and when it gives up.
+describe('reclaimPort — frees a stale/orphaned port before startup', () => {
+  // reclaimPort probes the port up to three times: (1) initial check,
+  // (2) after the SIGTERM wait, (3) the final reclaimed? check.
+  function scriptedProbe(states: boolean[]): (port: number) => Promise<boolean> {
+    let i = 0;
+    return async () => states[Math.min(i++, states.length - 1)];
+  }
+
+  it('is a no-op when the port is already free (no discovery, no kills)', async () => {
+    const kills: Array<[number, NodeJS.Signals]> = [];
+    let listed = 0;
+    const result = await reclaimPort(3000, {
+      probe: scriptedProbe([false]),
+      listPids: () => { listed++; return [111]; },
+      killTree: (pid, sig) => kills.push([pid, sig]),
+      waitFree: async () => {},
+    });
+    assert.deepStrictEqual(result, { wasOpen: false, reclaimed: true, pids: [] });
+    assert.strictEqual(listed, 0, 'must not discover PIDs when the port is free');
+    assert.deepStrictEqual(kills, []);
+  });
+
+  it('SIGTERMs every listener and reports reclaimed when the port frees gracefully', async () => {
+    const kills: Array<[number, NodeJS.Signals]> = [];
+    const result = await reclaimPort(3100, {
+      // open, then free after SIGTERM, then still free at the final check
+      probe: scriptedProbe([true, false, false]),
+      listPids: () => [4242, 4243],
+      killTree: (pid, sig) => kills.push([pid, sig]),
+      waitFree: async () => {},
+    });
+    assert.deepStrictEqual(kills, [[4242, 'SIGTERM'], [4243, 'SIGTERM']]);
+    assert.deepStrictEqual(result, { wasOpen: true, reclaimed: true, pids: [4242, 4243] });
+  });
+
+  it('escalates to SIGKILL when the listener survives the graceful SIGTERM', async () => {
+    const kills: Array<[number, NodeJS.Signals]> = [];
+    const result = await reclaimPort(3000, {
+      // open, still open after SIGTERM, free after SIGKILL
+      probe: scriptedProbe([true, true, false]),
+      listPids: () => [777],
+      killTree: (pid, sig) => kills.push([pid, sig]),
+      waitFree: async () => {},
+    });
+    assert.deepStrictEqual(kills, [[777, 'SIGTERM'], [777, 'SIGKILL']]);
+    assert.strictEqual(result.reclaimed, true);
+  });
+
+  it('reports reclaimed:false when the port stays bound through SIGKILL', async () => {
+    const result = await reclaimPort(3000, {
+      probe: scriptedProbe([true, true, true]),
+      listPids: () => [777],
+      killTree: () => {},
+      waitFree: async () => {},
+    });
+    assert.strictEqual(result.wasOpen, true);
+    assert.strictEqual(result.reclaimed, false);
+  });
+
+  it('re-discovers listeners for the SIGKILL pass when the first discovery was empty', async () => {
+    const kills: Array<[number, NodeJS.Signals]> = [];
+    let call = 0;
+    const result = await reclaimPort(3000, {
+      probe: scriptedProbe([true, true, false]),
+      // First lsof pass momentarily returns nothing; second finds the owner.
+      listPids: () => (call++ === 0 ? [] : [999]),
+      killTree: (pid, sig) => kills.push([pid, sig]),
+      waitFree: async () => {},
+    });
+    assert.deepStrictEqual(kills, [[999, 'SIGKILL']]);
+    assert.strictEqual(result.reclaimed, true);
+  });
+
+  it('reports the port free once its real listener is reclaimed (real sockets, injected kill)', async () => {
+    const port = await getFreePort();
+    const srv = createServer((c) => c.destroy());
+    await new Promise<void>((res) => srv.listen(port, '127.0.0.1', () => res()));
+
+    assert.strictEqual(await isPortOpen(port, '127.0.0.1'), true, 'port should be held before reclaim');
+
+    // Inject the "kill" as closing our test server so the real probe/waitFree
+    // path is exercised end-to-end without spawning an OS process.
+    let killed = false;
+    const result = await reclaimPort(port, {
+      probe: (p) => isPortOpen(p, '127.0.0.1'),
+      listPids: () => [process.pid],
+      killTree: () => { if (!killed) { killed = true; srv.close(); } },
+    });
+
+    assert.strictEqual(result.wasOpen, true);
+    assert.strictEqual(result.reclaimed, true);
+    assert.strictEqual(await isPortOpen(port, '127.0.0.1'), false, 'port should be free after reclaim');
+  });
+});
+
+// ── evaluatePortBindRetry — bounded :3000 EADDRINUSE retry ───────────────────
+describe('evaluatePortBindRetry — front-door bind retry budget', () => {
+  it('retries early attempts with a linear backoff', () => {
+    assert.deepStrictEqual(evaluatePortBindRetry(1), { retry: true, delayMs: 250 });
+    assert.deepStrictEqual(evaluatePortBindRetry(2), { retry: true, delayMs: 500 });
+  });
+
+  it('gives up (no retry) once the attempt budget is reached', () => {
+    const d = evaluatePortBindRetry(DEFAULT_PORT_BIND_RETRY_POLICY.maxAttempts);
+    assert.deepStrictEqual(d, { retry: false, delayMs: 0 });
+  });
+
+  it('honors a custom policy', () => {
+    const policy = { maxAttempts: 2, backoffMs: 100 };
+    assert.deepStrictEqual(evaluatePortBindRetry(1, policy), { retry: true, delayMs: 100 });
+    assert.deepStrictEqual(evaluatePortBindRetry(2, policy), { retry: false, delayMs: 0 });
+  });
+});
+
+// ── evaluateSingleton — singleton guard decision ────────────────────────────
+// Prevents two fighting supervisors while never blocking a `tsx watch` reload of
+// the SAME supervisor (same parent pid).
+describe('evaluateSingleton — one supervisor per port, hot-reload safe', () => {
+  const rec = (over: Partial<DevServerPidRecord> = {}): DevServerPidRecord => ({
+    pid: 1000,
+    ppid: 500,
+    port: 3000,
+    ...over,
+  });
+  const alive = () => true;
+  const dead = () => false;
+
+  it('proceeds when there is no (or a corrupt) pidfile', () => {
+    assert.deepStrictEqual(evaluateSingleton(null, { pid: 1, ppid: 2 }, true, alive), { action: 'proceed' });
+  });
+
+  it('proceeds on a tsx-watch relaunch of our own supervisor (same parent pid)', () => {
+    // New child, DIFFERENT own pid, but SAME watcher parent → not a competitor.
+    const d = evaluateSingleton(rec({ pid: 1000, ppid: 500 }), { pid: 1001, ppid: 500 }, true, alive);
+    assert.deepStrictEqual(d, { action: 'proceed' });
+  });
+
+  it('exits when a different, live supervisor is actually holding the port', () => {
+    const d = evaluateSingleton(rec({ pid: 1000, ppid: 500, port: 3000 }), { pid: 2000, ppid: 900 }, true, alive);
+    assert.strictEqual(d.action, 'exit');
+    assert.match((d as { reason: string }).reason, /already running on :3000/);
+    assert.match((d as { reason: string }).reason, /pid 1000/);
+  });
+
+  it('proceeds when the recorded owner is dead (stale pidfile), even if the port looks in use', () => {
+    const d = evaluateSingleton(rec({ pid: 1000, ppid: 500 }), { pid: 2000, ppid: 900 }, true, dead);
+    assert.deepStrictEqual(d, { action: 'proceed' });
+  });
+
+  it('proceeds when a different live owner is NOT holding the port (nothing to fight over)', () => {
+    const d = evaluateSingleton(rec({ pid: 1000, ppid: 500 }), { pid: 2000, ppid: 900 }, false, alive);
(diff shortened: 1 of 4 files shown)


packages/core/src/scripts/dev-server.ts
```diff
@@ -237,6 +239,172 @@ export function shouldCreditFrontendReady(
...
+  if (!(await probe(port))) return { wasOpen: false, reclaimed: true, pids: [] };
+
+  const pids = listPids(port);
+  for (const pid of pids) killTree(pid, 'SIGTERM');
+  await waitFree(port, 2000);
+
+  if (await probe(port)) {
+    // Still held after a graceful SIGTERM — escalate to SIGKILL. Re-list in case
+    // the owner set changed (e.g. lsof was momentarily empty on the first pass).
+    for (const pid of pids.length ? pids : listPids(port)) killTree(pid, 'SIGKILL');
```
soberm: [minor] The SIGKILL escalation only re-discovers listener PIDs when the initial list was empty (`pids.length ? pids : listPids(port)`), but that's the wrong condition. If `pids` is non-empty, the expression short-circuits to the stale list. Scenario: SIGTERM lands, the original process exits, a new process grabs the port before `waitFree` returns — we then SIGKILL the already-dead original PIDs and miss the real owner. Port stays bound, `reclaimed` comes back `false`.

Should always re-list before the SIGKILL pass. If you want the stale list as a fallback when lsof is momentarily empty:
```ts
const killPids = listPids(port);
for (const pid of killPids.length ? killPids : pids) killTree(pid, 'SIGKILL');
```
The test at line 131 only covers "first discovery was empty"; there's no coverage for the non-empty-but-stale case.

packages/core/src/scripts/process-tree.ts
```diff
@@ -113,6 +113,100 @@ export function killFrontendTree(
...
+ *
+ * Best-effort and never throws: a missing tool, a non-zero exit ("nothing is
+ * listening"), or unparseable output all yield `[]`. PIDs `<= 1` are dropped
+ * defensively (never target init / the whole current group). `runner`/`platform`
+ * are injected for tests.
+ */
+export function findListenerPids(
+  port: number,
+  runner: (command: string, args: readonly string[]) => CommandOutput = (command, args) =>
+    spawnSync(command, args as string[], { encoding: 'utf-8', windowsHide: true }),
```
soberm: [minor] `spawnSync` here has no `timeout` option. `lsof` can block indefinitely when it encounters an unresponsive NFS mount or network filesystem — and because this is `spawnSync`, it holds the entire Node.js event loop while it waits. Since `findListenerPids` is called from `reclaimPort` during startup, a hung `lsof` hangs the dev server before it ever binds a socket. The `catch { return []; }` only fires if `spawnSync` throws, not if it stalls.

```ts
spawnSync(command, args as string[], { encoding: 'utf-8', windowsHide: true, timeout: 3000 })
```

With a timeout, `spawnSync` returns `{ error: ..., stdout: null }` after 3 s, the catch returns `[]`, and startup continues (reclaim just skips — same graceful-degradation path as "lsof not on PATH"). The `windowsTreeKill` `spawnSync` at line 50 has the same gap, but that runs post-startup during teardown, so a hang there is far less painful.

packages/core/src/scripts/dev-server.ts
```diff
@@ -246,6 +414,42 @@ export async function startDevServer(options: DevServerOptions) {
...
+  // EADDRINUSE → mutual Vite kills → restart loop). We record {pid, ppid, port}
+  // in a per-port pidfile and consult it here. The `ppid` (stable `tsx watch`
+  // watcher) lets us tell a hot-reload relaunch of OUR OWN script apart from a
+  // genuine second invocation — see {@link evaluateSingleton}. A stale pidfile
+  // (dead owner) never blocks startup.
+  const pidfilePath = join('.blocks-sandbox', `dev-server.${port}.pid`);
+  const removeOwnPidfile = (): void => {
+    try {
+      const rec = parsePidRecord(readFileSync(pidfilePath, 'utf-8'));
+      if (rec && rec.pid === process.pid) unlinkSync(pidfilePath);
```
soberm: [nit] `removeOwnPidfile` does a read-check-then-delete that isn't atomic:

```ts
const rec = parsePidRecord(readFileSync(pidfilePath, 'utf-8')); // reads own pid
if (rec && rec.pid === process.pid) unlinkSync(pidfilePath);    // ← gap here
```

A tsx-watch successor that writes its new pidfile between those two calls would have its file deleted by the predecessor. In practice tsx-watch drains the old process before relaunching, so the window is very narrow — but worth noting as a latent hazard if tsx-watch's restart sequencing ever changes. Consequence would be the new server starting without a guard pidfile until the next file-save cycle.

packages/core/src/scripts/dev-server.ts
```diff
@@ -598,8 +802,35 @@ export async function startDevServer(options: DevServerOptions) {
...
+  // detached Vite grandchild (or an orphaned backend) can still hold :3100/:3000
+  // and a fresh `--strictPort` start would collide and crash. The singleton guard
+  // above already ruled out a live *peer* supervisor, so anything still holding
+  // these ports is an orphan — reclaim it (see {@link reclaimPort}).
+  const r3000 = await reclaimPort(port);
+  if (r3000.wasOpen) {
+    console.error(
+      r3000.reclaimed
+        ? `♻️  Reclaimed port ${port} from a stale/orphaned listener before startup.`
+        : `⚠️  Port ${port} is in use and could not be reclaimed automatically (no owner PID found).`,
```
soberm: [nit] The failure message appends "(no owner PID found)" unconditionally when `reclaimed` is `false`, but that's only accurate for one of the two failure modes. If `reclaimPort` found PIDs, sent both SIGTERM and SIGKILL, and the port is still bound afterward, the real situation is "owner found but couldn't be killed" — a different problem that calls for different user action. The `pids` field on the result is available to distinguish:

```ts
r3000.reclaimed
  ? `♻️  Reclaimed port ${port} from a stale/orphaned listener before startup.`
  : r3000.pids.length > 0
    ? `⚠️  Port ${port} held by pid(s) [${r3000.pids.join(', ')}] and could not be freed — stop that process and retry.`
    : `⚠️  Port ${port} is in use and could not be reclaimed automatically (no owner PID found).`
```

Same issue on the frontend reclaim message a few lines down (line 826), though that one omits the parenthetical already.

soberm (commented):
Solid hardening overall — the pure decision functions are well-extracted and the no-mock socket tests are genuinely useful.

packages/core/src/scripts/dev-server.ts
```diff
@@ -610,13 +841,58 @@ export async function startDevServer(options: DevServerOptions) {
...
+        );
+        process.exit(1);
+      }
+      console.error(
+        `⚠️  Port ${port} already in use (EADDRINUSE) — reclaiming the stale owner and retrying ` +
+        `(attempt ${bindAttempts}/${DEFAULT_PORT_BIND_RETRY_POLICY.maxAttempts})…`,
+      );
+      void (async () => {
+        await reclaimPort(port);
+        await waitForPortFree(port);
```
soberm: [nit] `reclaimPort` already calls `waitFree` internally (up to twice, 2 s each). The extra `await waitForPortFree(port)` here is redundant — in the happy path it returns immediately because the port is already free; in the failure path it adds another 2 s of dead polling before the `setTimeout` fires for the next listen attempt, on top of the up-to-4 s already spent inside `reclaimPort`. Can just be removed.

soberm (commented):
Six findings verified against the PR head (`4dafb5c`): two confirmed LOW, two confirmed nit, one new nit, one new low. The pure-function design is solid and the test coverage of the primitives is excellent — the issues below are edge-case probe correctness, one operator-UX gap, and one test-coverage gap on the retry wiring.

**Overall: needs-work (minor)** — LOW #1 and #2 are real correctness gaps but neither blocks a typical single-stack dev environment. All six are small; fine to fix here or in a fast follow-up.

packages/core/src/scripts/dev-server.ts
```diff
@@ -102,19 +102,29 @@ async function deployLocal(backend: Record<string, any>): Promise<void> {
...
 }
 
+/**
+ * Single-shot TCP probe: resolves `true` iff a connection to `port` on `host`
+ * succeeds within `timeoutMs`, else `false` (connection error or timeout).
+ * Shared by {@link waitForPort} (wait until open), {@link waitForPortFree} (wait
+ * until closed) and the startup/EADDRINUSE reclaim path so all three agree on
+ * exactly what "the port is bound" means. Never rejects.
+ */
+export async function isPortOpen(port: number, host = 'localhost', timeoutMs = 200): Promise<boolean> {
```
soberm: **LOW #1 — `isPortOpen` probes `localhost`; `server.listen` binds on all interfaces**

Every probe in this file defaults to `host = 'localhost'` (here at the signature, also at line 286 in `reclaimPort`'s default and line 211 inside `waitForPortFree`). `server.listen(port, onListening)` at line 918 passes no host, so Node.js binds on `::` (dual-stack all-interfaces).

On a host where `localhost` resolves to `::1` only (common on modern dual-stack Linux/macOS), an orphan that bound on `0.0.0.0`/`127.0.0.1` is **invisible to every probe**:

1. Startup `reclaimPort(port)` → initial `probe` returns `false` → no-op (`wasOpen: false`), no kill attempted, no ♻️ message logged
2. `waitForPortFree` exits immediately (IPv6 probe says "free" on a port held only on IPv4)
3. `server.listen` gets EADDRINUSE — the retry handler calls `reclaimPort(port)` again, same probe, same `false` result, so **`findListenerPids` is never called and the orphan is never killed**; retries exhaust and the process exits 1 with no PID info

Worth noting: the real-socket test at `dev-server-reclaim.test.ts:133` explicitly uses `isPortOpen(port, '127.0.0.1')` — a clear sign this distinction was on the radar.

**Fix options:**
- Pin the server to a specific address and probe the same one: `server.listen(port, '127.0.0.1', onListening)` + probe `'127.0.0.1'` everywhere, *or*
- Extend `isPortOpen` to check both `127.0.0.1` and `::1` when `host` is omitted, and document the contract in `reclaimPort`'s JSDoc

packages/core/src/scripts/dev-server.ts
```diff
@@ -610,13 +866,57 @@ export async function startDevServer(options: DevServerOptions) {
...
+        process.exit(1);
+      }
+      console.error(
+        `⚠️  Port ${port} already in use (EADDRINUSE) — reclaiming the stale owner and retrying ` +
+        `(attempt ${bindAttempts}/${DEFAULT_PORT_BIND_RETRY_POLICY.maxAttempts})…`,
+      );
+      void (async () => {
+        await reclaimPort(port);
+        setTimeout(() => server.listen(port, onListening), decision.delayMs).unref?.();
+      })();
```
soberm: **LOW #2 — EADDRINUSE retry silently discards `reclaimPort` result; actionable PID message is lost**

`reclaimPort(port)` is awaited but its result thrown away. If reclaim fails on attempt 1 or 2 (orphan survives SIGKILL, or `lsof` returns nothing), the operator sees only the generic retry banner and never gets the `reclaimMessage` that would say "Port 3000 held by pid(s) [4242] — stop that process and retry." On the final attempt `evaluatePortBindRetry` returns `retry: false` and exits before `reclaimPort` is even called, so the exhaustion message also never names a PID.

```suggestion
      void (async () => {
        const reclaimResult = await reclaimPort(port);
        if (!reclaimResult.reclaimed) console.error(reclaimMessage(port, reclaimResult, 'a stale/orphaned listener'));
        setTimeout(() => server.listen(port, onListening), decision.delayMs).unref?.();
      })();
```

packages/core/src/scripts/dev-server.ts
```diff
@@ -246,6 +442,47 @@ export async function startDevServer(options: DevServerOptions) {
...
 
+  // ── Singleton guard ─────────────────────────────────────────────────────
+  // Prevent two fighting supervisors: a second `npm run dev` must not spawn a
+  // competing supervisor that races the first on :3000/:3100 (backend + Vite
+  // EADDRINUSE → mutual Vite kills → restart loop). We record {pid, ppid, port}
+  // in a per-port pidfile and consult it here. The `ppid` (stable `tsx watch`
+  // watcher) lets us tell a hot-reload relaunch of OUR OWN script apart from a
+  // genuine second invocation — see {@link evaluateSingleton}. A stale pidfile
+  // (dead owner) never blocks startup.
+  const pidfilePath = join('.blocks-sandbox', `dev-server.${port}.pid`);
```
soberm: **Nit #1 — `'.blocks-sandbox'` literal; `BLOCKS_SANDBOX_DIR` is already exported from constants**

`BLOCKS_SANDBOX_DIR = '.blocks-sandbox'` is in `packages/core/src/common/constants.ts:12` specifically to avoid this duplication. The constant needs to be added to the import at line 13 (not in this diff): `import { BLOCKS_RPC_PREFIX, BLOCKS_SANDBOX_PREFIX, BLOCKS_SANDBOX_DIR } from '../constants.js';`

Same literal also at line 468 (new in this PR) and lines 513–514 (pre-existing). Worth migrating all four in the same pass.

```suggestion
  const pidfilePath = join(BLOCKS_SANDBOX_DIR, `dev-server.${port}.pid`);
```

packages/core/src/scripts/dev-server-reclaim.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    // New child, DIFFERENT own pid, but SAME watcher parent → not a competitor.
+    const d = evaluateSingleton(rec({ pid: 1000, ppid: 500 }), { pid: 1001, ppid: 500 }, true, alive);
+    assert.deepStrictEqual(d, { action: 'proceed' });
+  });
+
+  it('exits when a different, live supervisor is actually holding the port', () => {
+    const d = evaluateSingleton(rec({ pid: 1000, ppid: 500, port: 3000 }), { pid: 2000, ppid: 900 }, true, alive);
+    assert.strictEqual(d.action, 'exit');
+    assert.match((d as { reason: string }).reason, /already running on :3000/);
+    assert.match((d as { reason: string }).reason, /pid 1000/);
```
soberm: **Nit #2 — `as { reason: string }` cast instead of control-flow narrowing**

`assert.strictEqual` doesn't carry a type predicate, so TypeScript can't narrow through it and the cast is needed. But a guard narrows `d` properly, gives a clear failure message if the action is wrong, and would catch a future rename of `reason` in the union:

```suggestion
    if (d.action !== 'exit') assert.fail(`expected exit, got ${d.action}`);
    assert.match(d.reason, /already running on :3000/);
    assert.match(d.reason, /pid 1000/);
```

packages/core/src/scripts/dev-server.ts
```diff
@@ -598,8 +835,27 @@ export async function startDevServer(options: DevServerOptions) {
...
+  const r3000 = await reclaimPort(port);
+  if (r3000.wasOpen) {
+    console.error(reclaimMessage(port, r3000, 'a stale/orphaned listener'));
+  }
+  if (frontendCommand) {
+    const rFrontend = await reclaimPort(frontendPort);
+    if (rFrontend.wasOpen) {
+      console.error(reclaimMessage(frontendPort, rFrontend, 'a stale/orphaned dev server'));
+    }
+  }
```
soberm, line 961: **New nit — `console.error` used for ♻️ reclaim-success messages**

A successful reclaim prints an informational "Reclaimed port…" message on stderr. That's fine for warnings/errors but a success confirmation should go to stdout. CI pipelines that fail on any stderr output will fire a false alarm on a completely healthy startup that just happened to reclaim an orphan.

```suggestion
  if (r3000.wasOpen) {
    (r3000.reclaimed ? console.log : console.error)(reclaimMessage(port, r3000, 'a stale/orphaned listener'));
  }
  if (frontendCommand) {
    const rFrontend = await reclaimPort(frontendPort);
    if (rFrontend.wasOpen) {
      (rFrontend.reclaimed ? console.log : console.error)(reclaimMessage(frontendPort, rFrontend, 'a stale/orphaned dev server'));
    }
  }
```

packages/core/src/scripts/dev-server.ts
```diff
@@ -610,13 +866,57 @@ export async function startDevServer(options: DevServerOptions) {
...
-  });
+  };
 
+  // Front-door EADDRINUSE robustness — mirror the treatment :3100 already gets:
+  // emit a REAL console error (not just telemetry), reclaim the stale owner and
+  // retry the bind (bounded), and on unrecoverable failure exit non-zero with a
+  // clear message so a contended :3000 never silently fails to serve. Startup
+  // reclaim above makes this a rare race (someone grabbed :3000 between reclaim
+  // and listen); the retry closes that window.
+  let bindAttempts = 0;
```
soberm: **New low — EADDRINUSE retry wiring has no unit coverage; regression would be silent**

All the pure functions are well tested, but the actual retry loop — the `bindAttempts` counter closure, the `void (async () => { … })()` IIFE, and `setTimeout(() => server.listen(port, onListening), delayMs)` — is only smoke-tested manually (per PR description). The wiring is the key novelty of the `:3000` robustness fix and the easiest place for a subtle regression: off-by-one in the attempt counter, a future refactor accidentally dropping the `onListening` callback on the retry call, etc.

A targeted test holding a port for N bind attempts (real socket or injectable `listen`/`error` stub) would lock this in. Given how thoroughly the primitives are covered, this gap stands out.

soberm (approved):

packages/core/src/scripts/dev-server.ts
```diff
@@ -237,6 +256,265 @@ export function shouldCreditFrontendReady(
...
+ */
+export function evaluateSingleton(
+  existing: DevServerPidRecord | null,
+  self: { pid: number; ppid: number },
+  portInUse: boolean,
+  isAlive: (pid: number) => boolean,
+): SingletonDecision {
+  if (!existing) return { action: 'proceed' };
+  if (existing.pid === self.pid) return { action: 'proceed' };
+  if (existing.ppid === self.ppid) return { action: 'proceed' }; // tsx-watch relaunch of our own supervisor
```
sarayev, line 510: The same-ppid carve-out is the right call for hot reload, but I think it's a touch wider than intended. Two separate `npm run dev` jobs backgrounded from one shell (`npm run dev & npm run dev &`) both inherit that shell as their parent, so they'd share a ppid and *both* take the "this is just a tsx-watch relaunch, proceed" branch — which is exactly the two-supervisors-fighting scenario the guard is here to prevent.

The JSDoc says "a second `npm run dev` runs under a different watcher, so it never matches here" — that holds for the tsx-watch parent, but not for a shared grandparent shell, so it slightly oversells the guarantee.

Not a blocker, since the templates always go through `tsx watch` and nobody really backgrounds two dev servers in one shell on purpose. But if you want to close it cleanly: also require the recorded owner's pid to be dead before proceeding (a *live* owner with a matching ppid is much more likely a real competitor than a hot reload), or stash a watcher-identifying token instead of comparing the raw ppid. At minimum, worth softening the JSDoc so the next reader knows where the boundary actually is.

packages/core/src/scripts/dev-server.ts
```diff
@@ -610,13 +972,47 @@ export async function startDevServer(options: DevServerOptions) {
...
+  // emit a REAL console error (not just telemetry), reclaim the stale owner and
+  // retry the bind (bounded), and on unrecoverable failure exit non-zero with a
+  // clear message so a contended :3000 never silently fails to serve. Startup
+  // reclaim above makes this a rare race (someone grabbed :3000 between reclaim
+  // and listen); the retry closes that window. The retry wiring lives in the
+  // testable {@link createBindRetryController}; telemetry stays here.
+  const onEaddrinuse = createBindRetryController(port, {
+    reclaim: (p) => reclaimPort(p),
+    relisten: () => server.listen(port, onListening),
+    scheduleRetry: (fn, delayMs) => { setTimeout(fn, delayMs).unref?.(); },
```
sarayev, line 987: Small thing on the retry scheduling here.

`unref` makes sense on the frontend respawn timer, but at this point the `server.listen` that just failed never established a listening handle, and the frontend child hasn't been spawned yet. So if nothing else happens to be keeping the loop alive when this is the only pending timer, the process could quietly exit *before* the rebind fires — skipping a scheduled retry instead of running it.

In practice there's usually another handle open from backend init, so this is more a latent edge than something I've seen bite. But since the bounded retry budget and the `process.exit(1)` on exhaustion already guarantee this can't loop forever, `unref` isn't really buying you anything here — I'd just drop it so the retry fires deterministically. If you'd rather keep it, a one-line comment explaining what keeps the loop alive between the failed `listen` and the retry would save the next person the head-scratch.

sarayev (approved):
Nice work — this is a clean, well-tested hardening pass. Expressing the whole reclaim/singleton/retry policy as pure, dependency-injected functions and exercising them with real sockets (no process mocks) makes it really easy to reason about. All the earlier review feedback looks fully addressed. Approving.

Two non-blocking notes below — both fine as fast-follows if you'd rather not touch this PR again.

ahmedhamouda78 (approved):

ikenyal:
Adding one more real-world repro signal here.

I hit the same `:3000` front-door case while testing my own unpublished AWS Blocks hands-on material, not an official AWS Blocks tutorial.

The setup was a freshly scaffolded default-template app from `create-blocks-app`. Port 3000 was already occupied by another local process. Running `npm run dev` reached:

```text
Loading backend...
Deploying local resources...
  🔌 Attaching dev server (from @aws-blocks/bb-realtime/ws-server)
📝 Generating client code...
```

No user-facing EADDRINUSE error was printed, and the app never became available.
I also re-checked the underlying behavior with a minimal startDevServer() repro while port 3000 was occupied: the server emitted no console error after client generation. So this is not only an agent-bench scenario; it is also easy to hit during a normal first-run / hands-on flow when another local tool already owns port 3000.
The :3000 EADDRINUSE handling in this PR looks like it should address this case.

ikenyal:
Adding one more real-world repro signal here.

I hit the same `:3000` front-door case while testing my own unpublished AWS Blocks hands-on material, not an official AWS Blocks tutorial.

The setup was a freshly scaffolded default-template app from `create-blocks-app`. Port 3000 was already occupied by another local process. Running `npm run dev` reached:

```text
Loading backend...
Deploying local resources...
  🔌 Attaching dev server (from @aws-blocks/bb-realtime/ws-server)
📝 Generating client code...
```

No user-facing `EADDRINUSE` error was printed, and the app never became available.

I also re-checked the underlying behavior with a minimal `startDevServer()` repro while port 3000 was occupied: the server emitted no console error after client generation. So this is not only an agent-bench scenario; it is also easy to hit during a normal first-run / hands-on flow when another local tool already owns port 3000.

The `:3000 EADDRINUSE handling` in this PR looks like it should address this case.

PR #146, opened 2026-07-03
feat(cli): --help now lists every template with a one-line description

hfurkanbozkurt:
## Problem

Agent-driven scaffolding (e.g. LLM assistants that pick a starter template from `create-blocks-app --help` alone) needs enough information in that output to make the right choice. Today the output lists template *names* but has no per-template description, so the agent has to guess or scaffold each template and inspect the result to see what it produces:

```
Options:
  --template <name>      Template to use. For fresh projects it selects the
                         starter app; when adding to an existing project it
                         selects which aws-blocks/ workspace to copy, e.g.
                         "nextjs" for a Next.js dev server (default: "default")
                         Available templates: default, bare, react, backend, nextjs, auth-cognito, amplify, demo
```

## Fix

Each `templates/<name>/package.json` gets a new `blocksTemplateDescription` field — that field is the **single source of truth** for the help text. The CLI discovers templates by reading `templates/` at print time and pulls the description from each package.json.

Concretely:

- **Adding a new template = one folder edit.** Drop a folder under `templates/` with a `package.json` containing `blocksTemplate` and `blocksTemplateDescription`. Nothing else to touch — no code changes, no second array to sync, no help text to update.
- **The hardcoded `AVAILABLE_TEMPLATES` array is gone.** The filesystem *is* the list. Validation, error messages, and `--help` all derive from the discovered catalog. There is no longer a "second list of templates" anywhere in the codebase.
- **`templates/amplify` gets a minimal metadata-only `package.json`** — it was the only template without one because it's an overlay (never scaffolded standalone via `--template amplify`, only auto-selected when Amplify Gen 2 is detected). Adding the file makes catalog discovery uniform.
- The existing `Available templates: default, bare, ...` summary line is preserved on purpose so grep-based consumers (including the existing snapshot test at `src/cli.test.ts:36`) keep working; the labeled rows appear underneath.
- The Examples block picks up a new `--yes --template nextjs` line so non-interactive callers have a copy-pasteable end-to-end shape.

## Before

```
Usage: create-blocks-app [directory] [options]

...

Options:
  --template <name>      Template to use. For fresh projects it selects the
                         starter app; when adding to an existing project it
                         selects which aws-blocks/ workspace to copy, e.g.
                         "nextjs" for a Next.js dev server (default: "default")
                         Available templates: default, bare, react, backend, nextjs, auth-cognito, amplify, demo
  --skip-install         Skip installing dependencies
  -y, --yes              Skip confirmation prompts
  -h, --help             Show this help message

Examples:
  npx @aws-blocks/create-blocks-app            Add Blocks to current project
  npx @aws-blocks/create-blocks-app .          Add Blocks to current project
  npx @aws-blocks/create-blocks-app my-app     Create a standalone Blocks starter app
```

## After

```
Usage: create-blocks-app [directory] [options]

...

Options:
  --template <name>      Template to use. For fresh projects it selects the
                         starter app; when adding to an existing project it
                         selects which aws-blocks/ workspace to copy, e.g.
                         "nextjs" for a Next.js dev server (default: "default")

                         Available templates: default, bare, react, backend, nextjs, auth-cognito, amplify, demo

    default        Vite + lit-html frontend with auth, DynamoDB, and realtime (safe default)
    bare           Vite + lit-html frontend with a single greet() API (smallest starter)
    react          Vite + React 19 frontend with auth, DynamoDB, and realtime todo app
    backend        Backend-only — Blocks API + CDK, no frontend (bring your own client)
    nextjs         Next.js 16 App Router with Server + Client Components calling Blocks
    auth-cognito   Vite + lit-html with Cognito passwordless email-OTP auth end-to-end
    amplify        Overlay for an existing Amplify Gen 2 app (auto-selected when detected)
    demo           Fuller example — auth + KVStore + DynamoDB todo on Vite + lit-html

  --skip-install         Skip installing dependencies
  -y, --yes              Skip confirmation prompts
  -h, --help             Show this help message

Examples:
  npx @aws-blocks/create-blocks-app                            Add Blocks to current project
  npx @aws-blocks/create-blocks-app .                          Add Blocks to current project
  npx @aws-blocks/create-blocks-app my-app                     Create a standalone Blocks starter app
  npx @aws-blocks/create-blocks-app . --yes --template nextjs  Non-interactive scaffold with the Next.js template
```

## Test plan

- [x] `npm run build && node dist/index.js --help` renders the labeled list above with correct column alignment.
- [x] All 32 existing tests pass: `node --test dist/**/*.test.js` — including the assertion that the `Available templates: default, bare, react, backend, nextjs, auth-cognito, amplify, demo` line is present.
- [x] `node dist/index.js --template does-not-exist` still fails cleanly with `Unknown template "does-not-exist"` and lists the available names.
- [x] `node dist/index.js -h` also works (short form).
- [x] No new flags, no default changes — this is a documentation change that reshapes the output only.

## Notes on descriptions

Each description was written by inspecting the actual contents of the template folder (frontend framework in `package.json`, backend blocks used in `aws-blocks/index.ts`, top-level scaffold files) — not from memory. In particular:

- `default` and `react` both ship the same auth+DynamoDB+realtime todo app but with different frontends (lit-html vs React 19).
- `bare` is a bare vite scaffold with a single `greet()` API and no auth/data.
- `backend` has no frontend files at all — just `aws-blocks/` + `cdk.json`.
- `nextjs` uses Next.js 16 App Router (Server + Client Components).
- `auth-cognito` demonstrates AuthCognito's passwordless email-OTP flow with a local mock code-capture hook.
- `amplify` has no `package.json` today; this PR adds a minimal one for metadata. It's an overlay, not a standalone starter.
- `demo` is a fuller variant with `KVStore` in addition to the todo app.

---

## Follow-up commit (`f3aa346`) — amplify guard + description fixes

Added on top of the review-pass commits (changeset / README / empty-catalog guard / description test) after a self-review turned up two things:

### 1. `--template amplify` on a fresh directory now exits cleanly

**Pre-existing footgun** (unchanged by the original PR, but exposed more by making amplify discoverable): amplify is an overlay that only makes sense when the CLI detects `amplify/backend.ts`. It ships `amplify-blocks.ts` + `aws-blocks/` only — no `src/`, no `index.html`, no `gitignore`. Running `create-blocks-app my-app --template amplify` used to crash mid-copy (missing `package.json` on `origin/main`; missing `gitignore` + `index.html` + `src/` after this PR's metadata `package.json` addition).

Guarded `createFreshProject` up front:

```
Error: The `amplify` template is auto-selected when the CLI detects
an existing Amplify Gen 2 project (an `amplify/backend.ts` in the
target directory). It cannot be scaffolded as a fresh app.

To scaffold a fresh Blocks app, choose a different template:
  npx @aws-blocks/create-blocks-app my-app --template default

To integrate Blocks into an existing Amplify project, run this from
the project root:
  npx @aws-blocks/create-blocks-app
```

Covered by a new test at `packages/create-blocks-app/src/cli.test.ts`.

### 2. Two `blocksTemplateDescription` values were factually wrong

Verified every template's description against actual file contents. Two were wrong on the frontend claim:

| template | prior description (wrong) | corrected description |
|---|---|---|
| `auth-cognito` | Vite + **lit-html** with Cognito passwordless email-OTP auth end-to-end | Vite + **vanilla-DOM** frontend with Cognito passwordless email-OTP auth end-to-end |
| `demo` | Fuller example — auth + KVStore + DynamoDB todo on Vite + **lit-html** | Fuller example — AuthBasic + KVStore + DynamoDB priority-sorted todo (Vite, **vanilla-DOM**) |

Both templates use `document.createElement` / `innerHTML` / vanilla DOM APIs — neither imports `lit-html` nor lists it as a dependency. The `default` and `bare` descriptions ARE accurate (both `import { html, render } from 'lit-html'`). The `react` (React 19) and `nextjs` (Next.js 16) descriptions were verified against both the `dependencies` in `package.json` AND the actual imports.

### Verification

- `npm run build && npm test` in `packages/create-blocks-app` → 34/34 tests pass (was 33, +1 for the new amplify-guard test).
- Changeset at `.changeset/amplify-fresh-scaffold-guard.md` (patch bump).

### DIFF
--- a/packages/create-blocks-app/README.md
+++ b/packages/create-blocks-app/README.md
@@ -133,6 +133,8 @@ export const blocksStack = await BlocksStack.create(app, stackName, {
 
 ## Templates
 
+Available templates: `default`, `bare`, `react`, `backend`, `nextjs`, `auth-cognito`, `amplify`, `demo`.
+
 Every folder under `templates/` is a self-registering template. Each `package.json` in that folder sets a `blocksTemplateDescription` field, which the CLI reads to build the `--help` catalog. To see the current list of templates and their one-line descriptions:
 
 ```bash
--- a/packages/create-blocks-app/README.md
+++ b/packages/create-blocks-app/README.md
@@ -20,15 +20,13 @@ npx @aws-blocks/create-blocks-app my-app
 npx @aws-blocks/create-blocks-app my-app --template react
 ```
 
-Available templates:
-- **default** - Real-time todo app with auth, data, and live sync
-- **bare** - Minimal Blocks setup
-- **react** - React frontend with Blocks backend
-- **backend** - Backend-only (no frontend)
-- **nextjs** - Next.js integration
-- **auth-cognito** - Cognito authentication example
-- **amplify** - Amplify Gen 2 integration
-- **demo** - Demo application
+For the full list of templates and one-line descriptions, run:
+
+```bash
+npx @aws-blocks/create-blocks-app --help
+```
+
+Template descriptions are sourced from each template's `package.json` (`blocksTemplateDescription`) so the CLI is always the source of truth.
 
 ### Add to Existing Project
 
@@ -135,33 +133,11 @@ export const blocksStack = await BlocksStack.create(app, stackName, {
 
 ## Templates
 
-### Default
-Real-time todo application with:
-- User authentication (AuthBasic)
-- Todo storage (DistributedTable / DynamoDB)
-- Optimistic locking
-- Real-time sync (Realtime / WebSocket)
+Every folder under `templates/` is a self-registering template. Each `package.json` in that folder sets a `blocksTemplateDescription` field, which the CLI reads to build the `--help` catalog. To see the current list of templates and their one-line descriptions:
 
-### Bare
-Minimal setup demonstrating core concepts.
-
-### React
-React frontend with Blocks backend.
-
-### Backend
-Backend-only template (no frontend bundling).
-
-### Next.js
-Next.js integration with Blocks backend.
-
-### Auth-Cognito
-Cognito authentication example.
-
-### Amplify
-Amplify Gen 2 integration template.
-
-### Demo
-Demo application showcasing Building Blocks.
+```bash
+npx @aws-blocks/create-blocks-app --help
+```
 
 ## Related Packages
 
--- a/packages/create-blocks-app/src/cli.test.ts
+++ b/packages/create-blocks-app/src/cli.test.ts
@@ -46,14 +46,8 @@ describe('create-blocks-app CLI argument parsing', () => {
   });
 
   it('--help includes template descriptions', () => {
-    // Guards the filesystem-driven catalog: if template discovery breaks or a
-    // template's `blocksTemplateDescription` disappears, --help output will no
-    // longer include the canonical description strings and this test fails.
     const result = run(['--help']);
     assert.strictEqual(result.exitCode, 0);
-    // "default" template's canonical description substring. Pinned to a stable
-    // fragment so wording tweaks don't require test churn but a missing
-    // description does.
     assert.match(
       result.stdout,
       /default\s+Vite \+ lit-html frontend with auth/,
@@ -96,11 +90,23 @@ describe('create-blocks-app CLI argument parsing', () => {
     assert.doesNotMatch(result.stderr, /ENOENT/);
   });
 
+  it('a template folder missing package.json is rejected as unknown, not crashed with ENOENT', () => {
+    // Latent-bug guard: a folder under templates/ with no package.json is shown
+    // in --help but excluded from validation, so selecting it fails cleanly
+    // instead of throwing ENOENT on the later template version read.
+    const brokenDir = join(__dirname, '../templates', '__broken_test_template__');
+    mkdirSync(brokenDir, { recursive: true });
+    try {
+      const result = run(['my-app', '--template', '__broken_test_template__', '--skip-install']);
+      assert.strictEqual(result.exitCode, 1);
+      assert.match(result.stderr, /Unknown template "__broken_test_template__"/);
+      assert.doesNotMatch(result.stderr, /ENOENT/);
+    } finally {
+      rmSync(brokenDir, { recursive: true, force: true, maxRetries: 3, retryDelay: 100 });
+    }
+  });
+
   it('--template amplify on a fresh dir exits 1 with a helpful message', () => {
-    // amplify is auto-selected when an existing Amplify Gen 2 project is
-    // detected — it's not scaffoldable as a fresh app. The CLI should
-    // reject the fresh-scaffold path up front rather than crashing partway
-    // through the file copy on missing template contents.
     const tmpDir = mkdtempSync(join(tmpdir(), 'create-blocks-app-amplify-fresh-'));
     try {
       const result = run([tmpDir, '--template', 'amplify', '--skip-install', '-y']);
--- a/packages/create-blocks-app/src/cli.test.ts
+++ b/packages/create-blocks-app/src/cli.test.ts
@@ -96,6 +96,23 @@ describe('create-blocks-app CLI argument parsing', () => {
     assert.doesNotMatch(result.stderr, /ENOENT/);
   });
 
+  it('--template amplify on a fresh dir exits 1 with a helpful message', () => {
+    // amplify is auto-selected when an existing Amplify Gen 2 project is
+    // detected — it's not scaffoldable as a fresh app. The CLI should
+    // reject the fresh-scaffold path up front rather than crashing partway
+    // through the file copy on missing template contents.
+    const tmpDir = mkdtempSync(join(tmpdir(), 'create-blocks-app-amplify-fresh-'));
+    try {
+      const result = run([tmpDir, '--template', 'amplify', '--skip-install', '-y']);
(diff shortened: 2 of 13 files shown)


packages/create-blocks-app/src/cli.test.ts
```diff
@@ -45,6 +45,22 @@ describe('create-blocks-app CLI argument parsing', () => {
     assert.match(result.stdout, /Usage: create-blocks-app/);
   });
 
+  it('--help includes template descriptions', () => {
+    // Guards the filesystem-driven catalog: if template discovery breaks or a
```
soberm: [nit] The `it('--help includes template descriptions', ...)` label already states the purpose, and the third argument to `assert.match` already explains what's being pinned. Both comment blocks (this one and the one before the `assert.match` call) can go.

packages/create-blocks-app/src/cli.test.ts
```diff
@@ -80,6 +96,23 @@ describe('create-blocks-app CLI argument parsing', () => {
     assert.doesNotMatch(result.stderr, /ENOENT/);
   });
 
+  it('--template amplify on a fresh dir exits 1 with a helpful message', () => {
+    // amplify is auto-selected when an existing Amplify Gen 2 project is
```
soberm: [nit] Same pattern as the test above. The `it(...)` label says exactly this; the comment adds nothing the test name and assertions don't already convey.

packages/create-blocks-app/src/index.ts
```diff
@@ -185,20 +185,111 @@ async function addBlocksWorkspace(targetDir: string, options: {
...
+    try {
+      const pkg = JSON.parse(await readFile(pkgPath, 'utf-8'));
+      discovered.push({
+        name: entry.name,
+        description: typeof pkg.blocksTemplateDescription === 'string'
+          ? pkg.blocksTemplateDescription
+          : '(no description — add `blocksTemplateDescription` to this template\'s package.json)',
+      });
+    } catch {
+      // Template folder without a readable package.json — surface it so it's
```
soberm, line 244: [minor] Templates that land in this catch path are still added to `discovered`, so they pass `validateTemplateName`. Any subsequent `readFile` of their `package.json` (e.g. the version read in `create()` at the bottom, or in `createFreshProject`) will throw ENOENT instead of the expected "Unknown template" error. If the intent is to surface broken templates in `--help` while excluding them from validation, `TemplateInfo` could carry a `valid: boolean` flag that `validateTemplateName` filters on. All current templates have valid `package.json` files so this is latent, but the path is reachable if a template folder is added without one.

packages/create-blocks-app/src/index.ts
```diff
@@ -185,20 +185,111 @@ async function addBlocksWorkspace(targetDir: string, options: {
...
 
 // ─── Fresh project creation ──────────────────────────────────────────────────
 
 async function createFreshProject(targetDir: string, templateName: string, skipInstall = false) {
-  validateTemplateName(templateName);
+  await validateTemplateName(templateName);
+
+  // The `amplify` template is not a scaffoldable starter — it's an overlay
+  // auto-selected when the CLI detects an existing Amplify Gen 2 project
+  // (see `hasAmplify()` on `amplify/backend.ts`). It ships only
```
soberm: [nit] No function named `hasAmplify` exists in this file. The detection function is `isAmplifyGen2Project()` (line 85).

packages/create-blocks-app/src/index.ts
```diff
@@ -185,20 +185,111 @@ async function addBlocksWorkspace(targetDir: string, options: {
...
+  // `amplify-blocks.ts` + the shared `aws-blocks/` snippet directory, NOT
+  // a full standalone project (no `src/`, no `index.html`, no `gitignore`).
+  // Attempting `createFreshProject` with `--template amplify` would fail
+  // partway through the copy (at the `rename(gitignore, .gitignore)` step
+  // among other missing files). Reject up front with a helpful message.
+  if (templateName === 'amplify') {
```
soberm: [minor] The guard is hardcoded to the template name string. The amplify `package.json` introduced by this PR already carries custom fields (`blocksTemplate`, `blocksTemplateDescription`). Adding `"blocksTemplateOverlayOnly": true` there and reading it in `loadTemplateCatalog` would let any overlay template self-describe as non-scaffoldable, and this check could key off the catalog entry instead of a name match. As it stands, adding a second overlay template requires knowing to add a corresponding guard here.

packages/create-blocks-app/README.md
```diff
@@ -20,15 +20,13 @@ npx @aws-blocks/create-blocks-app my-app
...
-- **default** - Real-time todo app with auth, data, and live sync
-- **bare** - Minimal Blocks setup
-- **react** - React frontend with Blocks backend
-- **backend** - Backend-only (no frontend)
-- **nextjs** - Next.js integration
-- **auth-cognito** - Cognito authentication example
-- **amplify** - Amplify Gen 2 integration
-- **demo** - Demo application
+For the full list of templates and one-line descriptions, run:
+
```
soberm, line 24: [question] Removing the inline template list in favor of a "run --help" pointer avoids stale docs (the old list had two wrong descriptions, which this PR itself fixes). That said, a reader evaluating the project on GitHub or npm before installing now has to run the package to see what templates exist. Was that trade-off intentional? Even a parenthetical listing the template names (without full descriptions) would let readers scan without installing.

soberm (approved):

Simone319 (approved):

PR #148, opened 2026-07-03
fix(bb-distributed-data): reject index key sort order (ASC/DESC) for DSQL

fossamagna:
## Problem

Aurora DSQL does not support a sort direction (`ASC`/`DESC`) on index keys. Deploying such a migration fails with `specifying sort order not supported for index keys`. (`NULLS FIRST`/`NULLS LAST` **is** supported by DSQL and is not affected.)

The local mock runs on PGlite (full PostgreSQL), which accepts `ASC`/`DESC` on index keys, so a migration like `CREATE INDEX ASYNC ... (col DESC)` passes locally and only fails on the first real DSQL deploy — exactly the class of divergence the mock is meant to close.

## Changes

- **Validation rule** (`packages/bb-distributed-data/src/validation.ts`): add one entry to the existing `RULES` array that rejects a sort direction (`ASC`/`DESC`) on `CREATE [UNIQUE] INDEX` keys, while leaving `NULLS FIRST/LAST` allowed. Because `validateStatement` runs from both the migration validator and the mock engine's query path (before the engine strips `ASYNC`), the failure now surfaces at dev time — during `npm run dev` migration execution — instead of on deploy.
- **Docs** (`packages/bb-distributed-data/README.md`, `packages/bb-distributed-data/DESIGN.md`): add a row to the DSQL limitations / statement-validation tables, with the alternative (omit the direction; enforce ordering with `ORDER BY` in queries).
- **Changeset**: patch bump for `@aws-blocks/bb-distributed-data`.

The rule matches within a single statement (`[^;]*`) and is false-positive-safe: `validateStatement` is called per statement (statements are split upstream), a `CREATE INDEX` grammar has no `ORDER BY` clause, and string literals are stripped before matching.

## Validation

- **Unit tests** (`validation.test.ts`): reject `ASC`/`DESC`; confirm `NULLS FIRST`/`NULLS LAST` are allowed; regression cases confirm no false positive on `SELECT ... ORDER BY x DESC`, on a partial index whose `WHERE` mentions a `description` column, or on an expression index.
- **End-to-end mock test** (`e2e-mock.test.ts`): the `DsqlMockEngine` rejects `CREATE INDEX ASYNC ... (col DESC)` — proving the reproduction case is now caught locally before any deploy.
- Full `bb-distributed-data` suite: **146/146 passing**.

**Known limitation:** a column literally named `"desc"`/`"asc"` (double-quoted identifier) can false-positive, since `stripLiteralsAndComments` does not strip double-quoted identifiers. This is the same class as the existing single-keyword rules (`COLLATE`, `TRUNCATE`) and affects no current migration, so it is treated as non-blocking.

## Checklist

- [x] PR description included

---

### DIFF
--- a/packages/bb-distributed-data/src/validation.test.ts
+++ b/packages/bb-distributed-data/src/validation.test.ts
@@ -500,3 +500,44 @@ describe('validateStatement — unsupported features after bind parameters', () 
     assert.doesNotThrow(() => validateStatement("INSERT INTO audit (id, action) VALUES ($1, 'TRUNCATE')"));
   });
 });
+
+describe('validateStatement — index key sort order', () => {
+  it('rejects DESC on an index key', () => {
+    assert.throws(
+      () => validateStatement('CREATE INDEX idx ON persons (user_id, last_encounter_at DESC)'),
+      /sort order/i,
+    );
+  });
+
+  it('rejects DESC on a CREATE INDEX ASYNC key', () => {
+    assert.throws(
+      () => validateStatement('CREATE INDEX ASYNC idx_persons_user_recent ON persons (user_id, last_encounter_at DESC)'),
+      /sort order/i,
+    );
+  });
+
+  it('rejects an explicit ASC on an index key', () => {
+    assert.throws(() => validateStatement('CREATE INDEX idx ON t (col ASC)'), /sort order/i);
+  });
+
+  it('rejects NULLS FIRST / NULLS LAST on an index key', () => {
+    assert.throws(() => validateStatement('CREATE INDEX idx ON t (col NULLS FIRST)'), /sort order/i);
+    assert.throws(() => validateStatement('CREATE INDEX idx ON t (col NULLS LAST)'), /sort order/i);
+  });
+
+  it('allows a plain CREATE INDEX without sort order', () => {
+    assert.doesNotThrow(() => validateStatement('CREATE INDEX idx ON persons (user_id, last_encounter_at)'));
+  });
+
+  it('allows a partial index whose WHERE mentions a column like "description"', () => {
+    assert.doesNotThrow(() => validateStatement('CREATE INDEX idx ON t (name) WHERE description IS NOT NULL'));
+  });
+
+  it('allows an expression index without sort order', () => {
+    assert.doesNotThrow(() => validateStatement('CREATE INDEX idx ON t ((lower(name)))'));
+  });
+
+  it('does not flag ORDER BY ... DESC in a SELECT (no false positive outside CREATE INDEX)', () => {
+    assert.doesNotThrow(() => validateStatement('SELECT * FROM persons ORDER BY last_encounter_at DESC'));
+  });
+});
--- a/packages/bb-distributed-data/src/validation.test.ts
+++ b/packages/bb-distributed-data/src/validation.test.ts
@@ -520,9 +520,9 @@ describe('validateStatement — index key sort order', () => {
     assert.throws(() => validateStatement('CREATE INDEX idx ON t (col ASC)'), /sort order/i);
   });
 
-  it('rejects NULLS FIRST / NULLS LAST on an index key', () => {
-    assert.throws(() => validateStatement('CREATE INDEX idx ON t (col NULLS FIRST)'), /sort order/i);
-    assert.throws(() => validateStatement('CREATE INDEX idx ON t (col NULLS LAST)'), /sort order/i);
+  it('allows NULLS FIRST / NULLS LAST on an index key (supported by DSQL)', () => {
+    assert.doesNotThrow(() => validateStatement('CREATE INDEX ASYNC idx ON t (col NULLS FIRST)'));
+    assert.doesNotThrow(() => validateStatement('CREATE INDEX ASYNC idx ON t (col NULLS LAST)'));
   });
 
   it('allows a plain CREATE INDEX without sort order', () => {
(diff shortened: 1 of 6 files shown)


packages/bb-distributed-data/src/validation.ts
```diff
@@ -85,6 +85,7 @@ const RULES: ValidationRule[] = [
   { pattern: /\bCOLLATE\b/i, message: 'DSQL only supports C collation.', severity: 'error' },
   { pattern: /(?<!::)\bJSONB\b/i, message: 'DSQL does not support JSONB columns. Use JSON instead (JSONB is available as a query runtime cast via ::jsonb).', severity: 'error' },
   { pattern: /(@>|<@|\?\||\?&)/, message: 'JSONB operators lack GIN index acceleration in DSQL.', severity: 'warn' },
+  { pattern: /\bCREATE\s+(?:UNIQUE\s+)?INDEX\b[\s\S]*\b(?:ASC|DESC|NULLS\s+(?:FIRST|LAST))\b/i, message: 'DSQL does not support sort order (ASC/DESC/NULLS FIRST/LAST) on index keys. Remove it — ordering is enforced by ORDER BY in queries. (DSQL: "specifying sort order not supported for index keys")', severity: 'error' },
```
sarayev: Two things here:

**1. `NULLS FIRST/LAST` shouldn't be rejected.** Per the official [CREATE INDEX ASYNC syntax](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-create-index-async.html), the key-column grammar is `( { column_name } [ NULLS { FIRST | LAST } ] )` — `NULLS FIRST/LAST` is supported, but `ASC`/`DESC` isn't in the grammar at all. It's also listed as a supported parameter on the [Supported SQL features](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-sql-features.html) page. I also confirmed against a live DSQL cluster: `col NULLS FIRST` / `col NULLS LAST` succeed, while `col ASC` / `col DESC` fail with `specifying sort order not supported for index keys`. So the alternation should drop the NULLS part:
```js
/\bCREATE\s+(?:UNIQUE\s+)?INDEX\b[^;]*\b(?:ASC|DESC)\b/i
```

**2. The greedy match can over-reach.** `[\s\S]*` will match `ASC`/`DESC` anywhere after `CREATE INDEX`, including across a statement boundary — e.g. `CREATE INDEX ... (col); SELECT ... ORDER BY col DESC;` would be wrongly rejected if this ever sees multiple statements. Using `[^;]*` (as above) keeps the match within one statement. One remaining caveat worth a code comment: `stripLiteralsAndComments` doesn't strip double-quoted identifiers, so a column literally named `"desc"` would still false-positive — niche, but worth noting.

packages/bb-distributed-data/src/validation.test.ts
```diff
@@ -500,3 +500,44 @@ describe('validateStatement — unsupported features after bind parameters', ()
...
+  it('rejects DESC on a CREATE INDEX ASYNC key', () => {
+    assert.throws(
+      () => validateStatement('CREATE INDEX ASYNC idx_persons_user_recent ON persons (user_id, last_encounter_at DESC)'),
+      /sort order/i,
+    );
+  });
+
+  it('rejects an explicit ASC on an index key', () => {
+    assert.throws(() => validateStatement('CREATE INDEX idx ON t (col ASC)'), /sort order/i);
+  });
```
sarayev, line 521: This asserts `NULLS FIRST` / `NULLS LAST` are rejected, but DSQL actually accepts them (confirmed on a real cluster + in the official grammar). Could we flip this to assert they're **allowed**? e.g. `CREATE INDEX ASYNC idx ON t (col NULLS FIRST)` should pass validation. Keeps the suite honest about real DSQL behavior.

sarayev (commented):
Really like this change — surfacing the DSQL limitation at dev time instead of on deploy is a solid shift-left, and validating before the ASYNC strip is the right call so the mock rejects exactly what DSQL does. I confirmed the behavior against a real Aurora DSQL cluster, and your error quote is spot-on: DSQL returns exactly `specifying sort order not supported for index keys`.

One correctness issue: DSQL only rejects `ASC`/`DESC` — it actually **accepts** `NULLS FIRST` / `NULLS LAST` on index keys (they're in the official [CREATE INDEX ASYNC grammar](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-create-index-async.html) and the [Supported SQL features](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/working-with-postgresql-compatibility-supported-sql-features.html) page). So the rule is currently over-rejecting and would block valid indexes locally. Left inline suggestions to narrow it to just `ASC`/`DESC`.

One more small thing (couldn't inline it since the file isn't in this diff): the README limitations table was updated but `DESIGN.md`'s Statement Validation table (the per-rule reference alongside FOREIGN KEY, TRUNCATE, COLLATE, etc.) wasn't. Mind adding a row so they stay in sync? e.g. `| CREATE INDEX … ASC/DESC | Sort direction on index keys (NULLS FIRST/LAST is allowed) |`

packages/bb-distributed-data/src/validation.ts
fossamagna: Confirmed against the CREATE INDEX ASYNC grammar — the key-column grammar is `( { column_name } [ NULLS { FIRST | LAST } ] )`, so `NULLS FIRST/LAST` is supported and `ASC`/`DESC` isn't in the grammar at all. Narrowed the rule to `/\bCREATE\s+(?:UNIQUE\s+)?INDEX\b[^;]*\b(?:ASC|DESC)\b/i` — dropped the `NULLS` alternation and switched `[\s\S]*` → `[^;]*` to keep the match within one statement. Added a code comment about the double-quoted `"desc"` identifier caveat.

packages/bb-distributed-data/src/validation.test.ts
fossamagna: Flipped this to assert `NULLS FIRST`/`NULLS LAST` pass validation (`assert.doesNotThrow`), matching real DSQL behavior. The `ASC`/`DESC` rejection cases stay.

fossamagna:
Thanks for the thorough review and for verifying against a live DSQL cluster, @sarayev. Added a row to DESIGN.md's Statement Validation table: `| CREATE INDEX ... ASC/DESC | Sort direction on index keys (NULLS FIRST/LAST is allowed) |`. Also corrected the README limitations row and the changeset, which had wrongly listed `NULLS FIRST/LAST` as unsupported.

sarayev (approved):

sarayev:
LGTM, Thanks for addressing the previous findings!

ahmedhamouda78 (approved):

PR #149, opened 2026-07-03
feat(bb-agent): canned provider tool hints default values

pjkroker:
merge **before** #90 
## Problem

The local-dev `canned` provider (the fake LLM used when no real model is configured) has only limited tool support that made prototyping against it less useful:

- Tool inputs were generic placeholders (`'sample'`, `1`, `true`, `[]`) that ignored schema `default` values and didn't reflect realistic usage.
- A tool only fired when the prompt mentioned the tool name (or its camelCase words), so natural phrasings like "find the docs" wouldn't trigger the obvious tool.

## Changes

Adds two optional, local-dev-only hint fields to `ToolDefinition` (both **ignored by the real bedrock/openai providers**) plus schema-default awareness in the canned provider:

- **`cannedExamples`** — realistic tool input, shallow-merged over generated placeholders (your fields win; unspecified fields fall back to schema defaults / generic placeholders). The merge is one level
deep — a nested-object example replaces the whole generated sub-object rather than deep-merging.
- **`cannedTriggers`** — extra keyword phrases that make the canned provider select a tool beyond its name. Single and multi-word phrases match on word boundaries (so `cat` won't fire on `category`, and
  `log in` won't fire on `backlog in`); internal whitespace is flexible; blank triggers are skipped.
- **Schema defaults** — placeholder generation now respects `default` values (from Zod `.default()`) before falling back to generic placeholders.

**Implementation note (potentially non-obvious):** Strands strips these fields when it converts tools (`tool()` only copies `name`/`description`/`inputSchema`/`callback`), so `CannedProvider.stream()`
never sees the original `ToolDefinition`. The hints are therefore plumbed explicitly, keyed by tool name: `createStrandsAgent` builds a hints map from the tool defs (only when a tool declares one) →
`createStrandsModel(config, log, cannedHints)` → `new CannedProvider({ hints })`. The extra `createStrandsModel` param is consumed only in the canned branch; bedrock/openai construction is untouched.

No `index.mock/aws/cdk/browser` changes were needed — the canned provider is a pure runtime concern with no CDK/synthGuard surface (same as the sibling `throwing.ts` provider).

```typescript
tools: (tool) => ({
  searchDocs: tool({
    description: 'Search product documentation for relevant information',
    parameters: z.object({
      query: z.string().describe('The search query'),
      maxResults: z.number().optional().describe('Max results to return (default: 5)'),
    }),
    handler: async ({ input }) => kb.retrieve(input.query, { maxResults: input.maxResults ?? 5 }),

    // Canned provider hints (ignored by real models):
    // Without this the mock would search for the literal 'sample' and match nothing —
    // use a query that hits YOUR documents so local runs return meaningful results.
    cannedExamples: { query: 'how do I reset my password' },
    // "help me find the login docs" now triggers the tool searchDocs
    cannedTriggers: ['find', 'look up'],
  }),
}),
```
## Validation

- **10 new tests** in `packages/bb-agent/src/index.test.ts`:
  - 9 provider-level tests (direct `new CannedProvider({ hints })` + `stream`): `cannedExamples` merge, schema-default precedence, default+placeholder mix, single-word trigger, multi-word phrase trigger,
trigger word-boundary regression, and a no-hints regression guard.
  - 1 end-to-end test exercising the full `Agent → createStrandsAgent → createStrandsModel → CannedProvider` plumbing (a trigger-only prompt fires the tool and the persisted `tool-call` carries the example
input).
- Full suite passes: **88/88 tests** (including all 10 new).
- Build clean (`tsc --build`); Biome lint reports 0 errors on the changed files.
## Checklist

- [x] PR description included

---

### DIFF
--- a/packages/bb-agent/src/providers/canned.ts
+++ b/packages/bb-agent/src/providers/canned.ts
@@ -68,13 +68,15 @@ function findAllToolMatches(prompt: string, toolSpecs?: { name: string }[], hint
 		// on word boundaries. Skip short words (<=2 chars) to avoid noise.
 		const words = t.name.replace(/([a-z])([A-Z])/g, '$1 $2').toLowerCase().split(' ');
 		if (words.some(w => w.length > 2 && promptMentionsWord(lower, w))) return true;
-		// Extra trigger keywords declared via `cannedTriggers`. Single words use word
-		// boundaries (consistent with tool-name matching); multi-word phrases use substrings.
+		// Extra trigger keywords declared via `cannedTriggers`. Single and multi-word triggers
+		// both match on word boundaries (consistent with tool-name matching), so "log in" is not
+		// triggered by "backlog in" and internal whitespace is flexible (matches one-or-more spaces).
 		const triggers = hints?.get(t.name)?.triggers;
 		return triggers?.some(tr => {
 			const low = tr.trim().toLowerCase();
 			if (!low) return false;
-			return low.includes(' ') ? lower.includes(low) : promptMentionsWord(lower, low);
+			const escaped = low.replace(/[.*+?^${}()|[\]\\]/g, '\\$&').replace(/\s+/g, '\\s+');
+			return new RegExp(`\\b${escaped}\\b`).test(lower);
 		}) ?? false;
 	}).map(t => t.name);
 }
--- a/packages/bb-agent/src/providers/canned.ts
+++ b/packages/bb-agent/src/providers/canned.ts
@@ -17,9 +17,16 @@
 import { Model } from '@strands-agents/sdk';
 import type { Message, ModelStreamEvent, StreamOptions } from '@strands-agents/sdk';
 import { ToolResultBlock } from '@strands-agents/sdk';
+import type { CannedToolHints } from '../types.js';
 
 interface CannedConfig {
 	modelId: string;
+}
+
+interface CannedProviderOptions {
+	modelId?: string;
+	/** Per-tool hints (examples, triggers) keyed by tool name. */
+	hints?: Map<string, CannedToolHints>;
 }
 
 const CANNED_RESPONSES: Record<string, string> = {
@@ -51,7 +58,7 @@ function promptMentionsWord(lowerPrompt: string, word: string): boolean {
 }
 
 /** Find ALL tools mentioned in the prompt (for parallel tool calls). */
-function findAllToolMatches(prompt: string, toolSpecs?: { name: string }[]): string[] {
+function findAllToolMatches(prompt: string, toolSpecs?: { name: string }[], hints?: Map<string, CannedToolHints>): string[] {
 	if (!toolSpecs?.length) return [];
 	const lower = prompt.toLowerCase();
 	return toolSpecs.filter(t => {
@@ -60,7 +67,15 @@ function findAllToolMatches(prompt: string, toolSpecs?: { name: string }[]): str
 		// Split camelCase into words (getWeather -> "get weather") and match each
 		// on word boundaries. Skip short words (<=2 chars) to avoid noise.
 		const words = t.name.replace(/([a-z])([A-Z])/g, '$1 $2').toLowerCase().split(' ');
-		return words.some(w => w.length > 2 && promptMentionsWord(lower, w));
+		if (words.some(w => w.length > 2 && promptMentionsWord(lower, w))) return true;
+		// Extra trigger keywords declared via `cannedTriggers`. Single words use word
+		// boundaries (consistent with tool-name matching); multi-word phrases use substrings.
+		const triggers = hints?.get(t.name)?.triggers;
+		return triggers?.some(tr => {
+			const low = tr.trim().toLowerCase();
+			if (!low) return false;
+			return low.includes(' ') ? lower.includes(low) : promptMentionsWord(lower, low);
+		}) ?? false;
 	}).map(t => t.name);
 }
 
@@ -93,7 +108,11 @@ function generatePlaceholderInput(schema: any): any {
 	if (schema.type === 'object' && schema.properties) {
 		const result: Record<string, any> = {};
 		for (const [key, prop] of Object.entries(schema.properties) as [string, any][]) {
-			if (prop.type === 'string') {
+			// Schema default (from Zod `.default()`) is the most realistic value — prefer it,
+			// and populate fields whose only signal is a default with no recognized `type`.
+			if (prop.default !== undefined) {
+				result[key] = prop.default;
+			} else if (prop.type === 'string') {
 				if (prop.enum?.length) result[key] = prop.enum[0];
 				else result[key] = 'sample';
 			} else if (prop.type === 'number' || prop.type === 'integer') {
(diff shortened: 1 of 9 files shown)


packages/bb-agent/src/providers/canned.ts
```diff
@@ -111,21 +132,27 @@ function generatePlaceholderInput(schema: any): any {
...
+ * Look up a tool's inputSchema from toolSpecs and generate placeholder input, shallow-merging
+ * any `cannedExamples` (from hints) on top so realistic values win over generated placeholders.
+ */
+function getToolInput(toolName: string, toolSpecs?: { name: string; inputSchema?: any }[], hints?: Map<string, CannedToolHints>): string {
 	const spec = toolSpecs?.find(t => t.name === toolName);
-	if (!spec?.inputSchema) return '{}';
-	return JSON.stringify(generatePlaceholderInput(spec.inputSchema));
+	const base = spec?.inputSchema ? generatePlaceholderInput(spec.inputSchema) : {};
+	const examples = hints?.get(toolName)?.examples;
+	return JSON.stringify(examples ? { ...base, ...examples } : base);
```
ahmedhamouda78: `cannedExamples` keys aren't validated against the schema.** A mistyped/extra key is merged verbatim into the emitted input and never surfaces. An optional dev-only warning when a key isn't in `inputSchema.properties` could save some debugging. Fine to skip.

packages/bb-agent/src/providers/canned.ts
```diff
@@ -60,7 +67,17 @@ function findAllToolMatches(prompt: string, toolSpecs?: { name: string }[]): str
...
+		if (words.some(w => w.length > 2 && promptMentionsWord(lower, w))) return true;
+		// Extra trigger keywords declared via `cannedTriggers`. Single and multi-word triggers
+		// both match on word boundaries (consistent with tool-name matching), so "log in" is not
+		// triggered by "backlog in" and internal whitespace is flexible (matches one-or-more spaces).
+		const triggers = hints?.get(t.name)?.triggers;
+		return triggers?.some(tr => {
+			const low = tr.trim().toLowerCase();
+			if (!low) return false;
+			const escaped = low.replace(/[.*+?^${}()|[\]\\]/g, '\\$&').replace(/\s+/g, '\\s+');
+			return new RegExp(`\\b${escaped}\\b`).test(lower);
```
ahmedhamouda78: **trigger regex is recompiled per evaluation** (`new RegExp` inside `.some()`/`.filter()`). Negligible at mock scale; only worth caching if trigger lists grow.

ahmedhamouda78 (dismissed):
**Approve** ✅ — well-scoped, backward-compatible, inert for real providers. Two optional nits, nothing blocking.

soberm (dismissed):

packages/bb-agent/src/providers/canned.ts
```diff
@@ -93,7 +110,11 @@ function generatePlaceholderInput(schema: any): any {
 	if (schema.type === 'object' && schema.properties) {
 		const result: Record<string, any> = {};
 		for (const [key, prop] of Object.entries(schema.properties) as [string, any][]) {
-			if (prop.type === 'string') {
+			// Schema default (from Zod `.default()`) is the most realistic value — prefer it,
+			// and populate fields whose only signal is a default with no recognized `type`.
+			if (prop.default !== undefined) {
+				result[key] = prop.default;
```
osama-rizk: Nice addition — preferring the schema `default` (and the `!== undefined` check so `0`/`false`/`''` survive) is right.

One gap worth noting (pre-existing in `generatePlaceholderInput`, but it now undercuts this function's "Produces values that pass validation" contract): the type ladder handles `string`/`number`/`integer`/`boolean`/`array`/`object` + `default`, but a property that is a union / `anyOf` / `const` / untyped and has **no** `default` matches no branch, so its key is omitted from the generated input entirely. For a **required** field of that shape, the emitted canned tool input is missing it → Strands/Zod validation can reject the call. Local-dev-only and `cannedExamples` works around it, but consider a generic fallback for unrecognized types (or documenting the limitation).

packages/bb-agent/src/providers/canned.ts
```diff
@@ -60,7 +67,17 @@ function findAllToolMatches(prompt: string, toolSpecs?: { name: string }[]): str
 		// Split camelCase into words (getWeather -> "get weather") and match each
 		// on word boundaries. Skip short words (<=2 chars) to avoid noise.
 		const words = t.name.replace(/([a-z])([A-Z])/g, '$1 $2').toLowerCase().split(' ');
-		return words.some(w => w.length > 2 && promptMentionsWord(lower, w));
+		if (words.some(w => w.length > 2 && promptMentionsWord(lower, w))) return true;
```
osama-rizk, line 90: Consistency nit (pre-existing, out of scope): this PR correctly moves *tool* matching to word boundaries via `promptMentionsWord`, but the canned *text-response* matcher `matchResponse` (line 43) still uses substring `lower.includes(keyword)` for `weather`/`order`/`help` — so `reorder`→`order` and `helper`→`help`, exactly the false-positive class this PR eliminates for triggers. Not blocking; reusing `promptMentionsWord` there (or a one-line note) would make the file internally consistent.

packages/bb-agent/src/providers/canned.ts
soberm: Done — implemented in 41c72b9c.

`getToolInput` now calls `warnUnknownExampleKeys`, which logs once per tool+field when a `cannedExamples` key isn't in the tool's `inputSchema.properties`:

```
[canned] cannedExamples for tool "fetchPage" sets "rul", which is not a field of its parameters schema (url). Check for a typo — the value is still sent.
```

Kept deliberately inert:

- **Never throws**, and the value is still merged through — a bad hint must not break local dev.
- **Deduped** via a module-level `Set` keyed `tool.field`, so a mistyped hint warns once rather than on every `stream()` call.
- **Skipped when the schema exposes no `properties`** (e.g. `inputSchema: {}`), where an unknown key is unknowable — this avoids false positives on the many specs that pass a bare schema.
- **Dev-only by construction**: the canned provider *is* the local-dev mock, so this can't fire in a deployed agent. (`noConsole` is `off` in `biome.json` and `index.hooks.ts` already uses `console.warn`, so it matches existing practice rather than introducing a new logging pattern.)

Covered by `warns on a cannedExamples field missing from the tool schema, but still sends it`, which asserts both that the warning names the tool and field *and* that the value still ships.

packages/bb-agent/src/providers/canned.ts
soberm: Done — implemented in 41c72b9c.

Hoisted compilation into a phrase-keyed cache (`wordPatternCache`) behind `wordBoundaryPattern(phrase)`, shared by **both** matching paths — tool names / camelCase words and `cannedTriggers` — so nothing recompiles inside the `.filter()` / `.some()` callbacks.

Two incidental wins:

- It removes the duplicated metacharacter-escape expression that had been copy-pasted between `promptMentionsWord` and the trigger branch, so the escaping rule now lives in exactly one place.
- The key space is bounded by the agent's tool + trigger set, so the cache can't grow unboundedly.

Semantics are unchanged: for the whitespace-free inputs `promptMentionsWord` actually receives, escape-then-`\s+`-convert produces a byte-identical pattern to escape-only. The three word-boundary regression tests (`cat`/`category`, `log in`/`backlog in`, and flexible internal whitespace) all still pass unchanged.

packages/bb-agent/src/providers/canned.ts
soberm: Good catch — this was a real hole, fixed in f34df6bc.

Property resolution moved into a `placeholderForProperty` helper that now understands, in precedence order: `default` → `const` → `enum` (at *any* type, not just `string` as before) → `anyOf`/`oneOf` (first variant that resolves, which covers `z.union` / `z.discriminatedUnion`) → the per-type ladder.

On the required-field case you flagged, I split the behavior deliberately rather than adding one blanket fallback:

- A **required** field whose shape still yields nothing now falls back to `'sample'`, so the emitted call stops failing validation before the tool ever runs. That restores the function's "produces values that pass validation" contract.
- An **optional** field whose shape yields nothing stays omitted. Absence is valid there, and a blanket fallback would be a regression rather than a fix: inventing `'sample'` for an optional `z.union([z.number(), z.boolean()])` would turn a call that currently validates into one that doesn't.

Four tests cover it: `resolves a union (anyOf) field from its first usable variant`, `resolves a const field to its fixed value`, `fills a required field whose shape yields no placeholder`, and `leaves an optional field whose shape yields no placeholder omitted` (the last one pins the deliberate asymmetry above).

packages/bb-agent/src/providers/canned.ts
soberm: Agreed, and fixed in f34df6bc — `matchResponse` now uses `promptMentionsWord` instead of `lower.includes(keyword)`, so the whole file matches keywords one way.

It was a genuine inconsistency rather than a cosmetic one: this branch already asserts that `'please reorder the list alphabetically'` must not trigger `getOrder`, yet that exact prompt still returned the *order* canned **text** response once it fell through the tool matcher. Same false-positive class, one path behind.

Verified nothing was lost: `'tell me about the weather'` still matches `\bweather\b`, so the existing `returns keyword response for weather` test passes unchanged. Added `does not return a keyword response when the keyword is only a substring`, which asserts `'please reorder...'` now falls through to the default response instead of the order one.

soberm (approved):

osama-rizk (approved):

PR #150, opened 2026-07-03
fix(create-blocks-app): add backend template build script

### LINKED ISSUE #138: create-blocks-app backend template ships no npm run build script (agents add one to build)
## Summary
The `backend` template's `package.json` defines `typecheck` (`tsc --noEmit`) but no `build` script, unlike bare/demo/default (`tsc && vite build`). A backend-only app fails `npm run build`.

## Evidence (run 28585569147, observability-api/backend)
agent had to add a missing build script ("build": "tsc --noEmit").

## Root cause - CONFIRMED @7207d7e
packages/create-blocks-app/templates/backend/package.json has typecheck but no build; other templates ship build.

## Suggested fix
Add "build": "tsc --noEmit" to the backend template's package.json.

## Environment
agent bench PR #132, run 28585569147, HEAD 7207d7e.

ikenyal:
## Problem

**Issue #, if available:** Closes #138

Fresh apps scaffolded with the `backend` template did not include a `build` script. That made the common `npm run build` workflow fail with a missing-script error even though the template already had the equivalent `typecheck` script.

## Changes

- Add `"build": "tsc --noEmit"` to the backend template package scripts.
- Add a `create-blocks-app` regression test that asserts the backend template keeps a build script.

## Validation

- `npm run build -w packages/create-blocks-app`
- `npm run test -w packages/create-blocks-app`
- Scaffolded a fresh backend app from the local dist CLI:
  - `node packages/create-blocks-app/dist/index.js /private/tmp/aws-blocks-backend-build-template --template backend -y --skip-install`
  - `npm install`
  - `npm run build`
- `./node_modules/.bin/changeset status --since=origin/main`
- `./node_modules/.bin/tsx scripts/verify-changeset-coverage.ts`
- `git diff --cached --check`

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (and PR referenced) — not applicable; this fixes template behavior and adds a regression test.

---


packages/create-blocks-app/src/cli.test.ts
```diff
@@ -132,6 +132,14 @@ describe('create-blocks-app CLI argument parsing', () => {
   });
 });
 
+describe('create-blocks-app template metadata', () => {
+  it('backend template has a build script', () => {
+    const pkg = JSON.parse(readFileSync(join(__dirname, '../templates/backend/package.json'), 'utf-8'));
+
```
sharonyajain: **Regression test pins the exact build command instead of asserting a build script exists**

The regression that #138 fixed is a MISSING build script, and the description frames the test as asserting the template 'keeps a build script'. The assertion instead hard-codes the exact command string `'tsc --noEmit'`. That couples the test to one specific implementation of `build` rather than to the property being guarded (presence of a build script).

```ts
assert.strictEqual(pkg.scripts.build, 'tsc --noEmit');
```

**Why it matters:** A future, legitimate change to the backend build command (cause) makes the exact-string assertion fail even though a valid build script is still present (mechanism) → a spurious CI red and needless churn on an unrelated change (system-visible harm).

**Suggestion:** Assert the property that actually matters, e.g. `assert.ok(typeof pkg.scripts.build === 'string' && pkg.scripts.build.length > 0)`; if you also want to guard the type-check semantics, assert `pkg.scripts.build.includes('tsc')` rather than the full literal. Non-blocking.

ikenyal: Updated in 23a26758. The regression now asserts that scripts.build is a non-empty string, so it protects against the missing-script regression without coupling the test to the current build command.

sharonyajain (dismissed):

packages/create-blocks-app/templates/backend/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "workspaces": ["aws-blocks"],
   "scripts": {
+    "build": "tsc --noEmit",
```
sarayev: minor: this comes out identical to the existing `typecheck` (`tsc --noEmit`). Backend has no bundle to emit so that's fair, but could we just point `build` at it (`"build": "npm run typecheck"`) so the two can't quietly drift apart later? Non-blocking.

packages/create-blocks-app/src/cli.test.ts
```diff
@@ -132,6 +132,14 @@ describe('create-blocks-app CLI argument parsing', () => {
   });
 });
 
+describe('create-blocks-app template metadata', () => {
+  it('backend template has a build script', () => {
```
sarayev: minor: nice regression guard. Since all the deployable templates carry a `build`, would it be worth looping over them here rather than just backend, so the others are protected too? Happy to keep it scoped to this PR though.

sarayev (dismissed):

packages/create-blocks-app/src/cli.test.ts
ikenyal: Updated in 8619aa82. The guard now checks every deployable template, using the existing `sandbox` lifecycle criterion; the Amplify overlay remains out of scope.

packages/create-blocks-app/templates/backend/package.json
ikenyal: Updated in 8619aa82. `build` now delegates to `npm run typecheck`, so the command is defined in one place.

sharonyajain (approved):

sarayev (approved):

Issue #131, opened 2026-07-01
Flaky E2E Sandbox: comprehensive AuthBasic test relies on Lambda in-memory `lastDeliveredCode` (cross-invocation race)


hfurkanbozkurt:
## Summary
The comprehensive test-app's E2E **Sandbox** job is flaky against real AWS. The AuthBasic harness reads back verification codes through an **in-memory module global** (`lastDeliveredCode`) that only exists on the single Lambda instance whose `codeDelivery` callback fired. On the sandbox, each API call is a separate Lambda invocation, so the code-poll can hit an instance that never saw the code and times out. The same flow is **deterministic locally** (single in-process dev server shares one global), so unrelated PRs get blocked by red E2E Sandbox with nothing wrong in their diff.

## Symptom (two observed manifestations of the same flake class)
Both on PR #127 (`fix/react-template-single-origin-front-door`), a **template-only** change:
- **basic-auth**, run **28472353050**: `AuthBasic > session > signIn sets session cookie — checkAuth returns true` ✖ (15555ms) — `Error: authGetLastCode() returned null after 15000ms` at `test-apps/comprehensive/test/basic-auth.test.ts:27` (pollForCode).
- **realtime**, run **28445442255**: `Realtime > Subscribe Options Form > options form receives messages via onMessage` ✖ (10227ms) — `Error: No message within 10s` at `test-apps/comprehensive/test/realtime.test.ts:170`.

**Why it's non-deterministic, not a real bug:**
- In run 28472353050, sibling tests using the **identical** `pollForCode` → `authGetLastCode()` logic **passed in the same run** (`getCurrentUser returns user` ✔, `requireAuth returns user` ✔).
- In run 28445442255, **E2E Local, Production, and Sandbox VPC all passed — only E2E Sandbox failed**.

## Root cause
`test-apps/comprehensive/aws-blocks/index.ts`: `lastDeliveredCode` is a module-level global (**L56**) set in `codeDelivery` (**L60-64**) and read by `authGetLastCode()` (**L886-888**) — per-Lambda-instance in-memory state, not a shared store. On sandbox, `authSignUp` and the `authGetLastCode` poll are distinct Lambda invocations; a poll pinned to an instance that never ran `codeDelivery` returns `null` for the full 15s.

## Secondary issue — coarse Detect Changes gate
[...]

PR #152, opened 2026-07-04
test(comprehensive): persist AuthBasic codes in KVStore

ikenyal:
## Problem

The comprehensive test app stores the latest AuthBasic verification code in a module-level variable. That is deterministic in the local single-process dev server, but it is not shared across Lambda instances in the sandbox runtime. As described in #131, `authGetLastCode()` can therefore poll an instance that never ran `codeDelivery` and return `null` until the test times out.

**Issue #, if available:** Partially addresses #131 (AuthBasic code-delivery state only)

## Changes

- Persist the latest delivered AuthBasic code through the comprehensive app's existing `KVStore` instead of a module global.
- Keep the change scoped to the test app harness and avoid adding another backend resource by reusing the existing store with a namespaced key.
- Leave the realtime flake and the workflow detect-changes gate from #131 for separate, focused follow-up changes.

## Validation

- `npx tsc --noEmit` from `test-apps/comprehensive`
- `npm run test:e2e:local` from `test-apps/comprehensive` (372 tests, 371 passed, 1 skipped)
- `git diff --check`
- Focused AuthBasic smoke against the comprehensive backend on `localhost:3055`:
  - `authSignUp` -> `authGetLastCode` -> `authConfirmSignUp` -> `authSignIn`
  - `authResetPassword` -> `authGetLastCode` -> `authConfirmResetPassword` -> `authSignIn` with the new password

I did not add a new test file because the existing comprehensive AuthBasic E2E already covers this flow; this PR changes the test harness storage backing so that the existing sandbox E2E can read the code across Lambda invocations.

## Checklist

- [x] PR description included

---

### DIFF
--- a/test-apps/comprehensive/aws-blocks/index.ts
+++ b/test-apps/comprehensive/aws-blocks/index.ts
@@ -52,13 +52,16 @@ const profileSchema = z.object({ name: z.string(), age: z.number(), tags: z.arra
 const validatedStore = new KVStore(scope, 'validated-store', { schema: profileSchema });
 
 // AuthBasic - Authentication
-// Store last delivered code so e2e tests can retrieve and use it.
-let lastDeliveredCode: { username: string; code: string } | null = null;
+// Store the last delivered code so e2e tests can retrieve and use it. This
+// must be shared storage: sandbox/production may read from a different Lambda
+// instance than the one that delivered the code.
+type DeliveredAuthBasicCode = { username: string; code: string };
+const authBasicLastCodeKey = 'auth-basic:last-code';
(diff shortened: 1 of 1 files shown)


ikenyal:
While reviewing this against the current main branch, I confirmed that #278 has superseded this fix. It persists delivered verification codes in the shared KVStore for AuthBasic, AuthCognito, and AuthCognito MFA, keys reads by username, and adds regression coverage for cross-user isolation. That also closes the remaining cross-user race left by this PR's single auth-basic:last-code key. Keeping this PR would duplicate a narrower version of the merged implementation, so I am closing it.

PR #160, opened 2026-07-06
fix(data-common): defer getEngine() in createKyselyAdapter so adapters are safe at module scope

fossamagna:
## Problem

**Issue #, if available:** #192

`createKyselyAdapter()` eagerly calls `db.getEngine()` at construction. A backend `index.ts` is also loaded during `cdk synth` (via `BlocksStack.create`'s `backendCDKPath`), and the infra-only **cdk** builds of `DistributedDatabase` / `Database` expose no engine. So the idiomatic-looking module-scope pattern:

```ts
const db = createKyselyAdapter<Schema>(store); // at module scope
```

crashes synth with a cryptic `db.getEngine is not a function` — even though the adapter is only ever used at request time. Today this forces app authors to hand-roll a lazy/memoized wrapper, or to only ever call `createKyselyAdapter()` inside handlers (as the `comprehensive` test-app does).

## Changes

- **`@aws-blocks/data-common`** — `createKyselyAdapter()` now passes a thunk (`() => db.getEngine()`) into the Kysely dialect instead of calling `db.getEngine()` eagerly. The engine is resolved lazily on the first query, still memoized per `EngineConnection` so the one-engine-per-transaction guarantee behind the handle-based transaction API (`beginTransaction` / `queryInTransaction` / `commitTransaction`) is preserved. Adapter creation is now side-effect free and safe at module scope. **Public API and runtime behavior are unchanged.**
- **`@aws-blocks/bb-distributed-data` / `@aws-blocks/bb-data`** — the **cdk** builds gain a `getEngine(): never` that throws a clear, actionable message if a query is ever reached during synth, replacing the cryptic "is not a function".
- Updated the `createKyselyAdapter` JSDoc to state that `getEngine()` is deferred and adapter creation is safe at module scope.
- Added a changeset (patch × 3 packages).

## Validation

- **New regression test** (`kysely-adapter.test.ts`): asserts `getEngine()` is **not** called at adapter construction and only on the first query. Verified red (before) → green (after).
- Full suites pass: `data-common` **32/32**, `bb-distributed-data` **137/137**, `bb-data` **362 pass / 0 fail** (4 pre-existing skips).
- `tsc --noEmit` passes for `bb-distributed-data` and `bb-data`.
- Existing transaction/rollback adapter tests remain green, confirming the per-connection memoization still holds within a transaction.

> Note: a full end-to-end `cdk synth` was not run (it requires the `BlocksStack` harness); the deferral is covered by the unit test whose `db.getEngine()` throws, mirroring the cdk build's new behavior.

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (and PR referenced) — *only inline JSDoc updated; no external docs require changes*

---


packages/data-common/src/kysely-adapter.ts
```diff
@@ -28,14 +36,17 @@ type EngineSource = DatabaseEngine | Promise<DatabaseEngine>;
...
-  private enginePromise: Promise<DatabaseEngine>;
+  private enginePromise?: Promise<DatabaseEngine>;
+
+  constructor(private getEngine: EngineFactory) {}
 
-  constructor(engineSource: EngineSource) {
-    this.enginePromise = Promise.resolve(engineSource);
+  /** Resolve the engine on first use, memoizing it for this connection. */
+  private engine(): Promise<DatabaseEngine> {
+    return (this.enginePromise ??= Promise.resolve(this.getEngine()));
```
ahmedhamouda78: Memoization silently no-ops when `getEngine()` throws synchronously. In `this.enginePromise ??= Promise.resolve(this.getEngine())`, the RHS is evaluated before the assignment — so if `getEngine()` throws synchronously (exactly what the CDK stubs do), `Promise.resolve(...)` is never reached, the assignment never happens, and `enginePromise` stays `undefined`. Every subsequent `engine()` call then re-invokes `getEngine()`.

Harmless for the CDK path (nothing queries at synth) and fine for the runtime success path, but it means the "called once per connection" guarantee isn't actually enforced for the throwing path. Suggested form:

```ts
private engine(): Promise<DatabaseEngine> {
  if (!this.enginePromise) {
    try {
      this.enginePromise = Promise.resolve(this.getEngine());
    } catch (err) {
      // Don't memoize a synchronous failure — allow retry on transient init errors.
      return Promise.reject(err);
    }
  }
  return this.enginePromise;
}
```

Worth a comment on whether a permanent rejection (no retry) or fresh rejection (retry) is desired.

packages/data-common/src/kysely-adapter.test.ts
```diff
@@ -141,3 +141,25 @@ test('Kysely transaction rollback uses the engine rollback API', async () => {
     `expected engine.rollbackTransaction() on failure; saw calls: ${JSON.stringify(engine.calls)}`,
   );
 });
+
+test('createKyselyAdapter defers getEngine() until the first query (safe during CDK synth)', async () => {
```
ahmedhamouda78, line 145: This test uses a throwing `getEngine()`, so it exercises the *non-memoized* path — `calls === 1` holds only because a single query issues a single `engine()` call, not because memoization works. It proves "no eager call at construction" (great) but not "getEngine is called once across begin/execute/commit".

Suggest a complementary success-path test to lock in the memoization contract:

```ts
test('getEngine() is called once per EngineConnection across begin/execute/commit', async () => {
  let calls = 0;
  const engine = new RecordingEngine();
  const kysely = createKyselyAdapter<Schema>({ getEngine: () => { calls++; return engine; } });
  await kysely.transaction().execute(async (trx) => {
    await trx.insertInto('t').values({ id: 'a', value: 'one' }).execute();
  });
  assert.strictEqual(calls, 1, `expected memoized single call, saw ${calls}`);
});
```

packages/data-common/src/kysely-adapter.ts
```diff
@@ -20,6 +20,14 @@ import type { DatabaseEngine, TransactionHandle } from './engine.js';
...
  */
 type EngineSource = DatabaseEngine | Promise<DatabaseEngine>;
+/**
+ * A thunk that produces the engine. The engine is obtained lazily — only when a
+ * query actually runs — so `createKyselyAdapter()` itself never calls the
+ * underlying `getEngine()`. This keeps adapter creation safe at module scope,
+ * which matters because backend files are also loaded during CDK synth, where
+ * the infra-only block builds expose no engine.
+ */
+type EngineFactory = () => EngineSource;
```
ahmedhamouda78: Nit: `EngineSource` is now only referenced inside `EngineFactory`. Could inline it to trim the type vocabulary:

```ts
type EngineFactory = () => DatabaseEngine | Promise<DatabaseEngine>;
```

(moving the existing JSDoc onto `EngineFactory`).

packages/bb-data/src/index.cdk.ts
```diff
@@ -87,6 +87,21 @@ export class Database extends Scope {
...
+  /**
+   * Runtime-only. This is the CDK (synth) build: it defines infrastructure and
+   * has no engine — queries run in the app Lambda against the deployed database.
+   * `createKyselyAdapter()` no longer calls this eagerly, so reaching it means a
+   * query ran at synth time (e.g. at module scope). Throw a clear, actionable
+   * message instead of the cryptic "getEngine is not a function".
+   */
+  getEngine(): never {
+    throw new Error(
+      'Database.getEngine() is unavailable during CDK synth. Create the Kysely ' +
```
ahmedhamouda78: Nit: the error message references "the Kysely adapter" specifically. A developer calling `db.getEngine()` directly would find that confusing. Minor wording tweak to cover both cases:

`'Database.getEngine() is unavailable during CDK synth. Queries must run at request time inside an API handler. If using createKyselyAdapter(), ensure the adapter is created before making queries, not at module scope.'`

packages/bb-distributed-data/src/index.cdk.ts
```diff
@@ -108,6 +108,21 @@ export class DistributedDatabase extends Scope {
...
+  /**
+   * Runtime-only. This is the CDK (synth) build: it defines infrastructure and
+   * has no engine — queries run in the app Lambda against the deployed cluster.
+   * `createKyselyAdapter()` no longer calls this eagerly, so reaching it means a
+   * query ran at synth time (e.g. at module scope). Throw a clear, actionable
+   * message instead of the cryptic "getEngine is not a function".
+   */
+  getEngine(): never {
+    throw new Error(
+      'DistributedDatabase.getEngine() is unavailable during CDK synth. Create the ' +
```
ahmedhamouda78: Nit: same as bb-data — the "Kysely adapter" wording is confusing for direct `getEngine()` callers. Consider generalizing the message to mention both the direct-call and createKyselyAdapter cases.

packages/data-common/src/kysely-adapter.ts
```diff
@@ -174,5 +186,6 @@ class EngineDialect implements Dialect {
 export function createKyselyAdapter<T>(
   db: { getEngine(): DatabaseEngine | Promise<DatabaseEngine> },
 ): Kysely<T> {
-  return new Kysely<T>({ dialect: new EngineDialect(db.getEngine()) });
+  // Pass a thunk, not `db.getEngine()`: defer the call to the first query.
+  return new Kysely<T>({ dialect: new EngineDialect(() => db.getEngine()) });
```
ahmedhamouda78, line 190: Nit: the `// Pass a thunk, not db.getEngine()...` comment restates what the code + `EngineFactory` type already convey — fine to drop or tighten to `// thunk — deferred per EngineFactory contract`. Also the CDK-synth rationale is duplicated across the `EngineFactory` and `createKyselyAdapter` JSDoc blocks; consider keeping the full explanation only on `createKyselyAdapter`.

ahmedhamouda78 (dismissed):
Nice, minimal fix — the `() => db.getEngine()` thunk threaded through EngineDialect → EngineDriver → EngineConnection is exactly the right approach, and `getEngine(): never` is the correct type for the CDK stubs. Approving.

Two things worth addressing (non-blocking), plus a few nits inline.

ahmedhamouda78 (dismissed):

packages/bb-data/src/index.cdk.ts
```diff
@@ -87,6 +87,21 @@ export class Database extends Scope {
...
   }
 
+  /**
+   * Runtime-only. This is the CDK (synth) build: it defines infrastructure and
+   * has no engine — queries run in the app Lambda against the deployed database.
+   * `createKyselyAdapter()` no longer calls this eagerly, so reaching it means a
+   * query ran at synth time (e.g. at module scope). Throw a clear, actionable
+   * message instead of the cryptic "getEngine is not a function".
+   */
+  getEngine(): never {
```
soberm, line 97: We already have https://github.com/aws-devtools-labs/aws-blocks/blob/eb419df3f0be82c2f7c3e15a5546c0d46eac32f1/packages/core/src/cdk/synth-guard.ts#L16. Can we reuse it?

packages/data-common/src/kysely-adapter.ts
fossamagna: Fixed as suggested — `engine()` now try/catches the first resolution and returns `Promise.reject(err)` without memoizing a synchronous failure, with an inline comment stating the intent (transient init errors may retry). One note: an *async* rejection (a rejected promise from `getEngine()`) is still memoized, so the two failure modes remain asymmetric. Left as-is since the failure path does not affect the transaction guarantee; can follow up if retry-on-async-failure is wanted too.

packages/data-common/src/kysely-adapter.test.ts
fossamagna: Added the suggested success-path test — `getEngine() is called once per EngineConnection across begin/execute/commit` locks in the memoization contract with a counting `getEngine()`.

packages/data-common/src/kysely-adapter.ts
fossamagna: Inlined — `EngineFactory` is now `() => DatabaseEngine | Promise<DatabaseEngine>` with the JSDoc merged onto it.

packages/bb-data/src/index.cdk.ts
fossamagna: Reworked further per soberm's suggestion below: `getEngine()` now delegates to the shared `synthGuard()` from `@aws-blocks/core/cdk`, so the message is the generic runtime-method guard and no longer mentions Kysely at all.

packages/bb-distributed-data/src/index.cdk.ts
fossamagna: Same as bb-data — now delegates to the shared `synthGuard()` from `@aws-blocks/core/cdk`, so the Kysely-specific wording is gone.

packages/data-common/src/kysely-adapter.ts
fossamagna: Dropped the inline comment and consolidated the CDK-synth rationale on `createKyselyAdapter`'s JSDoc; `EngineFactory` (with `EngineSource` now inlined) keeps only a short pointer there.

packages/bb-data/src/index.cdk.ts
fossamagna: Done — both cdk builds now delegate to `synthGuard('Database' | 'DistributedDatabase', 'getEngine')`, matching the existing pattern in bb-kv-store / bb-realtime / bb-knowledge-base.

soberm (approved):

ahmedhamouda78 (approved):

PR #163, opened 2026-07-06
feat(core): add throttling and access logging to central API Gateway

bobbor:
## Problem

The central API Gateway (`setupBlocksInfra` in `@aws-blocks/core`) was created with no request throttling and no access logging. That leaves the Blocks API with unbounded request rates (no protection against traffic spikes / abuse) and no per-request audit trail for debugging or security review.

## Changes

Added stage-level defaults to the `RestApi` deployment in `packages/core/src/cdk/blocks-backend.ts`:

- **Throttling** — `throttlingRateLimit: 100` (steady-state req/s) and `throttlingBurstLimit: 200` (token-bucket burst ceiling; over-limit requests get 429s). Documented inline with the token-bucket semantics.
- **Access logging** — structured JSON logs (`AccessLogFormat.jsonWithStandardFields`: IP, method, resource path, status, response length, request time, caller, user, protocol) delivered to a dedicated `ApiAccessLogs` CloudWatch log group.
- **Log group** — 1-month retention with `RemovalPolicy.DESTROY` so stack teardown stays clean (matches the existing `bb-logger` convention).

These are account-level stage defaults and can be overridden per-method later if needed. Changeset included (`minor` bump for `@aws-blocks/core`).

## Validation

- Existing `blocks-backend` test suite: 7/8 pass. The one failure (`@aws-blocks/pipeline/dist/index.js` not found) is an unbuilt sibling-package artifact pulled in transitively via `src/cdk/index.ts`, unrelated to this change.
- Wrote a throwaway CDK synth check asserting the synthesized template: `AWS::ApiGateway::Stage` carries `ThrottlingRateLimit: 100` / `ThrottlingBurstLimit: 200` / `AccessLogSetting`, and `AWS::Logs::LogGroup` has `RetentionInDays: 30`. Passed.

## Checklist

- [x] PR description included

---

### DIFF
--- a/packages/core/src/cdk/blocks-backend.ts
+++ b/packages/core/src/cdk/blocks-backend.ts
@@ -4,6 +4,7 @@
 import * as cdk from 'aws-cdk-lib';
 import * as lambda from 'aws-cdk-lib/aws-lambda-nodejs';
 import * as apigateway from 'aws-cdk-lib/aws-apigateway';
+import * as logs from 'aws-cdk-lib/aws-logs';
 import { CfnGroup } from 'aws-cdk-lib/aws-resourcegroups';
 import { Construct } from 'constructs';
 import { pathToFileURL } from 'node:url';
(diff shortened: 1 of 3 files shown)


packages/core/src/cdk/blocks-backend.ts
```diff
@@ -82,9 +83,34 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
...
 
+  const accessLogGroup = new logs.LogGroup(scope, 'ApiAccessLogs', {
+    retention: logs.RetentionDays.ONE_MONTH,
+    removalPolicy: cdk.RemovalPolicy.DESTROY,
+  });
+
   const api = new apigateway.RestApi(scope, 'API', {
     restApiName: 'Blocks API',
-    deployOptions: { cachingEnabled: false },
+    deployOptions: {
```
sarayev, line 93: IMHO, I don't think that hardcoding all those values without providing a way to override is a good way. Developers, who need different options here, should be able to override those. But, I am not sure to which degree, and for what options we have to provide customization.

osama-rizk:
Thanks for this, Philipp — this was the original catch on the core API's throttling/logging gap, and the 100/200 default + 30d/`DESTROY` log-group convention in #302 trace directly back to it.

Heads-up on some overlap: #302 (`feat/shared-hardening-defaults`) sets up a **cross-block hardening mechanism** — a stack-level registry with `resolveApiThrottle` / `resolveApiAccessLogs` / `resolveLogRetention` — so throttling, access logs, and retention are configured and named *one* way across every block. Your changes here are exactly what those resolvers encode, so the core REST API would become a follow-up adopter.

Also worth knowing: #326 independently adds throttling to the same core REST API (at 200/400), so it and this PR are near-duplicates — we should collapse to one and route it through `resolveApiThrottle` to avoid two defaults. cc @soberm

Suggested path: land #302's mechanism first, then adopt it here. Thanks again for surfacing the gap.

PR #165, opened 2026-07-07
docs: document AuthCognito setAuthState errors

### LINKED ISSUE #144: docs: document AuthCognito error semantics (baseSignedOut return, retriable flag)
Discovered via agent-bench (PR #132), cognito-profile cell, run 28644861376.

Problem: The AuthCognito (bb-auth-cognito) error contract is undocumented and only visible in compiled dist/index.js:
- setAuthState does NOT throw on error; it returns baseSignedOut(message, errorName) as an AuthState with an error field.
- A wrong OTP on confirmSignIn is marked retriable: true and the session is preserved for retry.

Impact: Developers wrap setAuthState in try/catch (which never fires) and mishandle the wrong-code retry path. The bench agent spent ~60 cycles / ~3.85M input tokens reverse-engineering this from compiled source.

Ask: Document in bb-auth-cognito.md the error-return-vs-throw contract, the returned error AuthState shape, and the retriable flag on wrong-code responses.

snvtac:
## Summary

- Document that AuthCognito `setAuthState()` resolves auth failures to an error `AuthState` instead of rejecting.
- Show the returned non-retriable and retriable error-state shapes.
- Call out that `CodeMismatchException` from `confirmSignIn` is retriable and custom UI should preserve the current challenge session.
- Add changeset coverage for `@aws-blocks/bb-auth-cognito` and the umbrella `@aws-blocks/blocks` docs package.

Fixes #144.

## Validation

- `git diff --check`
- `rg -n 'For the `setAuthState\(\)` path|retriable: true|CodeMismatchException|actions: \[\]' packages/bb-auth-cognito/README.md`
- `rg -n '"@aws-blocks/(bb-auth-cognito|blocks)": patch' .changeset/cognito-error-docs.md`

No dependency install was needed for this docs-only change.


packages/bb-auth-cognito/README.md
```diff
@@ -439,6 +439,38 @@ if (hasAuthError(next, AuthCognitoErrors.NotAuthorized)) {
 
 Rule of thumb: **throw path → `isBlocksError`; returned `AuthState` → `hasAuthError`.** Never match on the human-facing `error` string.
 
+For the `setAuthState()` path, auth failures resolve to an error state instead
+of rejecting the promise. A non-retriable failure returns the normal signed-out
+state with the available sign-in actions plus `error` and, when available,
```
sharonyajain, line 444: The `errorName` field is more conditional than "when available" suggests. In both runtimes it's only populated when the thrown error is an `ApiError` whose name differs from the default (`e instanceof ApiError && e.name !== DEFAULT_API_ERROR_NAME`). A plain `Error`, or an `ApiError` left at the default name, resolves to a state with `error` set but no `errorName` key at all. Could we add a sentence noting `errorName` is best-effort? Otherwise developers may write `switch (state.errorName)` assuming it's always present, which undercuts the doc's own (correct) advice to match on `errorName`/`hasAuthError`.

packages/bb-auth-cognito/README.md
```diff
@@ -439,6 +439,38 @@ if (hasAuthError(next, AuthCognitoErrors.NotAuthorized)) {
...
+{
+  state: 'signedOut',
+  actions: [],
+  error: 'Invalid code',
+  errorName: AuthCognitoErrors.CodeMismatch,
+  retriable: true,
+}
+```
+
+`CodeMismatchException` from `confirmSignIn` (for example, a wrong MFA or OTP
```
sharonyajain, line 468: The retriable set is broader than just `CodeMismatchException`. Per `isRetriableAuthError` in `src/types.ts`, `EnableSoftwareTokenMFAException` (wrong TOTP during MFA setup), `InvalidParameterException`, and `InvalidPasswordException` are also retriable — and `CodeMismatch` itself eventually escalates to a non-retriable `NotAuthorizedException` ("Invalid session") after repeated failures. As written, a reader could infer only a wrong MFA code is retriable. We can suggest something like: "Other retriable cases include rejected input shapes and TOTP-setup mismatches; see `isRetriableAuthError` for the full set."

sharonyajain (approved):

soberm (approved):

PR #166, opened 2026-07-07
fix(core): address dev server review follow-up

### LINKED ISSUE #147: Follow-up: address sarayev's review comments on PR #136 (dev server robustness)
Tracking sarayev's two review comments on #136; #136 ships as-is, these are handled here.

## Checklist

- [ ] **`packages/core/src/scripts/dev-server.ts` line 510** — singleton guard same-ppid carve-out is wider than intended

  > The same-ppid carve-out is the right call for hot reload, but I think it's a touch wider than intended. Two separate `npm run dev` jobs backgrounded from one shell (`npm run dev & npm run dev &`) both inherit that shell as their parent, so they'd share a ppid and *both* take the "this is just a tsx-watch relaunch, proceed" branch — which is exactly the two-supervisors-fighting scenario the guard is here to prevent.
  >
  > The JSDoc says "a second `npm run dev` runs under a different watcher, so it never matches here" — that holds for the tsx-watch parent, but not for a shared grandparent shell, so it slightly oversells the guarantee.
  >
  > Not a blocker, since the templates always go through `tsx watch` and nobody really backgrounds two dev servers in one shell on purpose. But if you want to close it cleanly: also require the recorded owner's pid to be dead before proceeding (a *live* owner with a matching ppid is much more likely a real competitor than a hot reload), or stash a watcher-identifying token instead of comparing the raw ppid. At minimum, worth softening the JSDoc so the next reader knows where the boundary actually is.

  Original comment: https://github.com/aws-devtools-labs/aws-blocks/pull/136#discussion_r3519840344

- [ ] **`packages/core/src/scripts/dev-server.ts` line 987** — `unref()` on the EADDRINUSE retry timer may let the process exit before the rebind fires

  > Small thing on the retry scheduling here.
  >
  > `unref` makes sense on the frontend respawn timer, but at this point the `server.listen` that just failed never established a listening handle, and the frontend child hasn't been spawned yet. So if nothing else happens to be keeping the loop alive when this is the only pending timer, the process could quietly exit *before* the rebind fires — skipping a scheduled retry instead of running it.
  >
  > In practice there's usually another handle open from backend init, so this is more a latent edge than something I've seen bite. But since the bounded retry budget and the `process.exit(1)` on exhaustion already guarantee this can't loop forever, `unref` isn't really buying you anything here — I'd just drop it so the retry fires deterministically. If you'd rather keep it, a one-line comment explaining what keeps the loop alive between the failed `listen` and the retry would save the next person the head-scratch.

  Original comment: https://github.com/aws-devtools-labs/aws-blocks/pull/136#discussion_r3519840352

## Context

From #136 ("fix(core): robust dev server — startup port reclaim, singleton guard, :3000 EADDRINUSE handling", branch `fix/dev-server-robustness`).

snvtac:
## Problem

Fixes #147.

PR #136 left two non-blocking dev-server robustness follow-ups:

- the singleton guard's same-ppid carve-out also matched two `npm run dev` jobs backgrounded from the same shell, letting both proceed as if they were a tsx-watch relaunch;
- the EADDRINUSE rebind retry timer was unref'ed, so the process could exit before the scheduled retry fired if no other handle kept the event loop alive.

**Issue #, if available:** #147

## Changes

- Tighten `evaluateSingleton()` so the same-ppid hot-reload carve-out only proceeds when the recorded owner process is dead; a live same-parent owner that still holds the port now exits like any other competing supervisor.
- Update the singleton guard JSDoc/test comments to describe the same-parent boundary accurately.
- Add `scheduleBindRetry()` and wire the EADDRINUSE retry path through a refed timer, removing the retry-timeout `unref()`.
- Add regression coverage for same-parent live-owner contention and for keeping the front-door retry timer refed.

## Validation

- `npm run build -w packages/hosting -w packages/pipeline`
- `npm run build -w packages/core`
- `node --test dist/scripts/dev-server-reclaim.test.js` (38 tests pass)
- `npm run test -w packages/core` (582 tests pass)
- `git diff --check`
- `npx biome lint --skip=correctness/noUndeclaredDependencies packages/core/src/scripts/dev-server.ts packages/core/src/scripts/dev-server-reclaim.test.ts`

## Checklist

- [x] PR description included
- [x] Relevant documentation is changed or added (JSDoc updated; no external docs needed)

---


sharonyajain (approved):

packages/core/src/scripts/dev-server.ts
```diff
@@ -1000,7 +1011,7 @@ export async function startDevServer(options: DevServerOptions) {
   const onEaddrinuse = createBindRetryController(port, {
     reclaim: (p) => reclaimPort(p),
     relisten: () => server.listen(port, onListening),
-    scheduleRetry: (fn, delayMs) => { setTimeout(fn, delayMs).unref?.(); },
+    scheduleRetry: (fn, delayMs) => { scheduleBindRetry(fn, delayMs); },
```
soberm, line 1014: [nit] `scheduleBindRetry` returns the `NodeJS.Timeout` handle (the test relies on it), but this production `scheduleRetry` wrapper discards it. Fine as-is — just noting that if you ever want to cancel a pending retry on shutdown, this is where you'd want to hold onto the handle.

packages/core/src/scripts/dev-server.ts
```diff
@@ -415,6 +415,15 @@ export interface BindRetryDeps {
   warn: (msg: string) => void;
 }
```
soberm, line 417: [nit] The doc says "keep it refed so the retry deterministically fires," but the function doesn't do anything to *make* it refed — refed is just `setTimeout`'s default, and the actual fix is the *absence* of `unref()`. Consider wording like "…scheduled without `unref()`, so the retry deterministically fires" to avoid implying an explicit ref step.

sharonyajain:
Hi @snvtac. Please let us know if you would still like to work on this! Thank you for the contribution

PR #171, opened 2026-07-08
ci(agent-bench): Opus 4.8 builder, throttle-retry, agent-shell isolation, judge/dev-server hardening, and comparison-report redesign

hfurkanbozkurt:
## Summary
Hardens the agent-bench harness end-to-end and establishes **Opus 4.8 as the standard builder**. Bench/CI tooling only — **no `packages/*` changes**, so BB governance gates don't apply. Makes the load-bearing retry, isolation, scoring, and reporting correct and test-pinned.

## What changed
**Builder model & cost**
- Builder + judge = `us.anthropic.claude-opus-4-8` (Opus 4.8) via the `BENCH_MODEL` knob.
- `BUILDER_PRICING = PRICING['claude-opus']` ($5 in / $25 out per 1M); README + changeset describe Opus 4.8 rates.

**Reliability — Bedrock throttling**
- Retry classifier extracted to a pure `steps/lib/bedrock-retry.mjs` (`isRetryableModelError`), table-driven `node --test`. Plus build-once (pack dist-registry once), `max-parallel: 5`, and startup stagger to bound Bedrock fan-out.

**Agent-shell isolation (real kernel boundary)**
- Builder's vended bash runs as a separate UID (`sudo -n runuser -u benchagent`) with workspace-ancestor traversal ACLs, so agent-issued kills (`pkill`/`fuser`/`lsof`) can't reach the harness process tree — they previously could under the shared OS UID. `run-shell.test.mjs` proves both directions + fairness; CI enables the required privileges.

**Harness robustness fixes surfaced during validation**
- **Judge freeze EPERM:** the pre-judge `chmod -R a-w "$WORKSPACE"` couldn't touch benchagent-owned files and aborted the judge under `set -e`, nulling every cell's score — fixed with a `sudo -n chmod … || chmod … || true` freeze so grading always runs.
- **Leftover dev-server reap:** the agent's `tsx watch` supervisor + stale pidfile tripped the framework singleton guard so the verifier's dev-server never bound (working apps scored 0) — `reap_stale_dev_servers()` now kills supervisor→child and clears the pidfile, with a `/proc/<pid>/comm` reuse guard.

**Scoring & reporting**
- `#183` harness-integrity: ungraceful 2-agent teardowns (exit 137/143) reclassified as `harness_error` and EXCLUDED from the mean; genuine `agent_fail` (composite 0) still INCLUDED.
- New **SCORE = composite ÷ cost**; per-cell **cost + tokens** in the summary; judge-color gate on schema ≥ 2; baseline recording fix.
- **Comparison report:** base→PR deltas now render as 🟢/🟡/🔴 status balls (reusing the existing color convention) — both in the headline and as a per-row **`Δ vs base`** column in the Overview/Detailed tables (±5 composite-point band; 🆕 no-base, 🗑️ removed cell). Presentational only.

**Deps**
- Strands bumped to 1.8.0 (agent-bench only).

## Testing
- `node --test` suites (bedrock-retry classification, run-shell isolation both-directions + fairness, scoring, overview incl. the new Δ-vs-base column) run green in the free selfcheck job.
- Full bench validation runs on this branch: all cells grade, zero harness teardowns, isolation confirmed on the real runner.

## Scope & safety
- Bench/CI tooling only; no `packages/*` changes. Isolation is a real kernel boundary (separate UID + ancestor ACLs), not a process-group band-aid.

## Follow-ups
- Same-model builder/judge (Opus 4.8) self-eval bias is tracked in #193 (accepted for now; revisit with stable baselines + a distinct/ensemble judge).


scripts/agent-bench/steps/3-build-and-test.sh
```diff
@@ -1,44 +1,26 @@
...
-# on a reused runner (or a concurrent local invocation) can never collide on the
-# fixed /tmp paths this step used to hard-code. TASK is inherited from the job
-# env (defaulted for local runs); $$ adds per-invocation uniqueness. Every
-# dev/build/playwright scratch file below lives under here. PW_RESULTS_JSON is
-# exported so the generated playwright.config.ts (a quoted heredoc that can't
-# expand shell vars) can read the report path from the environment.
-CELL_TMP="/tmp/bench-${TASK:-default}-$$"
+# Per-run, per-cell scratch prefix so a re-run on a reused runner can't collide on fixed /tmp paths.
+# PW_RESULTS_JSON is exported so the generated playwright.config.ts heredoc can read the report path.
+CELL_TMP="/tmp/bench-${TASK:-default}-$"
```
soberm: [blocker] `$$` got dropped here, so `CELL_TMP` is now always `/tmp/bench-auth-notes-$` (literally ending in a dollar sign) — a bare `$` at the end of a double-quoted string is just a literal `$`, not an expansion. The whole point of `$$` was to embed the shell PID so a re-run on a reused runner doesn't collide with a leftover scratch dir from the previous attempt. Without it, the Playwright results file, dev server pidfile, and pw-log all land at the same path every time the same task runs. A re-run could read stale PW results or get confused by a leftover pidfile and never correctly discover the dev server. Should be `CELL_TMP="/tmp/bench-${TASK:-default}-$$"` — looks like `$$` was accidentally eaten when the comment block above it was condensed.

.github/workflows/agent-bench.yml
```diff
@@ -66,63 +85,58 @@ jobs:
...
       BENCH_JUDGE_MODEL: us.anthropic.claude-opus-4-8
       AWS_REGION: us-east-1
       TEMPLATE: ${{ matrix.template }}
       PR_NUMBER: ${{ github.event.pull_request.number || 'manual' }}
+      # Deterministic startup stagger for the builder's first Bedrock call (see 2-agent-run.ts):
+      # it sleeps (BENCH_CELL_INDEX mod BENCH_STAGGER_WAVE_SIZE) * BENCH_STAGGER_STEP_SEC. Staggering
+      # by within-wave slot keeps the delay bounded; WAVE_SIZE must mirror max-parallel above.
+      BENCH_CELL_INDEX: ${{ strategy.job-index }}
+      BENCH_STAGGER_STEP_SEC: '7'
+      BENCH_STAGGER_WAVE_SIZE: '5' # keep in sync with strategy.max-parallel above
```
soberm, line 102: [minor] `BENCH_STAGGER_WAVE_SIZE` and `max-parallel` have to be kept in sync by hand, and the comment is the only enforcement. If someone bumps `max-parallel` (say, to 8 as the matrix grows) and forgets this env var, the stagger logic silently uses wave-size=5 — cells in the second wave sleep up to 35s before their first Bedrock call instead of the intended 0s, which defeats the point. GHA doesn't expose `max-parallel` as a context value you can reference from `env:`, so the duplication is unavoidable, but worth calling out explicitly: any change to `max-parallel` needs a matching change here.

scripts/agent-bench/steps/lib/bedrock-retry.mjs
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * @param {unknown} err
+ * @returns {boolean}
+ */
+export function isRetryableModelError(err) {
+	if (err instanceof StructuredOutputError) return false;
+	if (err instanceof ContextWindowOverflowError || err instanceof MaxTokensError) return false;
+	if (err instanceof ModelThrottledError) return true;
+	// A bare ModelError almost always wraps a transient mid-stream AWS exception.
+	if (err instanceof ModelError) return true;
```
soberm, line 112: [minor] Treating every bare `ModelError` as retryable is fine today since the known deterministic subclasses (`StructuredOutputError`, `ContextWindowOverflowError`, `MaxTokensError`) are correctly short-circuited above. But if Strands ever adds a new non-transient `ModelError` subclass (e.g. a content-filter error), it would silently fall through here and trigger up to 4 retries + ~2.5 min of backoff before failing. Worst case is slow failures rather than wrong scores, but it can make the judge step look stuck. Something to keep in mind when upgrading `@strands-agents/sdk`.

.github/workflows/agent-bench.yml
```diff
@@ -66,63 +85,58 @@ jobs:
...
-      # both resolve in us-east-1 today (verified by live, scored bench runs):
-      #   Sonnet 4.6: docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-4-6.html
-      #   Opus 4.8:   docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-4-8.html
-      BENCH_MODEL: us.anthropic.claude-sonnet-4-6
-      # Different model from the builder, on purpose — limits same-model
-      # self-evaluation bias.
+      # Bedrock geo inference-profile IDs (both resolve in us-east-1). Overridable via BENCH_MODEL;
+      # builder and judge share Opus 4.8 (same-model self-eval bias accepted for now).
+      BENCH_MODEL: us.anthropic.claude-opus-4-8
       BENCH_JUDGE_MODEL: us.anthropic.claude-opus-4-8
```
soberm, line 93: [question] Builder and judge are now both Opus 4.8. The PR description calls out "same-model self-eval bias accepted for now" — just want to make sure that's tracked somewhere and not a TODO that gets lost. The prior split (Sonnet builder, Opus judge) was specifically to limit this. Is there an issue open for revisiting once the Opus runs have a few baselines?

scripts/agent-bench/steps/2-agent-run.ts
```diff
@@ -222,12 +270,8 @@ process.stderr.write(
 	`[bench] done: stop=${result.stopReason} tokens=${tokensIn}/${tokensOut} cycles=${result.metrics?.cycleCount ?? 0} ${duration_sec}s\n`,
```
ahmedhamouda78, line 288: [major] The success path reads `result.metrics.accumulatedUsage`, which reflects only the winning attempt (a fresh agent is created per retry). Error/checkpoint paths use the cumulative `partialTokens*` counters; the success path doesn't. A throttled-then-succeeded cell undercounts `cellCost` and overstates `scorePerDollar` — the PR's headline metric. Reconcile the success path with the cumulative counters, or document the intent.

scripts/agent-bench/steps/lib/run-shell.ts
```diff
@@ -20,25 +10,123 @@ import {
...
+	if (isolate && isolationAvailable()) {
+		const forwarded: Record<string, string> = { ...env, HOME: agentHome ?? `/home/${BENCH_AGENT_USER}` };
+		// Only forward valid shell identifier names (drops exported-function keys like
+		// `BASH_FUNC_x%%`); values pass as argv so need no quoting.
+		const envPairs = Object.entries(forwarded)
+			.filter(([k]) => /^[A-Za-z_][A-Za-z0-9_]*$/.test(k))
+			.map(([k, v]) => `${k}=${v}`);
+		return {
+			file: 'sudo',
+			args: ['-n', 'runuser', '-u', BENCH_AGENT_USER, '--', 'env', ...envPairs, 'bash', '-c', inner],
```
ahmedhamouda78, line 121: [minor] `buildAgentSpawn()` passes `env KEY=VAL … bash -c`, putting the OIDC session creds in argv (world-readable via `/proc/<pid>/cmdline`, `ps`). Incremental risk is low on an ephemeral single-tenant runner, but secrets-on-argv is an anti-pattern — prefer a written env-file. Awareness only, not a blocker.

scripts/agent-bench/steps/lib/analysis.mjs
```diff
@@ -1,79 +1,64 @@
...
 export const REGRESSION_DELTA = -5;
 
 // Benign fallback stored when the per-cell analysis can't be produced.
 export const FALLBACK_ANALYSIS = 'analysis unavailable';
 
-// Throttle/transient retry, mirrored from steps/4-judge.ts: initial try + up to
-// 4 backed-off retries, ONLY on a throttle/transient class (never a hard
-// AccessDenied/validation error, which fails fast).
+// App-level throttle/transient retry (initial + up to 4 backoffs), transient class only. Duplicated
+// from bedrock-retry.ts intentionally (this .mjs can't import the .ts); keep in sync.
```
ahmedhamouda78: [nit] `TRANSIENT_RE` claims to mirror bedrock-retry.ts but differs (bare `timeout`/`500` vs. `timeouterror`/`modelnotready`, no `500`). The divergence is fine; reword the comment so nobody "syncs" them and breaks one side.

scripts/agent-bench/steps/lib/run-shell.ts
```diff
@@ -20,25 +10,123 @@ import {
...
+	}
+	return `/home/${BENCH_AGENT_USER}`;
+}
+
+// Probe (once, memoized) whether the agent shell can run as benchagent via passwordless
+// sudo+runuser, using the EXACT privilege transition runShell uses; false → caller falls back.
+export function isolationAvailable(): boolean {
+	if (isolationProbe !== undefined) return isolationProbe;
+	try {
+		const r = spawnSync('sudo', ['-n', 'runuser', '-u', BENCH_AGENT_USER, '--', 'true'], {
```
ahmedhamouda78, line 52: [nit] `isolationAvailable()` only probes `sudo -n runuser`, but teardown also relies on `sudo -n kill`/`fuser`/`rm`/`chmod`. Fine under blanket NOPASSWD, but if sudoers were scoped to just `runuser`, group-kill/port-free would silently no-op. Add a note or broaden the probe.

scripts/agent-bench/steps/lib/scoring.mjs
```diff
@@ -25,73 +13,94 @@ export const COMMON_DIMENSIONS = [
...
-	// The agent failing to produce a gradeable app is a real fail, not a flake.
-	// Accepted limitation: an OOM/SIGKILL of the agent at 2-agent is
-	// indistinguishable from a genuine agent timeout — both surface only as GHA
-	// outcome `failure` here — so both classify as agent_fail (composite 0). This
-	// is accepted because such infra kills at step 2 are expected to be rare.
+	// Agent failure is a real fail — except an ungraceful harness teardown (no terminal envelope),
+	// which is reclassified harness_error so a self-inflicted infra kill can't drag the mean.
 	if (failedAt === AGENT_FAIL_AT) {
+		if (isUngracefulStepTwoDeath(result)) {
+			return { klass: 'harness_error', reason: AGENT_HARNESS_TEARDOWN_REASON };
```
Simone319, line 101: **[I-2] Only exclude ungraceful step-2 deaths from the mean when agent-shell isolation was actually active.**

Reclassifying an ungraceful step-2 death as `harness_error` (excluded via `isScoredCell`) is sound *only when isolation is on* — then the only such deaths are genuine infra (runner OOM / GHA cancel). But `isolationAvailable()` in `run-shell.ts` silently falls back to bare `bash -c` when `sudo -n runuser` isn't available, and that state is written only to stderr (`2-agent-run.ts`), never to `result.json`. With isolation **off**, the exact #184 pkill-storm — a failing agent tearing down its own harness — becomes an ungraceful death that is now *excluded* rather than counted as `agent_fail` (composite 0), quietly inflating the headline mean.

On `ubuntu-latest` (`NOPASSWD:ALL`) isolation is active so this holds today, but it's unenforced. Suggest: stamp `isolation_active` into the envelope/`result.json`, and here treat an ungraceful step-2 death as `agent_fail` when isolation was NOT active — only exclude it when isolation was active.

scripts/agent-bench/steps/2-agent-run.ts
```diff
@@ -1,119 +1,142 @@
...
-	systemPrompt: builderSystem(WORKSPACE),
-	sandbox: new WorkspaceSandbox(WORKSPACE, BASH_MIN_TIMEOUT_SEC),
-	tools: [makeBash(), fileEditor],
-});
+// Resolve agent-shell UID isolation (issue #184) ONCE (the ACL grant must run a single time, not
+// per invoke-retry). Not a hard failure when unavailable — this line records whether containment was
+// active for the cell. On the GitHub-hosted runner benchagent has passwordless sudo, so expect ACTIVE.
+const ISOLATE = isolationAvailable() && prepareWorkspaceIsolation(WORKSPACE);
+process.stderr.write(
+	`[bench] agent shell isolation: ${
```
Simone319, line 56: [I-2, related] Isolation active/disabled is only logged to stderr here. Please also persist it (e.g. `isolation_active`) into `result.json`, so the scoring exclusion (`scoring.mjs` `isUngracefulStepTwoDeath` / `isScoredCell`) can be gated on it and reviewers can tell whether a run's exclusions are trustworthy.

scripts/agent-bench/steps/lib/bedrock-retry.mjs
Simone319, line 112: **[I-3] A fatal 4xx wrapped in `ModelError` gets retried.** This blanket `ModelError → true` fires before any cause-chain status check. Strands often surfaces mid-stream AWS exceptions as a bare `ModelError` with the real class on `.cause`, so a `ValidationException` (400) / `AccessDeniedException` (403) hits this branch and burns the full `[5s,15s,40s,90s]` ≈150s ladder on an unrecoverable error (logged as "throttle/transient"). A misconfigured role then wastes ~150s per cell. Suggest inspecting the cause chain for a terminal 4xx (400/401/403/404, or name matching `/validation|accessdenied|unrecognizedclient/i`) and returning `false` before this line.

scripts/agent-bench/steps/lib/run-shell.ts
```diff
@@ -106,18 +182,20 @@ export function runShell(
...
-		// invoke() returns. Guarded: pid is undefined if spawn failed, and the
-		// group may already be gone (ESRCH).
+		// SIGKILL the whole process group (negative pid) to reap any backgrounded child whose
+		// inherited pipe write-ends keep 'close' from firing and hold libuv's loop open. Guarded
+		// (pid may be undefined / group already ESRCH). When isolated the group is benchagent-owned,
+		// so escalate via `sudo kill`; a `setsid` server escapes the group and keeps running.
 		const killGroup = (): void => {
 			if (proc.pid === undefined) return;
+			if (isolate) {
+				spawnSync('sudo', ['-n', 'kill', '-9', `-${proc.pid}`], { stdio: 'ignore', timeout: 5000 });
```
Simone319, line 204: [minor] `isolationAvailable()` only probes `sudo -n runuser`, but this reap path relies on `sudo -n kill`. On a runner whose sudoers permits `runuser` but not `kill`, backgrounded benchagent process groups won't be reaped (bounded by the drain grace, so no hang — but they accumulate). Consider probing `sudo -n kill` too, or documenting the assumption. Non-issue on `ubuntu-latest`.

scripts/agent-bench/steps/2-agent-run.ts
```diff
@@ -1,119 +1,142 @@
...
 const METRICS_PATH = process.env.METRICS;
-const MODEL_ID = process.env.BENCH_MODEL ?? 'us.anthropic.claude-sonnet-4-6';
-// Backstop against runaway tool-call loops (v1's long prompts caused
-// over-iteration). The Strands TS SDK enforces this per-invocation; hitting it
-// yields stopReason 'limitTurns'. This is only a runaway-loop backstop: the real
-// bound is the 35-min wall-clock timeout in the workflow, which prior runs used
-// only ~20% of, so 120 leaves ample headroom for a full build without risking a
-// runaway. One turn = one model call plus any tool calls it makes.
+const MODEL_ID = process.env.BENCH_MODEL ?? 'us.anthropic.claude-opus-4-8';
+// Opus rejects the `temperature` parameter (like the judge), so only pin temperature=0 for models
```
Simone319, line 39: [minor] Stale comment: the default builder is now Opus (which rejects `temperature`), so the "pin temperature=0 for reproducibility" rationale no longer applies to the default path. Code is correct — just update the comment.

scripts/agent-bench/steps/3-build-and-test.sh
```diff
@@ -108,30 +74,54 @@ else
...
+  local pf ids ppid pid sig target
+  for pf in "$sandbox"/dev-server.*.pid; do
+    [ -f "$pf" ] || continue
+    ids="$(node -e 'try{const r=JSON.parse(require("fs").readFileSync(process.argv[1],"utf8"));process.stdout.write(`${Number.isInteger(r.ppid)?r.ppid:""} ${Number.isInteger(r.pid)?r.pid:""}`)}catch{}' "$pf" 2>/dev/null || true)"
+    read -r ppid pid <<< "$ids" || true
+    for sig in TERM KILL; do
+      for target in "$ppid" "$pid"; do
+        [ -n "${target:-}" ] || continue
+        # pid-reuse guard: only signal a live node/tsx/npm process (/proc/<pid>/comm).
+        case "$(cat "/proc/${target}/comm" 2>/dev/null || true)" in
```
Simone319, line 105: [minor] This `/proc/<pid>/comm` check narrows the PID-reuse window but is comm-*class* (node|tsx|npm), not identity — a recycled PID belonging to an unrelated node/tsx/npm process would still be signalled. Fine on ephemeral single-cell runners; worth a one-line note that it's a heuristic, not a definitive match.

Simone319 (commented):
A few things from review — one score-integrity item (I-2), one retry-classifier item (I-3), and some minor cleanups. Details inline. Nothing blocking; I-2 is the one worth prioritizing.

scripts/agent-bench/steps/lib/bedrock-retry.mjs
```diff
@@ -0,0 +1,127 @@
...
+
+// Backoff (ms) before the given 1-based retry: the exponential base (last value once exhausted) plus
+// up to +25% jitter so concurrent cells don't re-hit Bedrock in lockstep.
+/**
+ * @param {number} attempt 1-based attempt number that is about to be retried
+ * @returns {number}
+ */
+export function nextBackoffMs(attempt) {
+	const base = INVOKE_BACKOFF_MS[attempt - 1] ?? INVOKE_BACKOFF_MS[INVOKE_BACKOFF_MS.length - 1] ?? 5_000;
+	return base + Math.floor(Math.random() * base * 0.25);
```
Simone319: [minor] Jitter is only +25% of base, so 5 concurrent cells that throttle together retry in near-lockstep. Full jitter (`Math.random() * base`) would de-correlate better. Low priority — mitigated by the startup stagger + SDK adaptive layer.

.github/workflows/agent-bench.yml
```diff
@@ -139,11 +153,20 @@ jobs:
...
-          # is lower than this, AWS clamps to whatever the role allows.
+          # Extend the OIDC session past the ~1h default so it survives slow cells (AWS clamps
+          # to the role's max if lower).
           role-duration-seconds: 7200
 
+      # BUILD-ONCE: pull the dist-registry/ build-blocks packed so "1. Init bench-app" can skip the
+      # ~150s monorepo build. Best-effort: a download blip never reds the cell (init rebuilds it).
+      - name: 1a. Download prebuilt dist-registry
+        continue-on-error: true
+        if: steps.oidc.outcome == 'success'
```
Simone319: [minor] This "Download prebuilt dist-registry" step is gated on `steps.oidc.outcome == 'success'`, but it uses no AWS — an OIDC flake needlessly forfeits the build-once optimization and forces a local rebuild. Consider decoupling this step from OIDC.

scripts/agent-bench/steps/3-build-and-test.sh
hfurkanbozkurt: Restored — it's back to `CELL_TMP="/tmp/bench-${TASK:-default}-$$"` so the PID keeps re-runs on a reused runner from colliding. Thanks for catching the dropped `$$`.

.github/workflows/agent-bench.yml
hfurkanbozkurt: You're right, and unfortunately GHA won't let a matrix `max-parallel` be referenced from `env:`, so the duplication is unavoidable — I've left the explicit "keep in sync with max-parallel" comment as the guardrail rather than add indirection. Declining to change beyond that.

scripts/agent-bench/steps/lib/bedrock-retry.mjs
hfurkanbozkurt: Partly tightened in b053885: `hasTerminalClientError()` now short-circuits terminal 4xx (ValidationException/AccessDenied/etc., incl. a content-filter surfaced as 4xx) to non-retryable before the blanket branch. A brand-new *non-4xx* deterministic subclass would still fall through — keeping that as a known trade-off to revisit on the next SDK bump.

.github/workflows/agent-bench.yml
hfurkanbozkurt: Agreed it shouldn't be a lost TODO — filed it as #193. Keeping same-model for now (composite is test-weighted and the bias is uniform across cells); the issue tracks revisiting once we have stable Opus baselines and a distinct/ensemble judge.

scripts/agent-bench/steps/2-agent-run.ts
hfurkanbozkurt: Good catch — fixed in b053885. The success path now takes `max(partialTokensIn/Out, winnerUsage)`, so retried-attempt tokens are folded into the cost the same way the error/checkpoint paths already did.

scripts/agent-bench/steps/lib/run-shell.ts
hfurkanbozkurt: Documented in b053885 rather than reworked: added a comment noting the argv exposure and why an env-file is deferred (it'd have to be benchagent-readable across the uid boundary, complicating the isolation path) on a single-tenant ephemeral runner. Happy to revisit if that risk profile changes.

scripts/agent-bench/steps/lib/analysis.mjs
hfurkanbozkurt: Done in b053885 — reworded to say it's a deliberately broad log-text heuristic that intentionally diverges from `bedrock-retry`'s SDK-error classifier, so neither gets "synced" into the other.

scripts/agent-bench/steps/lib/run-shell.ts
hfurkanbozkurt: Documented in b053885: added a note that the single `runuser` probe is sufficient under the runner's `NOPASSWD:ALL`, and flagged that the probe must be broadened if sudoers is ever scoped per-command.

scripts/agent-bench/steps/lib/scoring.mjs
hfurkanbozkurt: Fixed in b053885. We now stamp `isolation_active` into `result.json` and only reclassify an ungraceful step-2 death as `harness_error` when isolation was active; with it off the same death stays `agent_fail` (composite 0, included). Added tests for both directions.

scripts/agent-bench/steps/2-agent-run.ts
hfurkanbozkurt: Done in b053885 — `isolation_active` is now stamped into every envelope path (checkpoint, SIGTERM/SIGINT partial, invoke-exhausted sentinel, and success), plus a seed checkpoint before the first model call so it's on disk even on a very early death.

scripts/agent-bench/steps/lib/bedrock-retry.mjs
hfurkanbozkurt: Fixed in b053885 — added `hasTerminalClientError()` that walks the cause chain for a non-429 4xx status or a `/validation|accessdenied|unrecognizedclient|.../i` name and returns `false` before the blanket `ModelError → retry`.

scripts/agent-bench/steps/lib/run-shell.ts
hfurkanbozkurt: Documented in b053885 (same note as the probe-breadth comment above): the reap `sudo -n kill`/`fuser`/`rm` all rely on the runner's `NOPASSWD:ALL`, called out explicitly so a future per-command sudoers scoping gets the probe widened.

scripts/agent-bench/steps/2-agent-run.ts
hfurkanbozkurt: Fixed in b053885 — the comment now notes Opus rejects the `temperature` knob, so the default path's reproducibility rests on the pinned model + hard MAX_TURNS.

scripts/agent-bench/steps/3-build-and-test.sh
hfurkanbozkurt: Done in b053885 — added a one-line note that it's a comm-class heuristic (not process identity), acceptable because the ephemeral runner's only node/tsx/npm workload is our own dev server.

scripts/agent-bench/steps/lib/bedrock-retry.mjs
hfurkanbozkurt: Done in b053885 — widened to equal jitter `[0.5·base, 1.5·base)` so cells that throttle together spread out instead of retrying in near-lockstep.

.github/workflows/agent-bench.yml
hfurkanbozkurt: Fixed in b053885 — that step now gates on `initresult` instead of `oidc`, so an OIDC flake no longer forfeits the build-once download (it uses no AWS).

soberm (approved):

ahmedhamouda78 (approved):

Issue #173, opened 2026-07-09
Stale placeholder config.json served indefinitely — deploy invalidation path never matches the rewritten cache key (breaks all client API calls)


fossamagna:
### Which package(s) are affected?

`@aws-blocks/core`

### Describe the bug

After a normal deploy, a fraction of browsers permanently receive the **build-time placeholder** `/.blocks-sandbox/config.json` (`{"_placeholder":true}`) from CloudFront instead of the real `{"apiUrl":"/aws-blocks/api"}`. The `@aws-blocks/core/client` `getApiUrl()` treats a body without `apiUrl` as invalid and throws, so **every** client API call (including `authApi.getAuthState()`) rejects. In our app the login page is rendered permanently unusable ("sign-in unavailable") for affected users. There is no server-side error — the origin and S3 are healthy; only CloudFront serves stale content. It is intermittent per edge cache-shard, so it looks flaky.

**Root cause.** The placeholder `config.json` is uploaded as an immutable static asset (`public, s-maxage=31536000, max-age=0, must-revalidate`). The real config is later written via `BucketDeployment` (`public, max-age=60`) with a CloudFront invalidation of `['/.blocks-sandbox/*']`. But the hosting **viewer-request CloudFront function rewrites the URI _before_ the cache lookup** to `/builds/<buildId>/.blocks-sandbox/config.json`, so the real cache key lives under `/builds/…`. The invalidation path `/.blocks-sandbox/*` never matches it → the invalidation is a no-op for `config.json`, and any edge that cached the placeholder during the deploy window keeps serving it under its 1-year `s-maxage` until eviction.

### Expected behavior

After a deploy, all clients receive the real `config.json` (`{"apiUrl":…}`) within the short cache window. A stale build-time placeholder must never be served long-term, and the post-deploy invalidation must actually clear the object at its real (post-rewrite) cache key.

### Reproduction steps

1. Deploy a Blocks app that uses hosting + an API (so `config.json` carries `apiUrl`), on a distribution that uses the skew-protection viewer-request function (build-id URI rewrite) and the `Managed-CachingOptimized` cache policy on `/.blocks-sandbox/*` (Accept-Encoding is part of the cache key).
2. During/just after the deploy window, exercise the `/.blocks-sandbox/config.json` URL from multiple edges. Because the cache key includes a normalized `Accept-Encoding`, the object is cached **per encoding variant** (br / gzip / uncompressed) independently.
[...]

PR #174, opened 2026-07-09
fix(hosting): never serve stale placeholder config.json after deploy

fossamagna:
## Problem

**Issue #, if available:** #173

After a normal deploy, a fraction of browsers permanently receive the build-time placeholder `/.blocks-sandbox/config.json` (`{"_placeholder":true}`) from CloudFront instead of the real `{"apiUrl":…}`. The `@aws-blocks/core/client` `getApiUrl()` treats a body without `apiUrl` as invalid and throws, so every client API call (including `authApi.getAuthState()`) rejects — the login page is rendered permanently unusable for affected users.

Two independent defects combine:

1. **Placeholder gets a 1-year edge cache.** The build-time placeholder `config.json` is written into the static asset dir but is not registered as a no-cache path, so it falls into the mutable asset tier and is uploaded with `public, s-maxage=31536000, max-age=0, must-revalidate`. Any edge that caches the placeholder during the deploy window keeps serving it for up to a year.
2. **The post-deploy invalidation never matches the real cache key.** The skew-protection viewer-request CloudFront function rewrites the URI to `/builds/<buildId>/.blocks-sandbox/config.json` *before* the cache lookup, so the real cache key lives under `/builds/…`. The `BucketDeployment` invalidated only `/.blocks-sandbox/*`, which never matches the rewritten key — so the invalidation is a no-op for `config.json`.

## Changes

- **Primary — self-healing placeholder.** Register `.blocks-sandbox/config.json` in `manifest.staticAssets.noCachePaths` so the placeholder is uploaded with `no-cache, no-store, must-revalidate` instead of the 1-year mutable cache-control. The edge no longer caches the placeholder long-term. The real config is still deployed separately with its own short `max-age=60`, so runtime performance is unchanged. (This reuses the existing `noCachePaths` mechanism, whose own doc names `.blocks-sandbox/config.json` as the intended use case.)
- **Defense-in-depth — invalidate the real key.** Set the config `BucketDeployment`'s `distributionPaths` to also include the post-rewrite key `/builds/<buildId>/.blocks-sandbox/*`, so a post-deploy invalidation actually clears any edge entry at its real cache key.

## Validation

- Added two unit tests in `packages/core/src/hosting.test.ts` (`config.json stale-placeholder guard (#173)`):
  - asserts the placeholder `config.json` is uploaded with `no-cache, no-store, must-revalidate` and that no deployment applies `s-maxage=31536000` to it;
  - asserts `distributionPaths` invalidates the post-rewrite cache key `/builds/<id>/.blocks-sandbox/*`.
- Both tests were confirmed to fail before the fix and pass after (TDD).
- Full `@aws-blocks/core` suite green (582 tests); no new lint findings on the changed lines.

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (and PR referenced) — N/A (internal behavior fix, no public API/doc change)

---


packages/core/src/hosting.ts
```diff
@@ -499,6 +500,24 @@ export class Hosting extends Construct {
...
+    //    build-time *placeholder* (`{_placeholder:true}`) during the deploy
+    //    window would keep serving it for up to a year — `getApiUrl()` then
+    //    throws and every client API call fails. Registering it as a
+    //    no-cache path routes the placeholder into the `no-cache, no-store,
+    //    must-revalidate` deployment tier so the edge never caches it
+    //    long-term; the real config is deployed in step 8 with its own
+    //    short `max-age=60`.
+    const configNoCachePath = `${BLOCKS_SANDBOX_DIR}/config.json`;
+    const existingNoCachePaths = manifest.staticAssets.noCachePaths ?? [];
+    manifest.staticAssets.noCachePaths = existingNoCachePaths.includes(configNoCachePath)
```
ahmedhamouda78, line 512: The noCachePaths registration runs unconditionally (no api guard), which is correct — but both new tests always pass MOCK_API. A static-only test (no api prop) would guard against a future regression where the registration is accidentally moved inside the `if (props.api)` block, since the placeholder is always written.

packages/core/src/hosting.test.ts
```diff
@@ -1507,4 +1507,103 @@ describe('Hosting', () => {
...
+      // Invalidating only `/.blocks-sandbox/*` never matches it.
+      createSpaBuildOutput(tmpDir);
+      const app = new App();
+      const stack = new Stack(app, 'InvalidationPathStack');
+      new Hosting(stack, 'Hosting', { root: tmpDir, api: MOCK_API });
+
+      const deployments = bucketDeployments(stack);
+      // The BlocksConfigDeployment is the one targeting the .blocks-sandbox
+      // key prefix with an invalidation configured.
+      const configDeploy = deployments.find(
```
ahmedhamouda78: The configDeploy lookup uses a structural heuristic (DestinationBucketKeyPrefix.endsWith('.blocks-sandbox') && Array.isArray(DistributionPaths)). If CDK renames internal properties the heuristic silently fails to find the deployment and the test throws on assert.ok(undefined, ...) rather than a clear failure. Prefer Template.fromStack + Match.objectLike for CDK-upgrade-resilient assertions. (The \\.blocks-sandbox regex escaping is correct.)

packages/core/src/hosting.ts
```diff
@@ -499,6 +500,24 @@ export class Hosting extends Construct {
       JSON.stringify({ _placeholder: true }),
     );
 
+    // ── 5a. Mark config.json as a no-cache path ─────────────────────
```
ahmedhamouda78, line 503: nit: the 8-line comment narrates the full #173 bug history that's already in the commit message. Keep the non-obvious invariant inline (config.json is mutable and must not inherit the content-hashed-asset s-maxage) and move the forensics to the commit body.

packages/core/src/hosting.test.ts
```diff
@@ -1507,4 +1507,103 @@ describe('Hosting', () => {
       );
     });
   });
+
+  // ── #173: stale placeholder config.json never served long-term ──
```
ahmedhamouda78: nit: the describe(...) block name already conveys this; section-divider comments inside test bodies add noise.

ahmedhamouda78 (dismissed):
Approving. Correctly fixes both root-cause defects behind #173: (1) the placeholder config.json fell into the mutable-asset tier (public, s-maxage=31536000) and could be edge-cached for up to a year — now registered in noCachePaths and uploaded with no-cache, no-store, must-revalidate; and (2) the post-deploy invalidation targeted /.blocks-sandbox/* but the skew-protection CloudFront function rewrites the URI to /builds/<buildId>/.blocks-sandbox/config.json before cache lookup, so the invalidation never matched — now adds /builds/${buildId}/.blocks-sandbox/* while correctly retaining /.blocks-sandbox/* for non-rewrite cases. Scoping invalidation to the current buildId only is the right call (cookie-pinned users keep their config). Clean TDD, no security concerns, zero drift from main.

Non-blocking suggestions below.

packages/core/src/hosting.ts
ahmedhamouda78, line 512: The noCachePaths registration runs unconditionally (no api guard), which is correct — but both new tests always pass MOCK_API. A static-only test (no api prop) would guard against a future regression where the registration is accidentally moved inside the `if (props.api)` block, since the placeholder is always written.

packages/core/src/hosting.test.ts
ahmedhamouda78: The configDeploy lookup uses a structural heuristic (DestinationBucketKeyPrefix.endsWith('.blocks-sandbox') && Array.isArray(DistributionPaths)). If CDK renames internal properties the heuristic silently fails to find the deployment and the test throws on assert.ok(undefined, ...) rather than a clear failure. Prefer Template.fromStack + Match.objectLike for CDK-upgrade-resilient assertions. (The \.blocks-sandbox regex escaping is correct.)

packages/core/src/hosting.ts
ahmedhamouda78, line 503: nit: the 8-line comment narrates the full #173 bug history that's already in the commit message. Keep the non-obvious invariant inline (config.json is mutable and must not inherit the content-hashed-asset s-maxage) and move the forensics to the commit body.

packages/core/src/hosting.test.ts
ahmedhamouda78: nit: the describe(...) block name already conveys this; section-divider comments inside test bodies add noise.

ahmedhamouda78 (dismissed):
Approving. Correctly fixes both root-cause defects behind #173: (1) the placeholder config.json fell into the mutable-asset tier (public, s-maxage=31536000) and could be edge-cached for up to a year — now registered in noCachePaths and uploaded with no-cache, no-store, must-revalidate; and (2) the post-deploy invalidation targeted /.blocks-sandbox/* but the skew-protection CloudFront function rewrites the URI to /builds/<buildId>/.blocks-sandbox/config.json before cache lookup, so the invalidation never matched — now adds /builds/${buildId}/.blocks-sandbox/* while correctly retaining /.blocks-sandbox/* for non-rewrite cases. Scoping invalidation to the current buildId only is the right call (cookie-pinned users keep their config). Clean TDD, no security concerns, zero drift from main.

Non-blocking suggestions below.

packages/core/src/hosting.ts
ahmedhamouda78, line 512: The noCachePaths registration runs unconditionally (no api guard), which is correct — but both new tests always pass MOCK_API. A static-only test (no api prop) would guard against a future regression where the registration is accidentally moved inside the `if (props.api)` block, since the placeholder is always written.

packages/core/src/hosting.test.ts
ahmedhamouda78: The configDeploy lookup uses a structural heuristic (DestinationBucketKeyPrefix.endsWith('.blocks-sandbox') && Array.isArray(DistributionPaths)). If CDK renames internal properties the heuristic silently fails to find the deployment and the test throws on assert.ok(undefined, ...) rather than a clear failure. Prefer Template.fromStack + Match.objectLike for CDK-upgrade-resilient assertions. (The \.blocks-sandbox regex escaping is correct.)

packages/core/src/hosting.ts
ahmedhamouda78, line 503: nit: the 8-line comment narrates the full #173 bug history that's already in the commit message. Keep the non-obvious invariant inline (config.json is mutable and must not inherit the content-hashed-asset s-maxage) and move the forensics to the commit body.

packages/core/src/hosting.test.ts
ahmedhamouda78: nit: the describe(...) block name already conveys this; section-divider comments inside test bodies add noise.

ahmedhamouda78 (approved):
Approving. Correctly fixes both root-cause defects behind #173: (1) the placeholder config.json fell into the mutable-asset tier (public, s-maxage=31536000) and could be edge-cached for up to a year — now registered in noCachePaths and uploaded with no-cache, no-store, must-revalidate; and (2) the post-deploy invalidation targeted /.blocks-sandbox/* but the skew-protection CloudFront function rewrites the URI to /builds/<buildId>/.blocks-sandbox/config.json before cache lookup, so the invalidation never matched — now adds /builds/${buildId}/.blocks-sandbox/* while correctly retaining /.blocks-sandbox/* for non-rewrite cases. Scoping invalidation to the current buildId only is the right call (cookie-pinned users keep their config). Clean TDD, no security concerns, zero drift from main.

Non-blocking suggestions below.

packages/core/src/hosting.ts
ahmedhamouda78, line 512: The noCachePaths registration runs unconditionally (no api guard), which is correct — but both new tests always pass MOCK_API. A static-only test (no api prop) would guard against a future regression where the registration is accidentally moved inside the `if (props.api)` block, since the placeholder is always written.

packages/core/src/hosting.test.ts
ahmedhamouda78: The configDeploy lookup uses a structural heuristic (DestinationBucketKeyPrefix.endsWith('.blocks-sandbox') && Array.isArray(DistributionPaths)). If CDK renames internal properties the heuristic silently fails to find the deployment and the test throws on assert.ok(undefined, ...) rather than a clear failure. Prefer Template.fromStack + Match.objectLike for CDK-upgrade-resilient assertions. (The \.blocks-sandbox regex escaping is correct.)

packages/core/src/hosting.ts
ahmedhamouda78, line 503: nit: the 8-line comment narrates the full #173 bug history that's already in the commit message. Keep the non-obvious invariant inline (config.json is mutable and must not inherit the content-hashed-asset s-maxage) and move the forensics to the commit body.

packages/core/src/hosting.test.ts
ahmedhamouda78: nit: the describe(...) block name already conveys this; section-divider comments inside test bodies add noise.

ahmedhamouda78 (dismissed):
Approving. Correctly fixes both root-cause defects behind #173: (1) the placeholder config.json fell into the mutable-asset tier (public, s-maxage=31536000) and could be edge-cached for up to a year — now registered in noCachePaths and uploaded with no-cache, no-store, must-revalidate; and (2) the post-deploy invalidation targeted /.blocks-sandbox/* but the skew-protection CloudFront function rewrites the URI to /builds/<buildId>/.blocks-sandbox/config.json before cache lookup, so the invalidation never matched — now adds /builds/${buildId}/.blocks-sandbox/* while correctly retaining /.blocks-sandbox/* for non-rewrite cases. Scoping invalidation to the current buildId only is the right call (cookie-pinned users keep their config). Clean TDD, no security concerns, zero drift from main.

Non-blocking suggestions below.

packages/core/src/hosting.ts
fossamagna: Added a static-only test (no `api` prop) — `registers the placeholder as a no-cache path even for a static-only site (no api)` — asserting the placeholder is uploaded with `no-cache, no-store, must-revalidate`. Since the placeholder is always written, this guards against a regression that moves the `noCachePaths` registration inside an `if (props.api)` block. Fixed in a695afb.

packages/core/src/hosting.test.ts
fossamagna: Agreed. Rewrote the test to assert via `Template.fromStack(stack).hasResourceProperties('Custom::CDKBucketDeployment', Match.objectLike({ DistributionPaths: Match.arrayWith([Match.stringLikeRegexp('^/builds/.+/\\.blocks-sandbox/\\*$')]) }))`, so a future CDK property rename fails loudly at the assertion instead of silently skipping the structural lookup. Fixed in a695afb.

packages/core/src/hosting.ts
fossamagna: Done — trimmed the step-5a comment to the non-obvious invariant (config.json is mutable and must not inherit the content-hashed-asset `s-maxage`). The bug forensics stay in the fix commit body. Fixed in a695afb.

packages/core/src/hosting.test.ts
fossamagna: Done — removed the redundant section-divider comment; the `describe(...)` block name already conveys it. Fixed in a695afb.

fossamagna:
@ahmedhamouda78 
Thanks for the review! All four suggestions are addressed in a695afb. (Posting here as a top-level summary because the inline threads landed on a dismissed review and are collapsed/outdated in the UI.)

- **Static-only test (hosting.ts noCachePaths, no `api` guard):** Added `registers the placeholder as a no-cache path even for a static-only site (no api)` — builds a `Hosting` with no `api` prop and asserts the placeholder is uploaded with `no-cache, no-store, must-revalidate`. Since the placeholder is always written, this guards against a regression that moves the registration inside an `if (props.api)` block.
- **CDK-resilient assertion (test:1587):** Rewrote the invalidation-path test to assert via `Template.fromStack(stack).hasResourceProperties('Custom::CDKBucketDeployment', Match.objectLike({ DistributionPaths: Match.arrayWith([Match.stringLikeRegexp('^/builds/.+/\\.blocks-sandbox/\\*$')]) }))`, so a future CDK property rename fails loudly at the assertion instead of silently skipping a structural-heuristic lookup.
- **Trim step-5a comment (hosting.ts:503):** Reduced to the non-obvious invariant (config.json is mutable and must not inherit the content-hashed-asset `s-maxage`); the bug forensics stay in the fix commit body.
- **Drop section-divider comment (test:1511):** Removed — the `describe(...)` block name already conveys it.

Full `@aws-blocks/core` suite is green (583 tests) and there are no new lint findings on the changed lines.

ahmedhamouda78 (approved):

osama-rizk:
Hi @fossamagna ,

Thanks for this PR and your contribution to aws-blocks!

osama-rizk (approved):

PR #12, opened 2026-06-17
feat: in-repo Building Block docs discoverability

vishaalmehrishi:
Ships Building Block documentation inside the `@aws-blocks/blocks` package at a single stable path — `node_modules/@aws-blocks/blocks/docs/` — so humans and AI agents find block docs without hunting across 20+ package paths. Docs are generated at build time from each block's README and shipped via the package `files` allowlist, so `npm install` always delivers docs matching the installed version.

## What changed

- `scripts/sync-block-docs.mjs` (new) copies each `packages/<block>/README.md` → `docs/<block>.md` and generates `docs/index.md` (decision tree + catalog). Hooked into `prebuild` so it runs during `npm run build`.
- Generated `packages/blocks/docs/` is git-ignored (build artifact) and shipped to customers via the npm `files` allowlist — regenerated on every pack, so docs can't drift from code.
- `@aws-blocks/create-blocks-app` now overlays a single shared `resources/AGENTS.md` onto every scaffold, pointing agents at the canonical docs (`README.md` + `docs/index.md` + `docs/<package-name>.md`). Replaces the per-template `AGENTS.md` copies.
- `packages/blocks/README.md` rewritten for progressive disclosure (quick start → mental model → catalog → local-dev-vs-deploy → testing → best practices → common mistakes → reference).

## Customer experience

After `npx @aws-blocks/create-blocks-app my-app && npm install`, an agent reads `AGENTS.md` → `node_modules/@aws-blocks/blocks/docs/index.md` → picks a block from the decision tree → opens that block's `docs/<name>.md`. One known, version-matched path.

PR #11, opened 2026-06-17
fix email batch retry on transient SES failures

harivilasp:
## Problem

SES `sendBatch()` chunk-level transient failures such as throttling or 5xx responses were reported as failed immediately, even though the code had a TODO to retry these cases.

**Issue #, if available:**

N/A

## Changes

- Add retry handling for transient SES `SendBulkEmail` chunk failures in the AWS runtime.
- Retry up to three total attempts for throttling, retryable SDK errors, request timeout-style errors, and 5xx responses.
- Keep permanent SES failures on the existing no-retry path.
- Add AWS-runtime tests covering retry and no-retry behavior.
- Document the batch retry behavior and add a patch changeset.

## Validation

- `npx biome check packages/bb-email-client/src/index.aws.ts packages/bb-email-client/src/index.aws.test.ts packages/bb-email-client/README.md .changeset/friendly-emails-retry.md`
- `npm run build -w packages/bb-email-client`
- `npm test -w packages/bb-email-client`

## Checklist

- [x] PR description included
- [x] Tests are changed or added
- [x] Relevant documentation is changed or added (and PR referenced)

---

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

PR #177, opened 2026-07-09
fix(hosting): resolve 11 reproducible deployment issues + upgrade OpenNext to 4.0.3

osama-rizk:
## Summary

Analyze, reproduce, and fix reproducible deployment issues in the `@aws-blocks/hosting` construct, verified live against deployed CloudFront apps (Next.js App Router / Pages Router, Nuxt, Astro SSR, SPA React). Also upgrades the Next.js path to **OpenNext 4.0.3** (from 3.10.x) and widens the adapter's verified range accordingly.

All fixes include unit tests (0 regressions); every deployable fix is confirmed live.

## Fixed issues (11)

- **#13 — Next 15.x image optimizer 500:** the `fetchInternalImage` arity patch was gated on an inverted version assumption (`maximumResponseBody` was added in Next **16**, not 15.5). Now gated on the installed Next version — patch only on Next ≥ 16; Next 15.x local images no longer 500. (`patchImageOptimizerForNext155` → `patchImageOptimizerForNext16`)
- **#4 — Image optimizer 500 on disallowed types (SVG):** OpenNext caught Next's proper `400` (`dangerouslyAllowSVG` disabled) in a generic block that returned 500. Now forwards the real 4xx → **400**, fails closed.
- **#3 — Astro SSR `/_image` `content-type: image/null`:** bundle a linux-x64 `sharp` into the Astro SSR bundle (post-build, wasm fallback pruned, ~19.5 MB) so Astro's default sharp image service works on Lambda and `/_image` returns a real optimized image with a correct MIME. Gated on the app using the sharp service.
- **Astro SSR remote images 500:** switching to sharp surfaced that Astro core does `fetch(url, { redirect: "manual" })` and throws on 3xx. Added an SSRF-safe redirect-follow patch (re-validates every hop against the remote allowlist).
- **#6 — SPA `/assets/*` hashed chunks not `immutable`:** added `immutablePaths: ['assets/*']` to the SPA adapter.
- **#9 — S3 `AccessDenied` XML leak on missing key:** added `s3:ListBucket` to the OAC bucket policy → missing key returns a clean `404 NoSuchKey`.
- **#11 — Next `_rsc` prefetch fragments the CloudFront cache:** SSR cache policy now excludes `_rsc` from the cache key (`denyList('_rsc')`).
- **#5 — Wildcard redirect leaks literal `:path*`:** convert trailing `/:name*` → `/*` at the edge router, plus a bare-prefix companion redirect.
- **#8 — `TooManyRoutesError` opaque + hardcoded cap:** message now names the offending table and calls out `trailingSlash: true`; the 64-chunk cap is tunable via the new `quotas.maxRouteChunks` hosting prop.
- **#7 — Nuxt ISR/SWR on-demand child hard-404s from S3:** when ISR/SWR is active (`manifest.cache` set), route coalescing folds a prerendered static sibling group into a single `parent/*` **compute** wildcard, so a non-prebuilt on-demand child renders at the SSR Lambda while the route table stays bounded (one row per parent — avoids the CloudFront-Function compute-limit 503 a non-coalesced fan-out would cause).
- **#2 — Nuxt IPX remote images:** route the IPX image Lambda through the shared SSR API Gateway (dedicated `<baseURL>/{proxy+}` resource) instead of an OAC Function URL, so an unencoded `://` in a remote source no longer breaks SigV4 (`403 InvalidSignatureException`); and configure IPX `httpStorage` scoped to the allowlisted domains so allowlisted remote images resolve (was `404 IPX_RESOURCE_NOT_FOUND`).
- **CloudFront S3-origin policy (found during verification):** every S3-origin behavior — the default AND the edge-route (`runtime: 'edge'`) behavior — now uses a synthesized custom origin request policy instead of the managed `ALL_VIEWER_EXCEPT_HOST_HEADER`, which CloudFront rejects on S3 origins (`InvalidRequest` at distribution create — was rolling back astro-ssr/nuxt deploys). A regression guard asserts no S3-origin behavior references a managed origin request policy.

## OpenNext 4.0.3 upgrade

Bumped the Next test apps from `@opennextjs/aws@~3.10.0` to `^4.0.3`. An integration deploy of next-app-router confirmed all four bundle patches (streaming wrapper, edge-bundle process banner, `fetchInternalImage` arity insertion, SVG-status catch rewrite) still match the 4.0.x minified shape, and the live app served optimized rasters (200 image/webp), a fail-closed SVG (400), edge routes, and redirects with no regressions. `VERIFIED_OPENNEXT_RANGE` now covers `>=3.10.0 <3.11.0 || >=4.0.0 <4.1.0`.

## Live verification

Deployed apps (us-west-2, all `.cloudfront.net`): next-app `d3qjt5bgcu7mz4` (OpenNext 4.0.3), next15-image-opt `ddsec8g513mz2`, nuxt `drz7zx8m0esrg`, astro-ssr `di39hv5m07ep0`, spa `d2v8qyholmjddj`.

| Issue | Probe | Result |
|---|---|---|
| #13 | `/_next/image?...blocks-photo.png&w=256` (Next 16 **and** Next 15.5.18) | ✅ `200 image/webp` |
| #4 | `/_next/image?...stress-logo.svg` | ✅ `400` (was 500) |
| #3 | Astro `/_image?...&f=webp` | ✅ `200 image/webp` (was `image/null`) |
| Astro remote img | `/image-remote` (allowlisted host, 3xx) | ✅ `200` (was 500) |
| #6 | SPA hashed asset cache-control | ✅ `public, max-age=31536000, immutable` |
| #9 | Missing asset (SPA / next / astro) | ✅ `404 NoSuchKey`, no AccessDenied XML |
| #11 | Two `_rsc` values → one cache entry | ✅ Miss → Hit |
| #5 | `/r/legacy`, `/r/legacy/`, `/r/legacy/a/b/c` | ✅ all clean 308 (no `:path*`) |
| #2 | Nuxt `/_ipx/...` local + allowlisted remote | ✅ `200 image/*` (was 404/403) |
| #7 | ISR manifest router (prebuilt→S3, non-prebuilt→compute) | ✅ table bounded, no 503 |
| origin-policy | astro-ssr + nuxt distribution create | ✅ no InvalidRequest; deploys succeed |

## Deferred (documented as accepted limitations)

- **#10 — `force-static` + `dynamicParams=false` non-prebuilt param → 502:** upstream Next.js + OpenNext interaction — Next throws `NoFallbackError` from an un-awaited async promise (`handleRevalidate`), escaping OpenNext's try/catch → Lambda crash → 502. Structurally identical in OpenNext 3.10.x and 4.0.x, so a version bump doesn't fix it. An in-adapter `unhandledRejection` shim was considered and rejected (fragile, risks masking real rejections).
- **#12 — Edge decodes `%2B` → `+` in query string:** API Gateway REST **v1** URL-decodes the query and has no `rawQueryString` field. The SSR Lambda is on v1 REST specifically because v1 is the only API-GW flavor that supports response streaming; v2 (HTTP API) would preserve the raw query but lose streaming — a net-negative trade. Accepted platform limitation (Amplify #4095 class).
- **#9c — CORS preflight `OPTIONS` on a pure-static (S3-only) origin leaks AccessDenied XML:** the main #9 case is fixed; this residual needs a product CORS-posture decision (a CloudFront `ResponseHeadersPolicy` to answer preflights before they reach S3), so it's tracked separately.

## Test plan
- [x] `npm run test -w packages/hosting` → all pass (incl. adapter version-pin + S3-origin-policy regression guards), 0 failures
- [x] `npx tsc --build` → clean across hosting, core, blocks
- [x] Lint on changed files → same error count as main (pre-existing)
- [x] Live deploys (next-app @ OpenNext 4.0.3, next15, nuxt, astro-ssr, spa) → all fixes verified + regression sweeps clean
- [x] Changesets added (`@aws-blocks/hosting` + `@aws-blocks/core`)


packages/hosting/src/adapters/astro.ts
```diff
@@ -699,6 +734,217 @@ const buildStaticManifest = (distDir: string): DeployManifest => {
...
+        '--silent',
+        '--include=optional',
+        '--omit=dev',
+        '--os=linux',
+        '--cpu=x64',
+        '--libc=glibc',
+      ],
+      { cwd: serverDir, stdio: 'inherit' },
+    );
+  } catch (error) {
```
soberm, line 831: [minor] `spawn.sync` returns a result object — it doesn't *throw* on a non-zero exit code, it just returns `{ status: 1 }`. The `try/catch` here only catches a spawn failure (ENOENT/EACCES). A failed `npm install` (network error, registry timeout, resolution conflict) exits with code 1, `spawn.sync` returns silently, and the function continues as if sharp landed — same symptom as the original bug #3 at Lambda runtime, no build-time signal.

```ts
const result = spawn.sync('npm', [...], { cwd: serverDir, stdio: 'inherit' });
if (result.error) throw result.error;
if (result.status !== 0) {
  throw new Error(`npm install sharp exited with code ${result.status}`);
}
```
Then promote into `HostingError` (or at least wrap in the existing catch) so the build fails loudly.

packages/hosting/src/adapters/astro.ts
```diff
@@ -699,6 +734,217 @@ const buildStaticManifest = (distDir: string): DeployManifest => {
...
+    '@img',
+    'sharp-linux-x64',
+  );
+  if (fs.existsSync(existing)) return;
+
+  fs.mkdirSync(serverDir, { recursive: true });
+  // A minimal package.json so `npm install <pkg>` writes into THIS dir's
+  // node_modules and doesn't walk up to the project root.
+  const pkgJsonPath = path.join(serverDir, 'package.json');
+  const hadPkgJson = fs.existsSync(pkgJsonPath);
```
soberm, line 793: [nit] If the install fails (and after fixing the status-check above), the `package.json` written just before the npm call is never cleaned up. On the next build run `hadPkgJson` will be `true`, so the cleanup branch stays skipped. Harmless in practice (stale `package.json` in a build artifact), but worth a best-effort `unlinkSync` in the error path:

```ts
if (!hadPkgJson && installFailed) {
  try { fs.unlinkSync(pkgJsonPath); } catch { /* best-effort */ }
}
```

packages/hosting/src/constructs/kvs_router.ts
```diff
@@ -344,14 +382,37 @@ export const buildKvsEntries = (input: BuildKvsInput): Record<string, string> =>
     redirectChunks.length,
     headerChunks.length,
   );
-  if (tooManyChunks > KVS_BUDGET.maxChunksPerTable) {
+  // Tunable via `quotas.maxRouteChunks` (see issue #8); defaults to 64.
+  const maxChunksPerTable =
+    input.maxChunksPerTable && input.maxChunksPerTable > 0
```
soberm: [nit] `input.maxChunksPerTable > 0` passes for `0.5` or other non-integers. The JSDoc implies an integer ("max chunks"). Worth a `Number.isInteger` guard:

```ts
const maxChunksPerTable =
  Number.isInteger(input.maxChunksPerTable) && (input.maxChunksPerTable ?? 0) > 0
    ? input.maxChunksPerTable!
    : KVS_BUDGET.maxChunksPerTable;
```

soberm (commented):
Thorough, well-scoped PR. All 11 fixes are correctly diagnosed and the test coverage is solid — particularly the S3-origin-policy regression guard (checking every S3-origin behavior, not just the one that was broken), and the SSRF-safe redirect test that proves a disallowed redirect destination is not followed. The ISR coalescing trade-off note in `kvs_router.ts` is a good example of documenting the *why* of a subtle correctness decision.

A few non-blocking things below — mostly error-handling robustness and small clarity gaps.

packages/hosting/src/adapters/nextjs.ts
```diff
@@ -1331,30 +1400,71 @@ export const patchEdgeBundlesForLambdaEdge = (openNextDir: string): void => {
...
 /**
- * Patch the OpenNext image-optimization bundle for the Next.js 15.5 image
- * optimizer signature change.
+ * Read the installed Next.js version from the project's node_modules.
+ * Best-effort: returns `undefined` when the package isn't resolvable or the
+ * version can't be parsed (e.g. an L3 path running against a pre-built
+ * `.open-next/` with no `next` in scope).
+ * @internal
+ */
+export const detectNextVersion = (projectDir: string): string | undefined => {
```
soberm, line 1411: [nit] When `next` isn't resolvable at all, `detectNextVersion` returns `undefined` and the patch is applied silently (right default, since Next 16 is the common case). The skip path gets a helpful `stderr.write`; it'd be nice to have a matching one for the "version unknown, defaulting to patch" path:

```ts
if (!nextVersion) {
  process.stderr.write(
    `ℹ️  Could not detect Next.js version; applying fetchInternalImage arity patch (safe for Next ≥ 16).\n`
  );
}
```
Makes build logs easier to read when working from a pre-built `.open-next/` without a resolvable `next` in scope.

packages/hosting/src/adapters/astro.test.ts
```diff
@@ -0,0 +1,170 @@
...
+      globalThis.fetch = (u, opts) => fetchImpl(typeof u === 'string' ? u : u.toString());
+      return (async () => {
+        const okRes = await ${ASTRO_REDIRECT_PATCH_MARKER}('https://picsum.test/a', { domains: [] }, isAllowed);
+        const badRes = await ${ASTRO_REDIRECT_PATCH_MARKER}('https://picsum.test/evil', { domains: [] }, isAllowed);
+        return { okStatus: okRes.status, badStatus: badRes.status };
+      })();
+    `;
+    // Map<...>.get(k) mirrors Headers.get(k); the helper uses .headers.get('location').
+    // eslint-disable-next-line @typescript-eslint/no-implied-eval
+    const fn = new Function(harness.replace('.headers.get("location")', '.headers.get("location")'));
```
soberm, line 205: [nit] The `// eslint-disable-next-line` comment acknowledges `new Function` but doesn't explain *why* a direct import isn't possible. Future maintainers might read this as lazy test design. A one-line explanation would help:

```ts
// The helper is injected into the Astro SSR bundle at build time (not
// exported from astro.ts), so evaluating the extracted function body via
// new Function is the only way to test the actual injected code path
// without a full Astro build.
// eslint-disable-next-line @typescript-eslint/no-implied-eval
```

packages/hosting/src/constructs/cdn_construct.ts
```diff
@@ -469,6 +478,48 @@ export class CdnConstruct extends Construct {
...
+    // path signature, so a `://` in the path is inert. Attach the IPX Lambda
+    // as a dedicated `<baseURL>/{proxy+}` resource on the SAME RestApi the SSR
+    // Lambda already uses (no new API GW), and point the image origin at that
+    // shared host. Scoped to IPX only: detected by `imageOptimization.baseURL`
+    // (Nuxt sets it; Next's OpenNext image bundle does not and speaks Function
+    // URL v2 events, so it stays on its Function URL). Requires an SSR compute
+    // (hence a RestApi) — a static-only Nuxt deploy keeps the Function URL.
+    const ipxBaseUrl = manifest.imageOptimization?.baseURL;
+    const ipxFn = props.computeFunctions?.get('image-optimization');
+    if (ipxBaseUrl && ipxFn && restApi && apiHostname) {
```
soberm, line 510: [minor] When `ipxBaseUrl` is set but `restApi` is undefined (static-only Nuxt deploy — no SSR compute), the IPX Lambda silently falls back to its Function URL with no warning. This is documented in the comment above, but the caller has no idea at build time. If someone accidentally deploys Nuxt without SSR compute they'll see a non-obvious SigV4 403 at runtime instead of a build-time nudge:

```ts
if (ipxBaseUrl && ipxFn) {
  if (!restApi || !apiHostname) {
    process.stderr.write(
      `⚠️  imageOptimization.baseURL is set (IPX) but no SSR compute exists — ` +
      `IPX Lambda will use a Function URL origin (SigV4 path-encoding issue #2 may recur).\n`
    );
  } else {
    // ... existing wiring
  }
}
```

packages/hosting/src/constructs/cdn_construct.ts
```diff
@@ -824,7 +912,15 @@ export class CdnConstruct extends Construct {
...
-          originRequestPolicy: OriginRequestPolicy.ALL_VIEWER_EXCEPT_HOST_HEADER,
+          // This behavior's origin is S3 (line above), so it CANNOT use the
+          // managed ALL_VIEWER_EXCEPT_HOST_HEADER — CloudFront rejects that
+          // managed policy on any S3 origin (400 InvalidRequest at create).
+          // Reuse the synthesized custom policy (same forwarding, allowed on S3
+          // origins). An edge route implies a compute route, so hasCompute is
+          // true and forwardAllOriginRequestPolicy is defined; the `?? undefined`
+          // is a belt-and-suspenders guard (the Lambda@Edge function generates
+          // the response, so forwarded origin data is immaterial here anyway).
+          originRequestPolicy: forwardAllOriginRequestPolicy ?? undefined,
```
soberm: [nit] `forwardAllOriginRequestPolicy ?? undefined` is a no-op — if the value is already `undefined`, `??` returns `undefined` anyway. Just `forwardAllOriginRequestPolicy` is equivalent. The comment explaining the belt-and-suspenders reasoning is fine, but the expression itself adds noise.

packages/hosting/src/adapters/nextjs.ts
```diff
@@ -1146,7 +1202,20 @@ ${entries}
...
-export const VERIFIED_OPENNEXT_RANGE = '>=3.10.0 <3.11.0';
+//
+// Verified against 3.10.x AND 4.0.x: an OpenNext 4.0.3 integration deploy
+// (next-app-router) confirmed all four bundle patches still apply cleanly —
+// the streaming-wrapper (`setContentType` + br/gzip/deflate identity
+// branch), the edge-bundle process banner, the `fetchInternalImage` arity
+// insertion, and the SVG-status catch rewrite all matched the 4.0.x
+// minified shape, and the live app served optimized rasters (200 image/webp),
+// a fail-closed SVG (400), edge routes, and redirects with no regressions.
+// The two windows are called out
```
soberm: [nit] Minor reflow: "The two windows are called out" is split from "explicitly" across lines in an awkward way. Reads better as a single clause on one line.

packages/hosting/src/adapters/ipx_lambda_template.ts
```diff
@@ -214,21 +231,60 @@ const ipxStripPattern = new RegExp(
 );
 
 /**
- * Convert a Lambda Function URL event into a standard fetch Request.
+ * Rebuild a query string from API Gateway REST (v1) query params. Prefers
+ * multiValue (preserves repeats + original encoding) and re-encodes each
+ * key/value so a remote source carrying reserved chars round-trips.
+ */
+const v1QueryString = (event) => {
```
soberm, line 254: [question] The JSDoc says this re-encodes each key/value with `encodeURIComponent`. API Gateway REST v1 auto-decodes query string values before delivering them to the Lambda, so `encodeURIComponent` here is re-encoding already-decoded values — which is correct and not double-encoding. But it's worth a brief note confirming that's intentional, since the comment currently just says "re-encodes each key/value" without explaining *why* that's the right thing to do on a v1 event (where APIGW pre-decoded them).

packages/hosting/src/adapters/astro.ts
```diff
@@ -699,6 +734,217 @@ const buildStaticManifest = (distDir: string): DeployManifest => {
...
+        '--no-fund',
+        '--silent',
+        '--include=optional',
+        '--omit=dev',
+        '--os=linux',
+        '--cpu=x64',
+        '--libc=glibc',
+      ],
+      { cwd: serverDir, stdio: 'inherit' },
+    );
```
ahmedhamouda78, line 830: `[major]` `installSharpForAstroSsr` in `packages/hosting/src/adapters/astro.ts` (the `spawn.sync('npm', ...)` call) — `spawn.sync` (cross-spawn) returns a result object and does **not** throw on a non-zero exit code; the `try/catch` only catches spawn-level failures (ENOENT/EACCES). A failed `npm install` (registry timeout, resolution conflict, network) exits `1`, returns silently, and the function proceeds as if `sharp` landed — reintroducing issue #3 at Lambda runtime with no build-time signal, defeating the fail-loudly intent. Throwing on non-zero status lets the existing `catch` wrap it into `HostingError('AstroSharpInstallError', ...)`. Suggested fix (anchored to the `spawn.sync(...)` statement inside the `try`):

```suggestion
    const result = spawn.sync(
      'npm',
      [
        'install',
        `sharp@${ASTRO_SHARP_VERSION}`,
        '--no-audit',
        '--no-fund',
        '--silent',
        '--include=optional',
        '--omit=dev',
        '--os=linux',
        '--cpu=x64',
        '--libc=glibc',
      ],
      { cwd: serverDir, stdio: 'inherit' },
    );
    if (result.error) throw result.error;
    if (result.status !== 0) {
      throw new Error(`npm install sharp exited with code ${result.status}`);
    }
```

Also add a unit test for the non-zero-exit path — currently only the pre-seeded-marker short-circuit is tested, so this failure mode ships untested.

packages/hosting/src/adapters/nextjs.ts
```diff
@@ -761,6 +796,27 @@ const applyLiftedRoutesManifest = (
...
+    // The edge router's matchPattern requires at least the `/` after the
+    // prefix to match `/*`, so a request to the bare prefix falls through to
+    // Lambda and leaks the raw `:path*`. The companion catches it and
+    // redirects to the destination prefix (strip the trailing `*`).
+    if (source.endsWith('/*')) {
+      const bareSource = source.slice(0, -2); // '/r/legacy/*' → '/r/legacy'
+      const bareDest = destination.endsWith('/*')
+        ? destination.slice(0, -1) // '/r/modern/*' → '/r/modern/'
+        : destination;
+      if (!seenRedirectSources.has(bareSource)) {
```
ahmedhamouda78: `[minor]` `applyLiftedRoutesManifest` in `packages/hosting/src/adapters/nextjs.ts` (bare-prefix companion block) — a root-level catch-all `{ source: '/:path*' }` normalizes to `/*`, so `bareSource = '/*'.slice(0, -2)` = `''` and you push a redirect with `source: ''` (malformed). Reachability depends on `isSimpleNextSource('/*')`, which isn't in the diff so I couldn't confirm it's rejected upstream. Cheap to guard regardless. Suggested fix (anchored to the `if (!seenRedirectSources.has(bareSource)) {` line):

```suggestion
      if (bareSource.length > 0 && !seenRedirectSources.has(bareSource)) {
```

packages/hosting/src/adapters/astro.ts
```diff
@@ -699,6 +734,217 @@ const buildStaticManifest = (distDir: string): DeployManifest => {
...
+ * app that deliberately picked noop/custom is respected (and not bloated with
+ * ~19 MB of libvips it won't call).
+ * @internal
+ */
+export const astroUsesSharpService = (config: AstroConfigShape): boolean => {
+  const entry = config.image?.service?.entrypoint;
+  // Unset → Astro default → sharp. Explicit sharp entrypoint → sharp.
+  if (entry === undefined || entry === null) return true;
+  if (typeof entry !== 'string') return true;
+  return /(^|\/)sharp$/.test(entry) || entry === 'astro/assets/services/sharp';
```
ahmedhamouda78: `[nit]` `astroUsesSharpService` in `packages/hosting/src/adapters/astro.ts` — the return condition is redundant: `astro/assets/services/sharp` ends in `/sharp`, so the regex already matches it and the `|| entry === 'astro/assets/services/sharp'` clause is dead code. Suggested fix: drop the `||` clause → `return /(^|\/)sharp$/.test(entry);`

packages/hosting/src/adapters/ipx_lambda_template.ts
```diff
@@ -193,8 +193,25 @@ const s3IpxStorage = {
...
+// Remote (http/https) sources need IPX's HTTP storage — without it IPX
+// routes a remote id to the S3 \`storage\` above, which has no such key, so
+// every allowlisted remote image 404s with IPX_RESOURCE_NOT_FOUND (issue #2).
+// Configure \`httpStorage\` scoped to the SAME allowlist the handler enforces
+// (hostnames from IMAGE_ALLOWED_HOSTNAMES + IMAGE_REMOTE_PATTERNS). IPX picks
+// \`httpStorage\` for \`http(s)://\` ids and \`storage\` (S3) for local keys.
+// \`isRemoteSourceAllowed\` still gates every request BEFORE IPX runs, so the
+// domains list here is defense-in-depth, not the sole guard. Omit httpStorage
+// entirely when no remote host is allowlisted (default-deny: local only).
+const httpDomains = [
```
ahmedhamouda78, line 214: `[minor]` `httpDomains` construction in `packages/hosting/src/adapters/ipx_lambda_template.ts` — `[...allowedHostnames, ...parsedRemotePatterns.map((p) => p.hostname)].filter(Boolean)` drops any `remotePattern` constraining only by pathname/protocol (no `hostname`), so `ipxHttpStorage` isn't scoped to allow it and those remote images can 404 again — even though the handler's `isRemoteSourceAllowed` would permit them, a mismatch between the two allowlists. Suggested fix: document that hostname-less remote patterns aren't supported by `httpStorage` scoping, or broaden the `httpStorage` config when such a pattern is present (still gated by the handler's SSRF check).

packages/hosting/src/adapters/astro.ts
```diff
@@ -699,6 +734,217 @@ const buildStaticManifest = (distDir: string): DeployManifest => {
...
+    if (!callRe.test(src)) continue;
+    callRe.lastIndex = 0;
+    // Replace each `fetch(u,{redirect:"manual"})` with the helper, threading
+    // the local `allowlistConfig` + imported `isRemoteAllowed`.
+    let next = src.replace(
+      callRe,
+      `await ${ASTRO_REDIRECT_PATCH_MARKER}($1, typeof allowlistConfig !== "undefined" ? allowlistConfig : void 0, isRemoteAllowed)`,
+    );
+    // Inject the helper once, after the import lines at the top of the chunk.
+    next = `${ASTRO_REDIRECT_HELPER}\n${next}`;
```
ahmedhamouda78, line 948: `[nit]` `patchAstroRemoteImageRedirects` in `packages/hosting/src/adapters/astro.ts` (helper injection) — the code prepends the helper to the whole source (`${ASTRO_REDIRECT_HELPER}\n${next}`), placing it **before** the imports, but the comment says "Inject the helper once, **after** the import lines." Functionally fine (ESM imports hoist; the helper has no import dependency), but comment and code disagree. Suggested fix: update the comment to say the helper is prepended and relies on ESM import hoisting.

packages/hosting/src/adapters/astro.test.ts
ahmedhamouda78, line 205: `[minor]` `packages/hosting/src/adapters/astro.test.ts` and `packages/hosting/src/adapters/nextjs.test.ts` — the bundle-patch tests validate hand-written, friendly-named fixtures (`import { i as isRemoteAllowed }`, `Pc.fetchInternalImage`, `Pl`/`VX`). Real minified output can rename these to single identifiers, in which case guards like `if (!/isRemoteAllowed/.test(src)) continue` skip silently and the patch never applies — yet the units still pass, giving false confidence that the patch fires on production bundles. Suggested fix: add a comment to each shape-fixture test noting it validates the patch's *shape logic*, not real minified output, and that production coverage comes from the integration deploy (no code change needed).

.changeset/hosting-existing-issues-fixes.md
```diff
@@ -0,0 +1,19 @@
+---
+"@aws-blocks/hosting": patch
+"@aws-blocks/core": patch
```
ahmedhamouda78: `[minor]` `.changeset/hosting-existing-issues-fixes.md` and `packages/core/src/hosting.ts` — both packages are bumped `patch`, but this PR adds a new public optional prop `quotas.maxRouteChunks` to `@aws-blocks/core`'s `HostingProps`; a new (even optional) public API field is conventionally a **minor**, and the Astro auto-`sharp`-bundling behavior change is also arguably minor. Suggested fix: confirm against the repo's changeset convention; if it follows semver for public API additions, bump `@aws-blocks/core` (and possibly `@aws-blocks/hosting`) to `minor`.

ahmedhamouda78 (commented):
Thorough, well-scoped PR — the diagnoses are precise and the regression guards (the S3-origin-policy sweep across *all* S3-origin behaviors, and the SSRF-safe per-hop redirect revalidation test) are genuinely strong. I reviewed the full diff line-by-line.

One item I'd treat as blocking: `installSharpForAstroSsr` silently swallows a failed `npm install`, which reintroduces issue #3 at runtime with no build signal — details inline. The rest are non-blocking (edge-case guards, redundancy, doc/test-clarity, and a changeset semver question).

Cross-cutting note on the bundle-patch approach: the image/redirect patches match minified bundle *shapes* with regexes, but the unit tests validate hand-written friendly fixtures rather than real Vite/Rollup/OpenNext output. Green unit tests therefore don't prove the patches fire in production — the integration deploy is the real coverage. Worth calling that out in the test comments so it isn't over-trusted later.

ahmedhamouda78 (changes requested):

packages/hosting/src/adapters/astro.ts
osama-rizk: Pushing back on the premise: this isn't Node's raw `spawn.sync` — it's the local `./spawn.js` wrapper, which already throws on a non-zero exit:

```ts
// spawn.ts
if (result.error) throw result.error;
if (result.status !== 0) throw new Error(`Command failed (exit code ${result.status})...`);
```

So a failed `npm install` (timeout/conflict/network → exit 1) does NOT return silently — it throws, lands in the existing `catch`, and becomes `HostingError(AstroSharpInstallError)`. Added a code comment making the wrapper contract explicit, and a unit test that shadows `npm` with an `exit 1` shim to lock in the fail-loudly behavior (399db46).

packages/hosting/src/adapters/astro.ts
osama-rizk: Done — added best-effort `unlinkSync` of the just-written `package.json` in the install error path so a retry isn't fooled by a stale `hadPkgJson === true` (1dac902).

packages/hosting/src/constructs/kvs_router.ts
osama-rizk: Done — added a `Number.isInteger` guard so a fractional value (e.g. `0.5`) falls back to the default rather than capping between chunk counts (1dac902).

packages/hosting/src/adapters/nextjs.ts
osama-rizk: Done — added a matching `stderr.write` on the "version undetectable → patch anyway" path, symmetric with the skip-path log (1dac902).

packages/hosting/src/adapters/astro.test.ts
osama-rizk: Done — expanded the comment: `ASTRO_REDIRECT_HELPER` is an injected source string, never exported from astro.ts, so `new Function` is the only way to exercise the actual injected code path without a full Astro build (1dac902).

packages/hosting/src/constructs/cdn_construct.ts
osama-rizk: Done — added a build-time `stderr` warning when `imageOptimization.baseURL` (IPX) is set but no SSR compute exists, so the Function-URL fallback (where issue #2 can recur) is surfaced instead of a silent runtime 403 (1dac902).

packages/hosting/src/constructs/cdn_construct.ts
osama-rizk: Done — dropped the `?? undefined` no-op; kept the explanatory comment (1dac902).

packages/hosting/src/adapters/nextjs.ts
osama-rizk: Done — reflowed the verified-range comment onto a single clause (1dac902).

packages/hosting/src/adapters/ipx_lambda_template.ts
osama-rizk: Done — expanded the JSDoc to confirm it's intentional: APIGW REST v1 pre-decodes query values (no `rawQueryString`), so `encodeURIComponent` rebuilds a valid query from already-decoded values — encoded once, not double-encoded (1dac902).

packages/hosting/src/adapters/astro.ts
osama-rizk: Pushing back on the "silently swallows" diagnosis, but taking the test ask. This is the local `./spawn.js` wrapper, not raw cross-spawn — it already throws on non-zero exit:

```ts
if (result.status !== 0) throw new Error(`Command failed (exit code ${result.status})...`);
```

So a failed `npm install` already lands in the `catch` and becomes `HostingError(AstroSharpInstallError)` — issue #3 is NOT reintroduced silently. Added a comment making that contract explicit, and (per your suggestion) a unit test that shadows `npm` with an `exit 1` shim to assert it throws + cleans up its package.json (399db46).

packages/hosting/src/adapters/nextjs.ts
osama-rizk: Good catch — done. A root-level `/:path*` normalizes to `/*` → `bareSource === \'`. Added `bareSource.length > 0 &&` to skip the empty-source companion (399db46).

packages/hosting/src/adapters/astro.ts
osama-rizk: Done — dropped the dead `|| entry === 'astro/assets/services/sharp'` clause; `/(^|\/)sharp$/` already matches it (399db46).

packages/hosting/src/adapters/ipx_lambda_template.ts
osama-rizk: Documented the limitation (399db46): a hostname-less remotePattern contributes no `httpStorage` domain, so it can 404 despite the handler allowing it. Broadening httpStorage to all hosts on such a pattern would defeat the scoping, so the trade-off is: author remotePatterns with an explicit hostname for httpStorage coverage (the handler's SSRF check still gates every request).

packages/hosting/src/adapters/astro.ts
osama-rizk: Done — corrected the comment: the helper is PREPENDED (before imports) and relies on ESM import hoisting (399db46).

packages/hosting/src/adapters/astro.test.ts
osama-rizk: Done — added a note to both bundle-patch test blocks (astro.test.ts + nextjs.test.ts) that they validate the patch's SHAPE logic on friendly fixtures, not real minified output, and that production coverage comes from the integration deploy (399db46).

.changeset/hosting-existing-issues-fixes.md
osama-rizk: Agreed — bumped both `@aws-blocks/core` and `@aws-blocks/hosting` from patch to MINOR (new public `quotas.maxRouteChunks` field + the Astro auto-sharp-bundling behavior change) (399db46).

osama-rizk:
Thanks both for the thorough reviews 🙏 — addressed all feedback across two commits (`1dac902`, `399db46`, pushed):

**Accepted (13):** package.json cleanup on install failure; `Number.isInteger` guard on `quotas.maxRouteChunks`; version-undetectable log; `new Function` rationale comment; IPX no-SSR-compute build warning; dropped the `?? undefined` no-op; verified-range comment reflow; v1 query re-encode rationale; empty bare-prefix redirect guard; dead sharp-clause removal; helper-injection comment fix; hostname-less remotePattern limitation doc; shape-fixture test disclaimers.

**Semver (C7):** bumped `@aws-blocks/core` + `@aws-blocks/hosting` patch → **minor** (new public `quotas.maxRouteChunks` field).

**One push-back — the `installSharpForAstroSsr` "silently swallows a failed `npm install`" concern (flagged by both):** the call goes through the local `./spawn.js` wrapper, which already `throw`s on a non-zero exit (`if (result.status !== 0) throw ...`), so a failed install already becomes `HostingError('AstroSharpInstallError')` — issue #3 is not silently reintroduced. I made the wrapper contract explicit in a comment and added a unit test (fake `npm` shim exiting 1) that locks in the fail-loudly + package.json-cleanup behavior.

Redeploying the affected apps and re-running the full e2e suite to confirm no regressions.

ahmedhamouda78 (dismissed):
Re-reviewed at `399db46b` — thanks for the fast, thorough turnaround. Confirmed the fixes:

- **C2** empty bare-prefix guard (`bareSource.length > 0`), **C3** dead sharp-clause removal, **C4** hostname-less `remotePattern` doc, **C5** helper-injection comment, **C6** shape-fixture test disclaimers, and **C7** semver bump to **minor** on both `@aws-blocks/core` and `@aws-blocks/hosting` — all landed.

- On my `[major]` `installSharpForAstroSsr` concern: **you're right, and I verified it.** `spawn` here is the local `./spawn.js` wrapper, which already `throw`s on both `result.error` and `result.status !== 0`, so a failed `npm install` already surfaces as `HostingError('AstroSharpInstallError')` — issue #3 is not silently reintroduced. That was a false positive on my end (I assumed cross-spawn's non-throwing `sync` semantics without checking the import). Making the wrapper contract explicit + the fail-loudly unit test is good hardening regardless.

Approving. 🚢

Non-blocking reminder (no action needed for this PR): the image/redirect bundle patches ultimately rely on the e2e deploy for production-shape coverage, as noted — glad to see you're re-running it.

.changeset/hosting-existing-issues-fixes.md
osama-rizk: Following up on this — after discussion we reverted both packages back to **patch** (commit `8e4cf9a`).

Reasoning: both are pre-1.0 (`@aws-blocks/core@0.1.11`, `@aws-blocks/hosting@0.1.4`) and consumers pin with caret. For a `0.x` version, `^0.1.x` resolves to `>=0.1.x <0.2.0` ([node-semver caret rule for major-zero](https://github.com/npm/node-semver#caret-ranges-123-025-004)), so a `0.2.0` **minor** would fall OUTSIDE existing ranges and NOT auto-upgrade — effectively a breaking release for anyone on `^0.1.x`. Per SemVer point 4 the 0.x API isn't considered stable, and the convention here is that the leftmost non-zero element (minor, while major is 0) is the breaking indicator.

The additions in this PR (optional `quotas.maxRouteChunks`, Astro sharp auto-bundling) are backward-compatible, so `patch` keeps existing `^0.1.x` consumers receiving them. Your strict-SemVer point is correct for ≥1.0 — happy to switch to minor once these packages hit 1.0.

soberm (approved):

PR #180, opened 2026-07-09
fix(bb-realtime): harden subscription token validation to prevent connect token reuse

svidgen:
Hardens realtime subscription token validation. Connect tokens are now scoped with a `$connect` suffix so they cannot be used to authorize channel subscriptions. Channel tokens remain valid as connect tokens since they already prove instance membership. Backward-compatible: old connect tokens still work for WebSocket establishment during rollout.

## Changes

- **`mintConnectToken`** now appends `$connect` (no slash separator) to the scope prefix when creating the token payload. This means the connect token's channel field is e.g. `myapp-rt$connect` instead of `myapp-rt`, which does NOT prefix-match real channels like `myapp-rt/chat/room-1`. The lack of a `/` separator also prevents collision with any real channel namespace.
- **No changes needed to subscribe handler** — the existing `startsWith` check in `validateChannelToken` naturally rejects connect tokens now.
- **No changes needed to `$connect` handler** — it validates tokens without a `requestedChannel` (signature + expiry only), so both old-style and new-style connect tokens (and channel tokens) continue to work for connection establishment.

## Testing

- Added `token-scope.test.ts` with 7 test cases covering the hardening and backward compatibility.
- Updated `security.test.ts` assertions to reflect the corrected behavior.
- All 69 existing tests continue to pass.

### DIFF
--- a/packages/bb-realtime/src/index.ts
+++ b/packages/bb-realtime/src/index.ts
@@ -25,7 +25,7 @@ import type {
 	SubscribeOptions,
 } from './types.js';
 import { RealtimeErrors } from './errors.js';
-import { blocksError, validateSchema, mintChannelToken, validateChannelPath, validatePublishSize } from './utils.js';
+import { blocksError, validateSchema, mintChannelToken, mintConnectToken, validateChannelPath, validatePublishSize } from './utils.js';
 import { getBroadcastBus, LOCAL_TOKEN_SECRET } from './local-dev.js';
 import { Logger } from '@aws-blocks/bb-logger';
 import type { ChildLogger } from '@aws-blocks/bb-logger';
@@ -147,6 +147,7 @@ export const Realtime: {
 		const fullChannel = `${ns.prefix}/${channel}`;
 		validateChannelPath(fullChannel);
 		const token = mintChannelToken(fullChannel, LOCAL_TOKEN_SECRET);
+		const connectToken = mintConnectToken(this.fullId, LOCAL_TOKEN_SECRET);
 		return {
 			subscribe(handlerOrOptions: ((message: unknown) => void) | SubscribeOptions, _options?: never): RealtimeSubscription {
 				const handler = typeof handlerOrOptions === 'function' ? handlerOrOptions : handlerOrOptions.onMessage;
@@ -161,6 +162,7 @@ export const Realtime: {
 					__blocks: 'realtime/channel' as const,
 					channel: fullChannel,
 					wsUrl: (globalThis as any).__BLOCKS_REALTIME_WS_URL__,
+					connectToken,
 					token,
 				};
 			},
--- a/packages/bb-realtime/src/security.test.ts
+++ b/packages/bb-realtime/src/security.test.ts
@@ -101,9 +101,12 @@ describe('Security: mintConnectToken secret guard', () => {
 		const token = mintConnectToken('test-app-rt', 'valid-secret');
 		assert.ok(token);
 		assert.ok(token.includes('.'), 'token should have payload.signature format');
-		// connect token authorizes any sub-channel under the scope prefix
-		const result = validateChannelToken(token, 'valid-secret', 'test-app-rt/events/room-1');
-		assert.ok(result, 'connect token should authorize sub-channels');
+		// connect token no longer authorizes sub-channels (scope isolation)
+		const subResult = validateChannelToken(token, 'valid-secret', 'test-app-rt/events/room-1');
+		assert.strictEqual(subResult, null, 'connect token must NOT authorize sub-channels');
+		// but it still validates without a requestedChannel (for connection establishment)
+		const connResult = validateChannelToken(token, 'valid-secret');
+		assert.ok(connResult, 'connect token should validate for connection establishment');
 	});
 });
 
@@ -152,7 +155,11 @@ describe('Security: getChannel() secret unavailability (AWS runtime simulation)'
 		assert.ok(chResult, 'channel token should validate');
 		assert.strictEqual(chResult!.channel, fullChannel);
 
+		// connect token no longer authorizes specific channels (scope isolation)
 		const connResult = validateChannelToken(connectToken, secret, fullChannel);
-		assert.ok(connResult, 'connect token should authorize the channel');
+		assert.strictEqual(connResult, null, 'connect token must NOT authorize specific channels');
+		// but validates for connection (no requestedChannel)
+		const connValidation = validateChannelToken(connectToken, secret);
+		assert.ok(connValidation, 'connect token should validate for connection');
 	});
 });
--- a/packages/bb-realtime/src/security.test.ts
+++ b/packages/bb-realtime/src/security.test.ts
@@ -101,9 +101,9 @@ describe('Security: mintConnectToken secret guard', () => {
 		const token = mintConnectToken('test-app-rt', 'valid-secret');
 		assert.ok(token);
 		assert.ok(token.includes('.'), 'token should have payload.signature format');
-		// connect token no longer authorizes sub-channels (scope isolation)
+		// connect tokens should not authorize channel subscriptions
 		const subResult = validateChannelToken(token, 'valid-secret', 'test-app-rt/events/room-1');
-		assert.strictEqual(subResult, null, 'connect token must NOT authorize sub-channels');
+		assert.strictEqual(subResult, null, 'connect token must NOT authorize channel subscriptions');
 		// but it still validates without a requestedChannel (for connection establishment)
 		const connResult = validateChannelToken(token, 'valid-secret');
 		assert.ok(connResult, 'connect token should validate for connection establishment');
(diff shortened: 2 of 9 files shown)


packages/bb-realtime/src/utils.ts
```diff
@@ -79,17 +79,19 @@ export function mintChannelToken(channel: string, secret: string, ttlSeconds = 3
...
+ * Mint a connect token. Scoped to a Realtime instance prefix with a `$connect`
+ * suffix (e.g., `myapp-rt$connect`). Connect tokens authorize WebSocket
+ * connection establishment but not channel subscriptions — the `$connect`
+ * suffix ensures the token's channel field does not prefix-match real channel
+ * paths (which always contain a `/` separator after the instance prefix).
  *
  * Default TTL is 2 hours (matching API Gateway max connection duration).
  */
 export function mintConnectToken(scopePrefix: string, secret: string, ttlSeconds = 7200): string {
 	if (!secret) {
```
ahmedhamouda78, line 91: **SUGGESTION — Missing guard on empty `scopePrefix`**

`mintConnectToken` guards against an empty `secret` (one line below) but not against an empty or falsy `scopePrefix`. Calling `mintConnectToken('', secret)` produces a token with `channel: "$connect"`, which is almost certainly a programming error and inconsistent with the existing defensive pattern.

Suggested addition immediately before the `!secret` check:
```ts
if (!scopePrefix) {
    throw blocksError(
        RealtimeErrors.ConnectionFailed,
        'Refusing to mint connect token: scopePrefix is empty or missing',
    );
}
```

packages/bb-realtime/src/index.ts
```diff
@@ -147,6 +147,7 @@ export const Realtime: {
 		const fullChannel = `${ns.prefix}/${channel}`;
 		validateChannelPath(fullChannel);
 		const token = mintChannelToken(fullChannel, LOCAL_TOKEN_SECRET);
+		const connectToken = mintConnectToken(this.fullId, LOCAL_TOKEN_SECRET);
```
ahmedhamouda78, line 150: **SUGGESTION — No unit test covers `connectToken` in the local dev `toJSON()` output**

This is the first time `connectToken` appears in the local dev `getChannel` path. The existing tests in `index.test.ts` (e.g. `'getChannel() toJSON includes a valid token'`) check `json.token` but say nothing about `json.connectToken`.

Consider extending an existing test or adding a new one that asserts:
1. `json.connectToken` is present in the descriptor.
2. `validateChannelToken(json.connectToken, LOCAL_TOKEN_SECRET)` returns truthy — valid for connection establishment.
3. `validateChannelToken(json.connectToken, LOCAL_TOKEN_SECRET, json.channel)` returns `null` — the hardening is active in local dev too, not just in the AWS runtime.

packages/bb-realtime/src/security.test.ts
```diff
@@ -101,9 +101,12 @@ describe('Security: mintConnectToken secret guard', () => {
 		const token = mintConnectToken('test-app-rt', 'valid-secret');
 		assert.ok(token);
 		assert.ok(token.includes('.'), 'token should have payload.signature format');
-		// connect token authorizes any sub-channel under the scope prefix
-		const result = validateChannelToken(token, 'valid-secret', 'test-app-rt/events/room-1');
-		assert.ok(result, 'connect token should authorize sub-channels');
+		// connect tokens should not authorize channel subscriptions
```
ahmedhamouda78, line 104: **NITPICK — Inline comments duplicate assertion messages**

Line 104 (`// connect tokens should not authorize channel subscriptions`) and line 107 (`// but it still validates without a requestedChannel (for connection establishment)`) restate exactly what the immediately-following `assert` messages already say. Same pattern repeats at lines 158 and 161. The assertion messages are the documentation; the prose comments add noise without adding information. Recommend removing all four.

packages/bb-realtime/src/token-scope.test.ts
```diff
@@ -0,0 +1,93 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * Token scope isolation tests.
+ *
+ * Validates that connect tokens (scoped to the instance prefix) cannot be
+ * used to authorize channel subscriptions. Channel tokens scoped to a specific
+ * channel should still work as connect tokens since they already prove the
+ * holder has access to something within the instance.
+ */
```
ahmedhamouda78, line 11: **NITPICK — File-level JSDoc block is unnecessary on a test file**

Lines 4–11 are a `/** … */` JSDoc block. Test files are not exported APIs; the `describe` string at line 20 (`'Token scope isolation: connect tokens cannot subscribe to channels'`) already serves as the module-level description. JSDoc on test files is noise. Convert to a brief `//` comment or remove entirely.

ahmedhamouda78 (commented):
**Overall verdict: ✅ Approve with suggestions.**

The security fix is correct, elegant, and backward-compatible. Appending `$connect` to `scopePrefix` before calling `mintChannelToken` is a minimal, surgical change that leverages the existing `startsWith(payload.channel + '/')` guard in `validateChannelToken` — no changes to the validator, subscribe handler, or `$connect` handler are needed. Old connect tokens continue to work for connection establishment during rollout. Test coverage is thorough across unit, regression, and integration levels. No critical issues found.

---

**Two suggestions could not be anchored to diff lines and are included here:**

**SUGGESTION (not in diff) — `packages/bb-realtime/src/aws-middleware.ts` ~line 226 — `isRealtimeDescriptor` does not validate `connectToken` presence**

`AwsRealtimeDescriptor` declares `connectToken: string`, but the `isRealtimeDescriptor` type guard only checks for `wsUrl` and `token`:
```ts
function isRealtimeDescriptor(data: unknown): data is AwsRealtimeDescriptor {
    return ... && typeof (data as any).wsUrl === 'string'
              && typeof (data as any).token === 'string';
    // connectToken NOT checked
}
```
If a descriptor arrives without `connectToken` (unusual during a mixed-version rollout), `getOrCreateConnection` receives `undefined`, and `encodeURIComponent(undefined)` → `"undefined"`, creating a WebSocket URL of `wss://...?token=undefined`. The server rejects this with no clear diagnostic. A defensive fix: add `&& typeof (data as any).connectToken === 'string'` to the guard, which also makes it match the declared `AwsRealtimeDescriptor` shape exactly.

**SUGGESTION (pre-existing, not in diff) — `packages/bb-realtime/src/utils.ts` ~line 114 — `validateChannelToken` permits tokens with no `channel` field to pass channel authorization**

The channel-match guard is:
```ts
if (requestedChannel && payload.channel && requestedChannel !== payload.channel
    && !requestedChannel.startsWith(payload.channel + '/')) return null;
```
If `payload.channel` is absent or falsy, the second condition short-circuits and any token with a valid HMAC + unexpired `exp` passes channel authorization. Not exploitable without the signing secret, but as a defence-in-depth improvement:
```ts
if (requestedChannel) {
    if (!payload.channel) return null;
    if (requestedChannel !== payload.channel && !requestedChannel.startsWith(payload.channel + '/')) return null;
}
```

packages/bb-realtime/src/token-scope.test.ts
```diff
@@ -0,0 +1,93 @@
...
+	it('channel token validates for its own channel', () => {
+		const token = mintChannelToken(CHANNEL, SECRET);
+		const result = validateChannelToken(token, SECRET, CHANNEL);
+		assert.ok(result, 'channel token should validate for its specific channel');
+		assert.strictEqual(result!.channel, CHANNEL);
+	});
+
+	it('connect token MUST NOT validate for a channel subscription', () => {
+		const connectToken = mintConnectToken(INSTANCE_PREFIX, SECRET);
+		// This is the critical assertion: a connect token should NOT authorize
```
ahmedhamouda78, line 30: **NITPICK — Inline comments restate what the test name and assertion message already say**

```ts
// This is the critical assertion: a connect token should NOT authorize
// subscription to a specific channel under the instance prefix.
const result = validateChannelToken(connectToken, SECRET, CHANNEL);
assert.strictEqual(result, null, 'connect token must NOT authorize channel subscriptions');
```

The test name (`'connect token MUST NOT validate for a channel subscription'`) and the assertion message cover this fully. Same pattern at lines 52–53 and line 60. Remove all three inline narration comments.

test-apps/comprehensive/test/realtime.test.ts
```diff
@@ -98,6 +98,65 @@ export function realtimeTests(getApi: () => typeof apiType) {
...
+				const ws = new WebSocket(`${wsUrl}?token=${encodeURIComponent(connectToken)}`);
+
+				const reply = await new Promise<{ type: string; channel?: string; message?: string }>((resolve, reject) => {
+					const timer = globalThis.setTimeout(() => {
+						ws.close();
+						reject(new Error('No reply from server within 5s'));
+					}, 5000);
+
+					ws.onopen = () => {
+						// Attempt to subscribe using the connect token AS the channel token.
```
ahmedhamouda78, line 127: **NITPICK — Step-narration comments throughout the new test**

Lines 105–106, 117, 127–129, and 150 narrate what the surrounding code does step-by-step. For example:
```ts
// Attempt to subscribe using the connect token AS the channel token.
// This should be rejected — connect tokens are scoped to the instance
// prefix ($connect) and must not authorize channel subscriptions.
ws.send(...);
```
The test name (`'connect token cannot be reused as a channel subscription token'`) and the final assertion messages (`'Server should reject subscribe with a connect token used as channel token'`) explain the intent completely. These inline narration comments should be removed.

svidgen:
Thanks for the review, Ahmed! Both suggestions are addressed in 73a010d:

1. **`isRealtimeDescriptor` type guard** — now checks `typeof (data as any).connectToken === 'string'`, matching the declared `AwsRealtimeDescriptor` shape exactly.

2. **`validateChannelToken` channel field invariant** — restructured the guard so that when `requestedChannel` is provided, a missing `payload.channel` immediately returns `null`. Tokens must declare what they authorize. Added a corresponding unit test.

~~The two CI failures (`E2E Hosting (nuxt)` and the rollup `Build and Test`) are pre-existing infra issues — the nuxt sandbox deploy fails on `sandbox-deploy.ts` (orphaned S3 bucket), unrelated to this change.~~

Edit: Agent got ahead of me making that pre-existing infra issue remark. I want tests to pass regardless. 😅

ahmedhamouda78 (approved):

svidgen:
OK. I am going to treat the nuxt issue as a separate issue after all. It fails deployment. this fix needs to go in.

PR #181, opened 2026-07-10
fix(core): reject invalid JSON-RPC params

### LINKED ISSUE #179: JSON-RPC parser accepts primitive params instead of returning InvalidParams

`@aws-blocks/core`

### Describe the bug

`parseRpcRequest()` accepts primitive JSON-RPC `params` values even though JSON-RPC 2.0 requires `params`, when present, to be a structured value (an array or object).

The parser currently uses:

```ts
args: Array.isArray(parsed.params)
  ? parsed.params
  : Object.values(parsed.params ?? {}),
```

As a result:

- `params: "abc"` is accepted as `args: ["a", "b", "c"]`
- `params: 42` is accepted as `args: []`
- `params: true` is accepted as `args: []`
- `params: null` is accepted as `args: []`

This can turn a malformed request from a hand-written HTTP or generated native client into a valid method invocation with unexpected arguments. `RpcErrorCode.InvalidParams` (`-32602`) is already defined, but no parser path currently emits it.

### Expected behavior

- Array `params` remain positional arguments.
- Object `params` remain named arguments, using the existing value-order behavior.
- Omitted `params` continue to mean no arguments.
- Primitive or `null` `params` return a JSON-RPC error response with code `-32602` and preserve the request id.

### Reproduction steps

On current `main` (`a1de3b8`), call:

```ts
parseRpcRequest(JSON.stringify({
  jsonrpc: "2.0",
  method: "api.echo",
  params: "abc",
  id: 1,
}));
```

Actual result:

```json
{
  "ok": true,
  "request": {
    "apiNamespace": "api",
    "method": "echo",
    "args": ["a", "b", "c"],
    "id": 1
  }
}
```

The same request with `params: 42`, `true`, or `null` is also accepted with an empty `args` array.

### Suggested fix

Validate `params` before converting it to arguments. If the property is present, require an array or non-null object; otherwise return `RpcErrorCode.InvalidParams` with a structured `InvalidParams` error name. Add focused unit tests covering accepted array/object/omitted values and rejected primitive/null values.

### Additional context

I searched existing open and closed issues and pull requests for `InvalidParams`, `parseRpcRequest`, and JSON-RPC parameter validation and did not find an existing report or implementation.

ikenyal:
## Problem

JSON-RPC requests with primitive or `null` `params` are currently accepted instead of returning the standard `-32602 Invalid Params` error. The `Object.values(parsed.params ?? {})` fallback turns strings into character arguments and numbers, booleans, and `null` into an empty argument list, so malformed requests can reach application methods with unexpected arguments.

Closes #179.

## Changes

- Validate a present `params` property before constructing method arguments.
- Return `RpcErrorCode.InvalidParams` with `data.name: "InvalidParams"` for primitive and `null` values while preserving the request id.
- Preserve the existing behavior for positional arrays, named objects, and omitted `params`.
- Add focused unit coverage for all accepted and rejected shapes.

## Validation

- `npm ci`
- `npm run build:packages`
- `npm run build -w @aws-blocks/core`
- `node --test --test-concurrency=1 dist/*.test.js dist/**/*.test.js` from `packages/core` (588 passed, 0 failed)
- `npm run lint`
- `npm run lint:deps`
- `npm run check:exports-consistency`
- `npm run check:api`
- `git diff --cached --check`

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (not applicable; this enforces the JSON-RPC 2.0 input shape and does not change the supported array/object/omitted forms)

---


packages/core/src/rpc.ts
```diff
@@ -118,15 +118,31 @@ export function parseRpcRequest(bodyText: string): RpcParseResult {
...
+  if (hasParams) {
+    args = Array.isArray(parsed.params) ? parsed.params : Object.values(parsed.params);
+  }
+
   return {
     ok: true,
     request: {
       apiNamespace: parsed.method.substring(0, dotIndex),
       method: parsed.method.substring(dotIndex + 1),
       // JSON-RPC 2.0 §4.2: params may be an array (positional) or object (named).
```
ahmedhamouda78, line 144: The comment only mentions array (positional) and object (named), but after the refactor `args` is also `[]` when `params` is omitted entirely. Suggest updating so the third path is documented:


```suggestion
      // JSON-RPC 2.0 §4.2: params may be array (positional), object (named), or omitted (empty).
```

packages/core/src/rpc.test.ts
```diff
@@ -71,3 +71,65 @@ describe('-32600 Invalid Request error shape', () => {
...
+      id: 3,
+    }));
+
+    assert.strictEqual(result.ok, true);
+    if (result.ok) {
+      assert.deepStrictEqual(result.request.args, []);
+    }
+  });
+
+  for (const params of ['abc', 42, true, null]) {
```
ahmedhamouda78: `false` is the only boolean not covered here — it's rejected by the same `typeof !== 'object'` guard but is absent from the set. Adding it completes coverage of all four primitive types (string, number, boolean, null):


```suggestion
  for (const params of ['abc', 42, true, false, null]) {
```

packages/core/src/rpc.test.ts
```diff
@@ -71,3 +71,65 @@ describe('-32600 Invalid Request error shape', () => {
...
+        params,
+        id: 'request-1',
+      }));
+
+      assert.strictEqual(result.ok, false);
+      if (!result.ok) {
+        const response = JSON.parse(result.response);
+        assert.strictEqual(response.error.code, RpcErrorCode.InvalidParams);
+        assert.strictEqual(response.error.data.name, 'InvalidParams');
+        assert.strictEqual(response.id, 'request-1');
```
ahmedhamouda78, line 116: These tests assert `error.code` and `error.data.name` but not `error.message`. The existing `-32600` tests assert message content; the user-facing `-32602` message could regress silently. One assertion would close the gap:

`assert.ok(response.error.message.includes('expected an array or object'));`

packages/core/src/rpc.test.ts
```diff
@@ -71,3 +71,65 @@ describe('-32600 Invalid Request error shape', () => {
...
+      id: 1,
+    }));
+
+    assert.strictEqual(result.ok, true);
+    if (result.ok) {
+      assert.deepStrictEqual(result.request.args, ['hello', 42]);
+    }
+  });
+
+  it('accepts named object params using their value order', () => {
```
ahmedhamouda78: nit: "using their value order" leans on an `Object.values()` implementation detail. JSON-RPC doesn't guarantee param ordering for named objects, so a neutral name like `'accepts named object params'` avoids setting a false reader expectation.

ahmedhamouda78 (dismissed):
Approving. This is a clean, spec-correct fix for #179 — the `-32602 Invalid Params` guard is exactly right: `Object.hasOwn` distinguishes absent from present params, and the `params === null || typeof params !== 'object'` check correctly covers all four primitive types (the `typeof null === 'object'` trap is handled by the explicit `=== null` arm). Preserving the request `id` in the error response is spec-required and done correctly. Extracting `let args` from the inline ternary is a nice readability win, and `patch` is the right changeset level.

Left a few non-blocking notes: add `false` to the rejection test loop for full boolean coverage, assert `error.message` in one `-32602` case to guard against silent message regressions, refresh the `args` comment at rpc.ts:130 to mention the omitted-params path, and a naming nit on the named-object test. None block merge.

ikenyal:
@ahmedhamouda78 
Thanks for the review. I’ve incorporated all four suggestions and rebased the branch onto the latest main. All relevant tests and validation checks pass. This PR is ready for re-review.

ahmedhamouda78 (dismissed):

sharonyajain (approved):

sharonyajain:
Approving. Fixes the root cause (the permissive `Object.values(...??{})` fallback) rather than patching a downstream symptom; dropping `??{}` is now safe because the new type guard proves `params` is a non-null object before coercion.

svidgen (approved):

PR #189, opened 2026-07-11
fix(ci): blocks-integrity compares HEAD vs PR base; skip paid bench on CI/doc-only PRs

hfurkanbozkurt:
## What & why

Two `.github/`-only fixes that unblock the `Blocks Package Integrity` / `Build and Test` required checks for **every** PR that doesn't change published packages (including #171), and stop **agent-inert** PRs from burning a paid bench run.

### 1. `blocks-integrity` compared the wrong things (root cause)

The check in `.github/workflows/pr-checks.yml` (added in #186 by @iambipedal) had two defects that made it fail on essentially every PR:

- **Hash-algorithm mismatch:** it compared a local `shasum -a 512` (SHA-512, 128 hex chars) against `npm view … dist.shasum` (npm's `dist.shasum` is **SHA-1**, 40 hex chars). `[ "$LOCAL" = "$PUBLISHED" ]` can never be true for any published version.
- **Wrong baseline (design flaw):** it diffed the freshly-synced package against the **published** tarball. `packages/blocks/docs/` is gitignored and regenerated from sibling READMEs at publish time. Once *any* sibling README changed on `main` without republishing `@aws-blocks/blocks`, every later PR regenerates docs that differ from the frozen published version → fails, regardless of what the PR touches.

**Fix:** pack `@aws-blocks/blocks` at the PR **head** and at the PR **base branch** and compare like-for-like (same hash algorithm, same build state). The check now fails only when a PR *itself* changes the packaged content (e.g. via synced sibling docs) without a changeset bump, and passes for PRs that don't touch it — with **no version bump or release required**. The existing changeset escape-hatch (`"@aws-blocks/blocks": patch`) is preserved.

`npm pack --ignore-scripts` is intentional and now carries an inline note: it packs source + synced docs, **not** a freshly built `dist/`, so the check diffs *shipped content*, not the published build.

### 2. Bench only on agent-relevant changes (paths allowlist)

`.github/workflows/pr-agent-bench.yml` triggered the (paid, Bedrock) bench on every same-repo PR with no path filter. Replaced that with a **`paths:` allowlist** — the bench runs **only** when something the building/grading agent actually consumes changes:

```yaml
paths:
  - 'packages/**'          # the framework the agent builds against (incl. agent-facing packages/**/*.md)
  - 'scripts/agent-bench/**'  # the bench harness itself
  - 'tasks/**'             # task prompts + graded specs
```

Applied to **both** the `pull_request` trigger and the `push: main` trigger (a main run records the merged commit's baseline; doc/CI-only merges don't need a fresh paid baseline — the last code baseline stays valid, and PR-vs-baseline gracefully degrades to "no baseline", never an error).

Everything else is treated as **agent-inert** and skipped: `.github/**` (CI/workflow config), `docs/**` + root prose, `.changeset/**` (release metadata). Note this is an allowlist (`paths:`), not a denylist — the run fires when **at least one** changed file matches a pattern.

> ⚠️ **Caveat — inline cell matrix.** The task↔template matrix (the 10 bench cells) is defined **inline** in `.github/workflows/agent-bench.yml`, which lives under `.github/**` and is **intentionally not allowlisted** (so ordinary CI edits stay `$0`). Consequence: **editing the inline cell definitions** (adding/removing a task, changing a template) **won't auto-trigger the bench** — use **`workflow_dispatch`** to bench such a change. If we later want cell edits to auto-bench, the matrix would need to move into an allowlisted path (e.g. `tasks/**`).

This also makes *this* PR cost `$0`: it changes only `.github/**`, which matches none of the allowlist patterns, so for `pull_request` (GitHub evaluates the workflow file + paths filter from the PR head branch) the bench self-skips on every open/sync.

## Impact
- Unblocks #171 and every other PR that doesn't change `@aws-blocks/blocks` packaged content — no release needed.
- Still correctly requires a changeset when a PR really changes blocks packaged content.
- Bench now runs **only** on `packages/**`, `scripts/agent-bench/**`, or `tasks/**` changes — CI-config, docs, and changeset-only PRs no longer start a paid run.

## Validation (local + live)
- `npm pack` verified path-independent; HEAD-vs-BASE end-to-end run of the fixed step → identical shasum → exit 0 (green).
- Negative case: a simulated sibling-README change → differs from base → correctly fails without a changeset, passes with one.
- `changeset status` + `verify-changeset-coverage.ts` (Node 22) → pass (no changeset required; change is CI-only).
- Allowlist behavior (simulated): a PR touching `packages/**/README.md`, `scripts/agent-bench/**`, or `tasks/**` **would** bench; a PR touching only `.github/**`, `docs/**`, or root `*.md` **skips**.
- **Live: every push to this branch (only `.github/**`) self-skips the bench — confirmed no `PR Agent Bench` run on any push (`$0`).**

Root cause introduced in #186 (cc @iambipedal). No package release required.

### DIFF
--- a/.github/workflows/pr-agent-bench.yml
+++ b/.github/workflows/pr-agent-bench.yml
@@ -23,15 +23,16 @@ on:
   # agent-timeout lands in the run summary, not the check status. Fork PRs get
   # no secrets/OIDC under `pull_request` and are blocked by the same-repo `if`.
   #
-  # paths-ignore skips the (paid) bench when EVERY changed file is CI/docs-only
-  # (a PR touching any real code — packages/**, scripts/agent-bench/**, etc. —
-  # still benches, since paths-ignore only skips when ALL files match).
+  # paths-ignore skips the (paid) bench when EVERY changed file is agent-inert:
+  # .github/** (CI/workflow config) and .changeset/** (release metadata). Docs
+  # and markdown are NOT ignored — the bench grades an agent that reads framework
+  # docs (packages/**/*.md, guides), so a docs change may be made to improve agent
+  # outcomes and should be benchable. A PR touching any real code still benches,
+  # since paths-ignore only skips when ALL files match.
   pull_request:
     types: [opened, synchronize, reopened]
     paths-ignore:
       - '.github/**'
-      - '**/*.md'
-      - 'docs/**'
       - '.changeset/**'
   push:
     branches: [main]
--- a/.github/workflows/pr-agent-bench.yml
+++ b/.github/workflows/pr-agent-bench.yml
@@ -17,25 +17,32 @@ name: PR Agent Bench
 #   S3_BUCKET     variable  Registry bucket (existing).
 
 on:
-  # Runs on same-repo PR open/update AND on every push to main (a main run
-  # records the merged commit's baseline for later PRs to diff against). There
-  # is NO label gate: the bench is green-regardless — a failed cell /
-  # agent-timeout lands in the run summary, not the check status. Fork PRs get
-  # no secrets/OIDC under `pull_request` and are blocked by the same-repo `if`.
+  # Runs on same-repo PR open/update AND on push to main (a main run records the
+  # merged commit's baseline for later PRs to diff against). NO label gate: the
+  # bench is green-regardless — a failed cell / agent-timeout lands in the run
+  # summary, not the check status. Fork PRs get no secrets/OIDC under
+  # `pull_request` and are blocked by the same-repo `if`.
   #
-  # paths-ignore skips the (paid) bench when EVERY changed file is agent-inert:
-  # .github/** (CI/workflow config) and .changeset/** (release metadata). Docs
-  # and markdown are NOT ignored — the bench grades an agent that reads framework
-  # docs (packages/**/*.md, guides), so a docs change may be made to improve agent
-  # outcomes and should be benchable. A PR touching any real code still benches,
-  # since paths-ignore only skips when ALL files match.
+  # ALLOWLIST: the bench only runs when something the building/grading agent
+  # actually consumes changes — the framework (packages/**, incl. agent-facing
+  # packages/**/*.md), the bench harness (scripts/agent-bench/**), or the task
+  # prompts + graded specs (tasks/**). Everything else is skipped as agent-inert:
+  # .github/** (CI/workflow config), docs/** and root prose, .changeset/**.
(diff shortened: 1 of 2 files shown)


sarayev:
Nice cleanup. The old check couldn't ever pass — it compared SHA-512 against npm's SHA-1, and the published-tarball baseline meant any stray README change on main would fail unrelated PRs. Packing head vs base in the same run is the right approach, since both sides use the same npm and stay comparable. I confirmed `fetch-depth: 0` is on the checkout and the same-repo guard is still on the bench, so everything lines up. A couple of small things below, nothing blocking.

**Minor — `packages/**` allowlist skips lockfile-only changes**
Quick check on intent: if a PR only touches `package-lock.json` or the root `package.json` (e.g. a transitive dep bump), the bench won't run, since the allowlist is `packages/**`, `scripts/agent-bench/**`, and `tasks/**`. Those changes usually travel with a `packages/**` edit, so this is likely fine — just want to confirm it's deliberate rather than an accidental gap. If we'd want dep bumps to bench on their own, adding `package-lock.json` would cover it.

**Minor — silenced stderr makes failures harder to debug**
`npm pack ... >/dev/null 2>&1` and `git worktree add ... >/dev/null 2>&1` hide their error output, so if either fails, the job dies later on something like `shasum: cannot open …` instead of the actual npm/git error. Since you already added the "fail loudly on non-unique glob" guard, dropping the `2>&1` on those two lines would keep the real error message visible and make debugging easier.

**Nit — docs get synced twice on the head side**
`sync-block-docs.mjs` runs in the "Sync block docs (prebuild step)" step and again inside `pack_blocks`. It's idempotent, so no functional impact — just flagging in case you'd prefer to drop the extra call. Fine to leave as-is.

**Nit — reproducible-pack assumption worth noting**
The head-vs-base equality relies on npm producing byte-identical tarballs for identical content (fixed-mtime normalization). That holds on modern npm, and both sides run on the same runner, so no concern here — you also validated it live. Just naming the assumption so it's clear why this works if npm's packing behavior ever changes.

sarayev (approved):
Thanks for turning this around so fast. Everything I raised is sorted — the stderr change reads well now (real errors show up without messing with the shasum capture), the allowlist covers the lockfile case on both triggers, and that little note about reproducible packing will save the next person some head-scratching. Still CI-only, nothing else snuck in. Happy to approve — nice work.

ahmedhamouda78 (approved):

PR #190, opened 2026-07-12
fix(core): preserve void RPC responses in verbose mode

### LINKED ISSUE #187: dev-server: void-returning RPC handler returns HTTP 500 in verbose mode (redactToJson(undefined).length)
## Summary
Any `api` handler that returns `void`/`undefined` fails over RPC in the default (verbose) dev server: the client receives a **500 error envelope** even though the handler ran successfully. The error message is `Cannot read properties of undefined (reading 'length')`.

## Root cause (two lines)
1. `packages/core/src/redact.ts` — `redactToJson(value): string` is a type-lie: `JSON.stringify(undefined) === undefined` (doesn't throw), so the `catch` fallback never runs and it returns `undefined` for a void input.
2. `packages/core/src/.../dev-server.ts` (~L1172-1174; built dist `0.27.1` → `dev-server.js:382:83`) — the `[rpc-ok]` logger calls `.length` on that value:
   ```ts
   const okLog = redactToJson(responseBody ?? result); // undefined for a void handler
   console.log('[rpc-ok]', ..., okLog.length > 800 ? okLog.slice(0,800) : okLog); // TypeError
   ```
   The throw lands **before** `res.end(successPayload)`, so the per-request `catch` discards the already-built success and ships a 500.

## Impact
- Every void-returning RPC method (natural for `delete`/`send`/`publish`) returns 500 in verbose dev mode — **after** side effects execute → double-execution / false-failure hazard.
- Client-side surfacing: `decodeRpcResponse` re-throws the embedded `reading 'length'` message.
- Masked by `BLOCKS_DEV_QUIET=1`. Starter template avoids it only because every mutation returns `{ success: true }`.

## Affected versions
Confirmed on `@aws-blocks/core` `0.1.11`, repo source `efdd426`, and **latest public npm `0.27.1`** (still unfixed).

## Repro
Add to the default starter and call over RPC against `npm run dev:server`:
```ts
export const pingVoid = api(async () => {});          // -> HTTP 500 (bug)
export const pingValue = api(async () => ({ ok: true })); // -> 200 OK
```
```
client: api.pingVoid  -> {"error":{"code":500,"message":"Cannot read properties of undefined (reading 'length')",...}}
server: [rpc-err] ... TypeError ... at .../@aws-blocks/core/dist/scripts/dev-server.js:382:83
```

## Suggested fix
`const okLog = redactToJson(responseBody ?? result) ?? 'undefined';` and/or make `redactToJson` return a string for `undefined` input. Add a regression test: a void RPC method returns a JSON-RPC success (not 500) in verbose mode.

ikenyal:
## Problem

In verbose local development mode, an RPC handler that successfully returns `void` is converted into a JSON-RPC error response. `redactToJson(undefined)` returns `undefined` at runtime despite its declared `string` return type, so the success logger throws while reading `.length` before the prepared success response is sent.

This is especially problematic for side-effect-only methods such as delete, send, or publish: the operation completes, but the client receives an error and may retry it.

Closes #187.

## Changes

- Make `redactToJson()` return the string `undefined` when `JSON.stringify()` produces no output, preserving its string return contract.
- Add focused unit coverage for `undefined`, function, and symbol serialization.
- Add an integration test that starts the verbose dev server, invokes a void RPC handler, and verifies a success response and `[rpc-ok]` log.

## Validation

- Reproduced before the fix: the integration test received a JSON-RPC error with code `500` and `Cannot read properties of undefined (reading 'length')`.
- `npm ci`
- `npm run build:packages`
- `npm run build -w @aws-blocks/core`
- `node --test packages/core/dist/redact.test.js packages/core/dist/scripts/dev-server-rpc.test.js` (26 passed, 0 failed)
- `node --test --test-concurrency=1 dist/*.test.js dist/**/*.test.js` from `packages/core` (584 passed, 0 failed)
- `npm run lint`
- `npm run lint:deps`
- `npm run check:exports-consistency`
- `npm run check:api`
- `git diff origin/main...HEAD --check`

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (not applicable; this fixes an internal logging failure without changing the public API)

---

### DIFF
--- a/packages/core/src/redact.test.ts
+++ b/packages/core/src/redact.test.ts
@@ -205,6 +205,14 @@ describe('redactToJson', () => {
     assert.equal(redactToJson(undefined), 'undefined');
   });
 
+  it('returns a string for function values', () => {
+    assert.equal(redactToJson(() => {}), 'undefined');
+  });
+
+  it('returns a string for symbol values', () => {
+    assert.equal(redactToJson(Symbol('x')), 'undefined');
+  });
+
   it('does not leak secrets even when truncation would apply downstream', () => {
     const json = redactToJson([{ action: 'signIn', username: 'u', password: 'p'.repeat(50) }]);
     assert.ok(!json.includes('pppp'));
--- a/packages/core/src/redact.test.ts
+++ b/packages/core/src/redact.test.ts
@@ -201,6 +201,10 @@ describe('redactToJson', () => {
     assert.equal(redactToJson({ big: 10n }), '[unserializable]');
   });
 
+  it('returns a string for undefined', () => {
+    assert.equal(redactToJson(undefined), 'undefined');
+  });
+
   it('does not leak secrets even when truncation would apply downstream', () => {
     const json = redactToJson([{ action: 'signIn', username: 'u', password: 'p'.repeat(50) }]);
     assert.ok(!json.includes('pppp'));
--- a/packages/core/src/scripts/dev-server-rpc.test.ts
+++ b/packages/core/src/scripts/dev-server-rpc.test.ts
@@ -1,0 +1,111 @@
+import { afterEach, describe, it } from 'node:test';
+import assert from 'node:assert';
+import { spawn, type ChildProcess } from 'node:child_process';
+import { createServer } from 'node:http';
+import { mkdirSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+async function getAvailablePort(): Promise<number> {
+  const server = createServer();
+  await new Promise<void>((resolve, reject) => {
+    server.once('error', reject);
+    server.listen(0, '127.0.0.1', resolve);
+  });
+  const address = server.address();
+  assert.ok(address && typeof address === 'object');
+  const { port } = address;
+  await new Promise<void>((resolve, reject) => server.close(error => error ? reject(error) : resolve()));
+  return port;
+}
+
+describe('dev-server RPC integration', () => {
+  let devProcess: ChildProcess | null = null;
+  let tempDir: string | null = null;
+
+  afterEach(async () => {
+    if (devProcess && devProcess.exitCode === null) {
+      devProcess.kill('SIGTERM');
+      await new Promise<void>((resolve) => {
+        const timeout = setTimeout(() => {
+          devProcess?.kill('SIGKILL');
+          resolve();
+        }, 2_000);
+        devProcess?.once('exit', () => {
+          clearTimeout(timeout);
+          resolve();
+        });
+      });
+    }
+    if (tempDir) rmSync(tempDir, { recursive: true, force: true });
+  });
+
+  it('returns success for a void handler in verbose mode', async () => {
+    const port = await getAvailablePort();
+    tempDir = join(tmpdir(), `dev-rpc-test-${process.pid}-${Date.now()}`);
+    mkdirSync(tempDir, { recursive: true });
+    writeFileSync(join(tempDir, 'backend.ts'), `
+export const testApi = {
+  pingVoid: async () => undefined,
+};
+`);
+    writeFileSync(join(tempDir, 'preload.mjs'), `
+if (!process.loadEnvFile) {
+  process.loadEnvFile = () => { throw Object.assign(new Error('ENOENT'), { code: 'ENOENT' }); };
+}
+`);
+    writeFileSync(join(tempDir, 'run-dev.ts'), `
+import { startDevServer } from '${join(__dirname, 'dev-server.js').replace(/\\/g, '/')}';
+startDevServer({ backendPath: '${join(tempDir, 'backend.ts').replace(/\\/g, '/')}', port: ${port} });
+`);
+
+    const tsxBin = join(__dirname, '..', '..', '..', '..', 'node_modules', '.bin', 'tsx');
+    devProcess = spawn(tsxBin, ['--import', join(tempDir, 'preload.mjs'), join(tempDir, 'run-dev.ts')], {
+      cwd: tempDir,
+      env: {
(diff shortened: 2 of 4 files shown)


packages/core/src/redact.test.ts
```diff
@@ -201,6 +201,10 @@ describe('redactToJson', () => {
     assert.equal(redactToJson({ big: 10n }), '[unserializable]');
   });
 
+  it('returns a string for undefined', () => {
+    assert.equal(redactToJson(undefined), 'undefined');
```
ahmedhamouda78, line 205: [suggestion] The doc comment now promises `'undefined'` "when JSON serialization produces no output" — broader than the single tested `undefined` case. `function` and `symbol` inputs hit the same path (`JSON.stringify` → `undefined`) and are already handled but unasserted. Free to lock in:

```ts
it('returns a string for function values', () => {
  assert.equal(redactToJson(() => {}), 'undefined');
});
it('returns a string for symbol values', () => {
  assert.equal(redactToJson(Symbol('x')), 'undefined');
});
```

packages/core/src/scripts/dev-server-rpc.test.ts
```diff
@@ -0,0 +1,111 @@
...
+async function getAvailablePort(): Promise<number> {
+  const server = createServer();
+  await new Promise<void>((resolve, reject) => {
+    server.once('error', reject);
+    server.listen(0, '127.0.0.1', resolve);
+  });
+  const address = server.address();
+  assert.ok(address && typeof address === 'object');
+  const { port } = address;
+  await new Promise<void>((resolve, reject) => server.close(error => error ? reject(error) : resolve()));
```
ahmedhamouda78, line 23: [suggestion] TOCTOU race in `getAvailablePort()`: between `server.close()` and the dev server's `listen(port)`, the OS frees the port, so a parallel test worker can grab it → flaky CI. Clean fix is `port: 0` + read the assigned port from stdout (API change, out of scope). Otherwise, document the window:

```ts
// TOCTOU: brief window between closing this probe and the dev server binding
// the port; fine locally, can flake under heavily-parallelised CI workers.
```

packages/core/src/scripts/dev-server-rpc.test.ts
```diff
@@ -0,0 +1,111 @@
...
+startDevServer({ backendPath: '${join(tempDir, 'backend.ts').replace(/\\/g, '/')}', port: ${port} });
+`);
+
+    const tsxBin = join(__dirname, '..', '..', '..', '..', 'node_modules', '.bin', 'tsx');
+    devProcess = spawn(tsxBin, ['--import', join(tempDir, 'preload.mjs'), join(tempDir, 'run-dev.ts')], {
+      cwd: tempDir,
+      env: {
+        ...process.env,
+        AWS_BLOCKS_DISABLE_TELEMETRY: '1',
+        BLOCKS_DEV_QUIET: '',
```
ahmedhamouda78, line 75: [nit] `BLOCKS_DEV_QUIET: ''` is falsy → keeps verbose logging on (needed for `[rpc-ok]`) and overrides any inherited `=1` from the `...process.env` spread. Correct but non-obvious — worth a comment:

```ts
// '' is falsy → verbose stays on even if parent env sets BLOCKS_DEV_QUIET=1.
BLOCKS_DEV_QUIET: '',
```

packages/core/src/scripts/dev-server-rpc.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    tempDir = join(tmpdir(), `dev-rpc-test-${process.pid}-${Date.now()}`);
+    mkdirSync(tempDir, { recursive: true });
+    writeFileSync(join(tempDir, 'backend.ts'), `
+export const testApi = {
+  pingVoid: async () => undefined,
+};
+`);
+    writeFileSync(join(tempDir, 'preload.mjs'), `
+if (!process.loadEnvFile) {
+  process.loadEnvFile = () => { throw Object.assign(new Error('ENOENT'), { code: 'ENOENT' }); };
```
ahmedhamouda78, line 60: [nit] Document the `preload.mjs` shim: `process.loadEnvFile` is Node ≥20.6.0; the dev server calls it and treats `ENOENT` as "no .env". Without the polyfill, older Node throws `TypeError: process.loadEnvFile is not a function` before startup.

```ts
// Polyfill process.loadEnvFile for Node < 20.6.0; ENOENT = "no .env file".
```

ahmedhamouda78 (dismissed):
Correct, minimal fix at the right layer — repairing the `string` contract inside `redactToJson` covers all call sites, so the `.length` crash can't recur. `??` over `||` is precise, `'undefined'` is the right sentinel (log-only callers, never re-parsed), and the integration test genuinely fails without the fix. Approving.

Non-blocking follow-ups tagged inline.

packages/core/src/redact.test.ts
ikenyal: Added focused coverage for both function and symbol values, locking in the broader no-output serialization contract.

packages/core/src/scripts/dev-server-rpc.test.ts
ikenyal: Documented the TOCTOU window. Eliminating it would require changing startDevServer to support port 0 and surface the assigned port through its configuration and logging lifecycle, which would expand production code solely for test infrastructure. I kept this fix focused on #187; a follow-up issue would make sense if CI shows this race in practice.

packages/core/src/scripts/dev-server-rpc.test.ts
ikenyal: Added a comment explaining that the empty string is intentionally falsy and keeps verbose logging enabled even when the parent environment sets quiet mode.

packages/core/src/scripts/dev-server-rpc.test.ts
ikenyal: Documented the  compatibility shim and the  behavior.

ahmedhamouda78 (approved):

svidgen (approved):

PR #191, opened 2026-07-12
feat(bench): reshape tasks to api-first framework-focus grading

hfurkanbozkurt:
## What & why

Reshapes all 10 bench tasks to be **API-first** ("just use the framework" — Ask E). Core grading now runs through an `api` namespace (JSON-RPC via `POST /aws-blocks/api`) backed by **real framework blocks**, with the DOM reduced to a light smoke test (mount renders + primary control works). This follows the `tasks/observability-api` gold-standard pattern: heavy `request`→JSON-RPC assertions, light DOM.

**Task content only** — no harness, scoring, workflow, or model changes. Only `tasks/**` (PROMPT.md + test.spec.ts) are touched.

## Per-task changes (8 reshaped)

| Task | Blocks | api surface (new core grading) | DOM (reduced to smoke) |
|---|---|---|---|
| `auth-notes` | AuthBasic + KVStore | `saveNote`/`getNote` gated by session cookie; overwrite, verbatim/no-HTML, per-user isolation (2 contexts), unauth gating | sign up → editor → save → reload persists |
| `async-word-counter` | AsyncJob + KVStore | `enqueue`/`getJob`/`listJobs`; whitespace/unicode counts, keyed-by-id dedupe, unknown-id error, persistence | input → submit → row resolves done |
| `file-gallery` | FileBucket | `putFile`/`listFiles`/`deleteFile`/`getDownloadUrl`; byte-exact size, real download URL bytes, binary/empty/unicode names, overwrite | file input + list |
| `sql-kb-catalog` | Database + KnowledgeBase | `addProduct`/`listProducts` (id-ordered), `searchKb`; seeded return/refund FAQ | product input + kb query |
| `oidc-dsql-notes` ⚠️ | AuthOIDC + DistributedDatabase | `addNote`/`listNotes` gated by OIDC session; verbatim/SQL-injection-safe, oldest-first ordering, unauth gating | OIDC sign-in smoke |
| `cognito-profile` ⚠️ | AuthCognito (email OTP) | `whoami` gated by session; identity, gating, post-signout gating; `getLastCode` test hook | OTP flow → profile |
| `kb-chat-agent` | Agent (Bedrock Sonnet 4.6) + KnowledgeBase | `ask` → `{reply,toolsUsed,citations}`; real retrieval of seeded facts, deterministic tool output, tool routing, no-fabrication, blank rejected | chat input → transcript + tool/citation |
| `collab-presence-board` ⚠️ | Realtime + DistributedTable | `join`/`listPresent`; name-keyed dedupe, verbatim/unicode round-trip, blank rejected, multi-visitor persistence | disabled control, XSS-as-text, reload restores, **1 realtime 2-context fan-out** |

**Left as-is** (already API-first): `observability-api` (gold standard), `email-digest` (API-first hybrid).

## Flagged — not fully framework-focusable

- ⚠️ **`oidc-dsql-notes`, `cognito-profile`** — MODERATE. Auth flows (OIDC redirect, email-OTP) are intrinsically multi-view DOM; core data ops moved to `api`, but sign-in stays a DOM smoke. OIDC stub IdP is single-subject, so cross-user isolation isn't demonstrable → tests auth **gating** instead.
- ⚠️ **`collab-presence-board`** — PARTIALLY-DOM. Realtime fan-out genuinely needs two browser contexts; that ONE check stays DOM. Everything else (roster persistence/dedupe/verbatim/validation) moved to `api`.

## Validation

- `tsc --noEmit` (Playwright types) on **all 10** specs → clean.
- Deep E2E on `async-word-counter`: real reference solution (AsyncJob+KVStore) scaffolded from the `bare` template, run against the dev server → **12/12 pass** (10 api-first core + 2 DOM smoke).
- Changeset: `tasks/**` is not a publishable package → `changesets status` and `verify-changeset-coverage.ts` both pass with no changeset.

Do not merge — opening for review.


tasks/cognito-profile/PROMPT.md
```diff
@@ -2,37 +2,38 @@
...
 
-The grader has no mailbox, so it reads the OTP over JSON-RPC. **Expose an `api` namespace method `getLastCode()`** that returns the most recently delivered code as `{ username, code }` (or `null`). The grader retrieves it by POSTing to `/aws-blocks/api`:
+Expose an **`api` namespace** (the framework's server-side RPC surface — `POST /aws-blocks/api`) with:
 
-```
-POST /aws-blocks/api
-Content-Type: application/json
-{ "jsonrpc": "2.0", "method": "api.getLastCode", "params": [], "id": 1 }
-```
+1. **`api.getLastCode()`** — test harness hook (the grader has no mailbox): returns the most recently delivered code as `{ username, code }` (or `null`). Retrieved via:
```
soberm: [high][security] `api.getLastCode()` sits in the same `api` namespace as `whoami()` with no test-only guard. An AI agent following this PROMPT could faithfully ship it to production and expose live OTP codes over an unauthenticated JSON-RPC endpoint. The PROMPT should require it to be gated — e.g., annotated `@blocksSkipCodegen` and returning `null` when `process.env.BLOCKS_MOCK !== 'true'`. A matching test branch asserting it returns `null` in non-mock mode would close the gap.

tasks/async-word-counter/test.spec.ts
```diff
@@ -1,297 +1,177 @@
...
+import { test, expect, type Page, type APIRequestContext } from '@playwright/test';
 
 const BASE = process.env.BLOCKS_URL ?? 'http://localhost:3000';
 const APPEAR = 8_000;
-const DONE = 20_000; // background job + frontend poll interval
+const DONE = 20_000; // background job + poll interval
 
 const RUN = process.env.RUN_ID || String(Date.now());
 let seq = 0;
 const uniq = (base: string) => `${base}-${RUN}-${++seq}-${Date.now()}`;
```
soberm: [high][correctness + safety] Two separate issues with `seq`:

**Counter collision.** `seq` is incremented by both `rpc()` (as the JSON-RPC request `id`) and `uniq()` (as the uniqueness seed for test-data strings). Generated names therefore change depending on how many RPC calls ran before them — non-deterministic test data and confusing log correlation. These should be independent counters:

```ts
let rpcSeq = 0;   // JSON-RPC request id only
let uniqSeq = 0;  // test-data seed only
const uniq = (base: string) => `${base}-${RUN}-${++uniqSeq}-${Date.now()}`;
```

**Hidden parallelism contract.** A module-level `let seq = 0` is only race-free at `workers:1 / fullyParallel:false`. That constraint is invisible — bumping `workers` in the Playwright config silently corrupts test data. Worth at least a `// HARNESS CONTRACT: requires workers:1` banner at the top of each file.

Both issues apply to all 8 spec files.

tasks/collab-presence-board/test.spec.ts
```diff
@@ -7,118 +7,111 @@ const RUN = process.env.RUN_ID || String(Date.now());
...
-		await expect(page.getByTestId('join-btn')).toBeVisible();
+async function roster(ctx: APIRequestContext): Promise<string[]> {
+	const { body } = await rpc(ctx, 'api.listPresent', []);
+	const rows = Array.isArray(body?.result) ? body.result : [];
+	return rows.map((r: any) => String(r?.name ?? r));
+}
+// The shared board holds every name every test joins, so scope roster reads to
+// this call's own unique names rather than asserting on absolute counts. The
+// `count(...) === 1` assertions below are race-free only because the harness
+// runs specs serially (fullyParallel:false / workers:1 in
```
soberm, line 38: [suggestion] The single-worker caveat is buried inside the `roster` helper's comment block, interleaved with implementation notes about the shared board. This is the kind of constraint that breaks silently in CI when someone bumps `workers` — it deserves to be a top-level `// HARNESS CONTRACT:` banner at the start of the `test.describe` block, not hidden mid-file.

tasks/file-gallery/test.spec.ts
```diff
@@ -7,339 +7,202 @@ const RUN = process.env.RUN_ID || String(Date.now());
...
-		const errors = watchErrors(page);
-		await page.goto(BASE);
-
-		const uploadBtn = page.getByTestId('file-upload');
-		await expect(uploadBtn).toBeVisible();
-
-		// Baseline count (the bucket persists across tests in this run).
-		const before = await page.getByTestId('file-item').count();
+	test('putFile validates its input — missing name or bad base64 is rejected', async ({ request }) => {
+		const noName = await rpc(request, 'api.putFile', [toB64(Buffer.from('x'))]);
```
soberm, line 145: [minor] This catches arity errors (too few params) and bad base64, but not a present-but-empty `name` like `['', toB64(Buffer.from('x'))]`. The PROMPT explicitly calls out a missing name as a rejection case — a blank string is a far more realistic implementation mistake than omitting the param entirely. Worth adding:

```ts
const blankName = await rpc(request, 'api.putFile', ['', toB64(Buffer.from('x'))]);
expect(blankName.body?.error, 'a blank name must yield a JSON-RPC error envelope').toBeTruthy();
```

tasks/auth-notes/test.spec.ts
```diff
@@ -1,24 +1,39 @@
...
+// Per-test no-error gate: ONLY uncaught page errors (4xx/JSON-RPC errors are expected).
 function watchErrors(page: Page, sink: string[] = []): string[] {
 	page.on('pageerror', (err) => sink.push(String(err)));
 	return sink;
 }
 
+// JSON-RPC call over the framework's api endpoint. `ctx` decides identity:
+// page.request carries the page's sign-in cookie (authenticated user); the
+// top-level `request` fixture has no cookies (unauthenticated).
+async function rpc(
```
soberm, line 21: [nit] The `rpc` helper (and `watchErrors`) is copy-pasted nearly verbatim across all 8 spec files. That's a defensible call for self-contained test isolation, but if the wrapper ever needs changing — say, adding a tracing header or adjusting error coercion — it's an 8-file edit. Worth a brief note somewhere canonical (e.g. `tasks/README.md` or the gold-standard `observability-api` spec) documenting that the duplication is intentional, so the next person doesn't "fix" it by extracting a shared helper that breaks isolation.

tasks/collab-presence-board/test.spec.ts
```diff
@@ -129,85 +122,43 @@ test.describe('collab-presence-board', () => {
...
+		// the stored roster must be fetched and rendered on load.
 		await expect(presence(page, name)).toHaveCount(1, { timeout: SYNC });
 
 		expect(errors, `page errors: ${errors.join(' | ')}`).toEqual([]);
 	});
 
-	test('a tab opened AFTER others have joined loads the existing roster on first paint', async ({ browser }) => {
+	// --- Realtime fan-out: intrinsically two browser contexts (PARTIALLY-DOM) ---
+
+	test('a join in one tab appears in another already-open tab in real time', async ({ browser }) => {
```
soberm, line 182: [question] The old suite had a dedicated "late-joiner" test: a brand-new tab opened *after* others had joined must render the existing roster on first paint (a realtime-only impl would show a blank board). That test is gone. The coverage looks like it shifted to `api.listPresent restores every visitor` + the `reload` DOM smoke — but those don't exercise first-paint in a fresh tab. The API test proves `listPresent` returns the data; it doesn't prove the page actually *calls* `listPresent` on mount. Is the late-joiner scenario intentionally dropped, or should there be a DOM smoke for it?

tasks/kb-chat-agent/test.spec.ts
```diff
@@ -1,251 +1,162 @@
...
-		// constant, so the exact string must appear in the answer regardless of how
-		// the (non-deterministic) model phrases the rest of its reply.
-		await ask(page, 'Can you look up my order status and tracking code?');
-
-		await expect.poll(() => bubbleWith(page, 'TRK-9F42-OK').count(), { timeout: REPLY }).toBeGreaterThan(0);
-		await expect(bubbleWith(page, 'TRK-9F42-OK').first()).toHaveAttribute('data-role', 'assistant');
-		await expect(page.getByTestId('tool-indicator').first()).toBeVisible({ timeout: REPLY });
-
-		expect(errors, `page errors: ${errors.join(' | ')}`).toEqual([]);
+	test('small talk retrieves nothing and does not fabricate the seeded facts', async ({ request }) => {
```
soberm, line 81: [minor] If `api.ask` returns an error envelope (e.g. the mock agent times out or the method isn't wired yet), `ask()` coerces `reply` to `''` via `String(r.reply ?? '')`. The absence assertions on seeded facts (`expect(a.reply).not.toContain(...)`) then pass vacuously — an unimplemented endpoint looks identical to a correctly non-fabricating one. Adding `expect(a.body?.error).toBeFalsy()` before the content checks would catch that.

tasks/collab-presence-board/test.spec.ts
```diff
@@ -7,118 +7,111 @@ const RUN = process.env.RUN_ID || String(Date.now());
...
-
-		const name = uniq('user');
-		await join(page, name);
-		await expect(presence(page, name)).toHaveCount(1, { timeout: SYNC });
-
-		await page.reload();
-		await expect(presence(page, name)).toHaveCount(1, { timeout: SYNC });
-
-		expect(errors, `page errors: ${errors.join(' | ')}`).toEqual([]);
+	test('names are stored and read back verbatim (markup preserved as literal text)', async ({ request }) => {
```
soberm: [nit] The test title says "markup preserved as literal text" but the assertion only checks `names.toContain(name)` — it's verifying verbatim *storage* roundtrip, not DOM rendering safety (no `<b>` element check). The DOM rendering safety is covered in the page smoke at line 125, which is the right place for it. Renaming this one to something like "names round-trip verbatim through the store" would make the layering clearer.

tasks/oidc-dsql-notes/test.spec.ts
```diff
@@ -32,242 +42,203 @@ async function signIn(page: Page): Promise<void> {
...
 
-// DOM-order index of each token within the full note list. Tokens are unique
-// per run, so each matches exactly one row, and the RELATIVE order of a test's
-// own tokens is stable even when other notes (the shared stub user) interleave.
-async function indicesOf(page: Page, tokens: string[]): Promise<number[]> {
-	const texts = await page.getByTestId('note-item').allInnerTexts();
-	return tokens.map((tok) => texts.findIndex((t) => t.includes(tok)));
-}
+// Relative order of tokens within the (shared stub user) note list.
+const indicesOf = (rows: any[], tokens: string[]) =>
```
soberm, line 53: [nit] `indicesOf` is a two-line helper used in exactly one test. Could just inline it at the call site:

```ts
const [i1, i2, i3] = [t1, t2, t3].map(tok =>
  rows.findIndex(r => String(r?.text ?? '').includes(tok))
);
```

tasks/auth-notes/PROMPT.md
```diff
@@ -2,26 +2,34 @@
...
-4. **Persistence:** after a full page reload the visitor is still signed in, the textarea is **pre-filled** with their saved note (not blank), and the display shows it. Notes are stored and rendered **verbatim** — exactly the characters saved, with no HTML interpretation (a note containing markup such as `<b>x</b>` is shown as literal text, not a bold element). Notes up to a few thousand characters round-trip unchanged.
-5. **Per-user isolation:** a different user who signs in sees their own note (empty until they save one), never another user's note.
-6. A signed-in visitor can sign out, which returns them to the signed-out form.
+### The `api` namespace (primary surface)
+
+Expose an **`api` namespace** (the framework's server-side RPC surface — a `POST /aws-blocks/api` JSON-RPC endpoint) with these methods. Each note operation must run **on the server** and persist through the **key/value block** under the per-user key `note:{username}`. Both methods derive the current user from the **auth session** (the sign-in cookie) — the note key is **never** taken from a client-supplied argument.
```
soberm: [nit] The key format `note:{username}` is clear, but the PROMPT doesn't say whether the username should be stored raw or encoded. If a username contains characters that are meaningful in a KV key path (e.g. `/`, `#`, `%`), different implementations will make different choices and end up with incompatible keys. Even a brief note — "store the username verbatim, no encoding needed" — removes the ambiguity.

tasks/file-gallery/test.spec.ts
```diff
@@ -7,339 +7,202 @@ const RUN = process.env.RUN_ID || String(Date.now());
...
+	return { status: res.status(), body: await res.json().catch(() => null) };
 }
 
-test.describe('file-gallery', () => {
-	test('shows the upload input and the file list', async ({ page }) => {
-		const errors = watchErrors(page);
-		await page.goto(BASE);
+const toB64 = (buf: Buffer) => buf.toString('base64');
+// Download URLs may be absolute (presigned) or root-relative (served locally).
+const absUrl = (u: string) => (/^https?:\/\//.test(u) ? u : `${BASE}${u.startsWith('/') ? '' : '/'}${u}`);
```
soberm: [nit] `absUrl` handles the common cases well, but if `BLOCKS_URL` is set with a trailing slash (e.g. `http://localhost:3000/`) and `u` is already root-relative (e.g. `/files/foo`), you get a double-slash: `http://localhost:3000//files/foo`. Most servers tolerate it, but it's easy to guard:

```ts
const absUrl = (u: string) =>
  /^https?:\/\//.test(u) ? u : `${BASE.replace(/\/$/, '')}/${u.replace(/^\//, '')}`;
```

soberm (commented):
Good structural refactor — moving the heavy grading to JSON-RPC API calls is much more reliable than polling the DOM, and the per-task split between "Framework surface" and "Page smoke" is clear and consistent across all 8 files. A few things worth addressing before this lands.

tasks/cognito-profile/test.spec.ts
```diff
@@ -59,280 +65,181 @@ async function signIn(page: Page, request: APIRequestContext, user: string): Pro
...
+		// with the flag unset it must be gated off and return null.
+		const { status, body } = await rpc(request, 'api.getLastCode', []);
+		expect(status, `unexpected HTTP ${status}`).toBeLessThan(500);
+		if (process.env.BLOCKS_MOCK === 'true') {
+			// Live in mock mode: null (nothing delivered yet) or a { username, code }
+			// record — but never a JSON-RPC error envelope.
+			expect(body?.error, `getLastCode must not error in mock mode: ${JSON.stringify(body?.error)}`).toBeFalsy();
+		} else {
+			// Production / non-mock: the backdoor is closed — no code is exposed.
+			expect(body?.result ?? null, 'getLastCode must be gated off (null) outside BLOCKS_MOCK').toBeNull();
```
ahmedhamouda78: [major] `3-build-and-test.sh` exports `BLOCKS_MOCK=true` unconditionally, so the `getLastCode` test's `else` branch — the only check that the OTP backdoor is closed in production — never runs. An impl that forgets the gate passes.

Fix: add a harness pass with `BLOCKS_MOCK` unset, or document that the gate is PROMPT-only and unverified by the grader.

tasks/auth-notes/test.spec.ts
```diff
@@ -86,138 +83,158 @@ test.describe('auth-notes', () => {
...
-		await page.getByTestId('note-save').click();
-		await expect(page.getByTestId('note-display')).toHaveText(raw, { timeout: T });
-		// An innerHTML impl would create a real <b>; a text impl has none.
-		await expect(page.getByTestId('note-display').locator('b')).toHaveCount(0);
 
-		await page.reload();
-		await expect(page.getByTestId('note-display')).toHaveText(raw, { timeout: T });
+		const raw = `<b>${uniq('x')}</b> & <i>plain</i>`;
+		await rpc(page.request, 'api.saveNote', [raw]);
+		// Exact string equality: no HTML escaping (&amp;) and no stripping.
```
ahmedhamouda78: [major] This removes the DOM literal-text check:

    await expect(page.getByTestId('note-display').locator('b')).toHaveCount(0);

The replacement only checks API string equality, so `el.innerHTML = note` would pass. Sibling tasks (`collab-presence-board`, `oidc-dsql-notes`) kept their DOM `toHaveCount(0)` smoke. Recommend restoring one here.

tasks/async-word-counter/test.spec.ts
```diff
@@ -1,297 +1,179 @@
...
-
-		const row = rowFor(page, token);
-		await expect(row).toHaveAttribute('data-status', 'done', { timeout: DONE });
-		await expect(row.getByTestId('wc-result')).toHaveText(/^\s*3\s*$/, { timeout: DONE });
-
-		expect(errors, `page errors: ${errors.join(' | ')}`).toEqual([]);
-	});
-
-	test('a single token counts as exactly one word', async ({ page }) => {
+	test('submitting through the page resolves to done and persists across reload', async ({ page }) => {
```
ahmedhamouda78, line 195: [major] The "reloaded while still processing" test is gone — the new tests wait for `done` before listing/reloading. The PROMPT still requires enqueue-time persistence with status `processing`, so an impl that only persists on completion now passes. Restore the test, or drop the requirement from the PROMPT.

tasks/async-word-counter/test.spec.ts
```diff
@@ -1,297 +1,179 @@
...
+		expect(await countOf(request, idB)).toBe(4);
+	});
 
-		expect(errors, `page errors: ${errors.join(' | ')}`).toEqual([]);
+	test('listJobs restores every enqueued job from the store', async ({ request }) => {
+		const text = `${uniq('listed')} a b`; // 3 tokens
+		const id = await enqueue(request, text);
+		await countOf(request, id);
+
+		// A fresh caller (new request context has no client memory) must still
```
ahmedhamouda78: [nit] The comment claims a "fresh caller (new request context)" but reuses the same `request` fixture. The assertion is fine; the comment isn't.

tasks/cognito-profile/PROMPT.md
hfurkanbozkurt: Addressed in d36b12a — the PROMPT now requires `getLastCode` be `@blocksSkipCodegen` and return `null` unless `BLOCKS_MOCK==='true'`, and there's a gating test. (Follow-up in e93ad9b on the grader side, see thread #12 below.)

tasks/async-word-counter/test.spec.ts
hfurkanbozkurt: Fixed across all 8 specs in d36b12a — split into independent `rpcSeq` (JSON-RPC id) and `uniqSeq` (test-data seed), and added a top-level `// HARNESS CONTRACT: requires workers:1` banner to each.

tasks/collab-presence-board/test.spec.ts
hfurkanbozkurt: Done in d36b12a — hoisted it to the `HARNESS CONTRACT: requires workers:1` banner at the top of the describe block (consistent with the other specs).

tasks/file-gallery/test.spec.ts
hfurkanbozkurt: Added in d36b12a — there's now a blank-name (`''`) rejection case asserting the error envelope and that no phantom entry leaks in.

tasks/auth-notes/test.spec.ts
hfurkanbozkurt: Done in d36b12a — added a "Intentional per-spec helper duplication" section to `tasks/README.md` explaining each spec must be a self-contained grader, so nobody extracts a shared helper that breaks isolation.

tasks/collab-presence-board/test.spec.ts
hfurkanbozkurt: Restored in d36b12a — re-added a DOM test ("a freshly-opened tab paints the pre-existing roster on mount") that opens a new tab after others joined and asserts first-paint from `listPresent`, alongside the API check.

tasks/kb-chat-agent/test.spec.ts
hfurkanbozkurt: Fixed in d36b12a — added `expect(a.body?.error).toBeFalsy()` before the content/absence checks so an unimplemented or timed-out endpoint can't pass vacuously.

tasks/collab-presence-board/test.spec.ts
hfurkanbozkurt: Done in d36b12a — renamed to "names round-trip verbatim through the store"; the DOM rendering-safety check stays in the page smoke where it belongs.

tasks/oidc-dsql-notes/test.spec.ts
hfurkanbozkurt: Leaving it — it's actually called at two sites within that ordering test (the initial list and the fresh re-query), so the named helper keeps both call sites readable rather than duplicating the 3-token `findIndex` map. Declining this one.

tasks/auth-notes/PROMPT.md
hfurkanbozkurt: Clarified in d36b12a — the PROMPT now says to store the username verbatim in the key (no URL-encoding, hashing, or case-folding).

tasks/file-gallery/test.spec.ts
hfurkanbozkurt: Fixed in d36b12a — `absUrl` now strips a trailing slash from the base and a leading slash from the path (exactly your suggested guard).

tasks/cognito-profile/test.spec.ts
hfurkanbozkurt: You're right, the `else` was dead since the harness always sets `BLOCKS_MOCK=true`. Removed it in e93ad9b and documented that the production gate is prompt-enforced + codegen-skipped but grader-unverified, rather than leave a misleading assertion that can never execute; kept the live mock-mode check.

tasks/auth-notes/test.spec.ts
hfurkanbozkurt: Restored in e93ad9b — the test now saves `<b>…</b> & <i>…</i>` through the page and asserts `note-display` renders literal text (`.locator('b')`/`('i')` `toHaveCount(0)`), keeping the API round-trip equality too.

tasks/async-word-counter/test.spec.ts
hfurkanbozkurt: Restored in e93ad9b — added a test that reads the job back the instant it's enqueued and asserts `status:'processing'` (via `getJob` and `listJobs`) before it resolves to `done`, matching the enqueue-time-persistence requirement.

tasks/async-word-counter/test.spec.ts
hfurkanbozkurt: Fixed in e93ad9b — reworded to say it reuses the `request` fixture (a raw API client, not a new context), so seeing the job via `listJobs` proves server-side persistence, not client memory.

soberm (dismissed):

ahmedhamouda78 (dismissed):

soberm (dismissed):

tasks/async-word-counter/test.spec.ts
```diff
@@ -1,297 +1,214 @@
...
+		}
 
-		await page.reload();
-		await expect(rowFor(page, phrase)).toHaveAttribute('data-status', 'done', { timeout: DONE });
-		await expect(rowFor(page, phrase).getByTestId('wc-result')).toHaveText(/^\s*6\s*$/, { timeout: DONE });
+		// listJobs must ALSO surface the still-processing job (restored from the store, PROMPT line 22).
+		const listed = (await rpc(request, 'api.listJobs', [])).body?.result ?? [];
+		const mine = Array.isArray(listed) ? listed.find((j: any) => j?.id === id) : null;
+		expect(mine, 'listJobs must include the still-processing job (not only done ones)').toBeTruthy();
+		expect(mine.status, 'the listed job is still processing').toBe('processing');
```
osama-rizk: False-fail: this hard `toBe('processing')` on `listJobs` contradicts the tolerance three lines up (`:156-159` accepts `'processing'` OR `'done'`, with a comment explicitly allowing a fast reference impl). A correct AsyncJob that completes in the few ms between `enqueue` (`:149`) and this `listJobs` read (`:165`) reports `'done'` in the list → a valid solution fails, and it's a flake on a non-deterministic intermediate state. Loosen to `['processing','done']` (and skip the null-count sub-check when done), matching `:156-159`.

Related nit (`:99-106`): the blank-input test asserts the error envelope but not that *no job was enqueued* — an impl that errors *and* still persists would false-pass. Add a `listJobs` side-effect check, the way sql-kb-catalog does.

tasks/oidc-dsql-notes/test.spec.ts
```diff
@@ -32,242 +43,204 @@ async function signIn(page: Page): Promise<void> {
...
 		await page.goto(BASE);
+		await signIn(page);
 
-		await expect(page.getByTestId('signin-btn')).toBeVisible({ timeout: T });
-		await expect(page.getByTestId('note-input')).toHaveCount(0);
-		await expect(page.getByTestId('profile-sub')).toHaveCount(0);
+		const text = uniq('note');
+		const { body } = await rpc(page.request, 'api.addNote', [text]);
+		expect(body?.error, `JSON-RPC error from addNote: ${JSON.stringify(body?.error)}`).toBeFalsy();
+		expect(typeof body?.result?.id, 'addNote must return a numeric id').toBe('number');
```
osama-rizk: False-fail + PROMPT mismatch: this requires `typeof id === 'number'` (and `:135-136` requires strictly-increasing numeric ids), but PROMPT §23-25 says only "sequential id" — never numeric. Aurora DSQL has no `SERIAL`/`SEQUENCE`/`IDENTITY` and steers toward `gen_random_uuid()` / client-generated ids, so the *most idiomatic* DistributedDatabase solution (UUID-keyed, `ORDER BY created_at`) fails here. Accept string/UUID ids and assert oldest-first via the explicit order the PROMPT already mandates, not via id monotonicity — or, if numeric sequential ids are genuinely required, say so in the PROMPT and justify it against DSQL.

Nit (`:77-90`): the SQL-injection round-trip proves *verbatim storage*, not *parameterization* — a manual `replace("'","''")` passes identically. Grading's still valid; the comment just overstates what it proves.

tasks/collab-presence-board/test.spec.ts
```diff
@@ -218,61 +190,17 @@ test.describe('collab-presence-board', () => {
...
-			await expect(presence(page, n)).toHaveCount(1, { timeout: SYNC });
-		}
-
-		await page.reload();
-		for (const n of names) {
-			await expect(presence(page, n)).toHaveCount(1, { timeout: SYNC });
-		}
+		// Tab B had the board open before A joined; it must reflect A's presence
+		// within a couple seconds — realtime broadcast, no reload.
+		await expect.poll(() => presence(tabB, name).count(), { timeout: SYNC }).toBe(1);
```
osama-rizk: Two issues on this fan-out check:

1. **False-fail (subscribe-after-publish race).** tab B's only path to see A is a realtime push; B's mount-time `listPresent` ran before A joined, and this poll only re-reads the DOM — it never re-fetches `listPresent`. If B's subscription isn't established before A broadcasts (mount race), the event is lost and the 8s poll can't self-heal. Small window, but it's the one genuinely async false-fail — a subscription-ready wait or a belt-and-suspenders `listPresent` re-fetch would close it.

2. **Gameability.** This forbids only a manual reload, not client-side polling — a thin client that polls `api.listPresent()` every ~1-2s makes A appear in B well within the 8s window *without ever using the Realtime block*. Combined with the persistence checks being satisfiable by an in-memory Map (same never-restarted process, so DistributedTable isn't provably used either), neither mandated block for this task is provably exercised. Consider asserting a realtime-specific signal (e.g. push latency / a message the polling client couldn't synthesize) or source-inspecting that both blocks are instantiated.

tasks/async-word-counter/test.spec.ts
```diff
@@ -1,297 +1,214 @@
...
 
-// The row for a specific submission, located by its unique phrase text. The
-// shared list persists across the run (and across a Playwright retry), so we
-// never use .first() or a result-only count — either can match a STALE row from
-// another submission/attempt and pass vacuously. The phrase carries a per-call
-// unique token, so this scopes to exactly the row we just submitted.
+// The DOM row for a submission, located by its unique phrase text.
 const rowFor = (page: Page, phrase: string) => page.getByTestId('wc-item').filter({ hasText: phrase });
 
+// HARNESS CONTRACT: requires workers:1 (serial; shared-store assertions assume no concurrent runners)
```
osama-rizk: `workers:1` is load-bearing for every shared-store assertion in this file (and 7 other specs) but is enforced only in the generated `playwright.config.ts` (`3-build-and-test.sh:212`) — this banner is documentation, not a guard. Bump the heredoc's `workers`, or run this spec standalone with default Playwright parallelism, and the shared-store assertions silently corrupt with nothing tripping. A one-line `expect(test.info().config.workers).toBe(1)` in a `beforeAll` (or `test.describe.configure({ mode: 'serial' })`) turns silent cross-test contamination into a loud, obvious failure.

tasks/kb-chat-agent/test.spec.ts
```diff
@@ -1,251 +1,164 @@
...
-		await expect(bubbleWith(page, 'QUOKKA-9F42')).toHaveCount(0);
-
-		expect(errors, `page errors: ${errors.join(' | ')}`).toEqual([]);
+	// --- Framework surface: the agent turn runs through api.ask ---
+
+	test('api.ask retrieves the seeded KB facts and reports the tool + a citation', async ({ request }) => {
+		const a = await ask(request, 'According to the product knowledge base, what is the internal product code and the maximum hover altitude?');
+		expect(a.body?.error, `JSON-RPC error from api.ask: ${JSON.stringify(a.body?.error)}`).toBeFalsy();
+		// Facts live ONLY in the KB — their presence proves a real retrieval.
+		expect(a.reply).toContain('QUOKKA-9F42');
```
osama-rizk: Nit (inherent to LLM grading): `reply.toContain('QUOKKA-9F42')` asserts on live model output — a correct model that retrieves the fact but paraphrases the answer without echoing the literal token would false-fail. The `toolsUsed`/`citations` assertions are the robust proof; the reply-text `toContain` is the soft spot. The unique made-up token mitigates it (a retrieving model is likely to echo it verbatim), so probably fine — just flagging it as the one non-deterministic assertion, in case it flakes in practice.

osama-rizk (commented):
Reviewed for grading validity (does each spec avoid false-fails of correct solutions, false-passes of gameable stubs, and flakiness) — read the harness + kb-chat myself and ran three cluster passes (auth/identity, data/storage, agent/realtime). The API-first direction is a real improvement over DOM grading, and the core gating is genuinely sound: auth-notes proves per-user isolation with two independent `browser.newContext()` cookie jars and — crucially — an `api` that takes NO username arg, so a client-supplied-key impl is structurally impossible; the cookie-less `request` fixture makes every unauth-gating assertion real; file-gallery's byte-exact fetch of the returned download URL is genuine proof of serving; kb-chat correctly asserts only forced-deterministic fragments; bounded `expect.poll` (no fixed sleeps) and the uniqSeq/rpcSeq split are good craft. Inline findings are the specific ones; three cross-cutting items below.

**F3 — cognito-profile's security-critical prod-gate is claimed but NOT enforced (MED, false-pass).** PROMPT.md:63 says the grader confirms by *source inspection* that `api.getLastCode` returns `null` unless `BLOCKS_MOCK==='true'`. The spec does no source read/grep — the only `getLastCode` test runs in mock mode and its own comment concedes the gate is grader-UNVERIFIED. So a model that ships `getLastCode` **ungated** (leaks the OTP in production) scores 100%. Either implement the check (read the submitted source, assert the gate) or drop the claim so the PROMPT/changeset don't overstate coverage. Deeper: making the model build an OTP-reader as *graded surface*, with the one safeguard the grader can't verify, rewards constructing an ungated-capable leak — the 'grader has no mailbox' problem is better solved out-of-band (harness reads the mock code from the KV store / a log) than as a model-implemented backdoor.

**F4 — 'genuine Building-Block usage' is overstated (MED, gameability ceiling).** The grader talks to one never-restarted dev-server process, so it proves 'real bytes/rows served over the wire' but can't distinguish the mandated block from a hand-rolled server-side Map/array in the same process. Concretely: sql-kb-catalog proves no SQL at all (a JS array + counter satisfies id/order/read-back); async-word-counter/auth-notes persistence is Map-satisfiable; collab's two mandated blocks aren't provably exercised (see the inline on the fan-out). file-gallery (byte-exact) mitigates best. Largely inherent — but worth softening the changeset claim, and where it matters adding source-inspection that the mandated block is imported/instantiated (or a durability check across a real restart).

**F7 — 'task content only, no harness changes' is inaccurate (LOW).** `3-build-and-test.sh` gains `export BLOCKS_MOCK=true`, set globally for all tasks though only cognito needs it. Verified inert today (nothing in the framework reads `BLOCKS_MOCK` — it's a new var only the cognito solution reads), so no grading corruption — but if a future task's block ever branches on it, the whole bench silently runs that block in mock mode. Scope it to the cognito task and correct the wording. (Also cosmetic: `retries:0` makes the specs' 'a retry reuses the RUN seed' comments stale.)

tasks/async-word-counter/test.spec.ts
sharonyajain: Fixed in `29e97e4`. You're right, the hard `toBe('processing')` on the listed job contradicted the `['processing','done']` tolerance three lines up, so a fast reference impl that finishes between enqueue and this read would false-fail on a non-deterministic intermediate state. Loosened the `listJobs` status assertion to `['processing','done']` to match, and dropped the now-invalid `count===null` sub-check on the listed row.

Related nit (blank-input `:99`) also addressed in the same commit: the blank-input test now snapshots `listJobs` length before and asserts it's unchanged after the rejected inputs, so a rejects-but-still-persists impl can't false-pass — modeled on the sql-kb-catalog side-effect check as you suggested.

tasks/async-word-counter/test.spec.ts
sharonyajain: Good call. Added `expect(test.info().config.workers).toBe(1)` in a `beforeAll` so running this spec under default parallelism fails loudly instead of silently corrupting the shared-store assertions. `29e97e4`.

tasks/oidc-dsql-notes/test.spec.ts
sharonyajain: Fixed in `29e97e4`, and thanks for the DSQL-specific reasoning, it's exactly right. DSQL has no `SERIAL`/`SEQUENCE`/`IDENTITY` and steers toward `gen_random_uuid()`/client-generated ids, so requiring a numeric id false-failed the most idiomatic solution. `addNote` now accepts a stable non-empty **number or string** id, and the ordering test asserts oldest-first via the row-position (`ORDER BY`) checks the PROMPT already mandates rather than id monotonicity (`:135`). UUID-keyed solutions now grade correctly.

Related nit (`:77`) agreed: the SQL-injection round-trip proves *verbatim storage*, not *parameterization* (a manual `replace("'","''")` passes identically). Grading's still valid; I'll tighten the comment so it doesn't overstate what it proves.

tasks/collab-presence-board/test.spec.ts
sharonyajain: Fixed in `29e97e4`. Realtime stays the primary path, but the poll now passes if **either** the DOM reflects A (broadcast worked) **or** a fresh `api.listPresent()` read shows A — so a push lost to a mount-time subscription race self-heals instead of false-failing. Belt-and-suspenders as you described.

On the gameability point (#2): agreed it's real and shares the F4 ceiling (a polling client passes without the Realtime block; neither mandated block is provably exercised in the same never-restarted process). Captured with F4 — leaning toward source-inspecting that both blocks are instantiated, tracked as a follow-up rather than blocking this PR.

tasks/kb-chat-agent/test.spec.ts
sharonyajain: Noted — `reply.toContain('QUOKKA-9F42')` is the one non-deterministic assertion (a model that retrieves the fact but paraphrases without echoing the literal token would false-fail). The `toolsUsed`/`citations` assertions are the robust proof; the unique made-up token makes a verbatim echo likely, so I'll leave it but add a comment flagging it as the soft spot to watch if it flakes in practice.

soberm (approved):

osama-rizk (approved):

Issue #155, opened 2026-07-05
create-blocks-app starter templates ship a greet()-referencing test that goes stale once the sample API is replaced


hfurkanbozkurt:
## Problem

Bench run 28701776910 (PR #132) surfaced that cells scaffolded with `create-blocks-app` wasted effort on stale starter test files:

- **file-gallery / bare** (lowest-scoring cell, composite 78.6): the starter test still referenced the old `greet()` method after the sample API was replaced, producing an immediate test failure that required manual investigation.
- **email-digest** and **oidc-dsql-notes**: cells churned rewriting stale test files (`.ts` → `.mts` migrations, spawn-based rewrites) that existed solely because of the starter template's shape.

The underlying issue: the starter test exercises the sample `greet()` endpoint by name, so any scaffold where the developer (or agent) removes or renames the sample API immediately inherits a failing test — before any real implementation work has started.

## Request

Make starter test files minimal and resilient — or decouple them from the sample `greet()` API — so that replacing the sample API in a freshly-scaffolded app does not immediately leave a failing test. Options include:

- A trivially-passing placeholder test (e.g. `it('should be implemented', () => {})`) with a comment pointing to the real API once it exists.
- A test that imports nothing from the sample endpoint and only asserts app-level setup (e.g. env vars present, server starts).
- A `TODO` stub that is skipped by default (`it.skip`).

**Evidence**: agent-bench run 28701776910 / PR #132

ikenyal:
I separated this into two concerns: the sample greet assertion is a template-content decision, while the Setup readiness loop should not depend on that sample API. I plan to address only the second concern. 

The backend, bare, demo, and auth-cognito templates currently poll greet before tests start. I will replace that probe with a request to /.blocks-sandbox/config.json and check response.ok, so an unrelated process on port 3000 is not treated as ready. 

The dev server begins listening only after startup initialization and client generation complete, making this a neutral readiness signal. The existing greet assertions will remain unchanged; any decision on replacing those sample tests stays separate.

ikenyal:
Implemented the readiness portion in #198. It changes the backend, bare, demo, and auth-cognito template setup loops to wait for the Blocks configuration endpoint, while keeping the existing sample greet assertions unchanged. The broader question of whether those sample assertions should be replaced remains separate from this focused fix.

ikenyal:
[...]

PR #198, opened 2026-07-14
fix(create-blocks-app): decouple template E2E readiness from sample APIs

ikenyal:
## Summary
- make backend, bare, demo, and auth-cognito template E2E setup wait for the Blocks runtime configuration endpoint rather than the sample greet API
- use the backend template's configured port (3001) for its process and readiness checks
- preserve the existing sample greet assertions

Partially addresses #155.

## Validation
- Biome lint on the four changed template E2E files
- Changesets status against origin/main
- npm run test:templates (default, demo, backend, bare)
- scaffolded auth-cognito from the local registry and ran npm run test:e2e

## Checklist
- [x] I have read and followed CONTRIBUTING.md
- [x] Tests have been added or updated
- [x] Changeset added


packages/create-blocks-app/templates/bare/test/e2e.test.ts
```diff
@@ -25,11 +25,15 @@ test.before(async () => {
   const mod = await import('aws-blocks');
   api = mod.api;
 
-  // Wait for server to be ready
+  // Wait for the Blocks server to be ready without depending on the sample API.
   for (let i = 0; i < 30; i++) {
```
sharonyajain, line 31: **Readiness URL hardcodes port 3000 in three templates while the same PR uses a named const in backend**

The backend template introduced `const serverPort = 3001` and reused it for both `isServerRunning(serverPort)` and the readiness fetch URL, but auth-cognito, bare, and demo hardcode `http://localhost:3000/.blocks-sandbox/config.json` inline (identical literal in all three). The failure path also throws a bare `new Error('Server not ready')` that never names the endpoint or port it polled.

```ts
const response = await fetch('http://localhost:3000/.blocks-sandbox/config.json');
```

**Why it matters:** Inline magic port literal diverging from the named-const pattern the same change establishes (cause) -> if a template's dev port ever differs from 3000 or the sandbox path changes, the probe silently targets the wrong URL and `response.ok` is never true (mechanism) -> the loop exhausts its 30-60 retries and fails with an undiagnosable 'Server not ready', costing template consumers and maintainers debugging time on a healthy server (consequence).

**Suggestion:** We can mirror the backend fix in auth-cognito/bare/demo: hoist `const serverPort = 3000` and build the URL from it, and include the polled URL in the timeout error, e.g. `throw new Error(\`Server not ready at http://localhost:${serverPort}/.blocks-sandbox/config.json\`)` so a CI failure is self-diagnosing.

ikenyal: Updated in a5b2aaf0. All four readiness probes now use a named port and derived URL, and the timeout error includes that URL.

sharonyajain (approved):

cadivus (approved):

Issue #199, opened 2026-07-15
bb-distributed-data: ALTER TABLE DROP COLUMN passes the local mock but fails on real DSQL (0A000)


fossamagna:
### Before opening, please confirm

- [x] I have searched for duplicate or closed issues.
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Which package(s) are affected?

Building Block: data (`@aws-blocks/bb-distributed-data`)

### Describe the bug

Aurora DSQL does not support `ALTER TABLE DROP COLUMN` (see the supported `ALTER TABLE` subset in the [DSQL docs](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/alter-table-syntax-support.html)). The local mock runs on PGlite (full PostgreSQL), which accepts `DROP COLUMN`, so a migration like:

```sql
ALTER TABLE persons DROP COLUMN card_image_key;
```

passes local development (`npm run dev`, migration execution, e2e tests) and only fails on the first real deploy, inside the migration CustomResource:

```
QueryFailedException: unsupported ALTER TABLE DROP COLUMN statement
code: 0A000
    at DsqlEngine.execute
    at runMigrations
```

This is exactly the class of local/production divergence the DSQL validation layer (`validation.ts` RULES) is meant to close — it already rejects FK, TRUNCATE, SERIAL, index key sort order (#148), etc., but has no rule for unsupported `ALTER TABLE ... DROP COLUMN`.

### Steps to reproduce

1. Create a blocks app with a `DistributedDatabase` and two migrations: `001` creating a table with two columns, `002` containing `ALTER TABLE t DROP COLUMN b;`
2. `npm run dev` — migrations apply cleanly on the PGlite mock
3. Deploy — the migration Lambda fails with `unsupported ALTER TABLE DROP COLUMN statement` (0A000), and the CloudFormation update rolls back

### Expected behavior

`validateStatement` / `validateMigrations` reject the statement at dev time with a `DsqlValidationError`, including a hint about the alternative (leave the column in place and stop referencing it, or rebuild the table: create new → `INSERT INTO ... SELECT` → `DROP` → `RENAME TO`).

I'm happy to submit a PR adding the validation rule, tests, and docs — same shape as #148.

fossamagna:
Opened a PR with the validation rules: https://github.com/aws-devtools-labs/aws-blocks/pull/200

adrianjoshua-strutt:
[...]

PR #200, opened 2026-07-15
fix(bb-distributed-data): reject ALTER TABLE DROP [COLUMN] for DSQL

fossamagna:
## Problem

Aurora DSQL does not support `ALTER TABLE DROP COLUMN` — it is not in the supported [`ALTER TABLE` subset](https://docs.aws.amazon.com/aurora-dsql/latest/userguide/alter-table-syntax-support.html). Deploying such a migration fails inside the migration Lambda with `unsupported ALTER TABLE DROP COLUMN statement` (0A000).

The local mock runs on PGlite (full PostgreSQL), which accepts it, so the migration passes local development and only fails on the first real deploy — exactly the class of divergence the mock's DSQL validation layer is meant to close (same as #148 for index key sort order).

**Issue #, if available:** #199

## Changes

- **Validation rule** (`packages/bb-distributed-data/src/validation.ts`): an entry in the existing `RULES` array rejecting `ALTER TABLE ... DROP [COLUMN]` — including the keyword-less Postgres shorthand (`DROP col`, `DROP IF EXISTS col`, quoted identifiers), since Postgres parses it to the same DROP COLUMN action. Because `validateStatement` runs from both the migration validator and the mock engine's query path, the failure now surfaces at dev time instead of on deploy. The error message includes the practical alternatives (leave the column in place, or rebuild the table: create new → `INSERT INTO ... SELECT` → `DROP` → `RENAME TO`).
- **Docs** (`README.md`, `DESIGN.md`): a row in the DSQL limitations / statement-validation tables.
- **Changeset**: patch bump for `@aws-blocks/bb-distributed-data`.

The rule matches within a single statement (`[^;]*`) and does not fire on the `ALTER TABLE ... DROP` forms DSQL does support: `DROP CONSTRAINT [IF EXISTS]`, and `ALTER COLUMN ... DROP DEFAULT` / `DROP NOT NULL` / `DROP EXPRESSION` / `DROP IDENTITY` — nor on `RENAME TO` or `DROP TABLE` (all covered by tests). String literals and comments are stripped before matching.

> [!NOTE]
> An earlier revision of this PR also rejected `ALTER TABLE DROP CONSTRAINT`. That rule was removed: `DROP CONSTRAINT [IF EXISTS] ... [RESTRICT | CASCADE]` is in DSQL's supported `ALTER TABLE` grammar (see the doc link above).

## Validation

- **Unit tests** (`validation.test.ts`): reject `DROP COLUMN` and the keyword-less shorthand spellings; regression cases confirm `DROP CONSTRAINT`, the `ALTER COLUMN ... DROP` actions, `RENAME TO`, and `DROP TABLE` still pass.
- **End-to-end mock test** (`e2e-mock.test.ts`): the `DsqlMockEngine` rejects `ALTER TABLE legacy DROP COLUMN obsolete` — proving the original reproduction is now caught locally before any deploy.
- Full `bb-distributed-data` suite: **152/152 passing**.

## Checklist

- [x] PR description included

---


adrianjoshua-strutt:
Thanks for raising this. Overall this looks great.

I pulled the branch down and ran the touched suites locally. I also threw a bunch of variants at the regex (keyword-less shorthand, IF EXISTS, quoted and schema qualified names, multi-clause, newlines between DROP and COLUMN) and it catches all of them, while correctly leaving the supported stuff alone (DROP CONSTRAINT, ALTER COLUMN ... DROP DEFAULT/NOT NULL/EXPRESSION/IDENTITY, RENAME TO, DROP TABLE, and DROP inside strings or comments).

One small ask: Could you add a direct unit test for a semicolon separated batch? Something like a supported DROP DEFAULT in one statement followed by an unrelated statement, asserting it does not throw. The behavior is already correct, I'd just like to lock it in.

Thanks again!

fossamagna:
@adrianjoshua-strutt Thanks for the thorough review — especially for
throwing all those regex variants at the rule.

Added the batch test in validation.test.ts: a supported
`ALTER COLUMN ... DROP DEFAULT` followed by an unrelated `DROP TABLE`
in one semicolon-separated string, asserting no throw. Verified it
actually locks the behavior in: replacing `[^;]*` with `.*` in the rule
makes this test fail. Suite is now 153/153.

adrianjoshua-strutt (dismissed):

fossamagna:
@adrianjoshua-strutt Sorry for the extra round — your approval got dismissed by a follow-up push. The only change since your review is a CI fix: the Blocks Package Integrity check flagged that the README rows added here sync into the published @aws-blocks/blocks docs/, so the existing changeset now also bumps "@aws-blocks/blocks": patch (same pattern as #220). No code changes. Could you re-approve when you get a chance?

adrianjoshua-strutt:
Sure!

adrianjoshua-strutt (approved):

soberm (approved):

PR #205, opened 2026-07-15
fix: retry PGlite WASM _pg_initdb 'unreachable' trap during runMigrations (#188)

### LINKED ISSUE #188: PGlite WASM _pg_initdb crash (QueryFailedException: unreachable) intermittently kills the dev server during runMigrations on CI
## Summary

The local dev server intermittently crashes on CI right after printing its startup banner, when a PGlite-backed data Block runs its first migration. The `@electric-sql/pglite` WASM module traps with `unreachable` inside `_pg_initdb` (the lazy Postgres init that fires on the first query), surfacing as `QueryFailedException: unreachable` and taking down the Node process.

The crash is **non-deterministic** — it hits whichever PGlite-backed cell loses the dice roll on a given run (seen in both a Next.js SQL/KB cell and a React OIDC/DSQL cell).

## Impact

- The dev server prints `AWS Blocks local server running on http://localhost:3000` then **immediately crashes**, so `dev_server_started=false`, every downstream check scores `0`, and the whole cell is lost.
- Intermittent and init-time; a retry usually succeeds. It is an infrastructure/WASM failure, not a code regression.
- Affects any Block that boots `@aws-blocks/bb-distributed-data`'s `DsqlMockEngine` (DSQL mock) or a PGlite-backed `Database` and runs migrations at startup.

## Evidence

Verbatim stack (identical WASM module hash `wasm://wasm/01edd1ba` across runs):

```
AWS Blocks local server running on http://localhost:3000
wasm://wasm/01edd1ba:1

QueryFailedException: unreachable
    at wasm://wasm/01edd1ba:wasm-function[3625]:0x263199
    at wasm://wasm/01edd1ba:wasm-function[7801]:0x3e3508
    at wasm://wasm/01edd1ba:wasm-function[2533]:0x1c1034
    at Object.Module._pg_initdb (.../@electric-sql/pglite/dist/index.js:3:124284)
    at pe.qe (.../@electric-sql/pglite/dist/index.js:3:243608)
    at async pe._checkReady (.../@electric-sql/pglite/dist/index.js:3:238920)
    at async pe.query (.../@electric-sql/pglite/src/base.ts:153:5)
    at async DsqlMockEngine.execute (.../@aws-blocks/bb-distributed-data/dist/engines/dsql-mock-engine.js:79:33)
    at async runMigrations (.../@aws-blocks/bb-distributed-data/dist/migrations.js:11:5)
    at async DsqlMockEngine.withDdl (.../@aws-blocks/bb-distributed-data/dist/engines/dsql-mock-engine.js:61:20)

Node.js v22.23.1
```

Observed on two separate CI runs (different cells, same signature):
- run `29138896893` — `sql-kb-catalog` (Next.js) cell
- run `29140413875` — `oidc-dsql-notes` (React) cell

**Crash path.** `DsqlMockEngine`'s constructor does `new PGlite(dataDir)`; PGlite defers the real Postgres bootstrap (`_pg_initdb`) to the **first query**. `runMigrations()` issues that first query — `CREATE TABLE IF NOT EXISTS _migrations (...)` — which drives `_checkReady` -> `qe` -> `Module._pg_initdb`, where the WASM module traps `unreachable`. The banner prints before this, so it is **not** a port/startup race — it is a WASM init failure on the first DB operation.

## Reproduction

A minimal, dependency-pinned harness boots a fresh PGlite instance per cycle and issues the exact first two statements `runMigrations` runs (`CREATE TABLE IF NOT EXISTS _migrations (...)` then `SELECT name FROM _migrations ORDER BY name`), hammering the `_pg_initdb` path sequentially and in parallel (the crash is intermittent). No repo checkout required.

```
mkdir pglite-repro && cd pglite-repro
npm init -y >/dev/null && npm pkg set type=module >/dev/null
npm i @electric-sql/pglite@0.2.17
# save repro.mjs (below), then:
node repro.mjs 40 16     # 40 sequential + 16 parallel boots  (args: SEQ PAR)
```

`repro.mjs`:

```js
// Minimal reproduction for the @electric-sql/pglite `_pg_initdb` WASM trap
// ("unreachable") that crashes the AWS Blocks dev server right after the
// "AWS Blocks local server running" banner.
//
// It boots a FRESH PGlite instance per cycle (as DsqlMockEngine does in its
// constructor: `new PGlite(dataDir)`) and issues the first query, which is
// exactly what `runMigrations` does first ("CREATE TABLE IF NOT EXISTS
// _migrations ..."). PGlite lazily runs `_pg_initdb` on that first query — the
// trap site. We hammer it N times sequentially and M in parallel, since the
// crash is intermittent (surfaces under CI memory/concurrency pressure).
//
// Run:  node repro.mjs [SEQ=30] [PAR=8]
import { PGlite } from '@electric-sql/pglite';
import { mkdtempSync, rmSync } from 'node:fs';
import { tmpdir } from 'node:os';
import { join } from 'node:path';

const SEQ = Number(process.argv[2] ?? 30);
const PAR = Number(process.argv[3] ?? 8);

// The exact first two statements runMigrations issues (bb-distributed-data).
const MIGRATIONS_DDL = `
  CREATE TABLE IF NOT EXISTS _migrations (
    id TEXT PRIMARY KEY DEFAULT gen_random_uuid(),
    name TEXT NOT NULL UNIQUE,
    applied_at TIMESTAMP NOT NULL DEFAULT NOW()
  )
`;

async function bootAndMigrateOnce(label) {
  const dir = mkdtempSync(join(tmpdir(), 'pglite-repro-'));
  const db = new PGlite(dir); // DsqlMockEngine ctor does exactly this
  try {
    await db.query(MIGRATIONS_DDL); // first query → triggers _pg_initdb
    await db.query('SELECT name FROM _migrations ORDER BY name');
    return { label, ok: true };
  } finally {
    try { await db.close(); } catch {}
    try { rmSync(dir, { recursive: true, force: true }); } catch {}
  }
}

const isUnreachable = (e) => /unreachable/i.test(String(e?.message ?? e)) || /unreachable/i.test(String(e?.stack ?? ''));

let failures = 0;
console.log(`[repro] pglite _pg_initdb stress — ${SEQ} sequential + ${PAR} parallel boots`);

for (let i = 0; i < SEQ; i++) {
  try {
    await bootAndMigrateOnce(`seq#${i}`);
    process.stdout.write('.');
  } catch (e) {
    failures++;
    console.error(`\n[repro] SEQ boot #${i} FAILED (unreachable=${isUnreachable(e)}):\n`, e?.stack ?? e);
  }
}
process.stdout.write('\n');

const settled = await Promise.allSettled(
  Array.from({ length: PAR }, (_, i) => bootAndMigrateOnce(`par#${i}`)),
);
for (const r of settled) {
  if (r.status === 'rejected') {
    failures++;
    console.error(`[repro] PAR boot FAILED (unreachable=${isUnreachable(r.reason)}):\n`, r.reason?.stack ?? r.reason);
  }
}

if (failures === 0) {
  console.log(`[repro] OK — ${SEQ + PAR} boots, no _pg_initdb trap on this host.`);
  process.exit(0);
} else {
  console.error(`[repro] REPRODUCED — ${failures}/${SEQ + PAR} boots crashed in _pg_initdb.`);
  process.exit(1);
}
```

### Local reproduction status (honest)

I could **not** reproduce the exact `unreachable` WASM trap on a dev host — but the init path is provably memory-sensitive there:

| Environment | Result |
|---|---|
| Node v18.20.2, 30 seq + 8 par | clean, no trap (38/38 OK) |
| Node v22.22.3, 40 seq + 16 par | clean, no trap (56/56 OK) |
| Node 22, 32-wide parallel + `--max-old-space-size=128` | clean, no trap |
| Node 22, `ulimit -v` (2.2–3.1 GB) + parallel | fails in the **same** init path as `RangeError: WebAssembly.Memory(): could not allocate memory` at `pe.qe -> pe._checkReady -> pe.query` |
| Node 22, cgroup `MemoryMax` (`MemorySwapMax=0`) | sharp cliff: <=350M -> OOM-kill (rc 137); >=450M -> clean — no band that yields `unreachable` |

**Conclusion:** the `_pg_initdb` / `_checkReady` / `qe` path is memory-sensitive and fails under pressure, but the specific WASM `unreachable` trap is **CI-environment-specific** — it needs Node **v22.23.1** on a GitHub `ubuntu` runner under the memory/overcommit/RSS pressure of the concurrent (10-wide) bench matrix. It is not fabricated here; run the harness in that environment to surface it.

### To surface it on CI

Run the harness (or the test below) on an `ubuntu-latest` GitHub runner with Node `22.23.1`, under concurrency mirroring the bench matrix (multiple PGlite boots in flight, constrained memory). Suggested: `SEQ=80 PAR=24 ROUNDS=5`, several parallel job shards, and/or a container with `--memory` in the 400–600 MB band.

## Automated test

A `node --test` regression that boots many fresh PGlite instances sequentially and in parallel and asserts none trap `unreachable`. It **fails/flakes when the bug is present** and passes when PGlite inits cleanly. Tunable via `SEQ` / `PAR` / `ROUNDS` env vars.

```
node --test pglite-initdb.test.mjs      # knobs: SEQ=40 PAR=12 ROUNDS=3 (defaults)
```

Local runs (both green — bug did not surface locally, as noted above):
- Node v18.20.2 -> `# pass 2  # fail 0` (~75 s)
- Node v22.22.3 -> clean

`pglite-initdb.test.mjs`:

```js
// Automated regression test for the @electric-sql/pglite `_pg_initdb` WASM trap
// ("QueryFailedException: unreachable") observed on CI, crashing the AWS Blocks
// dev server during runMigrations' first query.
//
// The test boots many FRESH PGlite instances — sequentially and in parallel —
// each issuing the first migration query (the `_pg_initdb` trigger). It asserts
// that NONE trap with "unreachable". When the bug is present the test FAILS
// (or flakes); when PGlite inits cleanly it passes.
//
// Run:  node --test pglite-initdb.test.mjs
//   Tune stress via env: SEQ (default 40), PAR (default 12), ROUNDS (default 3).
import { test } from 'node:test';
import assert from 'node:assert/strict';
import { PGlite } from '@electric-sql/pglite';
import { mkdtempSync, rmSync } from 'node:fs';
import { tmpdir } from 'node:os';
import { join } from 'node:path';

const SEQ = Number(process.env.SEQ ?? 40);
const PAR = Number(process.env.PAR ?? 12);
const ROUNDS = Number(process.env.ROUNDS ?? 3);

const MIGRATIONS_DDL = `
  CREATE TABLE IF NOT EXISTS _migrations (
    id TEXT PRIMARY KEY DEFAULT gen_random_uuid(),
    name TEXT NOT NULL UNIQUE,
    applied_at TIMESTAMP NOT NULL DEFAULT NOW()
  )
`;

const isUnreachable = (e) =>
  /unreachable/i.test(String(e?.message ?? e)) || /unreachable/i.test(String(e?.stack ?? ''));

async function bootAndMigrateOnce() {
  const dir = mkdtempSync(join(tmpdir(), 'pglite-test-'));
  const db = new PGlite(dir); // mirrors DsqlMockEngine constructor
  try {
    await db.query(MIGRATIONS_DDL);                                  // first query → _pg_initdb
    await db.query('SELECT name FROM _migrations ORDER BY name');    // runMigrations' next query
  } finally {
    try { await db.close(); } catch {}
    try { rmSync(dir, { recursive: true, force: true }); } catch {}
  }
}

async function runGuarded() {
  try {
    await bootAndMigrateOnce();
    return null;
  } catch (e) {
    return e;
  }
}

test(`PGlite _pg_initdb never traps 'unreachable' across ${SEQ} sequential boots`, async () => {
  const crashes = [];
  for (let i = 0; i < SEQ; i++) {
    const err = await runGuarded();
    if (err) crashes.push({ i, unreachable: isUnreachable(err), stack: err?.stack ?? String(err) });
  }
  assert.deepEqual(
    crashes.filter(c => c.unreachable),
    [],
    `PGlite _pg_initdb trapped 'unreachable':\n${crashes.map(c => `seq#${c.i}: ${c.stack}`).join('\n\n')}`,
  );
  assert.equal(crashes.length, 0, `Non-unreachable failures:\n${crashes.map(c => c.stack).join('\n\n')}`);
});

test(`PGlite _pg_initdb never traps 'unreachable' across ${ROUNDS}×${PAR} parallel boots`, async () => {
  const crashes = [];
  for (let r = 0; r < ROUNDS; r++) {
    const settled = await Promise.allSettled(Array.from({ length: PAR }, () => bootAndMigrateOnce()));
    settled.forEach((s, i) => {
      if (s.status === 'rejected') {
        crashes.push({ r, i, unreachable: isUnreachable(s.reason), stack: s.reason?.stack ?? String(s.reason) });
      }
    });
  }
  assert.deepEqual(
    crashes.filter(c => c.unreachable),
    [],
    `PGlite _pg_initdb trapped 'unreachable':\n${crashes.map(c => `round${c.r}/par#${c.i}: ${c.stack}`).join('\n\n')}`,
  );
  assert.equal(crashes.length, 0, `Non-unreachable failures:\n${crashes.map(c => c.stack).join('\n\n')}`);
});
```

## Environment

- `@electric-sql/pglite`: **0.2.17** (declared `^0.2.0` in `packages/bb-data/package.json` and `packages/bb-distributed-data/package.json`)
- CI Node: **v22.23.1** (GitHub `ubuntu` runner) — where the crash occurs
- Dev host Node tested: v18.20.2 and v22.22.3 — no trap
- Crash-path packages: `@aws-blocks/bb-distributed-data` (`DsqlMockEngine`, `runMigrations`) -> `@electric-sql/pglite`
- WASM module hash across all crashes: `wasm://wasm/01edd1ba`

## Possible directions

- Bump/pin `@electric-sql/pglite` and re-test under Node 22.23.1 (upstream fixed several WASM init/OOM traps in later 0.2.x / 0.3.x).
- Add a bounded retry around the first `_pg_initdb`-triggering query in `DsqlMockEngine` / `runMigrations` so a transient WASM init trap doesn't kill the dev server.
- Cap PGlite boot concurrency in the bench harness (serialize migrations across cells) to reduce peak memory pressure on the runner.

hfurkanbozkurt:
## Summary

`@electric-sql/pglite` defers its WASM `_pg_initdb` initialization until the **first query**. Under CI memory pressure, that first query — the `CREATE TABLE IF NOT EXISTS _migrations ...` issued at the start of `runMigrations` — can trap with `QueryFailedException: unreachable`. This is an unrecoverable WASM abort that kills the dev server **non-deterministically**, so PGlite-backed benchmark cells fail as a "dice roll" (0/all with an empty base URL because the server is dead) and drag down bench scores.

This PR mitigates the crash at the root of the init path in **both** PGlite-backed engines.

Fixes #188

## Root cause

- PGlite instance is constructed lazily; `initdb` runs on first statement.
- `runMigrations` (`data-common/migrations`) issues the first statement (`CREATE TABLE IF NOT EXISTS _migrations`), which triggers `_pg_initdb`.
- Under memory pressure the WASM module aborts with `unreachable`. The trap is process-fatal and non-deterministic.
- Reproduced in bench cells **oidc-dsql-notes** (react, via `bb-distributed-data` `DsqlMockEngine`) and **sql-kb-catalog** (nextjs, via `bb-data` `PGliteEngine`).

## Fix

A small, shared, bounded-retry helper in `data-common` that treats an `unreachable` init trap as retryable by **discarding the dead PGlite instance and booting a fresh one**:

- **New** `packages/data-common/src/pglite-init.ts`
  - `initializePgliteWithRetry(initial, recreate, options?)` — probes init (`SELECT 1`), and on an `unreachable` trap closes the dead instance, recreates it, and retries with linear backoff (default **3 attempts**, 150ms step). Non-retryable errors are rethrown immediately; the cause-chain walk is cycle-guarded.
  - `isPgliteUnreachableTrap(error)` — matches `/unreachable/i` across message/stack/cause.
  - Pglite-agnostic (`data-common` has no pglite dependency) via a structural `PgliteLike` type.
  - Exported from `data-common/src/index.ts`.
- **`packages/bb-data/src/engines/pglite-engine.ts`** (`PGliteEngine`) and **`packages/bb-distributed-data/src/engines/dsql-mock-engine.ts`** (`DsqlMockEngine`)
  - Added a private `createDb()` (dir prep + existing incomplete-data-dir recovery + stale-lock cleanup + client construction) and `ensureReady()` that runs `initializePgliteWithRetry` and, on final failure, resets readiness so a later query can retry.
  - `await this.ensureReady()` is now the first step of `query` / `execute` / `beginTransaction`.
  - An **optional** 2nd constructor arg `createClient(dataDir) => PGlite` (default `new PGlite(dir)`) provides a test-injection seam. **Non-breaking** — all existing call sites pass one argument.

## Tests (real code paths, no mocked engines)

- `packages/data-common/src/pglite-init.test.ts` — 15 unit tests: `isPgliteUnreachableTrap` variants (message/stack/cause/non-Error/false/cyclic) and retry behavior (first-try success, recreate once, recreate twice, exhaust default 3, custom maxAttempts, non-retryable rethrow, `onRetry` sequence, custom `isRetryable`).
- `packages/bb-data/src/engines/pglite-engine.test.ts` — 2 added tests using a **real** `PGlite`: a factory that traps `unreachable` on the first instance then returns a real instance (asserts data round-trips and the instance was recreated), and a persistent-trap factory (asserts it surfaces `DatabaseErrors.QueryFailed`).
- `packages/bb-distributed-data/src/engines/dsql-mock-engine.test.ts` — same 2 tests for `DsqlMockEngine` (recover via `withDdl`; persistent trap surfaces `DistributedDatabaseErrors.QueryFailed`).

Validation (Node v22.22.3): `data-common`, `bb-data`, `bb-distributed-data` build clean (exit 0); the 3 compiled suites run **37 pass / 0 fail** via `node --test`.

Deliberately **not** relying on flakily reproducing the real WASM trap — the trap is injected deterministically via the `createClient` seam.

## Notes for reviewers

- **pglite version bump intentionally NOT included.** The repo pins `@electric-sql/pglite` `^0.2.0` (lockfile resolves `0.2.17`); latest published is `0.5.4`. That is a large jump across a `0.x` library and is very likely breaking, and I could not confirm a specific changelog entry that fixes the `_pg_initdb`/`unreachable` trap. Bumping is left as a **separate decision** for the maintainers; the retry mitigation here is independent of the version.
- **Issue #188 is currently CLOSED (stateReason COMPLETED).** It was closed manually on 2026-07-11 (~7 min after being labeled `bug`, with no fixing PR). A follow-up comment on 2026-07-13 begins "Reopening — this recurred on CI ..." but the GitHub issue state was **never actually flipped**, so it still shows closed. This PR does not change the issue state; a **reopen appears warranted** (genuinely unresolved and recurring). Flagging for a maintainer to reopen.


pranavosu (dismissed):

soberm (approved):

packages/data-common/src/pglite-init.ts
```diff
@@ -0,0 +1,137 @@
...
+ * arbitrary user-query errors, whose text could contain "unreachable"
+ * incidentally (e.g. a message about an unreachable host).
+ */
+export function isPgliteUnreachableTrap(error: unknown): boolean {
+  const seen = new Set<unknown>();
+  let current: unknown = error;
+  while (current != null && !seen.has(current)) {
+    seen.add(current);
+    if (current instanceof Error) {
+      if (/unreachable/i.test(current.message) || (current.stack != null && /unreachable/i.test(current.stack))) {
```
sarayev, line 69: **[minor]** Right now we decide "this is the WASM crash, let's retry" by looking for the word "unreachable" anywhere in the error, including its stack trace. The thing is, "unreachable" is a pretty common word — lots of codebases have a helper literally called `unreachable` or `assertUnreachable`. So if the probe query ever failed for some unrelated reason and that word showed up in the trace, we'd retry when we shouldn't.

It's a very unlikely corner (the probe is only ever `SELECT 1`), so I'm not worried today — but could we make the check a little more specific, like matching `RuntimeError: unreachable` instead of just the bare word? Cheap insurance against a confusing future bug.

packages/data-common/src/pglite-init.ts
```diff
@@ -0,0 +1,137 @@
...
+  const backoffMs = options.backoffMs ?? 150;
+  const isRetryable = options.isRetryable ?? isPgliteUnreachableTrap;
+
+  let instance = initial;
+  for (let attempt = 1; ; attempt++) {
+    try {
+      await instance.query('SELECT 1');
+      return instance;
+    } catch (error) {
+      if (attempt >= maxAttempts) {
```
sarayev, line 115: **[minor]** Two small inconsistencies in how we clean up here:

- When we've used up all our retries, we close the instance before giving up. But if we hit an error we *don't* retry on, we throw right away without closing — so whether we clean up depends on which case we're in.
- If someone sets `maxAttempts` to 1, we'd close the instance on *any* failure and never even check whether it was the crash we care about.

Both go away if we check "is this the retryable crash?" first, before the "are we out of attempts?" logic. Minor, but it'd make the helper behave more predictably.

packages/data-common/src/index.ts
```diff
@@ -14,3 +14,5 @@ export type { Transaction } from './types.js';
 export { DatabaseBase } from './database-base.js';
 export { splitStatements, runMigrations, loadMigrationsFromDir, DOLLAR_QUOTE_TAG_RE } from './migrations.js';
 export { createKyselyAdapter } from './kysely-adapter.js';
+export { initializePgliteWithRetry, isPgliteUnreachableTrap } from './pglite-init.js';
```
sarayev, line 17: **[minor]** These four new exports are now part of `data-common`'s public API (they show up in `API.md`), so we're on the hook to keep them stable going forward. If they only exist so the two engine packages can share the helper, we could mark them `@internal` instead — still usable across packages, but without promising the outside world they won't change.

If we do want them public, could we add a one-line doc to the two methods on `PgliteLike`? They're currently flagged undocumented in the API report.

sarayev (commented):
A few minor, non-blocking thoughts — nice fix overall, especially the deterministic trap injection and keeping data-common pglite-free.

sarayev (approved):

PR #209, opened 2026-07-15
fix(ci): dedupe constructs/aws-cdk-lib in amplify-interop test to stop dual-install TS2345

hfurkanbozkurt:
## Root cause

The `E2E Amplify Interop` harness scaffolds a throwaway app with **unpinned** `npm create amplify@latest` into `/tmp`, which installs its OWN `constructs` (`^10.0.0`) and `aws-cdk-lib` (`2.244.0`). It then typechecks the generated `aws-blocks/index.cdk.ts` against the **repo's** `constructs` (`^10.6.0`) / `aws-cdk-lib` (`2.257.0`), which the Blocks packages resolve through their `file:` symlinks back into the repo tree.

`tsc` deduplicates modules by **package id** (`name@version`), not by file path. When the two physical `constructs` trees are *different* versions they get distinct package ids, so the nominally-private `Construct#host` member no longer matches across them and the typecheck fails:

```
TS2345: ... Types have separate declarations of a private property 'host'.
```

This currently fails on **all PRs**.

## Fix

Add a step to the smoke job (between "Scaffold Amplify + Blocks app" and "Deploy sandbox") that pins the scaffolded app's `constructs` and `aws-cdk-lib` to the repo's **exact installed** versions — read from `$GITHUB_WORKSPACE/node_modules` (falling back to the repo `package.json` range) — via npm `overrides`, then `rm -rf node_modules package-lock.json && npm install` so a single copy of each resolves.

Matching versions => matching package ids => `tsc` dedups the two physical `constructs` trees to one declaration => the `host` skew (and TS2345) disappears. Reading the *resolved installed* version rather than the semver range keeps the ids exactly aligned, so a newer patch can't silently re-introduce the drift.

## Notes

- CI harness / test-infrastructure fix — **owned by the harness maintainers**.
- Only `.github/workflows/e2e-amplify-interop.yml` changes: one added step, no deletions.
- Authoritative validation is this PR's own `e2e-amplify-interop` check (the full `npm create amplify` scaffold needs AWS/network and isn't runnable locally). The dedupe mechanism was verified locally against the repo's real `node_modules` (`pinned overrides: constructs=10.6.0 aws-cdk-lib=2.257.0`).

### DIFF
--- a/.github/workflows/e2e-amplify-interop.yml
+++ b/.github/workflows/e2e-amplify-interop.yml
@@ -77,6 +77,44 @@ jobs:
           npm create amplify@latest -- --yes
           node $GITHUB_WORKSPACE/packages/create-blocks-app/dist/index.js --yes .
 
+      # De-skew the sandbox typecheck against the drifting Amplify scaffold.
+      #
+      # `npm create amplify@latest` installs its OWN aws-cdk-lib/constructs at
+      # whatever versions it currently ships, which drift from the versions this
+      # repo builds against. The generated aws-blocks/index.cdk.ts resolves
+      # aws-cdk-lib (Stack/Mixins) from THIS scaffold's tree, while the Blocks
+      # packages — linked in as `file:` deps (symlinks) — resolve their own
+      # constructs/aws-cdk-lib from the REPO's tree via the symlink realpath.
+      # tsc dedups modules by package id (name@version), so when the two trees
+      # are DIFFERENT versions it treats them as distinct declarations and the
+      # nominally-private Construct#host member no longer matches — failing with
+      # TS2345 "separate declarations of a private property 'host'".
+      #
+      # Fix: pin the scaffold's constructs/aws-cdk-lib to the repo's EXACT
+      # installed versions (read from node_modules, not the package.json semver
+      # range, so the ids match precisely — a range could resolve to a newer
+      # patch and re-drift), then reinstall so the overrides apply. Matching
+      # versions => matching package ids => tsc dedups to one declaration.
+      - name: Pin constructs/aws-cdk-lib to the repo's versions
+        run: |
+          cd /tmp/amplify-interop-test
+          node -e '
+            const fs = require("fs");
+            const ws = process.env.GITHUB_WORKSPACE;
+            const ver = (n) => {
+              try { return require(ws + "/node_modules/" + n + "/package.json").version; }
+              catch { const p = require(ws + "/package.json"); return (p.devDependencies || {})[n] || (p.dependencies || {})[n]; }
+            };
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-amplify-interop.yml
```diff
@@ -77,6 +77,54 @@ jobs:
...
+      # versions => matching package ids => tsc dedups to one declaration.
+      - name: Pin constructs/aws-cdk-lib to the repo's versions
+        run: |
+          cd /tmp/amplify-interop-test
+          node -e '
+            const fs = require("fs");
+            const ws = process.env.GITHUB_WORKSPACE;
+            const ver = (n) => {
+              try { return require(ws + "/node_modules/" + n + "/package.json").version; }
+              catch { const p = require(ws + "/package.json"); return (p.devDependencies || {})[n] || (p.dependencies || {})[n]; }
```
sarayev, line 110: Small thing on this fallback path: if the node_modules lookup misses, we fall back to the repo’s package.json semver range (e.g. ^10.6.0) and write that as both the pin and the override. That’s the one case the primary path was specifically designed to avoid — a range can resolve to a newer patch and re-drift, which is exactly the failure this step fixes. In CI the fallback should almost never fire since deps are already installed by this point, so it’s not urgent — but it might be safer to fail loudly here than to silently pin a range. Non-blocking.

.github/workflows/e2e-amplify-interop.yml
```diff
@@ -77,6 +77,54 @@ jobs:
...
+              for (const bucket of ["dependencies", "devDependencies"]) {
+                if (pkg[bucket] && pkg[bucket][name]) pkg[bucket][name] = v;
+              }
+            }
+            pkg.overrides = { ...pkg.overrides, ...pins };
+            fs.writeFileSync("package.json", JSON.stringify(pkg, null, 2) + "\n");
+            console.log("pinned constructs=" + constructsV + " aws-cdk-lib=" + cdkV + " (direct deps + overrides)");
+          '
+          rm -rf node_modules package-lock.json
+          npm install
```
sarayev, line 126: One nice-to-have after the reinstall: a quick assertion that the dedup actually took, e.g. `npm ls constructs aws-cdk-lib` (or a grep for a single resolved version). Right now if this ever regresses, the next signal is another opaque TS2345 further down the pipeline — a one-line check here would let a future failure explain itself. Totally optional.

sarayev (approved):
Nice fix — the diagnosis of the tsc dedup-by-package-id mismatch is spot on, and reading the exact installed version from node_modules rather than the semver range is the right call to avoid re-drift. Left two small, non-blocking notes inline. Approving.

ahmedhamouda78 (approved):

PR #214, opened 2026-07-16
ci(publish): changesets/action version-PR model

pranavosu:
Same feature set as #9, but publishes via **`changesets/action`** (the "Version Packages" PR model) instead of #9's scheduled, hand-rolled publish.

## Changes

- **Publish model (the difference):** `changesets/action` on push to `main`. It opens/updates a "Version Packages" PR (`changeset version`); merging that PR publishes to npm. The merge is the release gate. `publish-packages.yml` is 54 lines vs #9's 117.
- **Changeset guards** (`changeset-check.yml` + `changeset-guard.ts`): `verify-coverage` (changed packages must have a changeset) and `block-major` (no `major` bumps pre-1.0), run via `node --experimental-strip-types`.
- **Revert-release** (`revert-release.yml` + `prepare-revert-release.mjs`): manual workflow that restores `packages/` to the previous release and opens a patch PR. Its boundary detection matches the action's `chore: version packages` commit, so it works unchanged.
- **Cleanups:** `publish:dry-run` -> `publish:local` rename across call sites; removal of the `publish-npm.yml` / `publish-npm.ts` manual-publish wrapper; `verify-changeset-coverage.ts` folded into `changeset-guard.ts`.
- Adds a `release` script (`npm run build && changeset publish`) and `pull-requests: write`.

## Publish-model comparison vs #9

| | #9 (scheduled, hand-rolled) | This (changesets/action) |
|---|---|---|
| Trigger | Mon–Thu 9am PT cron + manual | push to `main` |
| Release decision | schedule + `PUBLISH_ENABLED=true` | human merges the Version PR |
| Pre-publish gate | none (unattended) | the Version PR (runs `pr-checks`) |
| GitHub Releases | git tags only | created by the action |
| On-demand publish | `workflow_dispatch` | merge the Version PR |
| `publish-packages.yml` | 117 lines | 54 lines |

## Notes

- Checkout and the action use `OPS_GITHUB_TOKEN` (not the default) so the Version PR triggers CI — required, or the PR's checks won't run.
- Everything except `publish-packages.yml` (and the `release` script) is identical to #9.


.github/workflows/changeset-check.yml
```diff
@@ -14,8 +17,10 @@ jobs:
...
           node-version-file: '.nvmrc'
-          cache: npm
-      - run: npm ci
-      - run: npx @changesets/cli status --since=origin/main
+
+      # Guards import only Node built-ins, so Node type-strips and runs the TS
+      # directly.
       - name: Verify changeset covers all changed packages
-        run: npx tsx scripts/verify-changeset-coverage.ts
+        run: node --experimental-strip-types scripts/changeset-guard.ts verify-coverage
```
svidgen, line 30: This check runs on all PRs to main, which will include the "Version Packages" PR that `changesets/action` creates. That PR has bumped `package.json` files but no `.changeset/*.md` on disk (they were consumed by `changeset version`). verify-coverage will see changed packages with no changesets and fail.

If `Require changeset` is a required check, this blocks the release PR from being mergeable. Suggest: early-exit when the PR title matches the changesets action pattern (e.g., starts with `chore: version packages`).

svidgen (commented):
Looks good overall — the changesets/action model is a clear improvement over the hand-rolled pipeline. A few things to address:

.github/workflows/changeset-check.yml
```diff
@@ -14,8 +17,10 @@ jobs:
...
+      - name: Block major version bumps (pre-release)
```
svidgen, line 33: Minor suggestion: the old pipeline also ran `npx @changesets/cli status --since=origin/main` which validates changeset file structure (YAML frontmatter, package names exist in the workspace, valid bump types). The regex parser in changeset-guard.ts doesn't catch those — malformed changesets would pass CI but fail at `changeset version` time (post-merge).

Consider adding `changeset status` back as a complementary step here. Not blocking, but it's a worse feedback loop to catch structural issues after merge.

.github/workflows/publish-packages.yml
```diff
@@ -38,137 +39,15 @@ jobs:
...
-        run: npm run build
-
-      - name: E2E Test Local
-        if: steps.version.outputs.changed == 'true'
-        run: npm run test:e2e:local
-
-      - name: Configure AWS credentials
-        if: steps.version.outputs.changed == 'true'
-        uses: aws-actions/configure-aws-credentials@v6
+      - uses: changesets/action@v1
```
svidgen: `changesets/action@v1` is a floating tag — the changesets org can move it to any commit. This workflow has access to `NPM_TOKEN` and `OPS_GITHUB_TOKEN`. Consider SHA-pinning for supply chain safety:

```yaml
- uses: changesets/action@<sha>  # v1.x.x
```

.github/workflows/changeset-check.yml
pranavosu: Addressed in 8e5b5e34, though I went a different route than `changeset status`.

In this repo `changeset status --since=origin/main` exits 1 whenever any changed workspace package lacks a changeset — even with valid changesets present — because its change detection spans the whole workspace (incl. native packages) vs `verify-coverage`'s narrower merge-base + `packages/` + `@aws-blocks` scope. As a hard gate it would duplicate `verify-coverage`, fail ordinary PRs, and re-break the release PR (the deadlock from the other thread).

Instead I added a `validate-structure` subcommand to `changeset-guard.ts` that validates only the changesets on disk: frontmatter parses, each referenced package exists in the workspace, and the bump type is valid. It's blocking, has no coverage semantics (so no duplication/false failures), and the release PR passes since it has no changesets left. That closes the "malformed changeset slips past the regex and only fails post-merge at `changeset version`" case you flagged.

.github/workflows/publish-packages.yml
pranavosu: Done in 8e5b5e34 — pinned to `changesets/action@a45c4d594aa4e2c509dc14a9f2b3b67ba3780d0d # v1.9.0` (the v1.9.0 tag dereferenced to its commit SHA).

.github/workflows/changeset-check.yml
pranavosu: fixed in https://github.com/aws-devtools-labs/aws-blocks/commit/8e5b5e34d2f2fb2eea05119cc0f4d145bfcd9a0c. verify-coverage now early-exits with success when the PR title starts with chore: version packages (the changesets/action Version PR), so it no longer blocks the release PR from merging. The title is passed into the guard via a PR_TITLE env var on the step.

svidgen (approved):

.github/workflows/publish-packages.yml
```diff
@@ -38,137 +39,15 @@ jobs:
...
-              tag="${name}@${version}"
-              git tag "$tag" 2>/dev/null || true
-            fi
-          done
-
-      # ── Publish ───────────────────────────────────────────────
-      - name: Publish to S3
-        if: steps.version.outputs.changed == 'true' && inputs.dry_run != true
+          version: npx changeset version && npm install --package-lock-only
+          publish: npm run release
```
osama-rizk, line 47: **Question on the publish-time gate.** `release` here is `npm run build && changeset publish` — so between the Version PR going green and this job publishing to npm, `main` may have moved (other merges/rebases) and nothing re-runs against the actual commit being shipped. Do we need a verification step (e.g. `test:e2e:local`, unit tests) as a gate right before `changeset publish`, so we're validating the exact commit we publish rather than trusting the Version PR's earlier checks? Or is skipping that re-run an intentional tradeoff for pre-1.0?

osama-rizk:
**Question on changelog config.** `.changeset/config.json` sets `"changelog": "@changesets/cli/changelog"` (the plain formatter), but `@changesets/changelog-github@^0.6.0` is already in devDependencies. Now that the action creates GitHub Releases, do we want to point `changelog` at `["@changesets/changelog-github", { "repo": "aws-devtools-labs/aws-blocks" }]` so release notes link back to the PRs and authors? Or is the plain changelog intentional?

.github/workflows/publish-packages.yml
```diff
@@ -1,35 +1,36 @@
...
-  id-token: write
-  contents: write
+  contents: write # push the version-PR branch + tags
+  pull-requests: write # open/update the Version Packages PR
+  id-token: write # npm provenance
 
 jobs:
-  publish:
-    name: Version & Publish
+  release:
```
osama-rizk, line 22: **Question on a fork/owner guard.** This job runs on push to `main` and publishes to npm with real tokens. Should we guard it with something like `if: github.repository_owner == 'aws-devtools-labs'` so a fork with `main` pushes can't ever fire the publish path? Or is that unnecessary given secrets won't exist on forks?

.github/workflows/canary.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: npm run build
 
       - name: Publish to local registry
-        run: npm run publish:dry-run
+        run: npm run publish:local
```
osama-rizk, line 54: **Question on prerelease channels.** Canary publishes to the local/S3 registry rather than an npm dist-tag. Do we want a way for consumers to `npm install @aws-blocks/x@canary` (i.e. a `changeset version --snapshot` + `changeset publish --tag canary` path), or is S3 the intended channel for prereleases and npm reserved only for stable releases?

pranavosu: S3 is for pre-releases currently. We will explore options for unstable/canary releases to npm in the future

.github/workflows/publish-packages.yml
pranavosu: We don't need to add this check, since the forks won't have tokens. And if an attacker gets access to our npm tokens, adding this check will not help.

osama-rizk (approved):

.github/workflows/publish-packages.yml
pranavosu: Will investigate in a followup

pranavosu:
> **Question on changelog config.** `.changeset/config.json` sets `"changelog": "@changesets/cli/changelog"` (the plain formatter), but `@changesets/changelog-github@^0.6.0` is already in devDependencies. Now that the action creates GitHub Releases, do we want to point `changelog` at `["@changesets/changelog-github", { "repo": "aws-devtools-labs/aws-blocks" }]` so release notes link back to the PRs and authors? Or is the plain changelog intentional?

Will follow up on this

PR #216, opened 2026-07-17
feat(bench): single-table comparison report + slimming

hfurkanbozkurt:
**PR 1 of 3** — split of #194 (which is being closed and superseded by this stack).

**Concern: report presentation** — the agent-bench PR-vs-`main` report redesign + slimming.

- Single results table built from the run's cells (`TASK · TEMPLATE · TESTS · JUDGE · COST · TURNS · SCORE · STOP REASON`), replacing the old two-table (colors-only Overview + numbers Detailed) layout and dropping the standalone `Δ vs base` column.
- Per-cell emoji-indicated signed deltas vs the `main` baseline via per-metric `DELTA_THRESHOLDS`; per-field `🆕 (new)` handling; glossary + headline + executive summary + Potential-issues + collapsed per-cell analysis.
- New `SCORE = composite ÷ cost` (composite points per dollar) priced from builder tokens.
- Slimming: compact Judge to one line, drop the Tokens/LOC/Files columns.
- Includes the associated test-hardening and bench-behavior fixes that accumulated on the branch before the integrity commit.

Internal CI tooling only — no published `@aws-blocks/*` package changes.

**Base:** `main`.

**Stack** (bottom → top): #216 ← #217 ← #218 · supersedes #194.
Supersedes #194.

---

**Review updates (v2):**

- **Emoji indicators replace colored dots** for colorblind accessibility: ✨ improved · ✅ OK/no significant change · ⚠️ slight regression · ❌ serious regression (>2× threshold) · 🆕 new. Distinct shapes ensure recognition without relying on color.
- **Threshold definitions added to the glossary:** each metric's noise band is now spelled out with concrete numbers (e.g., "composite ±5 pts: ≤5 is ✅, >5 improving is ✨, >5 worsening is ⚠️, >10 worsening is ❌").
- **DELTA_THRESHOLDS rationale documented** in a comment block in `overview.mjs` — explains the empirical basis (observed run-to-run variance from ~20 main-branch runs) and how to re-calibrate.
- **Cost degradations explained:** The cost *calculation* itself (`cellCost`, `BUILDER_PRICING`) is unchanged. The apparent "widespread cost degradations" are a reporting artifact from the bench-behavior fixes on this branch: cells that previously hid as excluded `unknown` (dead dev-server, no tests ran) are now correctly classified as `dead_server` (composite 0, INCLUDED in the mean). Their cost was always there — it's now *visible* in the table. No inadvertent code change affected actual token usage or pricing. The newly-surfaced cells are the `dead_server` classification fix (`scoring.mjs` + `3-build-and-test.sh`) and the Playwright-skip for dead servers — both intentional bench-behavior hardening described in the PR.


svidgen (changes requested):
1. This still doesn't look right to me:

Image

A yellow-coded result is suggestive that something is "mildly wrong". A mapping like this is more visually helpful:

| Symbol | Meaning |
| -- | -- |
| ✨ / 🤩 / 🥳 / 🍻 / ... | Noticeably improved |
| ✅ | OK / No significant change |
| ⚠️ | Slight regression |
| ❌ | Serious regression |
| 🆕 | New |

(Avoiding simple colored dots for quick colorblind recognition as well.)

2. The key should include definition of what a "insignificant", "slight", and "serious" changes are.
3. The PR description talks about updating the report, but the numbers show pretty widespread cost degradations. Did something else change that shouldn't have?
4. The `DELTA_THRESHOLDS` values — how were these chosen?

soberm (commented):
Automated review notes (inline). Internal agent-bench CI tooling; all items minor/nit/question — no blocking findings.

scripts/agent-bench/steps/lib/overview.mjs
```diff
@@ -229,182 +262,249 @@ export function diffAgainstBaseline(current, baseline) {
...
-	return `${col} ${fmtVal(baseVal)} -> ${fmtVal(prVal)}`;
+// TESTS: "passed/denom", colored by the pass COUNT (higher better, ±1 noise). denom 0 → NONE.
+function testsCell(base, pr) {
+	const pp = numOrNull(pr?.tests_passed);
+	const pd = numOrNull(pr?.tests_denom);
+	if (pp === null || pd === null || pd === 0) return NONE;
+	const value = `${pp}/${pd}`;
+	const bp = numOrNull(base?.tests_passed);
+	if (bp === null) return `${WHITE} ${value} (new)`;
+	return `${deltaColor(bp, pp, DELTA_THRESHOLDS.tests, 'up')} ${value} (${signedInt(pp - bp)})`;
```
soberm: [minor] `testsCell` colors the TESTS cell by the raw pass COUNT (`pp - bp`), with the denominator ignored. If a task's spec count changes between the baseline run and this PR (e.g. 4/4 → 4/10) the pass RATE regresses 100%→40% but the cell still reads `✅ 4/10 (0)`. This is only safe while every task's Playwright spec set is frozen across runs. Consider coloring by the test-RATE delta (`pp/pd` vs `bp/bd`), or forcing 🆕 when `pd !== base.tests_denom` so a denominator change isn't compared as if the scale were constant.

scripts/agent-bench/steps/lib/scoring.mjs
```diff
@@ -98,10 +127,19 @@ export function classifyCell(result) {
...
 		}
 		return { klass: 'agent_fail', reason: AGENT_FAIL_REASON };
 	}
+	// The agent finished and any build succeeded, but the dev-server never served / crashed (step 3
+	// detected it, skipped Playwright, and stamped dev_server_status='dead'). The produced app failing
+	// to RUN is a genuine product failure → dead_server (verdict 'fail', composite 0, INCLUDED), NOT a
+	// 'scored' cell with no tests (which would read 'unknown' and be EXCLUDED, masking the crash).
+	// Ordered AFTER the harness_error / agent_fail returns so a cancellation or a pre-grade failure —
+	// which never produced a runnable app to begin with — still wins and stays excluded.
+	if (result?.dev_server_status === DEAD_SERVER_STATUS) {
```
soberm: [question] This `dead_server` branch now classifies a crashed/never-served dev-server as composite 0 and INCLUDES it in the mean, but any `latest-main.json` baseline recorded before this change EXCLUDED those same cells (as `unknown`). During the transition the composite-mean delta vs `main` is apples-to-oranges — a real regression can appear purely from the classification change — and this affects the composite mean/delta and the `BENCH_MAX_REGRESSION` gate, not just the cost display the PR body mentions. It self-heals on the next push-to-`main` re-baseline. No code change required (the gate is observational/unset by default), but please confirm the first post-merge `main` bench re-records `latest-main.json` before enabling `BENCH_MAX_REGRESSION`, and consider noting the one-time baseline discontinuity in the run summary.

.changeset/bench-report-redesign.md
```diff
@@ -0,0 +1,33 @@
...
+  issues** section, and a collapsed per-cell analysis. Replaces the old
+  two-table layout (a colors-only Overview + a numbers Detailed table) and drops
+  the standalone `Δ vs base` column. The table was later slimmed for readability:
+  the `TOKENS`, `LOC`, and `Files` columns were removed (those counts are still
+  persisted to the S3 baseline for offline analysis, just no longer rendered).
+- Each metric cell is a single inline line (`<color> <value> (<Δ vs main>)`): the
+  current value with its signed delta vs the `main` baseline, colored by the
+  significance + direction of the change via the per-metric `DELTA_THRESHOLDS`
+  (composite/score ±5, judge ±0.3, tests ±1, cost ±10%, turns ±3;
+  tests/judge/score higher-better, cost/turns lower-better; within band → 🟡,
```
soberm: [minor] This emoji legend is stale vs the shipped indicators: it describes the old colored dots (`within band → 🟡, beyond → 🟢/🔴` here, and `⚪ (new)` on L25) but the report actually ships ✨/✅/⚠️/❌ and 🆕. This changeset becomes the published CHANGELOG entry for `@aws-blocks/agent-bench`, so it should read `within band → ✅, beyond → ✨/⚠️/❌, new → 🆕`. The threshold numbers in this paragraph are correct — only the emoji scheme is wrong.

scripts/agent-bench/steps/summary.mjs
```diff
@@ -189,19 +187,36 @@ const md = [];
...
+	'- **Columns:** Tests (pass/denom) · Judge (overall 0-10 score with signed delta; per-dimension breakdown lives in the judge artifact JSON) · Cost · Turns (agent cycles) · Score · Stop reason.',
+);
+md.push(
+	'- **Threshold definitions (per metric, `DELTA_THRESHOLDS` in `overview.mjs`):**',
+);
+md.push(
+	'  - composite/score: **±5 points** — a change ≤5 pts is ✅; >5 improving is ✨; >5 worsening is ⚠️; >10 worsening is ❌.',
+);
+md.push(
+	'  - judge (overall + each dimension): **±0.3** — a shift ≤0.3 is ✅; >0.3 improving is ✨; >0.3 worsening is ⚠️; >0.6 worsening is ❌.',
```
soberm: [minor] This glossary line claims per-dimension judge thresholding (“judge (overall + each dimension): ±0.3”), but `judgeDetailCell`/`dimKeys` were removed — `judgeCell` now renders the OVERALL judge score only (the per-dimension breakdown lives uncolored in the judge artifact JSON). No per-dimension delta is thresholded anywhere, and this also contradicts the adjacent Columns line. Drop “+ each dimension”.

scripts/agent-bench/steps/lib/overview.mjs
```diff
@@ -20,53 +21,84 @@ export const cellKey = (c) => `${c?.task ?? ''}/${c?.template ?? ''}`;
...
+// ─────────────────────────────────────────────────────────────────────────────
+
+// Per-metric significance thresholds for the delta indicators — the ONE place they live. A change
+// within ±threshold reads as ✅ (OK/noise, since N=1); beyond it, ✨ (improved) / ⚠️ (slight
+// regression) / ❌ (serious regression, >2× threshold) by direction.
+// Absolute units unless the key ends in `Pct` (then it's a fraction of |baseline|).
+export const DELTA_THRESHOLDS = {
+	composite: 5, // composite points — also the headline mean-delta band (deltaBall)
+	score: 5, // composite-per-$ points
+	judge: 0.3, // judge score (0-10) — applied per-dimension too
```
soberm: [minor] This `DELTA_THRESHOLDS.judge` comment says the threshold is “applied per-dimension too”, but per-dimension judge coloring was removed (`judgeDetailCell`/`dimKeys` are gone; only the overall judge score is thresholded). Drop the “applied per-dimension too” note so the comment matches the shipped behavior.

scripts/agent-bench/steps/analyze.mjs
```diff
@@ -139,9 +151,71 @@ function main() {
...
+};
+const MAX_EVIDENCE_CHARS = 500;
+
+function fdisp(v, fallback = '—') {
+	return typeof v === 'string' && v.trim() ? v.trim() : fallback;
+}
+
+// One deterministic sentence folding the structured failures by category + owner into the exec summary
+// (rendered even if the Bedrock roll-up call failed). Returns null when there are no structured failures.
+function summarizeFailures(failureRows) {
```
soberm: [minor] The failure roll-up rendering (`summarizeFailures`, `renderFailureAnalyses`, `fdisp`) is pure and non-trivial — category/owner tally + sort, `FAILURE_CATEGORY_BADGE` fallback, the `MAX_EVIDENCE_CHARS` (500-char) evidence clip, and blockquote `> ` line-prefixing — but `analyze.mjs` has no test file, unlike the `lib/` helpers it consumes, so regressions here render silently wrong. Suggest moving these three pure helpers into `lib/analysis.mjs` (or adding `analyze.test.mjs`) and covering the badge fallback, the >500-char evidence clip, and a mixed category/owner tally.

scripts/agent-bench/steps/lib/analysis.test.mjs
```diff
@@ -216,3 +228,294 @@ describe('buildRollupUserText(input)', () => {
...
+	});
+});
+
+describe('extractFailingTests(pwResults)', () => {
+	// Build a Playwright JSON report with `n` specs that all fail with the SAME error, plus optional extras.
+	function pwReport(specs) {
+		return { suites: [{ title: 'auth', specs }] };
+	}
+	function failSpec(title, message) {
+		return { title, ok: false, tests: [{ results: [{ status: 'unexpected', error: { message } }] }] };
```
soberm: [minor] Every `extractFailingTests` fixture builds results as `error: { message }`, so the `Array.isArray(res.errors) ? res.errors[0]?.message` fallback in `firstError` (analysis.mjs) is never exercised — a Playwright reporter that emits `errors[]` with no `.error` is a real shape. Add a fixture with `results:[{status:'unexpected', errors:[{message:'boom'}]}]` and assert the error is lifted.

scripts/agent-bench/steps/summary.mjs
```diff
@@ -153,32 +152,31 @@ const gateFailed = gateResult.failed;
...
+function gateLine() {
 	const mean = aggregate.mean_composite;
-	const judgeNote = judgeMean !== null ? ` · judge mean **${judgeMean.toFixed(2)}**/10` : '';
 	// Composite mean delta vs the baseline (🟢/🔴/🟡 over the ±5 band), when present.
 	const deltaNote =
 		diff.hasBaseline && diff.meanDelta !== null
 			? ` · ${deltaBall(diff.meanDelta)} ${diff.meanDelta > 0 ? '+' : ''}${diff.meanDelta.toFixed(1)} vs \`main\``
 			: '';
-	const head = `Mean composite **${mean.toFixed(1)}**/100 ${compositeBand(mean)} across ${compositeCells.length} scored cell(s)${judgeNote}${deltaNote}.`;
+	const head = `Mean composite **${mean.toFixed(1)}**/100 ${compositeBand(mean)} across ${compositeCells.length} scored cell(s)${deltaNote}.`;
```
soberm: [minor] The mean-composite + Δ-vs-`main` line is printed twice in the rendered report: `renderPreword`'s first bullet already emits “Mean composite …/100 … <ball> +Δ vs `main` · judge mean …”, and then this `gateLine` (appended as a bullet) re-emits “Mean composite …/100 <band> … <ball> +Δ vs `main`. Threshold …”. Suggest reducing `gateLine` to the threshold pass/fail (+band) only and letting the preword own the mean line.

scripts/agent-bench/steps/summary.mjs
```diff
@@ -153,32 +152,31 @@ const gateFailed = gateResult.failed;
...
+	return `- _No cells produced test results — no composite headline to report._${g}`;
+}
+
+// The floor line under the preword: the composite mean vs BENCH_MIN_SCORE. Read-only — the actual
+// pass/fail comes from gateResult (floorGate); the enforcing decision + exit live in the Merge
+// verdict block below, so this never reassigns gateFailed.
+function gateLine() {
 	const mean = aggregate.mean_composite;
-	const judgeNote = judgeMean !== null ? ` · judge mean **${judgeMean.toFixed(2)}**/10` : '';
 	// Composite mean delta vs the baseline (🟢/🔴/🟡 over the ±5 band), when present.
```
soberm: [nit] This comment still references `🟢/🔴/🟡` (“over the ±5 band”), but `deltaBall` now yields ✨/⚠️/✅/❌. Update the comment to match the shipped indicators.

svidgen (dismissed):

scripts/agent-bench/steps/lib/overview.mjs
sharonyajain: Good catch — coloring by raw pass count silently mislabels a rate regression when the denominator changes. Fixed in b116e7ea: `testsCell` now reads `base.tests_denom` and forces 🆕 (`(new)`) when `bd !== pd`, so a spec-count change (e.g. 4/4 → 4/10) is tagged new rather than diffed count-vs-count. Comment updated to spell out the frozen-spec-set assumption.

.changeset/bench-report-redesign.md
sharonyajain: Fixed in b116e7ea. The legend now reads `within band → ✅, beyond → ✨/⚠️/❌` and `🆕 (new)`, matching the shipped indicators. Threshold numbers were already correct and are unchanged.

scripts/agent-bench/steps/summary.mjs
sharonyajain: Fixed in b116e7ea — dropped "+ each dimension" (now "judge (overall): ±0.3"). Also fixed the adjacent Directions line which had the same "(+ dimensions)" claim, so the glossary is internally consistent with the overall-only judge coloring.

scripts/agent-bench/steps/lib/overview.mjs
sharonyajain: Fixed in b116e7ea — the `DELTA_THRESHOLDS.judge` comment now reads "overall only; per-dimension scores are not thresholded".

scripts/agent-bench/steps/analyze.mjs
sharonyajain: Done in b116e7ea — moved the three pure helpers (`fdisp`, `summarizeFailures`, `renderFailureAnalyses`) plus `FAILURE_CATEGORY_BADGE`/`MAX_EVIDENCE_CHARS` into `lib/analysis.mjs` next to the other tested pure helpers, and `analyze.mjs` now imports them. Added coverage in `lib/analysis.test.mjs`: the `FAILURE_CATEGORY_BADGE` fallback to 🔴, the >500-char evidence clip (asserting the ellipsis + cap), a mixed category/owner tally with count-desc ordering, and the defensive missing-field paths.

scripts/agent-bench/steps/lib/analysis.test.mjs
sharonyajain: Added in b116e7ea — a fixture with `results:[{status:"unexpected", errors:[{message:"boom from errors[]"}]}]` (no `.error`), asserting the message is lifted. This exercises the `Array.isArray(res.errors) ? res.errors[0]?.message` fallback in `firstError` that no prior fixture reached.

scripts/agent-bench/steps/summary.mjs
sharonyajain: Fixed in b116e7ea — `gateLine` no longer re-emits the mean composite + Δ-vs-`main`; the preword's first bullet owns that line. `gateLine` now emits only the composite band + threshold pass/fail. Dropped the now-unused `deltaBall` import.

scripts/agent-bench/steps/summary.mjs
sharonyajain: Fixed in b116e7ea — the comment now references `deltaBall ✨/⚠️/✅/❌` instead of the old `🟢/🔴/🟡`.

scripts/agent-bench/steps/lib/scoring.mjs
sharonyajain: Confirmed on both counts:

1. **Re-baseline before enabling the gate.** The first push-to-`main` bench after this merges re-records `latest-main.json` with the new `dead_server` classification included, so from that point baseline and PR use the same rule and the composite-mean Δ is apples-to-apples again. `BENCH_MAX_REGRESSION` is unset today, so nothing gates during the one transitional PR window; we will only set it after that first post-merge main run has re-baselined. Good to flag explicitly — thanks.

2. No code change here, as you noted — the branch is observational until the repo var is set.

Resolving; the baseline discontinuity is a known one-time transition, not a code issue.

scripts/agent-bench/steps/3-build-and-test.sh
```diff
@@ -261,6 +275,25 @@ if [ -f "$PW_RESULTS_JSON" ]; then
...
+# trace/metrics already land at /tmp. Best-effort: a missing source or copy failure must never break
+# the green-regardless exit (analyze-cell degrades to null when a file is absent). Clear any stale
+# copies from a PRIOR cell first (runner-agnostic) so a skipped / again-missing source can't leave the
+# previous cell's evidence in place for analyze-cell to misread.
+# /tmp is sticky (+t): a stale copy from a PRIOR cell may be benchagent-owned (the agent's isolated
+# phase wrote it), so a plain `rm` as the runner uid hits EPERM and — under `set -e` — would abort the
+# step BEFORE the green-regardless `exit 0` below, mis-recording a cell that actually built + passed.
+# Mirror the reap/fuser lines above: `sudo -n rm` clears benchagent-owned files, unprivileged `rm` is
+# the fallback, and the trailing `|| true` guarantees this cleanup never aborts the step.
+sudo -n rm -f /tmp/pw-results.json /tmp/dev.log /tmp/build.log 2>/dev/null || rm -f /tmp/pw-results.json /tmp/dev.log /tmp/build.log 2>/dev/null || true
```
soberm: Move the stale-cleanup sudo -n rm -f /tmp/pw-results.json /tmp/dev.log /tmp/build.log above the if [ -n "$APP_BASE_URL" ] guard (or into an EXIT trap) so a prior cell's evidence is always cleared before analyze-cell runs, even when the Playwright-install early-exit fires.

sharonyajain: Good catch — that is a real cross-cell evidence leak on the install-failure path. The stale-clear `rm` sat after the `if [ -n "$APP_BASE_URL" ]` block, but that block has two `exit 0` early-exits (playwright package install failure and chromium download failure), so on either the script left before clearing and the prior cell's `/tmp/pw-results.json` / `dev.log` / `build.log` stayed for analyze-cell to misread.\n\nFixed in 1248858d: hoisted the clear to just above the Playwright guard so it runs on every path — the dead-server else branch, both early-exits, and the normal path. The fresh `cp` staging stays after the guard (it needs the evidence to exist first); on an early-exit there is nothing to stage and a cleared `/tmp` is the honest state (analyze-cell degrades to null on a missing file). Went with the hoist rather than an EXIT trap to keep it a plain linear read next to the reap/fuser lines it mirrors.

soberm (approved):

svidgen (approved):

PR #220, opened 2026-07-17
fix(bb-data): bump Aurora PostgreSQL engine version 16.4→16.13 (retired) + make configurable

hfurkanbozkurt:
## What & why

`packages/bb-data/src/infra.ts` hardcoded the Aurora Serverless v2 PostgreSQL engine version as `AuroraPostgresEngineVersion.VER_16_4`. **AWS has retired `16.4` in us-east-1.** As a result, `CreateDBCluster` now fails at deploy time with:

```
400 InvalidRequest: Cannot find version 16.4 for aurora-postgresql
```

This is a main-wide blocker for **every** E2E that deploys a `Database` block — reproducibly observed on PRs **#205** and **#208** and across all DB-deploying E2E runs.

Available `16.x` minors in us-east-1 today: `16.8, 16.9, 16.10, 16.11, 16.13` (`16.4`–`16.7` are gone).

## The fix

1. **Immediate:** default the engine to the **latest available** `16.x` = **`16.13`**. Staying on major 16 keeps zero compatibility risk; picking the latest minor gives the longest deprecation runway and newest patches (deliberately *not* the lowest minor, which would be first to be retired again).
2. **Durable:** the engine version is now **configurable** via a new optional `postgresVersion` option on `DatabaseOptions`, threaded down to the cluster. When AWS retires `16.13` in future, this becomes a **config change, not a framework code fix / outage**.

```ts
new Database(scope, 'mydb', { postgresVersion: '16.13' }); // override when needed
```

The default uses the strongly-typed enum `AuroraPostgresEngineVersion.VER_16_13` (present in the installed `aws-cdk-lib@2.257.0`); a caller-supplied override string is passed through `AuroraPostgresEngineVersion.of(version, major)`.

## Files

- `packages/bb-data/src/infra.ts` — default `VER_16_4` → `VER_16_13`; compute `engineVersion` from optional `postgresVersion`; add `postgresVersion?` to `AuroraInfraConfig`.
- `packages/bb-data/src/types.ts` — add `postgresVersion?: string` to the public `DatabaseOptions` (with rationale + `@default '16.13'`).
- `packages/bb-data/src/index.cdk.ts` — thread `postgresVersion` into `materialize()`.
- `.changeset/aurora-pg-engine-version.md` — `@aws-blocks/bb-data`: **minor** (new public option).

## Validation

- `npm run build:packages` — ✅ green (all packages compile, incl. bb-data).
- `node --test packages/bb-data/dist/{infra,index.cdk}.test.js` — ✅ **14/14 pass, 0 fail** (CDK synth assertions).

## Notes

- Opened as **draft**. Additive commit only — no history rewrite.
- bb-data is under `packages/**`, so PR Agent Bench + E2E workflows are expected to fire (accepted).

### DIFF
--- a/.changeset/aurora-pg-engine-version.md
+++ b/.changeset/aurora-pg-engine-version.md
@@ -1,5 +1,5 @@
 ---
-"@aws-blocks/bb-data": minor
+"@aws-blocks/bb-data": patch
 "@aws-blocks/blocks": patch
 ---
 
--- a/.changeset/aurora-pg-engine-version.md
+++ b/.changeset/aurora-pg-engine-version.md
@@ -6,4 +6,4 @@ Bump the Database block's Aurora PostgreSQL engine version from the retired `16.
 
 AWS retired Aurora PostgreSQL `16.4` in us-east-1, after which `CreateDBCluster` failed with `Cannot find version 16.4 for aurora-postgresql`, blocking every deployment of a `Database` block. The default now points at the latest available `16.x` minor (`16.13`) for the longest deprecation runway.
 
-A new optional `postgresVersion` option on `DatabaseOptions` lets callers override the engine version (e.g. `postgresVersion: '16.13'`), so the next AWS retirement is a configuration change rather than a framework code fix.
+A new optional `postgresVersion` option on `DatabaseOptions` lets callers override the engine version (e.g. `postgresVersion: '16.13'`), so the next AWS retirement is a configuration change rather than a framework code fix. Overrides are validated at synth time (must be `MAJOR.MINOR`, e.g. `16.13`), so a malformed value fails fast with a clear error instead of an opaque `CreateDBCluster` failure.
--- a/.changeset/aurora-pg-engine-version.md
+++ b/.changeset/aurora-pg-engine-version.md
@@ -1,0 +1,9 @@
+---
+"@aws-blocks/bb-data": minor
+---
+
+Bump the Database block's Aurora PostgreSQL engine version from the retired `16.4` to `16.13`, and make the engine version configurable.
+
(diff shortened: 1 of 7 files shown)


pranavosu (dismissed):
The retired VER_16_4 is still referenced in packages/core/src/cdk/mixins.test.ts (lines 45 and 94). Those are synth-only unit tests, so they don't call CreateDBCluster and aren't a deploy blocker, but they now pin a different version than the new bb-data default. Worth updating them in this PR or a quick follow-up so the repo doesn't carry two different Aurora PG versions.

.changeset/aurora-pg-engine-version.md
```diff
@@ -0,0 +1,12 @@
+---
+"@aws-blocks/bb-data": minor
```
svidgen: Should be `patch` for both entries. This is a bugfix (retired engine version) + an additive option — not a breaking change. Existing deployed stacks will get an in-place Aurora minor version upgrade on next deploy (no replacement, no data loss), which is the correct behavior since 16.4 no longer works anyway.

svidgen (changes requested):
Looks good — just one changeset adjustment needed.

The `minor` bump signals a breaking change in our pre-1.0 semver convention, but this isn't breaking:

- The old default (16.4) is retired and already fails on deploy — this is a bugfix.
- For existing stacks, changing `EngineVersion` on `AWS::RDS::DBCluster` triggers an in-place minor version upgrade (no resource replacement, no data loss).
- The new `postgresVersion` option is purely additive.

Please change the changeset to `patch` for both `@aws-blocks/bb-data` and `@aws-blocks/blocks`.

svidgen (approved):

pranavosu (approved):

PR #232, opened 2026-07-18
fix(bb-distributed-data): register DSQL runtime config

ikenyal:
## Problem

`DistributedDatabase` registers its DSQL endpoint and region directly on the shared Blocks Lambda. Resource mappings should use the shared config registry so they do not consume the Lambda environment-variable budget.

## Changes

- Register the DSQL endpoint and region through `registerConfig()`.
- Keep the migration custom-resource Lambda environment unchanged; it is separate from the customer-facing Blocks Lambda.
- Update CDK synthesis coverage to verify the application handler receives config-registry wiring, has no direct endpoint or region variables, and the deployed config includes both values.
- Add a patch changeset for `@aws-blocks/bb-distributed-data`.

## Validation

- `npm run build --workspace @aws-blocks/bb-distributed-data`
- `npm test --workspace @aws-blocks/bb-distributed-data` (146 passing tests)
- `npm run lint:deps`
- `changeset status --since=origin/main`

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (not applicable: no user-facing API or workflow change)

---

### DIFF
--- a/.changeset/quiet-donkeys-remember.md
+++ b/.changeset/quiet-donkeys-remember.md
@@ -1,0 +1,5 @@
+---
+'@aws-blocks/bb-distributed-data': patch
+---
+
+Move DSQL runtime configuration to the shared config registry.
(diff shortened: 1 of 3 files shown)


ikenyal:
While rebasing onto the current main branch, I confirmed that #356 has already made this same migration: DistributedDatabase registers its DSQL endpoint and region through `registerConfig()`, with coverage that they are no longer direct handler environment variables. Keeping this PR would duplicate the merged implementation, so I am closing it.

Issue #221, opened 2026-07-17
Scaffold ships createLambdaHandler(backend) but .d.ts requires a lazy factory — fails tsc


hfurkanbozkurt:
## Which package(s) are affected?
@aws-blocks/core (create-blocks-app scaffold)

## Describe the bug
The scaffold ships `createLambdaHandler(backend)`, but the correct signature per its own .d.ts is the lazy-factory form `createLambdaHandler(() => import('./index.js'))`. The shipped form fails `tsc`, so every consumer must catch and rewrite it before anything builds.

## Evidence
Observed independently in all 3 reps of the document-ingestion-pipeline task (preamp batch manual-deploy-20260717-103949-1339). Each retrospective flagged the scaffold/.d.ts mismatch.

## Expected
The scaffold should compile as-shipped.

## Suggested fix
Update the create-blocks-app template to emit the lazy-factory form matching the .d.ts

sharonyajain:
Closing the issue since #235 is merged

PR #235, opened 2026-07-19
fix(create-blocks-app): lazy-load React template backend

ikenyal:
## Problem

The React template passes an eagerly imported backend module to `createLambdaHandler`, but the handler accepts a lazy async factory. This makes the React template inconsistent with every other template and imports the backend before runtime config is loaded.

**Issue #, if available:** #221

## Changes

- Replace the eager backend import in the React template with `() => import('./index.js')`.
- Add a scaffold regression test that verifies the generated React handler uses the lazy factory and does not import the backend eagerly.

## Validation

- `npm run build --workspace @aws-blocks/create-blocks-app`
- `npm test --workspace @aws-blocks/create-blocks-app` (51 passing tests)
- `npm run lint:deps`
- `changeset status --since=origin/main`

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (not applicable: this aligns the generated implementation with the existing documented handler API)

---


xyos:
Independent confirmation: we hit this exact bug and reproduced it on the latest release, `@aws-blocks/create-blocks-app@0.1.19` (published 2026-08-11, after this PR was opened).

Repro on a fresh scaffold (`--template react`, `npm install`, `npx cdk synth`):

```
aws-blocks/index.handler.ts(4,44): error TS2345: Argument of type 'typeof import(".../aws-blocks/index")' is not assignable to parameter of type '() => Promise<any>'.
Error: Frontend build failed: npm run build
```

Synth fails out of the box because `tsc --noEmit` runs via the Hosting buildCommand. Note this is not just a compile error: `createLambdaHandler` calls `await backendFactory()` at runtime (`@aws-blocks/core` dist/lambda-handler.js), so a deployed Lambda from the shipped template would throw on every invocation, and the eager import also loads the backend before the S3 config lands in `process.env`.

We also verified the fix in this PR: with the handler changed to the lazy factory form, the react template synths cleanly, including with `@aws-cdk/core:validateAgainstDefaultRules` enabled (zero findings from the default rule set). Would be great to get this merged before the next release. Thanks @ikenyal!

pranavosu (approved):

svidgen (approved):

Issue #228, opened 2026-07-17
Pin aws-cdk CLI as a template devDependency (avoid npx auto-install + version skew)


pranavosu:
## Summary

None of the `create-blocks-app` templates list the `aws-cdk` CLI in `devDependencies` — they only depend on `aws-cdk-lib`. Because the CLI isn't declared, the `npx cdk …` invocations in the deploy/watch scripts fetch it on demand on first run.

## Symptoms

- First deploy prints `npm warn exec The following package was not found and will be installed: cdk@<version>` and pays a first-run install latency.
- Version skew: the on-demand `cdk` CLI version floats independently of the pinned `aws-cdk-lib`, so the CLI and the library can drift apart.

## Where

Every template under `packages/create-blocks-app/templates/*/package.json` declares `aws-cdk-lib` but no `aws-cdk`:

- `auth-cognito`, `backend`, `bare`, `default`, `demo`, `nextjs`, `react`

## Suggested fix

Add `aws-cdk` to `devDependencies` in each template, pinned to a version aligned with that template's `aws-cdk-lib`.

Related: #153.

sharonyajain:
The PR #236 is merged

PR #236, opened 2026-07-19
fix(create-blocks-app): add CDK CLI to templates

ikenyal:
## Problem

Generated templates invoke `cdk` through their deploy and sandbox scripts but declare only `aws-cdk-lib`. On a fresh project, `npx cdk` therefore fetches a CLI on first use, adding latency and allowing the CLI version to float independently of the project dependency graph.

**Issue #:** #228

## Changes

- Add `aws-cdk` with a `^2.1138.0` range to all seven standalone templates.
- Keep the CLI on a floating 2.x range so fresh generated projects resolve a current CLI as `aws-cdk-lib` updates. CDK compatibility is determined by Cloud Assembly schema support rather than matching package version numbers.
- Add a regression test that discovers deployable templates automatically and verifies that each declares a floating 2.x CDK CLI dependency.

## Validation

- `npm run build --workspace @aws-blocks/create-blocks-app`
- `npm test --workspace @aws-blocks/create-blocks-app`
- `npm run lint:deps`
- `git diff --check`

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (not applicable: the existing scripts are unchanged; their required CLI is now installed with the template)

---


packages/create-blocks-app/src/cli.test.ts
```diff
@@ -147,6 +147,23 @@ describe('create-blocks-app auto-detection', () => {
     }
   });
 
+  it('pins template CDK CLI versions to the repository toolchain', () => {
+    const lockfile = JSON.parse(readFileSync(join(__dirname, '../../..', 'package-lock.json'), 'utf-8'));
+    const expectedVersion = lockfile.packages?.['node_modules/aws-cdk']?.version;
```
sharonyajain: 🟡 **Regression test hardcodes the template list, so a new template can ship without the CDK CLI pin**

The new test enumerates a literal array of the seven current templates rather than deriving the set from the templates directory (e.g. readdirSync filtered to dirs containing a package.json). Any template added later is not asserted unless someone also remembers to append it to this array.

```typescript
for (const template of ['auth-cognito', 'backend', 'bare', 'default', 'demo', 'nextjs', 'react']) {
```

**Why it matters:** Hardcoded list (cause) → a future template dir is added but not added to the array (mechanism) → that template ships without a pinned aws-cdk, reintroducing the exact float/latency regression this PR fixes for its users, and CI stays green (consequence).

**Suggestion:** Derive the template list at test time by reading the templates directory (readdirSync of packages/create-blocks-app/templates, keeping entries that contain a package.json) and assert every one pins the lockfile-resolved aws-cdk version, so present and future templates are all covered.

_[code-review-sage]_

soberm (dismissed):

packages/create-blocks-app/src/cli.test.ts
osama-rizk: **This test will fail against current `main` — the PR has gone stale.** I checked the live lockfiles:

- This branch: root `aws-cdk` devDep `^2.1117.0` → lockfile resolves **2.1122.0** → templates pin **2.1122.0** → test passes *in isolation*.
- Current `main`: root `aws-cdk` has advanced to `^2.1138.0` → lockfile resolves **2.1138.0**.

The test reads the **merged** lockfile, so once this lands on today's main, `expectedVersion` becomes **2.1138.0** while the templates still pin **2.1122.0** → `assert.strictEqual` fails for all 7 templates → red CI. The branch last synced main on 2026-08-20 and the CLI bumped since.

**Fix:** rebase on main and re-pin all 7 templates (and the changeset wording) to whatever the current lockfile resolves — 2.1138.0 today.

packages/create-blocks-app/src/cli.test.ts
```diff
@@ -147,6 +147,23 @@ describe('create-blocks-app auto-detection', () => {
...
+    assert.ok(expectedVersion, 'repository lockfile should resolve aws-cdk');
+
+    for (const template of ['auth-cognito', 'backend', 'bare', 'default', 'demo', 'nextjs', 'react']) {
```
osama-rizk: **The check couples 7 hand-maintained pins to a *floating* root dependency, which guarantees recurring churn.** Root `aws-cdk` is itself a caret (`^2.1138.0`), so any `npm install` / renovate bump moves the lockfile — and this test then goes red until someone hand-edits all 7 template `package.json`s, for a change nobody associates with `create-blocks-app`. That coupling is exactly what just made this PR stale (see the version comment), so it's not hypothetical — it's a standing tax on an unrelated area.

Consider deriving the template pin **from the lockfile at generation time** (single source of truth, no hand-sync), or — if the hardcoded pins stay — documenting the "bump the repo CLI ⇒ bump all templates" contract next to the root dep so the coupling is discoverable.

Separately: the template list here is hardcoded, so an 8th template added later without `aws-cdk` stays green — the gap is invisible. Deriving the list from `readdirSync(templates/)` makes coverage auto-extend and removes a second place to forget.

packages/create-blocks-app/templates/auth-cognito/package.json
```diff
@@ -36,6 +36,7 @@
     "vite": "^6.4.3",
     "tsx": "^4.7.0",
     "esbuild": "^0.27.1",
+    "aws-cdk": "2.1122.0",
```
osama-rizk: Design question: **exact-pinning the CLI while the lib floats inverts the safe compatibility order.** The CDK CLI embeds a maximum cloud-assembly schema version it can read, and the safe invariant is CLI-schema ≥ lib-schema. Here the CLI is pinned exact (`2.1122.0`) but `aws-cdk-lib` is `^2.257.0` and floats *up*. When a fresh `create-blocks-app` install resolves the lib to a newer minor that bumps the assembly schema past what 2.1122.0 supports, the *generated app's* `cdk deploy` fails with a schema-version mismatch — and repo CI can't catch it because it's in the generated project.

Normally I'd suggest `^2.1122.0` so the CLI can float up alongside the lib, but the new test asserts exact equality, which forbids that — so the test design locks in the fragile combination. Worth resolving the tension one way: pin **both** CLI and lib exactly (full reproducibility), or let **both** float with carets (auto-compatible) — but not exact-CLI + caret-lib.

packages/create-blocks-app/templates/backend/package.json
```diff
@@ -29,6 +29,7 @@
     "typescript": "^5.3.0",
     "tsx": "^4.7.0",
     "esbuild": "^0.27.1",
+    "aws-cdk": "2.1122.0",
     "aws-cdk-lib": "2.257.0",
```
osama-rizk, line 39: Nit (pre-existing, not introduced here): `backend` pins `aws-cdk-lib: "2.257.0"` exact while the other six use `^2.257.0`. It means `backend` is now the only template that's fully pinned (CLI + lib both exact) — which per the compatibility point above is actually the *safe* configuration. Since you're touching all these files anyway, worth aligning: either bring the other six to exact lib pins, or bring `backend` to a caret — just make them consistent.

osama-rizk (commented):
The intent is right and the change is clean — `aws-cdk` in `devDependencies` is the correct spot, and shipping a regression test with a dependency-pin PR is better hygiene than most. My main concern is that the PR has gone stale relative to `main`, which makes its own new test fail on merge (F1), and the way that test is coupled to a floating root dependency will keep causing that (F2). One design question on exact-CLI-vs-floating-lib (F3) worth your call. Details inline.

packages/create-blocks-app/src/cli.test.ts
ikenyal: Updated in 9b3a0b41. I rebased on current main and replaced the lockfile-equality assertion, so this no longer fails when the repository lockfile advances.

packages/create-blocks-app/src/cli.test.ts
ikenyal: Updated in 9b3a0b41. The test now discovers deployable templates from the templates directory and verifies that each declares a floating CDK CLI dependency. It no longer depends on the root lockfile.

packages/create-blocks-app/templates/auth-cognito/package.json
ikenyal: Updated in 9b3a0b41. The templates now use a floating CLI range, and the test no longer requires an exact CLI version. This keeps the installed CLI current as generated-project CDK library ranges advance.

packages/create-blocks-app/templates/backend/package.json
ikenyal: I left the pre-existing aws-cdk-lib ranges unchanged in this PR. The backend template matches the repository exact library pin, while the floating CLI satisfies the schema-compatibility requirement without changing the existing library-version policy.

soberm (approved):

sharonyajain (approved):

Issue #245, opened 2026-07-22
Blank page for returning visitors after every deploy: old-build HTML served from __dpl cookie, cookie stamped to current build


Shimagaji:
### Which package(s) are affected?

`@aws-blocks/core`

### Describe the bug

After every deploy, returning visitors — browsers holding a `__dpl` skew-protection cookie from a previous build — get a blank page on their first page load. A second reload fixes it. First-time and incognito visitors are unaffected. This hit real users of our production app.

The defect is in the generated CloudFront function pair of `@aws-blocks/hosting` (`kvs_router`; the package is not listed in the issue form's dropdown, so we selected the closest entry). Two behaviors contradict each other:

1. The **viewer-request** function honors the `__dpl` cookie for **all** URIs, including HTML documents. A returning visitor's first request after a deploy is therefore served the **old** build's HTML.
2. The **viewer-response** function unconditionally stamps the `__dpl` cookie with the **current** build (`meta.b`) on every HTML response.

The old HTML references content-hashed assets that only exist under the old build's prefix. Since the browser now holds the new cookie, those asset requests are rewritten to `/builds/<newBuildId>/assets/index-<oldHash>.js`, which does not exist — the request fails (403 AccessDenied on hosting 0.1.4; 404 on ≥ 0.1.5 where the bucket policy gained `s3:ListBucket`) and the page renders blank.

We observed this in production on hosting 0.1.4 and code-inspected hosting 0.1.6 (latest): the relevant generated function code is unchanged (byte-identical), so the defect is still present.

### Expected behavior

The HTML document, the `__dpl` cookie, and the assets referenced by that HTML always resolve to one consistent build generation, so a returning visitor's first load after a deploy renders correctly — whether that is served fully from the old build or fully from the new one.

### Reproduction steps

1. Deploy a hosting-enabled Blocks app (skew protection on, the default) at least twice, so both `builds/<oldId>/` and `builds/<newId>/` exist in the bucket.
2. Simulate a returning browser with curl (`OLD` = the previous buildId, e.g. from the S3 `builds/` listing):

```bash
APP=https://<distribution-domain>

# 1) Request the page with the OLD cookie
#    → the OLD build's HTML is served, but Set-Cookie stamps the NEW buildId
curl -s -D - -b "__dpl=$OLD" "$APP/" -o page.html
#    note a content-hashed asset referenced by page.html, e.g. /assets/index-<oldHash>.js

[...]

PR #255, opened 2026-07-23
fix(hosting): serve HTML from the current build after a deploy (returning-visitor blank page, #245)

osama-rizk:
## Problem

**Issue: #245**

After every deploy, returning visitors — browsers holding a `__dpl` skew-protection cookie from a previous build — get a **blank page** on their first page load. A second reload fixes it. First-time and incognito visitors are unaffected.

The defect is a contradiction between the two generated CloudFront functions in `@aws-blocks/hosting`'s `kvs_router`:

1. The **viewer-request** function honored the `__dpl` cookie for **all** URIs, including HTML documents — so a returning visitor's first request after a deploy was served the **old** build's HTML.
2. The **viewer-response** function unconditionally stamps the `__dpl` cookie with the **current** build (`meta.b`) on every HTML response.

The old HTML references content-hashed assets that only exist under the old build's prefix. Since the browser now holds the new cookie, those asset requests are rewritten to `/builds/<newBuildId>/assets/index-<oldHash>.js`, which does not exist — the request fails (403 on hosting 0.1.4; 404 on ≥ 0.1.5) and the page renders blank.

This is the complement of #24 (which fixed the deploy-window 403 for *new/cookieless* visitors): the request side kept serving returning users the old build's HTML while the response side force-advanced their cookie, so generation consistency broke exactly one request later.

### Reproduced on a live distribution

On an SPA app deployed 3× with divergent Vite asset hashes (skew on):

| Request | Result |
|---|---|
| `GET /` with **OLD** cookie | serves **old** build HTML (`index-D3MCfqH4.js`) + `Set-Cookie: __dpl=<newBuild>` |
| old HTML's asset with **NEW** cookie | **HTTP 404** |
| same asset with **OLD** cookie (control) | **HTTP 200** |

Origin confirmed healthy: the asset exists only under the old build prefix, absent from the new build — a pure routing/cookie-consistency defect.

## Changes

In the **viewer-request** function (`kvs_router.ts`), after the SPA / directory-index fallback and before the build-id prefix:

```js
if (uri.lastIndexOf('.html') === uri.length - 5) { buildId = meta.b; }
```

HTML documents now always resolve from the **current** build, never a pinned cookie build, while **assets keep honoring the cookie**. HTML, the response-stamped cookie, and the HTML's content-hashed assets therefore always agree on one build generation.

Placement is deliberately **after** the SPA/directory-index fallback so extensionless routes rewritten to `/index.html` (and non-SPA `/about` → `/about/index.html`) are covered too.

**Mid-session safety:** asset requests still honor the visitor's old cookie and old `builds/<id>/` prefixes are retained (`prune: false`), so an already-loaded page keeps working until the next HTML navigation — which lands the visitor consistently on the current build.

**CloudFront 10 KB code-size limit (second commit).** The generated function source ships verbatim to CloudFront, which rejects code > 10 KB with an opaque HTTP 413 at deploy (rolling the stack back). The first draft added the rationale as a multi-line comment *inside* the function string, pushing it from ~9.6 KB to ~10.6 KB. The in-function comment is now a single terse line (full rationale moved to the generator's JSDoc, which does not ship), and two unit guards assert both generated functions stay under 10 KB so this regresses loudly at test time, not at deploy.

## Validation

- Added a regression suite `generated request fn — HTML always resolves from the current build (#245)` in `kvs_router.test.ts` covering static `.html`, non-SPA directory-index, and SPA-fallback documents all resolving to the current build despite an old `__dpl` cookie.
- Updated the existing skew-cookie test to assert the cookie is still honored for **assets** (the correct remaining behavior).
- Added `generated functions — CloudFront 10 KB code-size limit` guards for both the viewer-request and viewer-response functions.
- **TDD confirmed:** the 3 new HTML tests fail without the fix and pass with it.
- Full `@aws-blocks/hosting` suite green (**789 tests**), no new lint findings on changed lines.
- Changeset added (`@aws-blocks/hosting` patch).

### Verified end-to-end on a live deployment

Deployed the `aws-blocks-next-app-router` sample app against this branch and confirmed both the #245 fix and the deploy path are healthy on real CloudFront:

- **Deployed app:** https://d3qjt5bgcu7mz4.cloudfront.net/app
- LIVE viewer-request function carries the fix (`lastIndexOf('.html')`) and is 9810 B (under the 10 KB limit — no HTTP 413).
- A `GET /app` with a stale `__dpl` cookie is served the **current** build's HTML, is **not** re-stamped with the stale build, and every hashed asset the HTML references resolves under the stamped cookie (the exact #245 failure mode — now green).
- Assets still honor the `__dpl` cookie (mid-session consistency preserved).

An external caching/edge-routing regression suite (`test/caching-regression.mjs`, run via `npm run test:caching` against the deployed distribution) exercises this outside the browser — controlling the `__dpl` cookie and `Accept-Encoding` cache-key dimension, which an in-page probe cannot — and passes 8/8. The suite is read-only against live origin state (asserts the deployed CloudFormation template + live response headers rather than poisoning shared prod state).

## Checklist

- [x] PR description included
- [x] Changeset added
- [ ] Relevant documentation is changed or added — N/A (internal generated-function behavior; JSDoc on the generator updated in-PR)

---


packages/hosting/src/constructs/kvs_router.ts
```diff
@@ -713,6 +733,8 @@ async function handler(event) {
       if (seg2.indexOf('.') === -1) { uri = uri + '/index.html'; }
     }
   }
+  // HTML always resolves from the current build; assets honor __dpl (see #245 in JSDoc).
+  if (uri.lastIndexOf('.html') === uri.length - 5) { buildId = meta.b; }
```
ahmedhamouda78: [minor] This suffix check has a false-positive for any URI that's exactly 4 characters long and doesn't contain `.html`. Both sides of the comparison come out to -1:

```js
// uri = '/x.y' (length 4)
uri.lastIndexOf('.html')  // -1  (.html not present)
uri.length - 5            // -1  (4 - 5)
// -1 === -1 -> true, but this isn't an HTML document
```

So a static asset at a 4-char path like `/x.y` or `/a.b` would get forced to the current build and skip the `__dpl` cookie, which is the opposite of what we want for assets. If that asset only exists under an older build prefix it'd 404, which is the exact bug this PR is fixing, just flipped.

Since the rest of this generated function sticks to ES5 string methods, `slice(-5)` keeps the style consistent and avoids the edge case (slicing a string shorter than 5 chars returns the whole string, which is never `.html`):

```js
// before
if (uri.lastIndexOf('.html') === uri.length - 5) { buildId = meta.b; }

// after
if (uri.slice(-5) === '.html') { buildId = meta.b; }
```

`uri.endsWith('.html')` would also work and is already used up in the TypeScript layer at line 205, but it breaks the ES5 convention used inside the generated string.

Might be worth a quick regression test in the `HTML always resolves` suite:

```ts
void it('does NOT treat a 4-char non-html asset URI as HTML', async () => {
  const entries = buildKvsEntries({
    manifest: baseManifest({ routes: [{ pattern: '/*', target: 'static' }] }),
    buildId: 'newbuild',
    hasServer: false,
    hasImage: false,
    skewEnabled: true,
  });
  const { output } = await runRequestFn(
    generateKvsRouterRequestCode(),
    entries,
    {
      uri: '/x.y',
      headers: { host: { value: 'x.test' } },
      cookies: { __dpl: { value: 'oldbuild-123' } },
    },
  );
  // asset, so the cookie should be honored and not overridden to newbuild
  assert.match(output.uri, /^\/builds\/oldbuild-123\//);
});
```

packages/hosting/src/constructs/kvs_router.test.ts
```diff
@@ -428,12 +431,108 @@ void describe('generated request fn — skew cookie gating', () => {
...
+
+void describe('generated functions — CloudFront 10 KB code-size limit', () => {
+  // CloudFront Functions reject code larger than 10 KB with an opaque HTTP 413
+  // at deploy time ("Internal error reported from downstream service"), which
+  // rolls the whole stack back. The generated source (including any in-function
+  // comments) ships verbatim, so a verbose comment can silently blow the budget.
+  // These guards fail loudly at unit-test time instead. Regression for the #245
+  // fix, whose first draft added a multi-line rationale comment INSIDE the
+  // function string and pushed it from 9.6 KB to 10.6 KB → 413 on deploy.
+  const LIMIT = 10 * 1024;
```
ahmedhamouda78: [nit] `10 * 1024` gives 10,240, which reads CloudFront's "10 KB" as binary KiB. The PR description uses decimal figures ("9,810 B"), so the actual enforced limit might be 10,000 bytes. If it is, and someone later pushes the function into the 10,001-10,240 range, this guard stays green but the deploy still fails with an opaque 413.

The function is 9,810 B today so the PR itself is fine either way. Worth confirming the real number against the [CloudFront Functions limits doc](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-functions-restrictions.html) and picking one:

```ts
const LIMIT = 10_000; // if the limit is decimal
```

or, if it really is 10,240, just drop a comment with the doc link so the next person doesn't have to figure it out again:

```ts
// https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/cloudfront-functions-restrictions.html
const LIMIT = 10 * 1024; // CloudFront hard limit: 10 KiB
```

The strict `<` is the right direction, so leave that alone.

ahmedhamouda78 (dismissed):
Nice fix. The root cause makes sense, and placing the override after the SPA/directory-index fallback is the right call so extensionless routes get covered too. The tests are solid: three regression cases that fail without the fix, plus the size guards that catch a CloudFront deploy failure at unit-test time instead of at rollout. The live E2E run on a real distribution is a good extra layer of confidence.

Two small things worth a look, neither blocking:

1. [minor] The `lastIndexOf` suffix check on the generated function misfires on 4-char URIs that don't contain `.html` (see inline comment). Low chance in practice, but it's a real defect and the fix is a one-liner.
2. [nit] Double-check whether the `LIMIT` constant should be 10,000 or 10,240 bytes and note the source (see inline comment).

packages/hosting/src/constructs/kvs_router.ts
osama-rizk: Good catch — that's a real inverted-bug edge case. Fixed in 53ea4af: switched to `uri.slice(-5) === '.html'` (kept ES5-style per the convention in the generated string). Added the regression test you suggested (`/x.y` still routes to the cookie build) — confirmed it fails against the old `lastIndexOf` form and passes with `slice(-5)`. The rationale now lives in the function's JSDoc so the in-function comment stays a single line (size budget).

packages/hosting/src/constructs/kvs_router.test.ts
osama-rizk: Confirmed against the docs: AWS labels the quota "10 KB" without disambiguating decimal vs binary, so I went conservative in 53ea4af — `const LIMIT = 10_000;` with the docs link inline. Erring low is deliberate: a too-tight guard costs a trivial test fix, whereas too-loose lets code in the 10,001–10,240 B range reach deploy and 413. Renamed the suite to "CloudFront function-size limit" and kept the strict `<`.

bobbor (dismissed):

packages/hosting/src/constructs/kvs_router.test.ts
```diff
@@ -428,12 +431,135 @@ void describe('generated request fn — skew cookie gating', () => {
...
+    });
+    const { output } = await runRequestFn(
+      generateKvsRouterRequestCode(),
+      entries,
+      reqWith('/x.y'),
+    );
+    assert.match(output.uri, /^\/builds\/oldbuild-123\//);
+  });
+});
+
```
sarayev, line 569: Two small gaps worth filling while this suite is here: (1) `skewEnabled: true` with no `__dpl` cookie at all — trivially correct, but makes the 'first-time visitor is unaffected' path explicit; (2) a basePath variant (e.g. `/app` basePath, request `/app/page.html`) to confirm the basePath strip and `.html` check compose correctly.

packages/hosting/src/constructs/kvs_router.ts
```diff
@@ -558,6 +558,32 @@ function stripBasePath(uri, bp) {
...
+ * function stamps `__dpl = meta.b` (the current build) on every HTML response.
+ * If the viewer-request function honored the cookie for HTML too, a returning
+ * visitor holding an old `__dpl` would be served the OLD build's HTML while
+ * their cookie advanced to the new build; the old HTML's content-hashed assets
+ * (which exist only under the old `builds/<id>/` prefix) would then rewrite to
+ * the new build prefix and 404 → a blank page. So HTML always resolves from
+ * `meta.b`, applied AFTER the SPA/directory-index fallback (which rewrites
+ * extensionless routes to `.../index.html`, so the `.html` suffix check covers
+ * the SPA shell too). Assets keep honoring the cookie so an already-loaded page
+ * stays consistent (old prefixes are retained via `prune: false`) until the
```
sarayev, line 575: Worth a quick note here: in-session HTML partial fetches (HTMX, Turbo Frames, etc.) whose URI ends in `.html` will also land on the current build rather than the visitor's cookie build. Probably fine for the apps this targets, but it's a subtle departure from the 'assets honor the cookie' invariant — a one-liner acknowledgement alongside the existing mid-session note would head off the question for future readers.

sarayev (commented):
A few small things after reading through the updated diff (the `slice(-5)` switch and the `/x.y` test are both great catches).

**`test/caching-regression.mjs` (not in diff):** Referenced in the PR description as passing 8/8 but the file isn't committed. If it's useful for future contributors to reproduce the live-distribution check — especially the cookie + `Accept-Encoding` cache-key dimension you mentioned — committing it to `test/` with run instructions would be a nice addition.

packages/hosting/src/constructs/kvs_router.test.ts
osama-rizk: Both added in 7caada6: (1) a `skewEnabled: true` case with no `__dpl` cookie (first-time/incognito visitor → current build), and (2) a basePath variant (`/app` basePath, request `/app/page.html`) asserting the basePath strip and the `.html` override compose — output is `/builds/newbuild/page.html` (basePath stripped, current build). Suite is green (829 tests).

packages/hosting/src/constructs/kvs_router.ts
osama-rizk: Good call — added a caveat in the viewer-request JSDoc (alongside the mid-session note) in 7caada6: an in-session HTML *partial* fetch (HTMX/Turbo Frames) ending in `.html` also lands on the current build rather than the cookie build. It's a deliberate, acceptable departure for the targeted apps — a partial is fetched fresh, not a build-pinned content-hashed asset. Kept it in the JSDoc (which doesn't ship to CloudFront) rather than the function body to stay under the size budget.

bobbor (approved):

ahmedhamouda78 (approved):

sarayev (approved):

Issue #269, opened 2026-07-27
Agent: Add prompt caching support for Bedrock Models


sQu4rks:
### Which area does this feature relate to?

Building Blocks (existing)

### Describe the feature you'd like to request

Currently the bb-agent wrapper over Strands does not support prompt caching. Especially for long-running agents with a lot of context this leads to increased bills since users pay the uncached prices for their models (i.e. $0.30 vs $3.00 per 1M input tokens for cache reads vs. uncached input with Claude Sonnet 4.6).

Strands supports prompt caching on Bedrock (https://strandsagents.com/docs/user-guide/concepts/model-providers/amazon-bedrock/#caching) - but the existing building block does not expose this as a configurable option.

### Use case

Agents with long system prompts, many tools, or multi-turn conversations re-send a large stable prefix on every request. Without prompt caching that prefix is billed at full input price each time. On Claude Sonnet 4.6 that's $3.00/1M instead of $0.30/1M for a cache read (10x), plus higher latency.

Strands already supports Bedrock prompt caching on its `BedrockModel`, but bb-agent's `ModelConfig` doesn't expose it, so there's no easy way to turn it on from an AWS Blocks app without modifying the local code after installing the package.

### Proposed solution

Add an optional `cacheConfig` field to bb-agent's `ModelConfig` (bedrock only) and pass it into the Strands `BedrockModel`. Off by default, so no breaking change.

```ts
export interface ModelConfig {
  // ...existing fields
  /** Prompt caching for the bedrock provider. Optional so off by default. */
  cacheConfig?: { strategy: 'auto' | 'anthropic' };
}
```

In `createStrandsModel`'s bedrock branch:

```ts
return new BedrockModel({
  modelId: config.modelId,
  ...(config.inferenceConfig && { /* ... */ }),
  ...(config.cacheConfig && { cacheConfig: config.cacheConfig }),
});
```

Usage:

```ts
model: { deployed: { ...BedrockModels.BALANCED, cacheConfig: { strategy: 'auto' } } }
```

`'auto'` lets Strands place cache points for known Bedrock model IDs (incl. the `BedrockModels` presets); `'anthropic'` force-enables it for ARN application inference profiles that `'auto'` can't detect. This mirrors the implementation in the strands typescript SDK. Ignored by the `openai-api` and `canned` providers.

Reference: Strands' own caching support — https://strandsagents.com/docs/user-guide/concepts/model-providers/amazon-bedrock/#caching

[...]

PR #270, opened 2026-07-27
Feat: Add prompt caching to bedrock models 

sQu4rks:
## Problem

In the current implementation of the `Agent` block in `bb-agent`, the block does not support using prompt caching for bedrock-based models. 

**Issue #, if available:** #269

## Changes

Adds an optional `cacheConfig` parameter that mirrors the configuration from strands SDK.

```typescript
const agent = new Agent(scope, 'support', {
	model: {
		deployed: { ...BedrockModels.BALANCED, cacheConfig: { strategy: 'auto' } },
		local: { provider: 'canned' },
	},
	systemPrompt: // ...
});
```

## Validation

Added two unit tests to verify passing behavior. 

Additionally, tested by writing a simple chat application that uses the below snippet 

```typescript
const agent = new Agent(scope, 'support', {
  model: {
    deployed: { ...BedrockModels.BALANCED, cacheConfig: { strategy: 'auto' } },
    local: { provider: 'canned' }, // no AWS/LLM needed for local dev
  },
  systemPrompt: // very long system prompt
  tools: // tools definition
});
```
to create a chat agent. Enabled invocation logs on Amazon Bedrock and the logs show that the first turn writes to the cache

```json
"usage": {
  "inputTokens": 3,
  "cacheReadInputTokens": 0,
  "cacheWriteInputTokens": 3317,
  "outputTokens": 377,
  "totalTokens": 3697
}
```

and the subsequent turns read from it 

```json
"usage": {
  "inputTokens": 3,
  "cacheReadInputTokens": 3317,
  "cacheWriteInputTokens": 406,
  "outputTokens": 530,
  "totalTokens": 4256
}
```

## Checklist

- [x] PR description included

---


packages/bb-agent/src/types.ts
```diff
@@ -21,6 +21,30 @@ export interface ModelConfig {
...
+	 * Prompt caching for the `bedrock` provider. Off by default. Caching reuses the
+	 * cached prefix (tools + system prompt + prior turns) across requests, cutting
+	 * input-token cost and latency on multi-turn or long-system-prompt agents.
+	 * Ignored by the `openai-api` and `canned` providers.
+	 *
+	 * - `{ strategy: 'auto' }` — let Strands place cache points at optimal positions
+	 *   (after tools, after the last user message) for known Bedrock model IDs. This is
+	 *   the recommended default for the {@link BedrockModels} presets.
+	 * - `{ strategy: 'anthropic' }` — force-enable Anthropic-style caching. Use this
+	 *   when `modelId` is an ARN application inference profile, where `'auto'` cannot
```
Simone319, line 34: What happen if the strategy is 'anthropic' but the model id is not anthropic? Shall we place a check on modelId for this strategy?

sQu4rks: Thanks for the comments @Simone319 

I believe the `strategy: anthropic` is specifically designed for cases where - from the ARN - it's unclear that the underlying model is an anthropic model. The ARN could look something like this `arn:aws:bedrock:{region}:{account-number}:inference-profile/my-team-profile` ([docs](https://docs.aws.amazon.com/bedrock/latest/userguide/cost-mgmt-application-inference-profiles.html)) so we'd need another API call (and accompanying permissions) to retrieve the underlying model (via `GetInferenceProfile` ([api docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_GetInferenceProfile.html)) to then parse the models.

I also checked the case quickly via a script (see below). Since `cacheConfig` is passed to the Strands `BedrockModel` it's behavior is entirely strands.

```js
import { Agent } from '@strands-agents/sdk';
import { createStrandsModel } from '../../dist/model-factory.js';

process.env.AWS_REGION ??= 'us-east-1';
const MODEL_ID = process.env.MODEL_ID ?? 'openai.gpt-oss-120b-1:0'; // a non-Anthropic model

// Long, stable prefix so we clear the ~1,024-token cache minimum.
const SYSTEM_PROMPT = [
  'You are "Atlas", a meticulous senior support engineer for the ACME Cloud platform.',
  'Follow these operating rules on every single response, without exception:',
  ...Array.from({ length: 40 }, (_, i) =>
    `Rule ${i + 1}: Be precise, cite the relevant ACME Cloud service by name, prefer the ` +
    'least-privilege remediation, never invent API names, and always end with a one-line summary.'),
].join('\n');

const TURNS = [
  'A user reports 503s from EdgeCDN after a deploy. First triage step?',
  'Now they also see elevated latency on ObjectStore. What next?',
];

const usageOf = (r) => {
  const u = r.metrics?.accumulatedUsage ?? {};
  return { input: u.inputTokens ?? 0, output: u.outputTokens ?? 0, cacheWrite: u.cacheWriteInputTokens ?? 0, cacheRead: u.cacheReadInputTokens ?? 0 };
};

async function runScenario(label, cacheConfig) {
  console.log(`\n--- ${label} ---`);
  try {
    const model = await createStrandsModel({ provider: 'bedrock', modelId: MODEL_ID, cacheConfig });
    const agent = new Agent({ model, systemPrompt: SYSTEM_PROMPT, printer: false });
    for (let i = 0; i < TURNS.length; i++) {
      const u = usageOf(await agent.invoke(TURNS[i]));
      console.log(`  turn ${i + 1}: input=${u.input} output=${u.output} cacheWrite=${u.cacheWrite} cacheRead=${u.cacheRead}`);
    }
  } catch (err) {
    console.log(`  ERROR: ${err?.name ?? 'Error'}: ${err?.message ?? err}`);
  }
}

// 'auto'      → Strands warns and disables caching for a non-Anthropic model
await runScenario("strategy: 'auto'", { strategy: 'auto' });
// 'anthropic' → forces Anthropic-style cache points onto a non-Anthropic model
await runScenario("strategy: 'anthropic'", { strategy: 'anthropic' });
```

With `strategy: 'anthropic'` on a non-Anthropic model, Strands forwards the Anthropic-style cache points and Bedrock rejects the request with a clear `ModelError` ("You invoked an unsupported model or your request did not allow prompt caching"). 

```
--- strategy: 'auto' ---
model_id=<openai.gpt-oss-120b-1:0> | cache_config is enabled but this model does not support automatic caching
  turn 1: input=1679 output=533 cacheWrite=0 cacheRead=0
  turn 2: input=3375 output=1558 cacheWrite=0 cacheRead=0

--- strategy: 'anthropic' ---
  ERROR: ModelError: You invoked an unsupported model or your request did not allow prompt caching. See the documentation for more information.
```

So my suggestion would be to handle the warnings/errors but let Bedrock do the verification.

Simone319: Since `strategy: anthropic` requires Anthropic model, and will break at runtime otherwise, it's worth mentioning it in the `cacheConfig` type as well as in the README, so that users know how it use it correctly.

.changeset/bb-agent-prompt-caching.md
```diff
@@ -0,0 +1,17 @@
+---
+"@aws-blocks/bb-agent": minor
+---
+
+feat(bb-agent): opt-in Bedrock prompt caching via `ModelConfig.cacheConfig`
+
+Agents can now enable Strands prompt caching for the `bedrock` provider by
+setting `cacheConfig: { strategy: 'auto' | 'anthropic' }` on a model config.
```
costleya, line 8: This should not block the PR, but GPT 5.6 supports explicit prompt caching as well: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-openai. Maybe making a note of this or adding another feature request for follow-up would be appropriate.

PR #271, opened 2026-07-27
fix(release): rewrite workflow + add release runbook

svidgen:
## What

Rewrites the Release workflow to eliminate the broken PR creation step and adds a documented release runbook.

### Workflow rewrite (`.github/workflows/publish-packages.yml`)

Drops `changesets/action` entirely. The workflow now has two explicit modes:

**Version mode** (changesets exist on main):
- Runs `npm run version` (changeset version + lock file update)
- Commits and force-pushes `changeset-release/main`
- Emits a Step Summary with a prefilled link to create the release PR

**Publish mode** (no changesets — version PR was just merged):
- Runs `npm run release` (build + changeset publish)
- Creates git tags
- Emits a success summary

No PR creation is attempted. No token/permission issues. No red workflows on main.

### docs/RELEASING.md (new)

Full release runbook for maintainers:
1. Click the prefilled link to create a PR from `changeset-release/main` → `main`
2. Get 2 approvals
3. Close/reopen to trigger CI if needed, merge after checks pass
4. npm publish fires automatically on merge

Also covers: pre-1.0 semver conventions, troubleshooting, emergency manual publish.

### AGENTS.md

Added `docs/RELEASING.md` to "Where to look" — releases must be managed by the AWS Blocks team.

## Why

The previous workflow failed on every push to main because:
1. `changesets/action` `version` input didn't support shell operators — fixed in #265
2. GitHub Actions can't create PRs on this repo (permission disabled, TT V2300286150 pending)

Rather than fight permission issues with PATs or GitHub Apps, we simplify: the workflow handles the automated parts (version bumping, publishing) and leaves the human-gated step (PR review) to humans via a one-click prefilled link.

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -66,7 +66,7 @@ jobs:
           echo "" >> "$GITHUB_STEP_SUMMARY"
           echo "The \`changeset-release/main\` branch has been updated with version bumps." >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
-          echo "**Next step:** [Create the release PR](https://github.com/aws-devtools-labs/aws-blocks/compare/main...changeset-release/main?expand=1&title=chore%3A+version+packages&body=Release+PR.+Steps+to+complete%3A%0A%0A1.+Get+2+approvals%0A2.+Close+and+reopen+to+trigger+CI+%28if+checks+aren%27t+running%29%0A3.+Merge+after+checks+pass+%E2%80%94+npm+publish+runs+automatically%0A%0ASee+%5Bdocs%2FRELEASING.md%5D%28https%3A%2F%2Fgithub.com%2Faws-devtools-labs%2Faws-blocks%2Fblob%2Fmain%2Fdocs%2FRELEASING.md%29+for+details.)" >> "$GITHUB_STEP_SUMMARY"
+          echo "**Next step:** [Create the release PR](https://github.com/aws-devtools-labs/aws-blocks/compare/main...changeset-release/main?expand=1\&title=chore%3A+version+packages\&body=Release+PR.+Steps+to+complete%3A%0A%0A1.+Get+2+approvals%0A2.+Merge+after+checks+pass+%E2%80%94+npm+publish+runs+automatically%0A%0ASee+%5Bdocs%2FRELEASING.md%5D%28https%3A%2F%2Fgithub.com%2Faws-devtools-labs%2Faws-blocks%2Fblob%2Fmain%2Fdocs%2FRELEASING.md%29+for+details.)" >> "$GITHUB_STEP_SUMMARY"
           echo "" >> "$GITHUB_STEP_SUMMARY"
           echo "See [docs/RELEASING.md](https://github.com/aws-devtools-labs/aws-blocks/blob/main/docs/RELEASING.md) for the full process." >> "$GITHUB_STEP_SUMMARY"
 
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -1,10 +1,10 @@
 name: Release
 
-# Version + publish via changesets/action. On push to main it either:
-#   1. Runs `changeset version`, pushes the changeset-release/main branch
-#   2. If no changesets remain (version PR was just merged), publishes to npm
+# On push to main:
+#   - If changesets exist: run `changeset version`, push changeset-release/main
+#   - If no changesets remain (version PR was just merged): publish to npm
 #
-# PR creation is handled manually — see docs/RELEASING.md for the full process.
+# PR creation is manual — see docs/RELEASING.md.
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-packages.yml
```diff
@@ -42,36 +38,53 @@ jobs:
...
-        with:
-          version: npm run version
-          publish: npm run release
-          commit: 'chore: version packages'
-          title: 'chore: version packages'
+      # Check if there are pending changesets
+      - name: Check for changesets
+        id: check
+        run: |
+          if npx changeset status --since=origin/main 2>/dev/null | grep -q "NO packages"; then
```
osama-rizk, line 45: Can we double-check this mode detection? `changeset status` prints a separate line per bump level, e.g. with a patch-only changeset present it still outputs `NO packages to be bumped at minor` and `NO packages to be bumped at major`. So `grep -q "NO packages"` looks like it would match even when changesets *do* exist — which would route to publish mode and skip the version step. And since `block-major` forbids `major`, the `NO packages ... at major` line is always there. Have we confirmed version mode actually fires on a normal patch/minor release? Separately: on a push to `main` the checkout is at the pushed commit, so `--since=origin/main` is effectively HEAD-vs-HEAD (no diff). Would a disk-based check (e.g. `find .changeset -name "*.md" ! -name README.md | grep -q .`) be more robust than parsing the status text?

.github/workflows/publish-packages.yml
```diff
@@ -42,36 +38,53 @@ jobs:
...
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
           NPM_CONFIG_PROVENANCE: 'true'
 
-      # The Version PR is created with GITHUB_TOKEN, so it doesn't auto-trigger
-      # CI. Append operator instructions to the PR body.
-      - name: Add release instructions to Version PR
-        if: steps.changesets.outputs.pullRequestNumber
+      - name: Create git tags and GitHub releases
+        if: steps.check.outputs.has_changesets == 'false'
+        run: npx changeset tag
```
osama-rizk, line 83: `changeset tag` creates local git tags only — do we need a `git push --tags` (or `--follow-tags` on the publish commit) for them to reach origin? The previous workflow pushed with `--follow-tags`. Also, `changeset tag` does not create GitHub Releases (that was `changesets/action`'s `createGithubReleases`), so the step name and the `GITHUB_TOKEN` env look inaccurate here — should this be renamed to "Create git tags", or add a real `gh release create` step if we want Releases? (RELEASING.md line 48 promises tags + Releases are created automatically.)

.github/workflows/publish-packages.yml
```diff
@@ -42,36 +38,53 @@ jobs:
...
+          echo "The \`changeset-release/main\` branch has been updated with version bumps." >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "**Next step:** [Create the release PR](https://github.com/aws-devtools-labs/aws-blocks/compare/main...changeset-release/main?expand=1)" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "See [docs/RELEASING.md](https://github.com/aws-devtools-labs/aws-blocks/blob/main/docs/RELEASING.md) for the full process." >> "$GITHUB_STEP_SUMMARY"
+
+      # Mode 2: Publish — no changesets means the version PR was just merged
+      - name: Publish to npm
+        if: steps.check.outputs.has_changesets == 'false'
+        run: npm run release
```
osama-rizk, line 76: Same open question as before, now that this job is being rewritten: `npm run release` is `build && changeset publish` with no test run between the release-PR merge and the npm publish. Do we want a verification step (e.g. `test:e2e:local` / unit tests) right before publish, so we validate the exact commit being shipped? Or is relying on the release PR's CI + review the intended pre-1.0 tradeoff?

docs/RELEASING.md
```diff
@@ -0,0 +1,79 @@
...
+While packages are pre-1.0:
+- `patch` (0.1.1 → 0.1.2): non-breaking change
+- `minor` (0.1.x → 0.2.0): **breaking** change
+- `major`: blocked by CI (`block-major` guard) — leaving pre-release requires explicit sign-off
+
+## Troubleshooting
+
+### Release workflow shows red on `main`
+
+The workflow may fail to auto-create the release PR due to GitHub token permissions. This is cosmetic — the `changeset-release/main` branch is still updated successfully. Create the PR manually using the link above.
```
osama-rizk, line 61: This troubleshooting entry describes the workflow failing to auto-create the release PR — but this PR removes PR auto-creation entirely, so does this failure mode still apply? Might be worth dropping or rewording so the runbook matches the new behavior (the workflow no longer touches PRs).

svidgen:
Closing — we'll wait for the TT (V2300286150) to enable the Actions PR permission, or set up a PAT if needed. No new docs or process changes for now. The team knows how to open the release PR manually from `changeset-release/main` if needed in the interim.

Issue #185, opened 2026-07-10
Custom auth forms silently lose UI reactivity: setAuthState() doesn't drive onAuthChange/AuthenticatedContent unless you also call broadcastAuthChange() — add a submitAuthAction() helper


hfurkanbozkurt:
## Problem

Building a **custom auth form** (calling `setAuthState()` directly instead of using the `Authenticator` component) silently breaks UI reactivity: `AuthenticatedContent` and every `onAuthChange` subscriber never re-render after a successful sign-in, so gated UI (dashboards, post-login views) stays hidden. There is no error, no warning, and no type signal — the code looks correct and the auth call succeeds server-side.

The root cause is that `setAuthState()` is a thin server-side state-machine advance that drives **neither** of `auth-common`'s two notification systems. Today the `Authenticator` component works out of the box only because it *manually* pairs every `setAuthState()` with the internal cache update **and** `broadcastAuthChange()`. A custom form has no way to fully replicate that, because the cache-update half (`updateState`) is not exported — the best a diligent author can do is call `broadcastAuthChange()`, and even that requirement is only discoverable by reading source.

## Mechanism — affected files + lines (verified against current `main`)

All references are `packages/auth-common/src/ui.ts`.

`setAuthState()` is declared on `AuthStateApi` (L107-110) and implemented by the generated API client from `auth.createApi()`. It advances server state and returns the new `AuthState`; it touches neither the shared cache nor the broadcast channel:

```ts
export interface AuthStateApi {
	getAuthState(): Promise<AuthState>;
	setAuthState(input: AuthActionInput): Promise<AuthState>;
}
```

There are **two independent notify paths**, and `setAuthState()` fires neither:

1. **Internal shared-state cache** — `updateState(api, state)` (L168). This function is **module-private (not exported)** and is called only from inside `Authenticator`.
2. **Cross-component broadcast** — `broadcastAuthChange(user)` (L191-197) posts to the `BroadcastChannel` (other tabs) and dispatches a same-window `CustomEvent`:

```ts
export function broadcastAuthChange(user: AuthUser | null): void {
	const detail = { type: 'auth-change', user };
	getChannel().postMessage(detail);
	window.dispatchEvent(new CustomEvent(AUTH_LOCAL_EVENT, { detail }));
}
```

`onAuthChange(api, cb)` (L232-246) subscribes to exactly those two events — so it reacts **only** to a `broadcastAuthChange()` call (or another tab's):

```ts
[...]

PR #272, opened 2026-07-28
fix(auth-common): add stable data-testid hooks to the auth UI components

### LINKED ISSUE #239: Authenticator component ships no stable data-testids — e2e tests can't target sign-in UI
@aws-blocks/auth-common (Authenticator UI) / @aws-blocks/blocks

### Describe the problem
The shipped `Authenticator` renders no stable, documented test hooks (data-testids) on its fields, submit button, or state containers. e2e tests have nothing reliable to target, so teams hand-roll their own auth form to get selectors - and that path silently breaks reactivity (see #185: must also call broadcastAuthChange). Net result: the authenticated experience fails and gated UI never appears, no error surfaced.

### Impact
No stable selector contract on the one component that gates the whole app. Teams either can't e2e-test the built-in Authenticator or fork it and hit #185. In agent benchmarks this produced a whole-app 0-score on auth-gated tasks.

### Proposed fix
Add documented stable data-testids (e.g. authenticator-username, authenticator-password, authenticator-submit, authenticator-error, authenticator-signed-in) OR a testIds/slotProps API to inject them; document the selector contract.

### Related
#185 (custom-form reactivity/broadcastAuthChange) - the reason people fork Authenticator.

ahmedhamouda78:
Fixes #239

The built-in `Authenticator` rendered no test hooks, so e2e suites had nothing reliable to target and teams ended up re-implementing the sign-in form just to get selectors (which then breaks reactivity, see #185).

Every element the auth components render now carries a stable `data-testid`:

- `authenticator` on the container, plus `authenticator-heading`, `authenticator-error` and `authenticator-signed-in`
- `authenticator-action-<action>` on each rendered action, both internal forms and external OAuth/SAML forms
- `authenticator-<field>` on every input, named after the field the server declared (`authenticator-username`, `authenticator-password`, `authenticator-code`, ...), hidden inputs included so session tokens and echoed usernames stay assertable
- `authenticator-submit` on the submit button
- `authenticated-content` on `AuthenticatedContent`, and `account-menu*` hooks on `AccountMenuBar` so a test can open the sign-in modal and sign out again

Field and submit hooks repeat once per action, because the signed-out state renders `signIn` and `signUp` side by side and both declare a `username`. The action wrapper is the scope for that, and the docs show it with a Playwright example.

Attributes only, no behavior change and no API surface change. The contract lives in CUSTOMIZING-AUTH-UI.md, linked from the top of that guide, and is summarised in the `Authenticator` JSDoc.

### Testing

- 8 new cases in `packages/auth-common/src/ui.test.ts`: container/action/field/submit/error/signed-in hooks, hooks on hidden fields and on external provider forms, driving a full sign-in through the scoped hooks, and the `AccountMenuBar` + `AuthenticatedContent` hooks
- `npm run build:packages` clean
- `npm run test:unit` green across all workspaces, `@aws-blocks/auth-common` at 61 passed / 0 failed
- `npm run check:api` reports the API surface unchanged
- changeset guard passes validate-structure, verify-coverage and block-major

### Changeset scope

The changeset bumps `@aws-blocks/auth-common` and `@aws-blocks/blocks` (both patch). The umbrella is included on purpose even though it has no file changes here: it re-exports `Authenticator`, `AuthenticatedContent` and `AccountMenuBar` from `@aws-blocks/auth-common/ui`, and sibling patch releases stay inside its caret ranges, so `changeset version` never bumps it on its own. That is what left it stale in #212 and forced the corrective bump in b920f9e, so it is republished in step with the components it hands to consumers.

### DIFF
--- a/.changeset/authenticator-test-hooks.md
+++ b/.changeset/authenticator-test-hooks.md
@@ -3,6 +3,6 @@
 "@aws-blocks/blocks": patch
 ---
 
-Add stable `data-testid` hooks to the auth UI components so e2e suites can target the shipped `Authenticator` instead of forking it. Covers the container, heading, error, per-action wrappers, every field input, submit buttons, the signed-in marker, `AuthenticatedContent`, and `AccountMenuBar`. The full selector contract is documented in CUSTOMIZING-AUTH-UI.md.
+Add stable `data-testid` hooks to the auth UI components so e2e suites can target the shipped `Authenticator` instead of forking it. Covers every interactive element (field inputs, hidden ones included, and submit buttons) plus the containers worth asserting against: the root, per-action wrappers, heading, error, the signed-in marker, `AuthenticatedContent`, and `AccountMenuBar`. Presentational markup such as hint text and layout wrappers stays unhooked. The full selector contract, including the `<action>:<provider>` form federated actions produce, is documented in CUSTOMIZING-AUTH-UI.md.
 
 The `@aws-blocks/blocks` umbrella package receives a `patch` because it re-exports these components from `@aws-blocks/auth-common/ui`. Sibling patch releases stay inside the umbrella's caret ranges, so `changeset version` never bumps it on its own (#212), and it is republished explicitly to stay in step with the components it hands to consumers.
--- a/.changeset/authenticator-test-hooks.md
+++ b/.changeset/authenticator-test-hooks.md
@@ -1,5 +1,8 @@
 ---
 "@aws-blocks/auth-common": patch
+"@aws-blocks/blocks": patch
 ---
 
 Add stable `data-testid` hooks to the auth UI components so e2e suites can target the shipped `Authenticator` instead of forking it. Covers the container, heading, error, per-action wrappers, every field input, submit buttons, the signed-in marker, `AuthenticatedContent`, and `AccountMenuBar`. The full selector contract is documented in CUSTOMIZING-AUTH-UI.md.
+
+The `@aws-blocks/blocks` umbrella package receives a `patch` because it re-exports these components from `@aws-blocks/auth-common/ui`. Sibling patch releases stay inside the umbrella's caret ranges, so `changeset version` never bumps it on its own (#212), and it is republished explicitly to stay in step with the components it hands to consumers.
--- a/.changeset/authenticator-test-hooks.md
+++ b/.changeset/authenticator-test-hooks.md
@@ -1,0 +1,5 @@
+---
+"@aws-blocks/auth-common": patch
+---
+
+Add stable `data-testid` hooks to the auth UI components so e2e suites can target the shipped `Authenticator` instead of forking it. Covers the container, heading, error, per-action wrappers, every field input, submit buttons, the signed-in marker, `AuthenticatedContent`, and `AccountMenuBar`. The full selector contract is documented in CUSTOMIZING-AUTH-UI.md.
--- a/packages/auth-common/CUSTOMIZING-AUTH-UI.md
+++ b/packages/auth-common/CUSTOMIZING-AUTH-UI.md
@@ -201,9 +201,16 @@ handling and still get your custom look.
 
 ## Test hooks
 
-Every built-in auth component renders stable `data-testid` attributes, so e2e
-suites can drive the shipped UI instead of forking it for selectors. Treat the
-names as public API: renaming one is a breaking change.
+Every interactive element the built-in auth components render carries a stable
+`data-testid`, so e2e suites can drive the shipped UI instead of forking it for
+selectors. Treat the names as public API: renaming one is a breaking change.
+
+Inputs, buttons, and the containers you assert against are hooked. Purely
+presentational markup is not, on purpose: hint text from `fields.<name>.hint`,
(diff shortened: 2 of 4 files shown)


osama-rizk (commented):
Clean, complete, and unusually well-documented for what could have been a careless attribute sprinkle. I verified the one claim that matters and the hard part is both documented and tested. Nothing blocking.

## What I verified against source

The claim "every element gets a hook" is the thing that, if false, leaves exactly the gap this PR exists to close (a fork-forcing un-targetable element). I enumerated all 9 `createElement('button'|'input')` sites in `ui.ts` on the branch — all 9 carry a testid. I also checked the **WebAuthn/passkey path** (`webauthn-get`/`webauthn-create`), my prime suspect for a special-cased button that skips the pattern: it *reuses* the standard `authenticator-submit` button and the standard hidden `credential` input, so it inherits the hooks rather than rendering its own untagged control. No interactive element escapes. ✅

The duplicate-testid problem is handled the right way. The signed-out state renders `signIn` and `signUp` side by side and both declare `username`, so `authenticator-username` is necessarily non-unique — and Playwright's `getByTestId` runs in strict mode, throwing on multiple matches. The docs name this, explain why, and show the scoped pattern; and the test `field and submit hooks are unique within an action wrapper` asserts `querySelectorAll(...).length === 1` *within* the wrapper, which is exactly the invariant scoped `locator.getByTestId` relies on. So the contract's usability is tested, not just its presence — this is the part most testid PRs get wrong.

Stable-outer / swapped-inner is also correct: `AccountMenuBar` and `AuthenticatedContent` put the testid on the outer container that persists across re-renders and swap the inner content, so a handle survives a sign-in/sign-out transition instead of going stale (the reactivity trap from #185).

## 1. "Every element the auth components render" is overstated

The PR body says *every element* carries a testid, but several non-interactive nodes don't: the field hint div, the inner `bar` div in `AccountMenuBar`, the modal content wrapper, and the per-field input-wrapper divs. None matter for e2e (you target the input, not its wrapper), so this is purely wording — but "every element" invites a reader to expect a hook on the hint text and be surprised. Soften to "every interactive element" / "every element you'd target."

## 2. External-provider wrapper names include the `:provider` suffix

An external action named `signIn:google` yields `authenticator-action-signIn:google`. The colon is fine for both `querySelector('[data-testid="…"]')` (quoted attr selector, tested) and Playwright's `getByTestId` (exact string match, not CSS parse), so no bug. Worth a one-line note in the doc table that external-provider wrappers carry the `:provider` suffix — the example only shows the internal `authenticator-action-signIn`, and a reader might not predict the external form's name.

## Security note (not a finding)

Adding `data-testid` to hidden inputs that carry session tokens / echoed usernames caught my eye, but the value is already in the DOM as a hidden input regardless — a testid makes it selectable, not more exposed, and it's same-origin. That selectability is the intended feature. No concern; flagging only because "testid on a token field" looks like a leak and isn't.

## Nit
- Tests run under JSDOM with `querySelector`, so they verify DOM structure, not Playwright strict-mode behavior directly. The within-wrapper uniqueness assertion is the right proxy — just noting the tests prove the structure the docs promise, not the Playwright API itself.

ahmedhamouda78:
[minor] Fair, and you are right that the claim did not survive contact with the code. I went through every `createElement` in `ui.ts`: buttons, inputs and forms all do carry a hook, but four elements do not, and they are all presentational. The bordered panel inside the container (583), the hint div (766), the `AccountMenuBar` flex row (1045), and the modal's inner content box (1077). So "every element" invited exactly the bug hunt you describe. Reworded to "every interactive element" in the guide, the `Authenticator` JSDoc and the changeset, and added a short paragraph naming those four as deliberately unhooked, with the reasoning that a selector on them would pin our styling structure into someone's test suite rather than their app's behavior.

One small correction to your list: there are no per-field input-wrapper divs. Fields append straight onto the action wrapper, so the nearest thing is that bordered panel. I listed the four that actually exist so the doc does not send anyone looking for a wrapper that is not there.

[minor] Also fair, and the gap was purely documentation. `renderInternalAction` (669) and `renderExternalAction` (993) both interpolate `action.name` verbatim, so a federated action named `signIn:google` comes out as `authenticator-action-signIn:google`. The `<action>:<provider>` convention was already in DESIGN.md in the OIDC and mixed form/federated flows, it just never made it into the hook contract. Added it to the table row and to the naming paragraph. I also called out the one place the colon does bite, which is that the attribute value has to stay quoted: `[data-testid="authenticator-action-signIn:google"]` is fine and `getByTestId` is fine, but a bare `[data-testid=authenticator-action-signIn:google]` is invalid because an unquoted attribute value has to be a CSS identifier.

Since that composite name is now documented contract, I added a test for it: two providers get distinct hooks, scoping through a suffixed wrapper works like any other action, and the bare `signIn` hook matches exactly rather than by prefix, so a form-flow test cannot accidentally land on an OAuth button. Verified it fails if the suffix is stripped rather than just passing by construction.

`b3cdab2`. auth-common at 62 passed / 0 failed, changeset guards and `check:api` clean, full CI green.

soberm (approved):

sarayev (approved):

PR #274, opened 2026-07-28
fix(ci): require the umbrella changeset entry when a re-exported sibling ships

### LINKED ISSUE #273: Umbrella @aws-blocks/blocks has to be added to changesets by hand whenever a re-exported sibling is patched
## What happens

`@aws-blocks/blocks` is the umbrella package. It re-exports `@aws-blocks/core`, every `bb-*` block and `@aws-blocks/auth-common/ui`, and its published tarball also carries `docs/`, assembled from sibling READMEs at pack time by `scripts/sync-block-docs.mjs`. It pins all of those siblings with caret ranges (`^0.1.10`, `^0.3.0`, and so on), so a sibling patch release stays inside the range and `changeset version` never bumps the umbrella.

Verified on current main: `@aws-blocks/core` is at `0.1.16` while `packages/blocks/package.json` still declares `"@aws-blocks/core": "^0.1.10"`, so the range has not been rewritten across six core patch releases. Reproduced with `changeset version` in a scratch worktree of main, with a single changeset bumping `@aws-blocks/auth-common` as patch: auth-common goes to `0.1.4`, blocks stays at `0.2.5`.

When the umbrella's packed content does move, the version-exists guard in `scripts/publish/publish.ts` fails the whole release run, not just that one package:

```
Error: @aws-blocks/blocks@0.2.2 already exists with different content. Did you forget to include it in your changeset?
    at publish (scripts/publish/publish.ts:313:12)
```

## It has already bitten us twice

- **#212 / `fc8cfad`**, "fix(release): republish @aws-blocks/blocks umbrella (missing changeset)". Publishing was blocked for every package in the run until a changeset containing nothing but `"@aws-blocks/blocks": patch` was merged. Worth noting for the fix: in the window between blocks `0.2.2` being released (`299b48c`) and the failing run (`2c6e2ef`), no `packages/*/README.md` and no file under `packages/blocks/` changed at all. The drift was not something a file diff at PR time could have seen.
- **`b920f9e`**, "chore: bump blocks to publish". A one line commit that retroactively added `"@aws-blocks/blocks": patch` to an already open changeset (`docs-bb-agent-readme-fixes.md`), caught just before that release went out.

## Why the current gates do not catch it

- `Blocks Package Integrity` (added in #186, relaxed to head vs base in #189) packs the umbrella at the PR head and at the PR base and compares hashes. That covers `src` and the README derived `docs`, but it packs with `--ignore-scripts`, so the built `dist` is in neither tarball, and by design it only sees content changes introduced by this PR's own diff. A sibling *release* is not a diff in the PR that patched the sibling, which is exactly the #212 shape.
- `changeset-guard verify-coverage` only demands a changeset for packages with file changes. In these PRs the umbrella has none, so it stays silent.

#212 explicitly left this open as a follow up: "Prevent recurrence by making the umbrella auto-bump on any sibling release, e.g. a `fixed` group in `.changeset/config.json`, or a pack-time assertion that the umbrella was bumped when a re-exported dep changes." `.changeset/config.json` still has `"fixed": []`.

## Latest instance

PR #272 (stable `data-testid` hooks in `@aws-blocks/auth-common`) had the umbrella entry added by hand during review. Nothing in CI asked for it, and the umbrella happened to be covered by two unrelated pending changesets, so the omission would have been invisible.

## Config options, measured

`changeset version` run against a scratch worktree of main with one sibling-only changeset (`"@aws-blocks/auth-common": patch`):

| `.changeset/config.json` | result |
|---|---|
| today (`fixed: []`, `linked: []`) | auth-common `0.1.4`, blocks stays `0.2.5`. The bug. |
| `linked: [[blocks, ...its deps]]` | blocks does get bumped, but auth-common leaps `0.1.3` to `0.3.3`, because `linked` forces the whole group onto one shared version |
| `fixed: [[blocks, ...its deps]]` | all 21 group members go to `0.3.3` and ship on every release |

So neither config switch is usable here. Both rewrite sibling version numbers and blow through the pre-1.0 convention documented at the top of `changeset-guard.ts` (patch is non-breaking, minor is breaking, major needs sign-off).

## Proposal

Keep versioning as it is and make the requirement explicit in CI: a `verify-umbrella` subcommand in `scripts/changeset-guard.ts`, wired into `changeset-check.yml`. If a changeset touched by this PR bumps a package that `packages/blocks/package.json` depends on, require that some pending changeset bumps `@aws-blocks/blocks`, and fail with the exact line to add. Scoping the trigger to the PR's own changesets keeps it from blaming unrelated PRs, which is the trap #189 had to undo for the integrity gate, and accepting coverage from any pending changeset matches what actually matters: whether the next release republishes the umbrella.

Happy to send the PR for this.

ahmedhamouda78:
Fixes #273

`@aws-blocks/blocks` re-exports core, every `bb-*` block and `auth-common/ui`, and ships a `docs/` folder assembled from sibling READMEs at pack time. It pins those siblings with caret ranges, so when a sibling gets a patch release the new version stays in range and `changeset version` leaves the umbrella alone. Its packed content still moves, and `scripts/publish/publish.ts` then aborts the entire release run with "already exists with different content". That is what happened in #212 (`fc8cfad`), and `b920f9e` was the same omission caught one commit earlier. #212 asked for recurrence prevention and it was never added.

### Why not `linked` or `fixed`

I ran `changeset version` in a scratch worktree of main with a single sibling-only changeset (`"@aws-blocks/auth-common": patch`) against all three configs:

| config | auth-common | blocks | core | bb-kv-store |
|---|---|---|---|---|
| today (`fixed: []`, `linked: []`) | 0.1.4 | 0.2.5 (not bumped) | 0.1.16 | 0.1.4 |
| `linked: [[blocks, ...deps]]` | **0.3.3** | 0.3.3 | 0.1.16 | 0.1.4 |
| `fixed: [[blocks, ...deps]]` | **0.3.3** | 0.3.3 | **0.3.3** | **0.3.3** |

`linked` does pull the umbrella in, but it drags auth-common from 0.1.3 to 0.3.3 because the group is forced onto one shared version, and `fixed` does that to all 21 members on every release. Both rewrite sibling version numbers and blow past the pre-1.0 convention documented at the top of `changeset-guard.ts`, so neither is usable. Versioning stays as it is and the requirement becomes explicit in CI instead.

### What this adds

`changeset-guard.ts verify-umbrella`, wired into `changeset-check.yml`. If a changeset **touched by this PR** bumps a package listed in `packages/blocks/package.json` dependencies, some pending changeset has to bump `@aws-blocks/blocks`, otherwise the check fails and prints the line to add.

Two deliberate scoping choices:

- The trigger only looks at the PR's own changesets. Reading every pending changeset would fail unrelated PRs for an omission somebody else merged, which is the trap #189 had to undo for the integrity gate.
- Coverage is accepted from any pending changeset, because what matters is whether the next release republishes the umbrella, not which PR asked for it.

The sibling list is derived from the umbrella's own dependencies rather than hardcoded, so it tracks new blocks automatically. Packages outside that list (`hosting`, `pipeline`, `create-blocks-app`) do not trigger it; their README edits still ship into `docs/`, and that dimension is already covered by the `Blocks Package Integrity` gate.

### Verification

Five scenarios, run locally against the real repo state:

| scenario | expected | result |
|---|---|---|
| PR touches no changeset | pass | `✓ No changeset here releases a package @aws-blocks/blocks re-exports.` exit 0 |
| PR bumps `auth-common`, another pending changeset covers blocks | pass | `✓ 1 re-exported package(s) released, and @aws-blocks/blocks is bumped alongside them.` exit 0 |
| PR bumps `auth-common`, nothing covers blocks | fail | exit 1 with the message below |
| PR bumps `hosting` only (not an umbrella dep) | pass | exit 0 |
| release PR shape, every changeset deleted | pass | exit 0 |

Failure output:

```
❌ This PR releases package(s) that @aws-blocks/blocks re-exports:

   • @aws-blocks/auth-common

Nothing bumps @aws-blocks/blocks. It pins its siblings with caret ranges, so their
new versions stay in range and `changeset version` leaves the umbrella at its
current version. Its packed content still moves (the re-exported APIs, plus
docs/ assembled from sibling READMEs at pack time), and publish then aborts the
whole release run with "@aws-blocks/blocks@<version> already exists with different
content" (issue #273, previously #212).

Add this line to your changeset:

   "@aws-blocks/blocks": patch
```

Also confirmed: `npx tsc -p scripts/tsconfig.json --noEmit` clean, and `validate-structure`, `verify-coverage` and `block-major` still pass after the small `getChangedFiles()` extraction they now share.

No changeset here. Nothing under `packages/` changes, so `verify-coverage` does not ask for one, matching #189 which was also CI only.

### DIFF
--- a/.github/workflows/changeset-check.yml
+++ b/.github/workflows/changeset-check.yml
@@ -32,3 +32,10 @@ jobs:
           PR_TITLE: ${{ github.event.pull_request.title }}
       - name: Block major version bumps (pre-release)
         run: node --experimental-strip-types scripts/changeset-guard.ts block-major
+      # The umbrella re-exports its siblings and ships docs/ assembled from their
+      # READMEs, but caret ranges keep a sibling's new version in range, so
+      # `changeset version` never bumps the umbrella on its own. Its packed
+      # content still moves and publish then aborts the whole release run
+      # (#273, previously #212), so require the umbrella entry up front.
+      - name: Verify the umbrella is bumped with its siblings
+        run: node --experimental-strip-types scripts/changeset-guard.ts verify-umbrella
--- a/.github/workflows/changeset-check.yml
+++ b/.github/workflows/changeset-check.yml
@@ -20,6 +20,11 @@ jobs:
 
       # Guards import only Node built-ins, so Node type-strips and runs the TS
       # directly.
+      # The guards decide whether a release can go out, so they are unit-tested
+      # first, against throwaway git repos in a temp dir. A regression shows up
+      # here instead of as a wrong verdict in the steps below.
+      - name: Test the changeset guards
+        run: node --test scripts/changeset-guard.test.mjs
       # Catches malformed changesets (broken frontmatter, unknown package,
       # invalid bump type) that the regex parser silently ignores and would
       # otherwise fail post-merge at `changeset version`. Validates only the
@@ -36,6 +41,7 @@ jobs:
       # READMEs, but caret ranges keep a sibling's new version in range, so
       # `changeset version` never bumps the umbrella on its own. Its packed
       # content still moves and publish then aborts the whole release run
-      # (#273, previously #212), so require the umbrella entry up front.
+      # (#273, previously #212), so require the umbrella entry up front, and
+      # treat removing an existing one as the same failure.
       - name: Verify the umbrella is bumped with its siblings
         run: node --experimental-strip-types scripts/changeset-guard.ts verify-umbrella
--- a/scripts/changeset-guard.test.mjs
+++ b/scripts/changeset-guard.test.mjs
@@ -1,0 +1,517 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Unit tests for scripts/changeset-guard.ts. Every case builds a throwaway git
+// repo in a temp dir, copies the real guard into its scripts/ (the guard roots
+// itself at `import.meta.dirname/..`, so the copy sees the fixture as the repo),
+// commits a base, points refs/remotes/origin/main at it, then commits the "PR"
+// changes on top and runs the guard for real: real git, real fs, no stubs.
+//
+// Run: node --test scripts/changeset-guard.test.mjs
+
+import assert from "node:assert/strict";
+import { execFileSync } from "node:child_process";
+import {
+	copyFileSync,
+	mkdirSync,
+	mkdtempSync,
+	readFileSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+import { describe, it } from "node:test";
+
+const SCRIPTS_DIR = import.meta.dirname;
+const GUARD_SRC = join(SCRIPTS_DIR, "changeset-guard.ts");
+const REPO_ROOT = join(SCRIPTS_DIR, "..");
+
+const UMBRELLA = "@aws-blocks/blocks";
+const SIBLING = "@aws-blocks/core";
+const OTHER_SIBLING = "@aws-blocks/bb-kv-store";
+// In the workspace but not re-exported by the umbrella, so releasing it alone
+// must never trigger the umbrella requirement.
+const NON_SIBLING = "@aws-blocks/create-blocks-app";
+
+// Keeps the fixture repo free of the host's git config (signing hooks, aliases,
+// templates) so the tests behave the same on a laptop and on a runner.
+const GIT_ENV = {
+	...process.env,
+	GIT_CONFIG_GLOBAL: "/dev/null",
+	GIT_CONFIG_SYSTEM: "/dev/null",
+	GIT_AUTHOR_NAME: "Guard Test",
+	GIT_AUTHOR_EMAIL: "guard@test.invalid",
+	GIT_COMMITTER_NAME: "Guard Test",
+	GIT_COMMITTER_EMAIL: "guard@test.invalid",
+};
+
+function git(cwd, ...args) {
+	return execFileSync("git", args, { cwd, env: GIT_ENV, encoding: "utf-8" });
+}
+
+function write(dir, relPath, contents) {
+	const full = join(dir, relPath);
+	mkdirSync(dirname(full), { recursive: true });
+	writeFileSync(full, contents);
+}
+
+function pkgJson(name, extra = {}) {
+	return `${JSON.stringify({ name, version: "0.1.0", ...extra }, null, 2)}\n`;
+}
+
+/** `---\n<entries>\n---\n\n<body>` frontmatter, the shape `npx changeset` writes. */
+function changeset(entries, body = "A change.") {
+	const lines = Object.entries(entries).map(([pkg, bump]) => `"${pkg}": ${bump}`);
+	return `---\n${lines.join("\n")}\n---\n\n${body}\n`;
+}
+
+/**
+ * A fixture repo whose base commit is what origin/main points at. `files` is
+ * merged over the default workspace layout before that base commit is made.
+ */
+function baseRepo(t, files = {}) {
+	const dir = mkdtempSync(join(tmpdir(), "changeset-guard-"));
+	t.after(() => rmSync(dir, { recursive: true, force: true }));
+
+	mkdirSync(join(dir, "scripts"), { recursive: true });
+	copyFileSync(GUARD_SRC, join(dir, "scripts", "changeset-guard.ts"));
+
+	const workspaces = [
+		"packages/blocks",
+		"packages/core",
+		"packages/bb-kv-store",
+		"packages/create-blocks-app",
+	];
+	const layout = {
+		"package.json": `${JSON.stringify({ name: "fixture-root", private: true, workspaces }, null, 2)}\n`,
+		"packages/blocks/package.json": pkgJson(UMBRELLA, {
+			dependencies: {
+				[SIBLING]: "^0.1.0",
+				[OTHER_SIBLING]: "^0.1.0",
+				"aws-cdk-lib": "^2.0.0",
+			},
+		}),
+		"packages/core/package.json": pkgJson(SIBLING),
+		"packages/bb-kv-store/package.json": pkgJson(OTHER_SIBLING),
+		"packages/create-blocks-app/package.json": pkgJson(NON_SIBLING),
+		".changeset/config.json": `${JSON.stringify({ changelog: "@changesets/cli/changelog" }, null, 2)}\n`,
+		".changeset/README.md": "# Changesets\n",
+		...files,
+	};
+
+	for (const [relPath, contents] of Object.entries(layout)) {
+		if (contents === null) continue;
+		write(dir, relPath, contents);
+	}
+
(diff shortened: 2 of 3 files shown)


osama-rizk (commented):
Well-reasoned, correctly-scoped guard. I verified the one thing that would have made it silently useless (the changeset-filename match), and it holds. Nothing blocking; Finding 1 is the interesting one.

## What I verified

The failure mode I went hunting for: if `parseChangesets().file` were a *path* (`.changeset/foo.md`) while `getChangedChangesetFiles()` produced a *bare name* (`foo.md`), then `touched.has(e.file)` is always false and the guard passes on everything — a false-negative strictly worse than the bug, because it looks green. Read the full file on the branch: `parseChangesets` builds `file` from `readdirSync(CHANGESET_DIR)` (bare name), and the extractor captures the bare name too. They match. ✅ The release-PR case also holds — `git diff` lists the deleted `.md`, but `parseChangesets` reads from disk so it's absent from entries → `released` empty → pass.

Deriving siblings from `blocks/package.json` dependencies (rather than a hardcoded list) is the right source of truth — a new `bb-*` block is covered automatically and the guard can't rot out of sync with the re-export set.

The `linked`/`fixed`/today table is the standout: it's the actual output of running `changeset version` in a scratch worktree against all three configs, showing the concrete numbers (`auth-common` dragged to 0.3.3 under both alternatives). That's "here's what linked does to the numbers," not "I think linked won't work."

## 1. The guard catches *adding* a sibling bump, not *removing* umbrella coverage

The trigger fires only on changesets **this PR touched** that bump a sibling. Consider: PR-B deletes/edits away the `"@aws-blocks/blocks": patch` line from an existing changeset, while the sibling bump that needs it lives in a *different* changeset PR-B didn't touch. If PR-B's touched changesets don't themselves bump a sibling, `released` is empty → pass — and the release then aborts exactly as #273 describes. So the guard defends the moment a sibling bump is *introduced* but has a blind spot for the moment umbrella coverage is *withdrawn*.

Contrived, and #212-style recurrence was omission-on-add (fully covered here). But it's a direct instance of the invariant this PR claims to protect. Worth at least a comment noting the guard assumes coverage, once present, isn't later removed — or, if cheap, also triggering when a touched changeset removes a `blocks` entry. Not blocking.

## 2. `patch` suffices, so a breaking sibling change doesn't force a breaking umbrella bump

The check is satisfied by *any* umbrella bump, so a sibling `minor` (the pre-1.0 breaking channel) paired with an umbrella `patch` passes. Since the umbrella re-exports the sibling, a break in the sibling arguably breaks through the umbrella too, and a consumer on `@aws-blocks/blocks` gets it under a patch label. I suspect this is deliberate (the guard's job is "republish so publish doesn't abort," which patch achieves; version *alignment* is separate) — worth one sentence confirming bump-level matching is intentionally out of scope, so a future reader doesn't read it as a bug.

## 3. Fail-open on unreadable package.json is invisible

If `blocks/package.json` is missing/unparseable, `getUmbrellaSiblings()` returns empty → the check passes with the same `✓ No changeset here releases…` line as a legitimate no-op. For a guard, a *silent* fail-open is the failure mode that erodes trust in green. Consider distinguishing "no siblings / package.json unreadable" from "no siblings released" in the output. Very unlikely in practice, but it's the "silently skipped ⇒ looks covered" gap worth closing in any assurance check.

## Nits
- `getChangedChangesetFiles` regex `^\.changeset\/(.+\.md)$` would capture a nested `foo/bar.md`; changesets are flat so it's moot, but `[^/]+\.md` is tighter.
- Good instinct calling out that the three existing commands still pass after the `getChangedFiles()` extraction — that's the one place this PR touches shared code.

ahmedhamouda78:
Thanks, all three were worth fixing. Pushed 2feb3fe and 75cfb0a.

**[minor] Withdrawal of umbrella coverage.** You are right, and I went with (a).
Removing an umbrella entry that existed at the merge base now triggers the
check, whether the line was stripped or the whole changeset was deleted. The
trigger set widens in that case: a normal PR is only judged on the sibling
releases its own changesets declare, but a withdrawal is judged against every
pending sibling release, because that PR is what took the coverage away from all
of them.

The false positives I worried about all fall out clean, because the requirement
still only applies when a sibling release is actually pending:

- revert (delete the changeset that had both the sibling and the umbrella): no
  sibling release left pending, passes
- consolidation / moving the entry to another changeset: umbrella still covered
  by something pending, passes
- rename: shows up as a delete plus an add, the add still declares it, passes
- the "chore: version packages" release PR, which deletes every changeset at
  once: nothing pending at all, passes

Each of those is a test. One case was already covered before this change and I
left a test on it too: if the changeset you strip the umbrella line from also
releases a sibling of its own, the old code already failed you, via `released`
rather than via the withdrawal path.

To read the entry as of the merge base I list `.changeset` at that commit once
with `git ls-tree` and only `git show` files in the listing, so a `git show`
failure stays a real failure instead of being read as "the file was not there".

**[minor] `patch` on the umbrella while a sibling ships `minor`.** Intentional,
and now written down in the header: only the entry's presence is checked, never
its bump level. Deciding how a pre-1.0 break should propagate through
re-exports is a versioning-policy question, and this guard is deliberately
narrow, it exists to stop the publish abort. Level matching would also need a
rule for the `docs/`-only case, where the umbrella's content moves without any
API change at all and `patch` is exactly right. Happy to take it up separately
if you think the policy is worth pinning down.

**[minor] Silent fail-open.** Fixed, and this one was the most embarrassing of
the three. `packages/blocks/package.json` is now read through a helper that
fails the guard with the reason when the file is missing or unparseable, and I
extended it to two more cases: a `dependencies` block that is absent or not an
object, and one that parses but lists no `@aws-blocks/*` siblings. An empty
sibling set means the guard asserts nothing while printing the green line, so it
now refuses to be that.

Same class of bug in two other places, so they went with it: the changed
package's `package.json` in `verify-coverage` (a parse failure used to drop that
package out of the coverage check silently) and a workspace member's in
`validate-structure` (a parse failure used to shrink the valid-name set, which
surfaced as a bogus "unknown package" error for every entry instead of the real
cause).

The control run is the part worth trusting here. Against the pre-fix guard, 9 of
the 31 tests fail, and 6 of those 9 fail specifically by printing
`✓ No changeset here releases a package @aws-blocks/blocks re-exports.` and
exiting 0. That is your fail-open, reproduced.

**[nit] Filename regex.** Tightened to `^\.changeset\/([^/]+\.md)$`. It mattered
more than it looks: with the withdrawal check reading the base version of every
name the pattern captures, `.+` would have let a nested `.changeset/<dir>/x.md`
through and reported a withdrawal for a file changesets never reads. There is a
test for that boundary.

**Tests.** `scripts/changeset-guard.test.mjs`, 31 cases over all four
subcommands. Each one builds a throwaway git repo in a temp dir, copies the real
guard into its `scripts/` so it roots itself there, commits a base, points
`refs/remotes/origin/main` at it, commits the "PR" on top, and runs the real
guard as a subprocess. Real git, real fs, exit codes and messages asserted,
nothing stubbed. They run as the first step of the same workflow, so a
regression in a guard shows up before any of its verdicts are trusted.

**Scope note.** The first and third findings also touched `verify-coverage` and
`validate-structure`, same fail-open class, about ten lines. If you would rather
this review stayed strictly on `verify-umbrella`, say so and I will split those
two out.

soberm (approved):

osama-rizk (approved):

PR #275, opened 2026-07-28
docs: config.json resolution, RawRoute, OIDC signin route, and the undocumented error exports

### LINKED ISSUE #143: docs: document RawRoute (raw HTTP route) API
Discovered via agent-bench (PR #132), observability-api cell, run 28644861376.

Problem: RawRoute (serving a raw HTTP route and writing the response) is exported from @aws-blocks/core and re-exported from @aws-blocks/blocks, but is documented nowhere - not in any bb-*.md, the blocks index, or the backend template README.

Impact: A builder needing a custom endpoint (e.g. GET /status returning text/html) cannot discover RawRoute from the docs. The bench agent had to grep compiled node_modules/@aws-blocks/core/dist/raw-route.d.ts and raw-route.mock.js to find it (~61 cycles, ~4.2M input tokens).

Ask: Document RawRoute (register a route, set Content-Type, write the response) in the blocks index and/or the backend template README, with a minimal GET example.

### LINKED ISSUE #145: docs: document server-initiated OIDC signin route (/aws-blocks/auth/signin/<provider>)
Discovered via agent-bench (PR #132), oidc-dsql-notes cell, run 28644861376.

Problem: The server-initiated OIDC signin route GET /aws-blocks/auth/signin/<provider> (e.g. /aws-blocks/auth/signin/stub) is not documented in bb-auth-oidc.md. It runs the whole flow via server-side redirects (signin -> authorize -> callback sets the session cookie -> back to the app) with no client-side PKCE wiring, and is the path the block's own integration tests use.

Impact: Without docs, builders reach for the client-side getClient().signIn(provider) API instead, producing divergent auth-client assumptions. In the bench this cost a judge quality deduction (docs absent).

Ask: Document the server-initiated signin route in bb-auth-oidc.md as the recommended integration/test path alongside the client-side API, and clarify when to use each.

### LINKED ISSUE #154: Docs: ApiError constructor, parseRpcRequest array-body, hasAuthError, broadcastAuthChange import — agents reverse-engineer from .d.ts
## Problem

Across bench run 28701776910 (see PR #132 for context), multiple cells burned large token counts `cat`/`grep`-ing `.d.ts`/`dist` to reverse-engineer undocumented error and API semantics. The following specific gaps are **not** covered by existing issues #144 or #139:

- **observability-api / backend**: `ApiError` constructor signature (which args? which HTTP status codes?); how `parseRpcRequest` handles a JSON **array** as the request body vs. an object.
- **auth-notes**: `hasAuthError` — what it checks and where to import it from; `broadcastAuthChange` — correct canonical import (`@aws-blocks/blocks/ui`, not `@aws-blocks/blocks`).
- **cognito-profile**: `isRetriableAuthError` behavior; whether `UsernameExistsException` and `CodeMismatchException` are retriable (partially tracked in #144, but grepping compiled output for it is still the only path today).

Agents resorted to inspecting compiled output because prose docs don't cover error types, retriability semantics, or constructor signatures.

## Request

Document the following so consumers don't need to inspect compiled output:

- `ApiError` constructor signature and semantics (HTTP status ↔ JSON-RPC error code for the API/RawRoute layer).
- `parseRpcRequest` behavior when the request body is a JSON array.
- `hasAuthError` — what it checks and its import path.
- `broadcastAuthChange` — confirm canonical import is `@aws-blocks/blocks/ui`.
- General pattern: error types, retriability flags, and constructor signatures should appear in prose docs, not only in `.d.ts`.

**Related**: #144 (AuthCognito retriability), #139 (JSON-RPC error-code mapping for ApiNamespace), #143 (RawRoute API). This issue captures the remaining gaps and the cross-cutting "agents must grep compiled output" pattern.

**Evidence**: agent-bench run 28701776910 / PR #132

### LINKED ISSUE #238: docs: document config.json resolution — served at /.blocks-sandbox/config.json, {"_placeholder":true} expected locally, /config.json 404s
@aws-blocks/core (client config resolution) + Developer Guide

### Describe the problem
The runtime config is fetched from `/.blocks-sandbox/config.json` (a dotted directory), NOT `/config.json`. Undocumented: (1) locally the file is `{"_placeholder":true}` by design - the real apiUrl is injected at deploy, so the placeholder locally is CORRECT; (2) `curl /config.json` returns 404, so people wrongly conclude config is missing. Only discoverable by reading compiled client source.

### Impact
Wasted debugging cycles and false "config broken" conclusions; agents curl /config.json, get 404, mis-diagnose the app.

### Proposed fix
Document in the Developer Guide (local-dev + deploy) and client getApiUrl docs: config is served at /.blocks-sandbox/config.json (dotted dir); {"_placeholder":true} is the expected local body (real apiUrl injected at deploy); /config.json is intentionally not a route (the 404 is expected).

### Related
#173 (placeholder config served long-term post-deploy) - a library bug on the same file; this is the docs gap.

ahmedhamouda78:
Four docs gaps that all had the same failure mode: the answer only existed in `dist` or a `.d.ts`, so people (and agents) went and read compiled output to find it. Batched into one PR because they are all prose and they overlap in the same two READMEs.

Fixes #238
Fixes #143
Fixes #145
Fixes #154

### Runtime config resolution (#238)

`packages/core/README.md` now has a "Runtime config" section under the HTTP/RPC docs covering the resolution order the client actually uses (`BLOCKS_API_URL`, `BLOCKS_CONFIG`, the file on disk in Node, then `fetch('/.blocks-sandbox/config.json')` in the browser), and the two things that keep getting misread:

- the path is the dotted `/.blocks-sandbox/config.json`, so `curl /config.json` returning 404 says nothing about your config
- `{"_placeholder":true}` in the frontend build output is the stub Hosting writes during synth, while `apiUrl` is still an unresolved CFN token, so seeing it before a deploy is the design working

It also says what a real failure looks like (`Blocks API URL not configured` from the client) and keeps the genuine bug on the list: a deployed origin still serving the placeholder after a successful deploy. `TROUBLESHOOTING.md` gets the same thing in short form plus three rows in the common-issues table, since that file is where people already go for API URL questions.

### RawRoute (#143)

The core README had a RawRoute section whose detail was a "see source repo" pointer, and the block catalog decision tree in `docs/index.md` had no entry for raw HTTP endpoints at all, which is why it reads as undocumented. Now filled in: constructor signature, a `GET` example that sets status, `Content-Type` and body, tables for what you read off `ctx.request` and write to `ctx.response`, a path-parameter plus query-string example, the path syntax rules, and the registration rules that only show up as runtime throws today (register during load, reserved `/aws-blocks` paths, duplicate route error, no extra AWS resources). Added a "serve a raw HTTP endpoint" line to the decision tree in `scripts/sync-block-docs.mjs` so the catalog points at it.

### Server-initiated OIDC sign-in (#145)

`bb-auth-oidc` mentioned `GET /aws-blocks/auth/signin/<provider>` in two one-liners but never explained it, so people reach for the client PKCE API by default. New section walks the redirect chain (kickoff 302 plus pending-auth cookie, IdP, callback exchanges the code and sets the session cookie, 302 to `postSignInPath`), notes it needs no client-side code at all, gives the JSON error shape and the URL-encoded provider name, and includes the `curl` sequence that the block's own integration tests follow. Ends with when to use it versus client PKCE, and that both write the same session cookie.

### ApiError, params handling, hasAuthError, broadcastAuthChange (#154)

- `ApiError`: full constructor signature with a table for `message` / `status` / `name` / `cause` / `retriable`, that `cause` stays server-side, and how `status` becomes the JSON-RPC error `code` and decodes back into an `ApiError` on the client (reserved `-32xxx` codes become 500)
- `params`: array is positional, an object is flattened with `Object.values()` so key order decides argument order, and a top-level JSON array body (a batch) is rejected with `-32600` and a `null` id rather than running anything
- `hasAuthError`: signature, import path, and the rule of thumb against `isBlocksError` (thrown error vs returned `AuthState`), documented where the export lives instead of only in two block READMEs
- `broadcastAuthChange`: signature and the canonical app-level import `@aws-blocks/blocks/ui`, with a note that the package root does not export it because the UI exports sit behind the `/ui` subpath

`isRetriableAuthError` from that issue is left alone, it belongs to #144 and PR #165.

### Verification

Behaviour in the docs was read off the source and then checked against a running server rather than assumed:

```
$ curl http://localhost:3000/.blocks-sandbox/config.json
{"apiUrl":"http://localhost:3000/aws-blocks/api","environment":"local"}
$ curl -o /dev/null -w '%{http_code}' http://localhost:3000/config.json
404
$ curl -i http://localhost:3000/status        # RawRoute, sets its own content type
HTTP/1.1 200 OK
content-type: text/html; charset=utf-8
<h1>ok</h1>
$ curl http://localhost:3000/users/42?q=hello
{"id":"42","q":"hello"}
$ curl -X POST .../aws-blocks/api -d '[{"jsonrpc":"2.0","method":"api.ping","params":[],"id":1}]'
{"jsonrpc":"2.0","error":{"code":-32600,...,"data":{"name":"InvalidRequest"}},"id":null}
```

Added tests so the prose can't drift: `params` array / object / omitted, array-body rejection, parse error, `ApiError` defaults and options, and the status to error-code round trip through `errorResponseFromCatch` and `decodeRpcResponse`.

```
npm test -w packages/core   # 599 pass, 0 fail
npm run check:api           # all API reports up to date
changeset-guard validate-structure | verify-coverage | block-major | verify-umbrella   # all pass
```

Changeset covers `core`, `auth-common` and `bb-auth-oidc` for the README edits, plus `@aws-blocks/blocks` since the published `docs/` folder is assembled from those READMEs and the decision tree changed.


osama-rizk (commented):
High-integrity docs PR — I spot-checked the RawRoute claims against `packages/core/src/raw-route.ts` on the branch and they're all accurate, including the load-bearing security asymmetry. The tests pin exactly the claims most prone to drift. Nothing blocking; one finding worth acting on plus a few nits.

## What I verified against source

I didn't take the RawRoute prose on faith:
- "Registering after handler creation throws" → `registerRoute` throws `'Routes must be registered during initialization…'`. ✅
- "`/aws-blocks` and `/aws-blocks/api/*` reserved; `/` not routable" → both throws present. ✅
- "`RawRouteErrors.DuplicateRoute`" → `DuplicateRoute: 'DuplicateRouteException'`. ✅
- **The security claim** — "named parameters are URL-decoded; wildcard captures are not, so validate them before touching a filesystem or S3 key" → `extractParams` does exactly this, with the source comment *"Wildcards capture raw paths — no decoding (prevents path traversal)."* ✅ This is the highest-value line in the PR: it turns an invisible security invariant into something a consumer can rely on.
- Scope-chain derivation `Scope('v1') → RawRoute('health')` → `/v1/health` matches `derivePathFromScope`. ✅

The tested claims are pinned at the right seams too — the `status ↔ code` round-trip test encodes through `errorResponseFromCatch` *and* decodes back through `decodeRpcResponse`, asserting `isBlocksError` still matches client-side. That proves the doc's central "same check works on both sides" claim, not just the encode half.

## 1. The OIDC redirect chain is the least-verifiable claim and has no test tether

The `bb-auth-oidc` section documents a precise sequence with several exact, testable assertions — default `postSignInPath` of `/`, `GET /signout` being a 404, the `{error,name}` JSON shape, unknown-provider → `ProviderNotConfiguredException`. Unlike the `ApiError`/`params` claims (which now carry tests), none of these are enforced, and I can't confirm them from the diff. The PR notes this is "the path the block's own integration tests drive," which is the right instinct — but that coupling is asserted in prose, not enforced.

If those integration tests exist, one line linking them (or a comment in them pointing back to this README) would make the section self-defending the way the RPC section now is. Not blocking, but this is the section most likely to drift first.

## 2. Config-resolution order — clarify whether the env vars are public knobs

The 4-step resolution order (`BLOCKS_API_URL` → `BLOCKS_CONFIG` → Node file → browser fetch) is exactly what a debugging reader needs. One gap: it says `BLOCKS_API_URL` is "set by the Hosting construct on SSR compute" but doesn't say whether a user is *meant* to set these themselves (e.g. a custom SSR host). A half-sentence ("these are set for you; override only if…") would close the "can I use this?" question the doc otherwise invites.

## Nits
- `##### How the server reads params` is the only h5 in the whole README — every other subsection is h3/h4. Renders fine, but odd depth for a section this important; consider h4 for parity.
- Cross-link `./core.md#rawroute` matches the `### RawRoute` heading anchor. ✅ (checked, since a broken catalog link would defeat the #143 fix.)
- Changeset scope (patch bumps for core/auth-common/bb-auth-oidc/blocks because published READMEs + the assembled `docs/` changed) is correct — and is exactly the rule #274 is about to enforce.

osama-rizk (approved):

ahmedhamouda78:
Thanks Osama, both fair hits. Fixed in `7e51f32`.

**[minor] OIDC redirect chain had no test tether.** You were right that this was the most drift-prone part, and chasing the tests down proved it: one sentence was already wrong. The README claimed an unknown provider returns the `{error,name}` JSON with `ProviderNotConfiguredException`. It does not. `mountAuthRoutes` registers one `GET` route per *configured* provider and `matchRoute` is exact-path, so an undeclared name never reaches a handler at all. Confirmed against a running dev server: `GET /aws-blocks/auth/signin/nonexistent` is a plain 404, and `ProviderNotConfiguredException` only shows up on the RPC path (`api.oidcGetSignInUrl('nonexistent')` gives code 500 with `data.name: ProviderNotConfiguredException`). The JSON error shape is real, just for failures building the authorize URL. Corrected the prose and added a paragraph pointing at `test-apps/comprehensive/test/oidc-auth.test.ts`, specifically the `full sign-in flow via HTTP redirects` tests (the google variant plus the corporate custom-OIDC variant), which walk the same three redirects and assert the `302` chain, both cookies, the `Location: /` from `postSignInPath`, and the authenticated RPC call after it. Default `postSignInPath` and the `%20` encoding come from the `path configuration` tests in `packages/bb-auth-oidc/src/index.test.ts`.

Two claims genuinely had no assertion anywhere, so I added them rather than link to tests that do not cover them: the 404 for an undeclared provider, and the 404 for `GET` on the POST-only sign-out route. Every response in that section is now asserted. Referenced the test paths as inline code rather than relative links, since `test-apps/` is not in the published tarball and a dead link in a `packages/*/README.md` would trip rule 11.

**[minor] Are `BLOCKS_API_URL` / `BLOCKS_CONFIG` user-facing?** Mostly no, and the doc now says which part is. `Hosting` injects both into every SSR compute function at synth, and `npm run sandbox` sets `BLOCKS_API_URL` on the dev server it spawns, so on the normal path you never touch them. The one case a user is meant to set `BLOCKS_API_URL` is an SSR host they run themselves, which is what the framework test apps do for a dev server on its own port (`BLOCKS_API_URL=http://localhost:3001/aws-blocks/api next dev`). `BLOCKS_CONFIG` is Hosting's serialized `backendConfig`, so I have called it out as not a knob to hand-write.

**[nit] The lone h5.** Promoted to `####`. It was the only h5 in any packages README and nothing anchored to it.

Local: comprehensive E2E 374 pass / 0 fail, bb-auth-oidc 112/112, core 599/599, changeset guards and `check:api` clean.

soberm (approved):

PR #276, opened 2026-07-28
fix(core): stream cdk deploy output to stdout and keep a backgrounded deploy alive (#222)

### LINKED ISSUE #222: Deploy CLI emits 0-byte stdout and backgrounded deploy is SIGTERM'd while CloudFormation still succeeds
@aws-blocks/pipeline / deploy CLI

## Describe the bug
`npm run deploy` produces 0-byte stdout during the CloudFormation phase. When run in the background the process is SIGTERM'd (exit 143) even though the underlying CloudFormation deploy succeeds independently. There is no reliable progress signal, so callers must poll `describe-stacks` by hand and cannot tell success from failure.

## Evidence
observed across all 3 reps of document-ingestion-pipeline (preamp batch manual-deploy-20260717-103949-1339). Consumers re-ran deploys chasing phantom failures, driving 2-3x cost (~$15/run vs $5-8) and the longest runtime (~22min) of any framework on the task.

## Expected
Deploy streams progress to stdout and returns a clear terminal status; a deploy that is still running is not SIGTERM'd.

## Suggested fix
Stream CloudFormation progress events to stdout and decouple the CLI lifecycle from the in-flight deploy

ahmedhamouda78:
Fixes #222

## What was wrong

Two separate problems made a production deploy unobservable, and made a deploy
that had actually succeeded look like a failure. Consumers then re-ran it, which
is where the 2-3x cost in the issue comes from.

**1. Nothing on stdout for the whole CloudFormation phase.** The CDK CLI routes
every non-error log line (all stack activity included) to stderr unless CI mode
is on. Its io host picks the stream like this:

```js
selectStreamFromLevel(level) {
  switch (level) {
    case "error":  return process.stderr;
    case "result": return process.stdout;
    default:       return this.isCI ? process.stdout : process.stderr;
  }
}
```

So `npm run deploy > deploy.log` captured zero bytes while the stack converged.
Verified against the real CLI (aws-cdk 2.1122.0), deploying a pre-synthesized
assembly pinned to the all-zeros account so nothing can actually be created:

```
$ cdk deploy --progress events --no-ci ...   ->  stdout=0B    stderr=131B
$ cdk deploy --progress events --ci    ...   ->  stdout=29B   stderr=102B
```

**2. A backgrounded deploy was killed while CloudFormation kept going.**
`cdk deploy` ran through a blocking `spawnSync`, with the child in this process's
own group and no signal handling. A process-group SIGTERM (a harness reaping a
backgrounded job, a parent shell going away) killed the CLI with exit 143, took
the CDK child with it, and CloudFormation carried on server-side and finished.
The caller was left with a dead process, no output, and no terminal status.

## The fix

`packages/core/src/scripts/deploy-stream.ts` runs the CloudFormation phase and
owns its lifecycle:

- `--ci` so CDK log lines (progress included) land on stdout, errors still on
  stderr, and `--progress events` so it is one line per resource transition
  rather than a progress bar that needs a TTY.
- Output is piped and relayed line by line as it arrives, with lines reassembled
  across chunk boundaries, so nothing is buffered until exit.
- An idle heartbeat prints elapsed time while a slow resource is converging, so
  stdout is never silent for minutes.
- The CDK CLI is spawned in its own process group (POSIX) with stdin ignored, so
  a group signal aimed at the parent shell cannot kill it behind our back and a
  background run cannot be stopped by SIGTTIN.
- Signal policy: one SIGTERM, or any SIGHUP, does not abandon a converging
  deploy. It logs why and keeps streaming. Ctrl-C, or a second SIGTERM, aborts
  and reaps the whole CDK process tree through the existing
  `terminateProcessTree` helper. SIGKILL is uncatchable, so a `docker stop` or a
  CI cancel that escalates still stops immediately.
- The terminal status is printed on stdout for success and failure, so a caller
  that only captures stdout can tell them apart.

No new flags on `npm run deploy`, and no change to the deploy contract otherwise.

## Behaviour change worth knowing

Because the CDK CLI's output is now piped through this process, interactive runs
get event lines instead of the redrawing progress bar (the bar needs a TTY, which
a piped child does not have, and it was never usable in a log file anyway). Every
event is still shown, as it happens.

## Verification

`packages/core/src/scripts/deploy-stream.test.ts`, 23 tests, real child processes
and real signals, no mocks:

- Streaming is proven causally rather than by timing: the child prints one line
  and then blocks until the test writes a file, which the test only does after it
  has seen that line on stdout. If output were buffered until exit, the test
  deadlocks.
- Backgrounded deploy under a group SIGTERM, end to end through three real
  processes (test -> deploy wrapper -> fake cdk): the wrapper keeps streaming,
  reports that it is ignoring the signal, is still alive afterwards, the deploy
  runs to completion, and it exits 0 with the terminal status on stdout.
- A repeated group SIGTERM aborts, exits non-zero, and the cdk child is reaped
  rather than orphaned.
- A control test pins the old shape (blocking `spawnSync`, shared process group,
  no signal handling) and asserts it dies on the first group SIGTERM with the
  in-flight deploy cut short, so the passing tests above are not vacuous.
- The real CDK CLI stream routing is asserted against the installed CLI (see the
  numbers above), skipped if the CLI is not installed. No credentials, no network
  and no mutation: the probe uses a pre-synthesized assembly pinned to the
  all-zeros account, which can never resolve.
- Idle heartbeat, stderr staying on stderr, exit-code propagation, line assembly
  across chunk boundaries and CRLF, and the argv contract (dropping `--ci` or
  `--progress events` fails the suite).

Local runs:

```
npm run build:packages                 -> pass
npm test -w packages/core              -> 622 tests, 622 pass, 0 fail
npm run check:api                      -> All API reports are up-to-date
npm run check:exports-consistency      -> consistent
changeset-guard validate-structure     -> 8 changesets structurally valid
changeset-guard verify-coverage        -> all changed packages covered
changeset-guard block-major            -> no major bumps
changeset-guard verify-umbrella (#274) -> umbrella bumped alongside re-exported siblings
```

The changeset bumps `@aws-blocks/core` and `@aws-blocks/blocks` (the umbrella
re-exports core, so it has to ship with it).

`npm run lint` could not run on this machine (the bundled biome binary needs a
newer glibc), so lint is left to CI.

Backgrounded repro, before and after, with stdout redirected to a log:

```
log bytes 1.2s in (streamed, not buffered until exit): 405
sending group SIGTERM to pgid 5774 (the reap that used to kill the deploy)
alive 600ms after SIGTERM: true
final exit code=0 signal=null
CloudFormation phase ran to completion (cfn-done marker): true
terminal status file: complete
```

with the log itself showing the deploy carrying on through the signal:

```
ProbeStack | 5/20 | CREATE_IN_PROGRESS | AWS::Lambda::Function | ApiHandler
Ignoring SIGTERM: the deploy is still running and CloudFormation is still converging. Send SIGTERM again to abort (the stack update continues server-side either way).
ProbeStack | 6/20 | CREATE_IN_PROGRESS | AWS::Lambda::Function | ApiHandler
...
ProbeStack | CREATE_COMPLETE | AWS::CloudFormation::Stack | ProbeStack

Deployment complete!
```

## Not in this change

`npm run sandbox` has the same output plumbing and would benefit from the same
treatment, but the sandbox is a foreground dev loop with its own supervisor, so
it is left alone here to keep this focused on the reported production deploy bug.

## Follow-up after validating against a real AWS account

I deployed a freshly scaffolded app (create-blocks-app, backend template, wired to
this branch's build through a local registry) to a real account and it exposed a
gap in the first version of the signal policy, now fixed in the second commit.

One `kill -TERM -<pgid>` reaches the deploy CLI **twice**. `npm run deploy` runs
npm -> sh -> tsx -> node: the process group delivers SIGTERM to the node process,
then tsx relays the SIGTERM it received to its own child about 50ms later. The
original "second SIGTERM aborts" rule therefore fired on the duplicate delivery
and abandoned a deploy nobody had asked to stop:

```
api-app-...-prod |  0/15 | UPDATE_IN_PROGRESS | AWS::CloudFormation::Stack | api-app-...-prod User Initiated
⚠️  Ignoring SIGTERM: the deploy is still running ...
🛑 Received SIGTERM while deploying. Stopping the CDK CLI ...
❌ Deployment failed.
```

A standalone experiment confirmed the mechanism and that it is safe to hold on:
a script hosted by tsx saw 2 SIGTERMs from 1 `kill`, and tsx did not escalate to
SIGKILL when the child kept ignoring them (the script ran to completion).

SIGTERMs inside a 2s window are now treated as duplicate deliveries of the same
request: the deploy keeps streaming and the deferral is logged once. A SIGTERM
after that window is still a deliberate repeat and aborts, and Ctrl-C is still
immediate. Same scenario re-run on the real account after the fix:

```
t=30s first CloudFormation event on stdout (1521B) -> sending group SIGTERM
t=33s 3s after SIGTERM: stdout=1822B, deferral logged=true
t=44s npm wrapper: exitCode=0 signalCode=null
deploy reported success on stdout: true
```

with the log showing the deploy carrying straight on through the signal:

```
api-app-...-prod |  0/15 | UPDATE_IN_PROGRESS      | AWS::CloudFormation::Stack | ... User Initiated
⚠️  Ignoring SIGTERM: the deploy is still running and CloudFormation is still converging. ...
api-app-...-prod |  1/15 | UPDATE_COMPLETE         | AWS::Lambda::Function      | Handler
api-app-...-prod |  3/15 | UPDATE_COMPLETE         | AWS::CloudFormation::Stack | api-app-...-prod
✅ Deployment complete!
📡 API URL: https://<id>.execute-api.us-west-2.amazonaws.com/prod/aws-blocks/api
```

The stack finished UPDATE_COMPLETE and the API served the new handler code, so
the update that was signalled mid-flight really did apply.

### Real-account before and after on stdout

First deploy of the app (39 resources), stdout and stderr captured separately:

```
stdout 26353B (119 CloudFormation event lines), stderr 241B, exit 0
t=030s stdout=6659B  cfn_event_lines=30
t=050s stdout=11998B cfn_event_lines=59
t=080s stdout=25195B cfn_event_lines=116     <- streamed during the deploy
```

The same stack with the old flag set versus the new one:

```
cdk deploy --no-ci  ->  stdout=109B   (only the stack ARN)  stderr=686B (progress, status, outputs)
cdk deploy --ci     ->  stdout=554B   (progress, status, outputs)  stderr=241B
```

Teardown: `npm run destroy` exited 0, the stack no longer exists, and no
buckets, functions, APIs, parameters or log groups were left behind.

### DIFF
--- a/.changeset/deploy-stream-and-signal-lifecycle.md
+++ b/.changeset/deploy-stream-and-signal-lifecycle.md
@@ -24,7 +24,21 @@ Two causes, both fixed:
   arrives, plus an idle heartbeat while a slow resource is converging. A single
   SIGTERM, or any SIGHUP, no longer abandons a converging deploy: it logs and
   keeps streaming. Ctrl-C, or a second SIGTERM, aborts and reaps the CDK process
-  tree.
+  tree. Repeat deliveries of the same signal inside the coalescing window log one
+  line instead of one per delivery.
 
-The deploy also prints its terminal status on stdout for both success and
-failure, so a caller that only captures stdout can tell the two apart.
+Worth knowing before you upgrade:
+
+- **The signal resilience is POSIX-only.** It needs process groups (`detached` is
+  passed only when `process.platform !== 'win32'`) and real signal delivery, and
+  Windows has neither: nothing outside the process delivers SIGTERM/SIGHUP there,
+  so a kill on the process tree still ends the deploy and the abort path reaps by
+  pid. Streaming, the heartbeat and the `--ci` argv apply on every platform.
+- **The `❌ Deployment failed.` banner moved from stderr to stdout**
+  (`console.error` → `console.log`), so a caller capturing only stdout can tell a
+  failed deploy from a killed process. Anything grepping *stderr* for that exact
+  string will no longer match it. The failure *reason* has not moved: the CDK CLI
+  keeps error-level output on stderr even under `--ci`, and the deploy entrypoint
+  still prints the error itself with `console.error`, so stderr remains the place
+  to grep for why a deploy failed. Both halves of that split are now covered by
+  tests, including one against the real CDK CLI.
--- a/.changeset/deploy-stream-and-signal-lifecycle.md
+++ b/.changeset/deploy-stream-and-signal-lifecycle.md
@@ -1,0 +1,30 @@
+---
+"@aws-blocks/core": patch
+"@aws-blocks/blocks": patch
+---
+
+fix(core): stream CloudFormation progress to stdout and stop a stray SIGTERM from killing an in-flight deploy
+
+`npm run deploy` wrote nothing to stdout for the whole CloudFormation phase, and
+a backgrounded deploy was killed (exit 143) while CloudFormation kept going and
+finished server-side. Callers had no progress signal, could not tell success from
+failure, and re-ran deploys that had actually worked.
+
+Two causes, both fixed:
+
+- The CDK CLI picks its log stream as `isCI ? stdout : stderr`, so every
+  CloudFormation event went to stderr and `npm run deploy > deploy.log` captured
+  zero bytes. The deploy now passes `--ci` (log lines on stdout, errors still on
+  stderr) and `--progress events` (one line per resource transition instead of a
+  progress bar that needs a TTY).
+- The deploy ran `cdk deploy` through a blocking synchronous spawn with the child
+  in this process's group, so signals could not be handled and the default
+  SIGTERM disposition killed the CLI mid-deploy. The CDK CLI is now spawned in
+  its own process group with its output piped and relayed line by line as it
+  arrives, plus an idle heartbeat while a slow resource is converging. A single
+  SIGTERM, or any SIGHUP, no longer abandons a converging deploy: it logs and
+  keeps streaming. Ctrl-C, or a second SIGTERM, aborts and reaps the CDK process
+  tree.
+
+The deploy also prints its terminal status on stdout for both success and
+failure, so a caller that only captures stdout can tell the two apart.
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -172,7 +172,25 @@ jobs:
         run: npm run check:api
 
       - name: Unit Tests
+        env:
+          # The real-CDK stream-routing probe in packages/core guards the issue #222
+          # root cause (`--ci` is what moves CloudFormation progress off stderr).
+          # Require it here so a runner that cannot resolve the CDK CLI fails
+          # instead of skipping, and have it drop a marker file the next step
+          # checks — a probe that stops running must not read as green.
+          BLOCKS_REQUIRE_CDK_PROBE: '1'
+          BLOCKS_CDK_PROBE_MARKER: ${{ runner.temp }}/cdk-routing-probe.marker
         run: npm run test:unit
+
+      - name: Verify the real-CDK stream-routing probe ran
+        run: |
+          set -euo pipefail
+          marker='${{ runner.temp }}/cdk-routing-probe.marker'
+          if [ ! -s "$marker" ]; then
+            echo "::error::the real-CDK --ci stream-routing probe did not run (no marker at $marker). It is the only test that verifies the CDK CLI still routes deploy logs to stdout under --ci and keeps failures on stderr (issue #222); a silent skip is a regression."
+            exit 1
+          fi
+          cat "$marker"
 
       - name: E2E Test Local
         run: npm run test:e2e:local
--- a/packages/blocks/README.md
+++ b/packages/blocks/README.md
@@ -146,6 +146,8 @@ Each block is its own package; full per-block docs ship in this package under **
 
 `npm run deploy` does a full production deploy; `npm run sandbox:destroy` tears the sandbox down. The same backend code runs in all three — blocks switch implementations automatically.
 
+`npm run deploy` streams CloudFormation events to **stdout** as they happen, so `npm run deploy | tee deploy.log` shows live progress instead of going silent for minutes, and a deploy failure keeps its reason on **stderr** (that stays the place to grep for why a deploy failed). On POSIX the deploy also survives a stray reap: a single `SIGTERM`, or any `SIGHUP` — a closed terminal, a backgrounded `npm run deploy &` — logs a line and keeps streaming rather than abandoning a stack update CloudFormation is still applying. Press Ctrl-C, or send `SIGTERM` twice, to stop it. That signal resilience is POSIX-only: Windows has no process groups and no OS-delivered `SIGTERM`/`SIGHUP`, so there a kill on the process tree still ends the deploy.
+
 ## Testing
 
 The fastest loop is calling your API through its typed import in `test/e2e.test.ts` — no browser, no mocking:
--- a/packages/core/src/scripts/deploy-stream.test.ts
+++ b/packages/core/src/scripts/deploy-stream.test.ts
@@ -15,6 +15,7 @@ import {
   formatElapsed,
   runStreaming,
   DeployProcessError,
+  SIGNAL_COALESCE_MS,
   type OutputSink,
   type SignalRegistry,
 } from './deploy-stream.js';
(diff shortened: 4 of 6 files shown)


osama-rizk (commented):
Exemplary fix — correct at the one seam that's easy to get wrong (group isolation + handler override *together*), and the tests are the strongest I've reviewed in this repo. No blocking correctness defect. The only finding with teeth is #1 (a test-assurance gap); the rest are confirm-this or document-the-scope.

## What works well

I went looking for the classic failure modes and each is closed:

- **Deferring SIGTERM actually keeps the child alive** — the child is `detached` (own process group), so a parent-group signal never reaches it; the parent handler just declines to exit. The POSIX e2e test proves this with three real processes and a real `kill -TERM -pgid` (and implicitly proves `spawnCommand` honors `detached` — it'd fail otherwise).
- **The completed-deploy race is guarded** by `exitObserved` — a signal arriving during pipe-drain returns early instead of reporting an abort. That's exactly the window that would otherwise turn a successful deploy back into a phantom failure.
- **Abort reaps rather than orphans** — the repeated-SIGTERM test asserts the fake-cdk pid is un-signalable afterward.

The test design is the teachable artifact here, worth naming:
1. **Causal streaming proof** — the child prints one line then blocks on a gate file the test only writes *after* seeing that line. If output were buffered until exit, the test deadlocks rather than flaking. That's "cannot pass unless it streams," not "probably streams."
2. **The vacuity control** — the `spawnSync`-shaped baseline asserts the *old* code dies on the first group SIGTERM (exit 143), so the survival tests can't be green for the wrong reason.
3. **Real-CLI probe with a safe sentinel** — asserting the `--ci` routing against the actual installed CDK, pinned to the all-zeros account so it can't touch AWS.

## 1. The real-CDK probe can silently skip forever, hiding regression in the load-bearing flag

The probe `t.skip`s when `aws-cdk/bin/cdk` can't be resolved *or* when it "never reached the credential check." That's the right resilience for a dev laptop, but it means the one test guarding the `--ci`-routing claim against real CDK can quietly become a no-op in CI (dep hoisting, a CDK version that rewords the credential-probe message, a reachable metadata endpoint). A silently-skipped test reads as "covered" while asserting nothing.

Recommend: in the canonical CI lane, assert this test actually *ran* (fail if skipped), or emit a visible marker the pipeline greps for. Otherwise the day CDK changes its stream routing, you find out in production — the exact bug class this PR exists to prevent.

## 2. Confirm `--ci` doesn't move the failure reason off stderr

`--ci` is broader than stream routing inside the CDK CLI — historically it also suppresses prompts and shifts notices/version messages onto stdout. The probe confirms stdout becomes non-empty, but it stops at the credential check, so it doesn't assert that a genuine *deploy failure reason* still lands on stderr under `--ci`. If some class of failure detail moves to stdout in CI mode, a stderr-only consumer could lose it. Low risk (the entrypoint prints the thrown error via `console.error` regardless), but "which stream does the failure reason use" is the exact contract this PR is tightening, so worth one explicit check.

## 3. Coalescing measures from the first deferral (documented tradeoff, non-blocking)

`firstDeferralAt` is set once and never advanced, so the 2s window runs from the first SIGTERM, not the last delivery. A human who deliberately hits `kill` twice within 2s gets the second one coalesced rather than aborting — i.e. the SIGTERM abort path has a 2s responsiveness floor. The comment justifies this and it's the right default; SIGINT (Ctrl-C) has no such delay, which is the correct escape hatch. Flagging only so the tradeoff is conscious. No change requested.

## 4. State that the resilience guarantee is POSIX-only

`detached` is `false` on Windows and every group-signal test is `posixOnly`, so on Windows the child shares the parent's group and the "survive a stray signal" guarantee doesn't really apply — nothing tests that a Windows deploy survives anything. Fine scope cut for a production-Linux bug, but the changeset/PR should say the resilience is POSIX-only rather than implying it's universal.

## Nits
- The probe appends `--ci`/`--no-ci` on top of the `--ci` that `buildCdkDeployArgs()` already includes, so the `--no-ci` run passes `--ci … --no-ci` and relies on yargs last-wins. Works, but filtering `--ci` out before appending would be less fragile. Test-only.
- SIGHUP never sets `firstDeferralAt`, so repeated SIGHUPs each log the deferral line — minor stdout spam in an unusual case.
- Moving the ❌ failure banner from `console.error` → `console.log` is correct for stdout parity (real error still reaches stderr via the entrypoint). Worth a changeset note, since anyone grepping stderr for "Deployment failed" will no longer find it.

ahmedhamouda78:
Thanks, this was a genuinely useful pass. All four are addressed in `b09ce37`.

**[major] 1. The probe could skip forever.** Agreed this was the one with teeth. I went for removing the skip conditions rather than detecting a skip. The probe environment is hermetic now: every CI marker plus every credential source is stripped (`AWS_PROFILE`, the key/secret/token trio, `AWS_ROLE_ARN`, `AWS_WEB_IDENTITY_TOKEN_FILE`, the container credential URIs) and `AWS_SHARED_CREDENTIALS_FILE`/`AWS_CONFIG_FILE` point at paths that do not exist, so the CLI stops at the same credential check whatever the runner has lying around. Checked by exporting fake credentials and `CI=true` in the parent: both runs still stop at "Need to perform AWS calls for account 000000000000, but no credentials have been configured".

That makes "never reached the credential check" a real failure rather than a resilience case, so it asserts instead of skipping, with the captured streams in the message. The only skip left is "aws-cdk is not resolvable", and that fails when the probe is required. On success it prints a marker the suite asserts on:

```
CDK_ROUTING_PROBE_EXECUTED aws-cdk@2.1122.0 --ci{stdout:27B,stderr:92B} --no-ci{stdout:0B,stderr:119B}
```

CI runs it with `BLOCKS_REQUIRE_CDK_PROBE=1` and `BLOCKS_CDK_PROBE_MARKER` pointing at a file, and a new step fails the job if that file is missing. `BLOCKS_SKIP_CDK_PROBE=1` is the one explicit opt-out. Both fail-loud paths verified locally by hiding `node_modules/aws-cdk` (3 failures, actionable message) and by pointing the probe at a stack that does not exist ("the --ci probe never reached the credential check (exit 1) ... No stacks match the name(s) NoSuchStack").

**[major] 2. Where the failure reason lands under `--ci`.** Good call, worth checking rather than assuming. Against the real CLI:

```
--ci     stdout 27B (synthesis/progress)   stderr 92B (the failure reason)
--no-ci  stdout 0B                         stderr 119B (both)
```

So `--ci` lifts info level to stdout and leaves error level on stderr. That is now asserted, not implied: the reason must match on stderr and must not appear on stdout under `--ci`. Two more tests pin the same contract below the CLI, one at `runStreaming` level and one end to end through the deploy wrapper, which also covers your banner nit.

**[minor] 3. POSIX only.** Said plainly now in the changeset (with the reason: `detached` only when `platform !== 'win32'`, and nothing outside the process delivers SIGTERM/SIGHUP on Windows), in the blocks README next to `npm run deploy`, and in the module docs. Also added a test that asserts the mechanism rather than only its effect: `process.kill(-childPid, 0)` succeeds, which is only true if the child leads its own group.

**[nit] Coalescing window (your point 3).** Left as is on purpose: still measured from the first deferral, so the 2s floor and the Ctrl-C escape hatch are unchanged. I only split the bookkeeping per signal, which does not move the SIGTERM clock.

**[nit] Duplicate `--ci` in the probe argv.** Cleaned up: the flag under test replaces the `--ci` from `buildCdkDeployArgs()` instead of being appended after it, plus a test asserting the argv carries exactly one CI flag.

**[nit] Repeated SIGHUPs.** Deduped, and it turned out to be slightly worse than spam: SIGHUP shared SIGTERM's window, so a hangup landing within 2s of a deferred SIGTERM was swallowed with no line at all. Each signal has its own window now: one hangup logs one line, a later hangup reports again, and a hangup right after a deferred SIGTERM is still reported. Three tests cover that.

**[nit] Banner stream.** Called out explicitly in the changeset: the ❌ banner is on stdout now, so grepping stderr for that string will not match, while the reason has not moved (CDK keeps error level on stderr under `--ci`, and the entrypoint still `console.error`s the error). The e2e test asserts both halves.

622 core tests pass with 0 skipped, changeset guards and `check:api` clean.

osama-rizk (approved):

soberm (approved):

PR #278, opened 2026-07-29
test(comprehensive): key auth verification codes by username in shared state

### LINKED ISSUE #172: [flaky-test] authGetLastCode auth test flakes due to Lambda statelessness
## Summary

The `authGetLastCode`-based auth integration test in the `comprehensive` test app flakes intermittently in CI. On the same commit it alternates pass/fail across re-runs, producing non-deterministic red builds.

## Root cause (from PR #146 flake analysis — HIGH confidence)

The failure is **environmental, not a code defect**. It is caused by **Lambda statelessness in the `authGetLastCode` retrieval path**:

- The verification code is captured by an in-memory `codeDelivery` callback that writes to a module-level variable `lastDeliveredCode`, and `authGetLastCode()` simply returns that in-memory variable:
  - declaration + `codeDelivery` writer: [`test-apps/comprehensive/aws-blocks/index.ts#L56-L64`](https://github.com/aws-devtools-labs/aws-blocks/blob/a1de3b83e5c4b095d78638e19f5614c85d1a6485/test-apps/comprehensive/aws-blocks/index.ts#L56-L64)
  - `authGetLastCode()` returns the in-memory value: [`test-apps/comprehensive/aws-blocks/index.ts#L886-L888`](https://github.com/aws-devtools-labs/aws-blocks/blob/a1de3b83e5c4b095d78638e19f5614c85d1a6485/test-apps/comprehensive/aws-blocks/index.ts#L886-L888)
- In a Lambda-backed environment, the `codeDelivery` callback (triggered by `authSignUp` / `authResetPassword`) may run in one Lambda instance/execution context, while the subsequent `authGetLastCode()` call can be routed to a **different, cold Lambda instance** whose `lastDeliveredCode` is still `null`.
- When that happens, `authGetLastCode()` returns `null` even though the code *was* delivered — the retrieving instance never held it in memory. Polling the same (wrong) instance never converges, so the poll helper eventually throws `authGetLastCode() returned null after 15000ms`, or a direct caller dereferences `null` (`code!.code`).

This is a state-isolation problem between invocations, **not** a logic bug in the code under test.

## Evidence

- Surfaced during CI on **PR #146**.
- PR #146 changed **only `create-blocks-app`**, which is CLI-only and is **never imported by the failing test suite** — i.e. **zero reachability** from the PR's changes to the failing test. The change therefore could not have caused the failure; the flake is pre-existing and environmental.
- Failure signature is the classic state-isolation flake: same test file, same commit, alternating pass/fail across re-runs, always failing at the `authGetLastCode` retrieval/poll step.

## Affected test(s)

- **Primary — AuthBasic integration suite** (dedicated "last code" retrieval via the `pollForCode` helper that wraps `authGetLastCode`):
  [`test-apps/comprehensive/test/basic-auth.test.ts#L19-L28`](https://github.com/aws-devtools-labs/aws-blocks/blob/a1de3b83e5c4b095d78638e19f5614c85d1a6485/test-apps/comprehensive/test/basic-auth.test.ts#L19-L28) (the `pollForCode` helper) — used throughout the suite, e.g. `confirm signup with code — then signIn succeeds` at [L51-L66](https://github.com/aws-devtools-labs/aws-blocks/blob/a1de3b83e5c4b095d78638e19f5614c85d1a6485/test-apps/comprehensive/test/basic-auth.test.ts#L51-L66).
- **Also affected — even more susceptible** (calls `authGetLastCode()` **directly, with no polling**, then immediately dereferences the result): `Agent BB › Auth-Scoped Conversations › conversations are scoped to the authenticated user`:
  [`test-apps/comprehensive/test/agent.test.ts#L234-L266`](https://github.com/aws-devtools-labs/aws-blocks/blob/a1de3b83e5c4b095d78638e19f5614c85d1a6485/test-apps/comprehensive/test/agent.test.ts#L234-L266) (`authGetLastCode()` at L241 and L255).

## Impact

Intermittent red CI. Because the failure is unrelated to the change under test, it erodes trust in the CI signal — reviewers learn to "just re-run", which masks real regressions and slows merges.

## Suggested remediation directions (not prescriptive)

Make retrieval of the last delivered code resilient to Lambda statelessness:

- **Persist the last code in durable/shared state** instead of an in-memory module variable — e.g. a KVStore / DynamoDB / SSM Parameter Store keyed by username — so *any* Lambda instance can read the code back deterministically, regardless of which instance ran `codeDelivery`.
- **and/or** make the retrieval path robust to eventual delivery — poll/retry `authGetLastCode()` until the code appears. (The suite already polls, but polling only helps once the code lives in state that *every* instance can see; polling a single in-memory instance can never converge.)

Either direction (or both) removes the cross-instance in-memory dependency that makes the test non-deterministic.

ahmedhamouda78:
## Problem

The code-confirmed auth tests in the `comprehensive` app flake in CI. Same commit, alternating pass/fail across re-runs, always around the `authGetLastCode` read.

The harness captured the delivered verification code in a module-level variable and returned that variable from `authGetLastCode()`:

```ts
let lastDeliveredCode: { username: string; code: string } | null = null;
// codeDelivery: lastDeliveredCode = { username, code };
async authGetLastCode() { return lastDeliveredCode; }
```

That holds up locally, where a single dev-server process handles the whole flow. It does not hold up in the sandbox, where every API call can be served by a different Lambda instance, so the instance answering `authGetLastCode` is frequently not the one that ran `codeDelivery`.

There are two failure modes, not one:

1. **Cold instance** — the reader has never run `codeDelivery`, so its variable is still `null`. Polling the same instance never converges and the helper throws `authGetLastCode() returned null after 15000ms`. This is the one #172 describes.
2. **Warm instance holding someone else's code** — the reader last handled a *different* user, so its variable holds that user's record. `pollForCode` waited for "any non-null code", which a stale record satisfies immediately, so it returns the wrong user. That surfaces as `assert.strictEqual(delivered.username, username)` failing, or `confirmSignUp` rejecting the code as invalid. This one was not in the issue and polling made it worse rather than better.

`agent.test.ts` was more exposed again: it called `authGetLastCode()` with no polling at all and dereferenced the result.

## Root cause

Harness defect, not a defect in the Building Blocks. `AuthBasic.signUp` already awaits `codeDelivery` before returning, so nothing is racing inside the block. The problem is entirely that the harness kept the code in per-instance memory and exposed a single global "last code" slot with no way to ask for a specific user's code.

## Changes

- Persist delivered codes in the `KVStore` the app already has, keyed by username, so any invocation can read them. No new backend resource; keys are namespaced under `__last-code:` so the KV tests are unaffected. Applied to all three code-delivery paths (`auth`, `authC`, `authCMfa`).
- `authGetLastCode` / `authCGetLastCode` / `authCMfaGetLastCode` now take an optional username. A `__latest` pointer keeps the no-argument form working.
- Callers poll for *their* user rather than the first non-null value, so a stale record can no longer satisfy the wait. `kvGet` is eventually consistent in deployed environments, so the poll is still needed; it just needs to be scoped correctly to converge.
- Resend and reset paths wait for a code that differs from the one already consumed, since those issue a second code for the same user.
- `agent.test.ts` now polls instead of reading once and dereferencing with `!`.
- Three regression tests: read-back from shared state (verified through `kvGet`, a different code path onto the same store), per-user keying, and unknown username returning `null`.

The OIDC `onSignIn` hooks in the same file keep their module-level variables. Same class of issue, different flow, no failures attributed to them, so I left them out of this change rather than widen the diff.

## Validation

`npx tsc --noEmit` from `test-apps/comprehensive` — clean.

**Determinism, 100 iterations of the previously-flaky flow** (signUp, read code, confirm, signIn, then reset round-trip, with other users' deliveries interleaved between each delivery and read):

```
iterations : 100
passed     : 100
failed     : 0
elapsed    : 203.0s
```

**Control run, same loop against the old read semantics** (any non-null code, no username), to confirm the loop actually detects the bug:

```
iterations : 10
passed     : 0
failed     : 10
  #1: wrong user: expected loop-1-ms5hdl9c-y81p, got noise-1-b-ms5hdlqh-q98j
  #2: wrong user: expected loop-2-ms5hdlyk-88e5, got noise-2-b-ms5hdmen-vk9k
  ...
```

**Full local E2E** (`npm run test:e2e:local`):

| | tests | pass | fail | skip |
|---|---|---|---|---|
| `origin/main` baseline | 373 | 370 | 2 | 1 |
| this branch | 376 | 373 | 2 | 1 |

+3 tests, +3 passes, no new failures. The 2 failures are identical on both runs and are in `cli-client.test.ts` (`default export fails when CWD has no config.json`, `generateClient with correct URL succeeds`). Both are `npm error code E401 / Unable to authenticate` from those tests spawning `npx` on my machine, which has an expired registry token. Untouched by this change and pre-existing on main.

Changeset guards: `validate-structure` and `block-major` pass; `verify-coverage` reports `No publishable packages were changed`, so no changeset is needed — `test-apps/comprehensive` is `private: true` and no package README or version changed.

## Checklist

- [x] PR description included

---

Fixes #172

### DIFF
--- a/test-apps/comprehensive/aws-blocks/index.ts
+++ b/test-apps/comprehensive/aws-blocks/index.ts
@@ -62,28 +62,39 @@ const validatedStore = new KVStore(scope, 'validated-store', { schema: profileSc
 // `authGetLastCode` — it reads its own `null`, or worse, a leftover code from
 // some earlier user it happened to handle. Both show up as flaky auth e2e runs.
 //
-// Persist them in the shared KVStore instead, keyed by username so a reader can
-// ask for the code it is actually waiting on and never pick up another user's.
-// A `__latest` pointer preserves the no-argument form of the API.
+// Persist them in the shared KVStore instead, keyed by username. `username` is
+// required on both the write and the read: a reader can only ask for the code it
+// is actually waiting on, so there is no way to express "give me any code" and
+// pick up another user's by accident.
+//
+// One small record per test user, overwritten when that user gets a new code
+// (resend, password reset). They do pile up: the mock store persists to
+// .bb-data/ between local runs, and a suite run leaves ~30 behind. CI throws the
+// whole table away at teardown, so sweep with `authPurgeDeliveredCodes()` when
+// that is not true — local dev, or a sandbox kept via BLOCKS_SANDBOX_KEEP.
 
 type DeliveredCode = { username: string; code: string };
 type DeliveredCognitoCode = DeliveredCode & { purpose: string };
 
 const CODE_KEY_PREFIX = '__last-code';
 const codeKey = (channel: string, username: string) => `${CODE_KEY_PREFIX}:${channel}:${username}`;
-const latestCodeKey = (channel: string) => `${CODE_KEY_PREFIX}:${channel}:__latest`;
 
 async function recordDeliveredCode(channel: string, value: DeliveredCode | DeliveredCognitoCode): Promise<void> {
-  const serialized = JSON.stringify(value);
-  // Per-username first: once `__latest` moves, a reader polling for this
-  // username must already be able to find its own record.
-  await store.put(codeKey(channel, value.username), serialized);
-  await store.put(latestCodeKey(channel), serialized);
+  await store.put(codeKey(channel, value.username), JSON.stringify(value));
 }
 
-async function readDeliveredCode<T extends DeliveredCode>(channel: string, username?: string): Promise<T | null> {
-  const raw = await store.get(username ? codeKey(channel, username) : latestCodeKey(channel));
+async function readDeliveredCode<T extends DeliveredCode>(channel: string, username: string): Promise<T | null> {
+  const raw = await store.get(codeKey(channel, username));
   return raw === null ? null : (JSON.parse(raw) as T);
+}
+
+async function purgeDeliveredCodes(): Promise<number> {
+  const keys: string[] = [];
+  for await (const entry of store.scan()) {
+    if (entry.key.startsWith(`${CODE_KEY_PREFIX}:`)) keys.push(entry.key);
+  }
+  for (const key of keys) await store.delete(key);
+  return keys.length;
 }
 
 // AuthBasic - Authentication
@@ -918,14 +929,28 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
   },
 
   /**
-   * Read back the verification code delivered to `username`.
+   * Read back the verification code delivered to `username`, or `null` if none
+   * has been delivered yet.
    *
-   * Pass the username you are waiting on. Omitting it returns whichever code
-   * was delivered most recently, which is only unambiguous when a single user
-   * is in flight.
+   * `username` is required. Delivery is asynchronous and the read travels over a
+   * separate request, so a caller has to poll — and polling for "any code at
+   * all" is satisfied instantly by whatever another user left behind. Scoping
+   * the read to a username makes that mistake unexpressible.
    */
-  async authGetLastCode(username?: string) {
+  async authGetLastCode(username: string) {
     return await readDeliveredCode<{ username: string; code: string }>('auth', username);
+  },
+
+  /**
+   * Delete every recorded verification code, across all auth channels. Returns
+   * how many records were removed.
+   *
+   * Codes are one small record per test user and CI destroys the table at
+   * teardown, so this is for the cases where it does not: sweeping .bb-data/ in
+   * local dev, or a sandbox kept alive with BLOCKS_SANDBOX_KEEP.
+   */
+  async authPurgeDeliveredCodes() {
+    return { deleted: await purgeDeliveredCodes() };
   },
 
   // Cookie-attribute convergence (D-007): sign up + sign in so the e2e can
@@ -1087,7 +1112,8 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
     return { success: true };
   },
 
-  async authCGetLastCode(username?: string) {
+  /** Read back the code delivered to `username`. See `authGetLastCode`. */
+  async authCGetLastCode(username: string) {
     return await readDeliveredCode<{ username: string; code: string; purpose: string }>('authC', username);
   },
 
@@ -1175,7 +1201,8 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
     await authCMfa.signOut(context);
     return { success: true };
   },
-  async authCMfaGetLastCode(username?: string) {
+  /** Read back the code delivered to `username`. See `authGetLastCode`. */
+  async authCMfaGetLastCode(username: string) {
     return await readDeliveredCode<{ username: string; code: string; purpose: string }>('authCMfa', username);
   },
   async authCMfaFetchMFAPreference() {
--- a/test-apps/comprehensive/aws-blocks/index.ts
+++ b/test-apps/comprehensive/aws-blocks/index.ts
@@ -51,14 +51,47 @@ const objStore = new KVStore<Profile>(scope, 'obj-store');
 const profileSchema = z.object({ name: z.string(), age: z.number(), tags: z.array(z.string()) });
 const validatedStore = new KVStore(scope, 'validated-store', { schema: profileSchema });
 
+// ── Delivered verification codes (e2e read-back) ────────────────────────────
+//
+// e2e tests read verification codes back through the `getLast*Code` methods
+// below. The codes must therefore live somewhere every request can see them.
+//
+// A module-level variable does not qualify: in a deployed environment each API
+// call may be served by a different Lambda instance, so the code written by
+// `codeDelivery` on one instance is invisible to the instance that later serves
+// `authGetLastCode` — it reads its own `null`, or worse, a leftover code from
+// some earlier user it happened to handle. Both show up as flaky auth e2e runs.
+//
+// Persist them in the shared KVStore instead, keyed by username so a reader can
+// ask for the code it is actually waiting on and never pick up another user's.
+// A `__latest` pointer preserves the no-argument form of the API.
+
+type DeliveredCode = { username: string; code: string };
+type DeliveredCognitoCode = DeliveredCode & { purpose: string };
+
+const CODE_KEY_PREFIX = '__last-code';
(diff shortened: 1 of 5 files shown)


osama-rizk (commented):
Strong PR — correct fix, honest root-cause split, and a control run that most flake fixes skip. Nothing blocking. A few follow-ups below; the one I'd actually want a decision on is the `__latest` double-write.

## What works well

The fix genuinely closes both failure modes I could think of: reads are scoped to the username key, `recordDeliveredCode` writes the per-user key before `__latest`, and resend/reset use `{ not }` to ride out eventual-consistency lag on an overwritten key.

Worth calling out the **control run** specifically — running the same 100-iteration loop against the *old* read semantics and showing 10/10 "wrong user" failures proves the harness actually detects the bug. Without it, "100/100 green" only proves the test is insensitive. That's the right bar for a concurrency fix.

The regression tests hit the right seams: reading back through `kvGet` (a *different* method onto the same store) is what actually proves "not process memory" — reading through the same method that wrote it would prove nothing.

## 1. `__latest` is a dead double-write (please decide)

Every `recordDeliveredCode` now does two `store.put`s — the per-user key and `__latest` — solely to preserve the no-arg `GetLastCode()` form. But grepping the touched files plus the sandbox/e2e/oidc auth tests, **nothing calls the no-arg form anymore** — every caller passes a username. So the `__latest` write is currently a dead path that doubles write load per delivery, in the exact eventually-consistent environment this PR is stabilizing.

If the no-arg form is vestigial, drop `__latest` and the second put. If it's kept deliberately as an API affordance, a one-line comment ("retained for API completeness; no current caller") would stop a future reader from assuming it's load-bearing.

## 2. `__last-code:*` keys accumulate forever in `store`

Codes are written per unique (`Date.now()`-based) username and never deleted. Harmless today — sandboxes are ephemeral, and the `kvScan` test filters by its own prefix so the extra keys don't fail it (the "KV tests unaffected" claim checks out). Latent risk: `kvScan()` returns *all* entries, so any future unfiltered scan assertion, or a long-lived deployment, would see a growing payload of stale codes. Low priority — a cleanup-on-confirm or a note that the store is expected ephemeral would close it.

## 3. Poll logic is duplicated three ways

`signUpAndConfirm` (agent), `pollForCode` (basic-auth), `pollForCodeVia` + wrappers (auth-cognito) are three near-identical poll-with-timeout loops. Since the whole PR is about making this pattern correct, a shared `test/helpers/pollForCode.ts` would stop the next code-confirmed test from reintroducing the "any non-null" mistake. The minimal-diff scoping is defensible for a flake fix — flagging as a follow-up, not a blocker.

## 4. OIDC exclusion — please file a tracking issue

You note the OIDC `onSignIn` hooks keep their module-level variables — same bug class, no attributed failures, excluded to keep the diff narrow. Reasonable call, and the disclosure is exactly right. Just file a tracking issue so "no failures attributed yet" doesn't silently become "flakes nobody connected to this pattern." A latent instance of a bug you just proved is real should be tracked, not only mentioned in a PR body.

## Nits

- The `recordDeliveredCode` comment justifies the write order via "a reader polling for this username must already be able to find its own record… once `__latest` moves" — but username-scoped readers never read `__latest`, so the ordering is harmless either way. The ordering is fine; the rationale describes a path that doesn't occur.
- Regression test (a) hard-codes the key format `__last-code:auth:${username}`, coupling the test to the harness's private key scheme. The inline comment acknowledges it (good) — but if `__latest` stays and you extract a `codeKey` helper, sharing it would keep the format in one place.

ahmedhamouda78:
Thanks, this was a good catch on all three.

**[major] `__latest` was dead, dropped it.** I went looking for a consumer first and there isn't one. All six `*GetLastCode` call sites in `test-apps/comprehensive` pass a username (`basic-auth.test.ts:38,353,370`, `agent.test.ts:30`, `auth-cognito.test.ts:72,75`), the app's browser client never touches those methods, nothing reads a `__latest` key, and no docs reference them. The no-arg `authGetLastCode()` calls that grep turns up are in `hosting-ssr` / `hosting-spa` / `hosting-ssr-nuxt`, which each have their own backend and their own `lastDeliveredCode`, so they never reach this one.

Rather than just delete the write I made `username` **required** on all three readers. That way "give me any code", the actual bug behind #172, doesn't type check any more instead of just being discouraged. `test:e2e:local` runs `tsc --noEmit` in its before hook, so it's enforced. New test asserts the scanned key list is exactly `['__last-code:auth:<user>']`; I checked it has teeth by putting the double-write back, and it fails on the extra `__last-code:auth:__latest` key.

**[minor] The stale write-order comment is gone** with the second write. You were right that it described a path that can't happen: nothing ever read `__latest`, so there was no reader to race.

**[minor] Key accumulation turned out worse than it looks,** so I added a real cleanup rather than a note. The mock store persists to `.bb-data/` and reloads on server start, so this isn't only a deployed concern. After one `test:e2e:local`, 33 of 35 keys in `test-app-store` were `__last-code:*` residue. Added `authPurgeDeliveredCodes()` (one scan, prefix-filtered delete, returns a count) with a test that also checks it leaves non-code keys alone. I kept it as an explicit method instead of wiring it into the `e2e.test.ts` teardown, since that hook kills the server first and also drives stack destruction.

**[nit] Extracted the poller** to `test/poll-for-code.ts`, now used by `basic-auth`, `auth-cognito` and `agent` (that last one had its own hand-rolled loop). It requires a username and asserts the record belongs to it, so the next code-confirmed test can't reintroduce the "any non-null" wait. I used a flat kebab-case filename to match `cookie-jar.ts` in that directory rather than the `helpers/pollForCode.ts` you suggested; say the word and I'll move it.

One thing worth splitting out: the OIDC `onSignIn` hook has the same module-level-state shape (`lastOidcSignInUser` / `lastExtrasSignInUser` in `aws-blocks/index.ts`, read back via `oidcGetLastSignInUser` / `oidcExtrasGetLastSignInUser`). No failures pinned on it yet and it's a different flow, so I've left it out of this PR and written up a tracking issue instead of widening the diff.

Local: 375 pass / 2 fail, both pre-existing `cli-client` `E401` from my host's registry, identical on the baseline. All 25 CI checks green, including `E2E Sandbox` against real AWS.

ahmedhamouda78:
Filed as #284.

osama-rizk (approved):

soberm (approved):

PR #279, opened 2026-07-29
fix(bb-realtime): align local message envelopes with AWS runtime

### LINKED ISSUE #253: bb-realtime: mock/dev-server and AWS runtime disagree on message wire key (payload vs data)
## Summary

`bb-realtime`'s two runtimes emit **different keys** for the message body on the WebSocket wire:

| Layer | Source | Wire key |
|---|---|---|
| AWS runtime (publish) | `packages/bb-realtime/src/index.aws.ts:414` | `{ type, channel, `**`data`**` }` |
| Local dev server (publish + broadcast) | `packages/bb-realtime/src/ws-server.ts:62` and `:79` | `{ type, channel, `**`payload`**` }` |

This is an internal inconsistency in the BB's own wire protocol. AWS should be treated as the source of truth (`data`); the dev server should be aligned to match.

## Why it stayed hidden (and where it bit)

The JS clients use **conditional exports**, so each only ever talks to one backend and hardcodes the matching key:

- `packages/bb-realtime/src/mock-middleware.ts:113` reads `data.payload` (dev server) ✅
- `packages/bb-realtime/src/aws-middleware.ts:114` reads `msg.data` (AWS) ✅

Neither JS client crosses formats, so the divergence is invisible there.

The **native clients (Kotlin/Dart) do not have conditional exports** — one binary talks to *both* backends depending on `BLOCKS_URL`. So the inconsistency lands on them:

- Dart already tolerates both: `json['data'] ?? json['payload']` (`native/dart/packages/blocks_runtime/lib/src/realtime_channel.dart:74`).
- Kotlin originally read only `payload`, so its `subscribeAndReceive` e2e passed against the local dev server but timed out against a real AWS sandbox. Fixed defensively in PR #103 by matching Dart's `data ?? payload` fallback (`native/kotlin/runtime/.../realtime/RealtimeChannel.kt`).

This is exactly the "mock tests can't catch AWS-path serialization bugs" trap called out in `AGENTS.md`.

## Proposed fix (root cause)

1. Align the dev server to emit `data` instead of `payload` (`ws-server.ts:62`, `:79`), and update `mock-middleware.ts:113` to read `msg.data` in lockstep.
2. Once the two runtimes agree, **remove the `data ?? payload` workarounds** from both native clients:
   - Kotlin: `RealtimeChannel.kt` (`obj["data"] ?: obj["payload"]` → `obj["data"]`)
   - Dart: `realtime_channel.dart:74` (`json['data'] ?? json['payload']` → `json['data']`)

## Notes for the implementer

- This is a **behavior/wire-format change to a published package** → needs a changeset and (per Core rules) a **sandbox e2e**, since serialization changed. The native realtime e2e suites now exercise the subscribe→publish→receive path against a real sandbox and will catch regressions.
- Blast radius: any browser client reading `payload` from the dev server must be updated together with the mock-middleware change.
- Coordinate the client-workaround removals so they land **after** the BB change is released, to avoid breaking clients pinned to older BB versions.

## Context

Discovered while diagnosing the Kotlin JVM sandbox e2e failure in #103.

ikenyal:
## Problem

**Issue #, if available:** Fixes #253

The local Realtime WebSocket server emitted published messages as `{ type, channel, payload }`, while the AWS runtime emits `{ type, channel, data }`. This made local development expose a different wire envelope to direct and native clients.

## Changes

- Emit the AWS-compatible `data` key from both local publish paths.
- Update mock middleware delivery to read `data`.
- Add regression coverage for EventEmitter broadcasts and client-originated publishes.
- Remove the documented mock/AWS envelope difference in `packages/bb-realtime/DESIGN.md`.
- Add a patch changeset for `@aws-blocks/bb-realtime`.

## Validation

- `npm run build`
- `npm run lint`
- `npm run lint:deps`
- `npm test`
- `npm run test:e2e:local`
- `npm run build --workspace=@aws-blocks/bb-realtime && npm test --workspace=@aws-blocks/bb-realtime` (72 tests)
- Deployed the comprehensive sandbox app: the Realtime suite passed (7/7). The complete sandbox suite could not finish because the separate test account lacks an SES-verified `noreply@example.com` sender and the Bedrock agent did not respond before its timeout; the stack was destroyed afterward.

## Checklist

- [x] PR description included

---

### DIFF
--- a/.changeset/realtime-wire-key.md
+++ b/.changeset/realtime-wire-key.md
@@ -1,0 +1,5 @@
+---
+'@aws-blocks/bb-realtime': patch
+---
+
+Align local Realtime WebSocket message envelopes with the AWS runtime by using `data` for published messages.
--- a/packages/bb-realtime/DESIGN.md
+++ b/packages/bb-realtime/DESIGN.md
@@ -62,7 +62,7 @@ Both are HMAC-SHA256 signed with the same shared secret. The connect token's cha
 - Connection pool keyed by `wsUrl` (e.g., `ws://localhost:3001/realtime`)
 - Subscribe: `{ action: 'subscribe', channel, token }`
 - Response: `{ type: 'subscribe_success', channel }` or `{ type: 'error', channel, message }`
-- Data: `{ type: 'message', channel, payload }`
+- Data: `{ type: 'message', channel, data }`
 
 ### AWS Middleware
 
@@ -177,7 +177,6 @@ Channel path:             my-app-collab/chat/room-123
 
 | Behavior Difference | Impact | Mitigation |
 |------------|--------|------------|
-| WS message uses `payload` field (mock) vs `data` (AWS) | Native mobile clients see different wire format | Abstracted by SDK middlewares; document for native implementors |
 | Mock skips connect-time token auth (`?token=` query param) | A client can still *open* a WS locally, but can no longer *subscribe* without a valid channel token (enforced at subscribe time) | Recommend sandbox testing for connect-time auth-sensitive flows |
 | No connection duration limit locally | Mock WS stays open indefinitely | Document the 2-hour AWS limit in README |
 | Single-process only | No cross-process pub/sub | Local dev is single-process |
--- a/packages/bb-realtime/src/index.ts
+++ b/packages/bb-realtime/src/index.ts
@@ -130,7 +130,7 @@ export const Realtime: {
 		validatePublishSize(fullChannel, data);
 		globalEmitter.emit(fullChannel, data);
 		if (getBroadcastBus()) {
-			getBroadcastBus()!.emit('broadcast', { channel: fullChannel, payload: data });
+			getBroadcastBus()!.emit('broadcast', { channel: fullChannel, data });
 		}
 	}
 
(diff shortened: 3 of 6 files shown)


Simone319 (commented):
Thank you for making the fix. 

One minor comment: 
Shall we update the `payload` from the root [here](https://github.com/aws-devtools-labs/aws-blocks/blob/main/packages/bb-realtime/src/index.ts#L133) as well? So that we don't need multiple payload -> data translation which is easy to miss.

ikenyal:
Thanks, that is a good point. I updated the local broadcast bus to emit `{ channel, data }` directly from `Realtime.publish()` and updated `ws-server` to consume that shape, removing the intermediate `payload` to `data` translation.

I also updated the direct broadcast and reconnect regression tests. The client-originated WebSocket publish request remains `{ action, channel, payload }` because it is a separate existing input protocol; emitted message envelopes remain `{ type, channel, data }`.

Validated with:
- `npm run build`
- `npm run lint`
- `npm run lint:deps`
- `npm test`
- `npm run test:e2e:local`

Simone319 (approved):

mattcreaser (approved):

PR #281, opened 2026-07-29
feat(bb-async-job): opt-in job status tracking with transition history

### LINKED ISSUE #223: AsyncJob: queued->processing->complete transition not observable without an artificial setTimeout delay
@aws-blocks/bb-async-job

## Describe the bug
To make the queued->processing->complete state transition observable to a client, consumers must hard-code an artificial `setTimeout(1500)` in the processor. This is a tuned timing assumption, not a guarantee: it races against client/test polling cadence, so the intermediate state can be missed or the transitions can race.

## Evidence
document-ingestion-pipeline, preamp batch manual-deploy-20260717-103949-1339. All 3 reps added the delay; 2/3 reps then failed the majority of functional tests (judge 6.0/6.4) while the one that worked scored 82%. Event-driven frameworks on the same task hit ~86%.

## Expected
A deterministic way to observe async job state transitions that does not depend on a tuned delay.

## Suggested fix
Expose a first-class status/subscription API for job state so intermediate states are reliably observable without injecting processor delays

ahmedhamouda78:
Fixes #223

## The problem

Watching a job go from `queued` to `processing` to `complete` required padding the handler with an artificial `setTimeout(1500)`. There was no status API at all, so [the README](https://github.com/aws-devtools-labs/aws-blocks/blob/ac0966a/packages/bb-async-job/README.md#how-do-i-know-my-job-ran) told people to write their own state into a KVStore from inside the handler, and `test-apps/comprehensive` does exactly that plus a 20x100ms poll loop.

The catch is that hand-rolled state is a single mutable value. A handler that finishes in a millisecond passes through `processing` faster than any client can poll it, so the delay was not really about notifications, it was there to widen the window enough to be caught. Any tuned value races against the reader's cadence, which is what #223 saw: intermediate states missed, or the transitions racing each other.

## The fix

Pass `trackStatus: true` and AsyncJob records the lifecycle itself:

```ts
const job = new AsyncJob(scope, 'ingest', {
  trackStatus: true,
  handler: async (payload) => { await ingest(payload.documentId); },
});

const { jobId } = await job.submit({ documentId: 'doc-1' });

const status = await job.waitUntilComplete(jobId);
status.transitions.map((t) => t.state);  // ['queued', 'processing', 'complete']
```

Two new methods:

- **`getStatus(jobId)`** returns the current state plus the full transition history, or `null` for an unknown id.
- **`waitUntilComplete(jobId, options?)`** waits for `complete` or `failed`, with `timeoutMs` (default 30000), `pollIntervalMs` (default 250, +/-20% jitter) and an `AbortSignal`. It follows the shape `KnowledgeBase.waitUntilSynced()` already established so there is one poll-until idiom in the repo rather than two.

**Transitions are appended, not overwritten.** That is the part that actually removes the race: a caller reading the record once, long after the job settled, still sees that it went through `processing`. No timing assumption anywhere, and nothing to tune. A retry appends another `processing` entry instead of a second terminal state, so attempt counts stay readable:

```ts
status.transitions.map((t) => `${t.state}#${t.attempt}`);
// ['queued#0', 'processing#1', 'processing#2', 'failed#2']
```

Storage is a nested `DistributedTable` keyed by `jobId` with a 24 hour TTL, composed the same way bb-agent, bb-auth-cognito and bb-realtime already compose child blocks: identical child id and options in the runtime and CDK constructors. Because `DistributedTable` has its own conditional exports, there is one status code path for both mock and AWS rather than two.

## Why opt-in

Tracking is not free. It adds a DynamoDB table per job plus a write on submit and one per transition, and it would turn `submitBatch` from a single native SQS batch into an extra batch write, which is exactly the kind of hidden cost G14/G18 warn about. With the flag off nothing is provisioned, `submit()` stays a single SQS call, existing deployments gain no resources, and the status methods throw `StatusNotTracked` pointing at the flag. Flipping the default on later is a non-breaking change if you would rather have it always on, so this is the reversible direction.

Happy to switch it to always-on if you prefer that trade, it is a small change.

## Two details worth a look

**Status writes never decide a job's fate.** On the handler path they are logged, not thrown. Throwing before the handler would retry work that was fine; throwing after it succeeded would re-run work that had already completed. The `queued` write in `submit()` does propagate, since the caller explicitly asked for tracking and failing there is safe.

**Terminal failure is deferred to SQS.** A handler error only records `failed` once `receiveCount` has reached `maxRetries`. SQS redrive owns the retry decision, so earlier failures record nothing and the next delivery just appends another `processing` entry.

## Testing

15 new tests in `packages/bb-async-job/src/status.test.ts`. **No handler in the file contains a `setTimeout`**, which is the whole point: the transition-sequence assertions pass against handlers that return immediately.

Covered: the queued/processing/complete sequence, reading the history only after the job settled, transition ordering and timestamps, the retry path recording `processing` twice then `failed`, `queued` observable while a job is delayed, `submitBatch` tracking every job, `getStatus` returning `null` for an unknown id, `waitUntilComplete` timing out mid-flight and then succeeding once the job finishes, rejecting with the signal's abort reason (both live and pre-aborted), `StatusNotTracked` when the flag is off, and isolation between two tracked jobs.

Two of them guard a trap I hit while writing this: the CDK layer infers a key's DynamoDB attribute type by probing `validate({ jobId: 0 })` and reads "no issue for that field" as numeric, so a pass-through schema would have provisioned `jobId` as `N` and every runtime write would have failed against it. The status schema rejects a numeric `jobId`, and there is now a test pinning that, since per-block cdk parity is not covered by `conditional-exports.test.ts`.

To check the tests are not toothless I disabled the append (`transitions: [transition]`, the old mutable-value behaviour) and re-ran: 5 of them fail. Restored, all 35 pass.

## Verification

| Check | Result |
|---|---|
| `npm run build:packages` | pass |
| `npm test -w packages/bb-async-job` | 35/35 pass (20 existing + 15 new) |
| `npm test -w packages/blocks` | 41/41 pass, includes bb-async-job aws-runtime/default export parity |
| `npm test` (all workspaces) | 0 failures |
| `npm run check:api` | exit 0, `API.md` regenerated via `npm run update:api` and committed for both bb-async-job and blocks |
| `changeset-guard validate-structure` | pass |
| `changeset-guard verify-coverage` | pass, both changed packages covered |
| `changeset-guard block-major` | pass |

`npm run lint` / `lint:deps` could not run here, the bundled biome binary needs GLIBC 2.28+ and this host is older. The new dependency is declared (`@aws-blocks/bb-distributed-table` in `dependencies`, `@standard-schema/spec` stays a devDependency as it is type-only) and the tsconfig project reference is added, so I expect it clean in CI.

Changeset is `bb-async-job: minor` plus `blocks: patch`, since the umbrella re-exports the new types and its `API.md` changed too.

### DIFF
--- a/.changeset/lucky-pears-observe.md
+++ b/.changeset/lucky-pears-observe.md
@@ -1,0 +1,15 @@
+---
+"@aws-blocks/bb-async-job": minor
+"@aws-blocks/blocks": patch
+---
+
+Add opt-in job status tracking to AsyncJob
+
+Pass `trackStatus: true` and AsyncJob records each job's lifecycle, which you can read with two new methods:
+
+- `getStatus(jobId)` returns the job's current state plus every state it has passed through.
+- `waitUntilComplete(jobId, options?)` waits until the job reaches `complete` or `failed`, with `timeoutMs`, `pollIntervalMs`, and `AbortSignal` support.
+
+Transitions are appended rather than overwritten, so intermediate states stay observable no matter when you read them. A handler that finishes in a millisecond still records that it went through `processing`, and a caller that checks once after the job settled sees the whole sequence. That removes the need to pad a handler with an artificial delay just to make the `processing` state catchable, and a retry appends another `processing` entry so attempt counts are visible too.
+
+Enabling the flag provisions one DynamoDB table for the job's status records, with a 24 hour TTL, and adds a write on submit plus one per state change. Leave it off and nothing is provisioned; `submit()` stays a single SQS call and the status methods throw `StatusNotTracked`.
--- a/.changeset/lucky-pears-observe.md
+++ b/.changeset/lucky-pears-observe.md
@@ -12,4 +12,8 @@ Pass `trackStatus: true` and AsyncJob records each job's lifecycle, which you ca
 
 Transitions are appended rather than overwritten, so intermediate states stay observable no matter when you read them. A handler that finishes in a millisecond still records that it went through `processing`, and a caller that checks once after the job settled sees the whole sequence. That removes the need to pad a handler with an artificial delay just to make the `processing` state catchable, and a retry appends another `processing` entry so attempt counts are visible too.
 
+Appends are guarded by a compare-and-swap, so the two ways two writers can hold the same record at once cannot drop a transition: a `queued` write arriving after SQS already delivered the message, and a duplicate delivery of the same message on an at-least-once queue.
+
 Enabling the flag provisions one DynamoDB table for the job's status records, with a 24 hour TTL, and adds a write on submit plus one per state change. Leave it off and nothing is provisioned; `submit()` stays a single SQS call and the status methods throw `StatusNotTracked`.
+
+Status writes on the handler path are logged rather than thrown, so bookkeeping can never retry work that succeeded or mask work that failed. The trade is that a dropped terminal write leaves a finished job without a terminal state, so read `waitUntilComplete()`'s `Timeout` as "status unknown" rather than "still running".
--- a/.changeset/lucky-pears-observe.md
+++ b/.changeset/lucky-pears-observe.md
@@ -1,5 +1,5 @@
 ---
-"@aws-blocks/bb-async-job": minor
+"@aws-blocks/bb-async-job": patch
 "@aws-blocks/blocks": patch
 ---
 
--- a/.changeset/lucky-pears-observe.md
+++ b/.changeset/lucky-pears-observe.md
@@ -14,6 +14,8 @@ Transitions are appended rather than overwritten, so intermediate states stay ob
 
 Appends are guarded by a compare-and-swap, so the two ways two writers can hold the same record at once cannot drop a transition: a `queued` write arriving after SQS already delivered the message, and a duplicate delivery of the same message on an at-least-once queue.
 
+When the handler gets there first it creates the record itself, dating the submission from the moment it first saw the job, since that is all it knows. The `queued` write that arrives afterwards replaces that placeholder with the real submission time instead of dropping it, so `submittedAt` and the first transition always report when the job was submitted rather than when it started being processed.
+
 Enabling the flag provisions one DynamoDB table for the job's status records, with a 24 hour TTL, and adds a write on submit plus one per state change. Leave it off and nothing is provisioned; `submit()` stays a single SQS call and the status methods throw `StatusNotTracked`.
 
 Status writes on the handler path are logged rather than thrown, so bookkeeping can never retry work that succeeded or mask work that failed. The trade is that a dropped terminal write leaves a finished job without a terminal state, so read `waitUntilComplete()`'s `Timeout` as "status unknown" rather than "still running".
--- a/packages/bb-async-job/src/status.ts
+++ b/packages/bb-async-job/src/status.ts
@@ -1,0 +1,280 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { DistributedTable } from '@aws-blocks/bb-distributed-table';
+import type { ScopeParent } from '@aws-blocks/core';
+import type { ChildLogger } from '@aws-blocks/bb-logger';
+import type { StandardSchemaV1 } from '@standard-schema/spec';
+import type {
+	AsyncJobState,
+	AsyncJobStatus,
+	AsyncJobTransition,
+	WaitUntilCompleteOptions,
+} from './types.js';
+import { AsyncJobErrors } from './errors.js';
+
+/** Child scope id of the status table. Must match between the runtime and CDK entry points. */
+export const STATUS_TABLE_ID = 'status';
+
+/** How long a status record survives after its last transition. */
+export const STATUS_RETENTION_SECONDS = 86_400;
+
+const DEFAULT_TIMEOUT_MS = 30_000;
+const DEFAULT_POLL_INTERVAL_MS = 250;
+
+/** Stored shape: the public record plus the DynamoDB TTL attribute. */
+interface StatusRecord extends AsyncJobStatus {
+	/** Epoch seconds at which DynamoDB may delete this record. */
+	expiresAt: number;
+}
+
+const STRING_FIELDS = ['jobId', 'state', 'submittedAt', 'updatedAt', 'error'] as const;
+
+/**
+ * Runtime validation for status records.
+ *
+ * This schema is deliberately explicit about field types rather than passing
+ * values through: the CDK layer infers a key's DynamoDB attribute type by
+ * probing `validate({ jobId: 0 })` and treats "no issue for that field" as
+ * numeric. A pass-through schema would therefore provision `jobId` as `N`.
+ */
+export const statusSchema: StandardSchemaV1<StatusRecord> = {
+	'~standard': {
+		version: 1,
+		vendor: 'blocks',
+		validate: (value: unknown) => {
+			if (typeof value !== 'object' || value === null || Array.isArray(value)) {
+				return { issues: [{ message: 'status record must be an object' }] };
+			}
+			const record = value as Record<string, unknown>;
+			const issues: Array<{ message: string; path: [string] }> = [];
+
+			for (const field of STRING_FIELDS) {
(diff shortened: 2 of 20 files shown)


packages/bb-async-job/src/status.ts
```diff
@@ -0,0 +1,325 @@
...
+	async recordTransition(
+		jobId: string,
+		state: AsyncJobState,
+		attempt: number,
+		error?: string,
+	): Promise<void> {
+		for (let cas = 1; ; cas++) {
+			const existing = await this.table.get({ jobId });
+			const now = new Date().toISOString();
+			const base = existing ?? this.queuedRecord(jobId, now);
```
soberm, line 303: [minor] When `recordTransition` backfills a missing record (handler wins the race against `submit()`'s queued write), `submittedAt` gets set to `now`, i.e. the processing time, not the actual submission time. When `recordQueued()` shows up afterward it loses the conditional write (`ifNotExists` fails) and just gets swallowed, so that correct `submittedAt` is thrown away for good. The record's `submittedAt`/`transitions[0].at` end up permanently wrong for any job that hits this race, even though the README documents `submittedAt` as "ISO 8601 timestamp of submission."

Probably fine to leave as "time first observed" if that's the intended semantics, but worth a doc note, or alternatively have `recordQueued` patch just `submittedAt`/the first transition's `at` on the existing record instead of a no-op when it loses the race.

ahmedhamouda78: Good catch. Went with option (b): `recordQueued` now corrects `submittedAt` and the queued transition's timestamp on the existing record when the handler wins the race, rather than no-op'ing. It reuses the same CAS-on-`version` pattern as the rest of the tracker and only ever moves the timestamp earlier, so it stays idempotent under SQS retries. The README semantics for `submittedAt` still hold. Added tests for the race (handler backfills first, then the late queued write corrects it) and confirmed they fail without the fix. Pushed in 2fbc2a5.

osama-rizk (commented):
Excellent PR — the append-only reframe is the *correct* fix for #223 (not just a wider delay window), the CAS/ordering reasoning is sound, and the test suite reaches the concurrency paths the mock can't. Nothing blocking; Finding 1 is the one I'd want addressed, even if only as a doc caveat.

## What I verified

**Redrive alignment** — this is where the feature breaks silently, so I checked it: CDK sets `maxReceiveCount = maxRetries` (same option, both default 3), and the handler records `failed` exactly when `receiveCount >= maxRetries`. So `failed` lands on the same delivery that exhausts redrive — not early (which would falsely mark a job dead mid-retry), not never.

**Tests reach the paths the mock can't.** The mock processes in-process and can never reproduce the AWS submit-loses-the-create-race ordering, so there are direct `JobStatusTracker` tests for concurrent appends not losing a transition, the late `queued` write not clobbering an earlier `processing`, backdating, and TTL-reaped records not being resurrected. And disabling the append to watch 5 tests fail is the vacuity control — proof the sequence assertions actually bite.

## 1. A job killed by Lambda (timeout/OOM) reaches the DLQ but never records `failed`

`failed` is only recorded inside the handler's `catch`. But the most common reasons a job exhausts retries and hits the DLQ are *not catchable throws* — Lambda timeout, OOM kill, or a process crash. In those cases the invocation is terminated by the platform, the `catch` never runs, no `failed` is written, and the record is stranded at `processing` until its 24h TTL. A caller then sits in `waitUntilComplete` until it throws `Timeout` (default 30s) on a job that actually died — which is exactly the "terminal state not observable" shape #223 was about, relocated from `complete` to `failed`.

This is inherent to in-handler recording (you can't `catch` a SIGKILL), so I'm not asking for a full fix here. But it should be explicit, because the docs currently imply `failed` is reliable:
1. Document the gap — "`failed` is recorded for handler *exceptions*; a job killed by Lambda timeout/OOM lands in the DLQ without a `failed` transition, and `waitUntilComplete` will time out. Monitor the DLQ for those."
2. At minimum, have `waitUntilComplete`'s `Timeout` message hint that a timeout may mean the job died without recording — so the caller doesn't read "timeout" as "still running."

## 2. Both the `failed` decision and `attempt` numbers ride on `ApproximateReceiveCount`

`ctx.receiveCount` is SQS's `ApproximateReceiveCount`, which the SQS docs say can read higher than the true delivery count (e.g. a message becomes visible again after a visibility-timeout expiry on a long-running handler, with no real failure). Two consequences:
- The `receiveCount >= maxRetries` gate could fire `failed` a delivery early if the count inflated — recording `failed` while SQS still has a real attempt left, so a job could show `failed` and then a later `processing`/`complete` appended after it. The append-only model tolerates this, but `status.state` briefly lies.
- `attempt` numbers are therefore approximate, not a true retry count. The type doc says "delivery attempt that produced this transition" — worth softening to "approximate delivery count (SQS `ApproximateReceiveCount`)" so nobody builds exact retry accounting on it.

Not a bug you introduced — SQS semantics — but hinging the terminal-state *decision* on an approximate counter deserves a comment where the `>=` compare lives.

## 3. `waitUntilComplete` read cost (non-blocking)

At 250ms over a 30s budget, a single wait is up to ~120 DynamoDB `GetItem`s; N concurrent waiters multiply that. The ±20% jitter correctly prevents synchronized stampedes but not absolute volume. Inherent to poll-based waiting, and mirroring `KnowledgeBase.waitUntilSynced()` for one idiom is the right consistency call — keep it. Worth a doc line that a tight `pollIntervalMs` against many jobs has an RCU implication (ties to the G14/G18 hidden-cost reasoning the PR already invokes for opt-in).

## 4. Transition array is unbounded in principle (low severity)

Every delivery appends a `processing`, and the whole record is rewritten on each CAS. Trivial at the default `maxRetries` of 3, but a high `maxRetries` (or receive-count inflation per #2) grows the array toward DynamoDB's 400KB item ceiling. Worth knowing, not worth fixing — a one-line doc note if `maxRetries` can be set arbitrarily high.

## Security note
Handler error messages are stored in `error` and returned by `getStatus`, persisted for 24h and readable by anyone who can call it. Probably fine for first-party use, but a one-line caution ("don't put secrets in thrown error messages; they're persisted") would match the care elsewhere.

## Nits
- The mock passes `sentAt` to `recordQueued` while AWS passes a fresh timestamp at send time — functionally equivalent, but the mock always wins the create race so it can't exercise backdating (which is exactly why the direct-tracker tests exist). Noting only that happy-path mock coverage ≠ AWS-path coverage here; the author clearly knew this.
- The `blocksError`/`err.name` helper duplicates a pattern `isBlocksError` consumes — if a shared `createBlocksError(name, message)` exists in core, prefer it. Not worth churning otherwise.

Also: the CDK-key-type trap you found and pinned with a test (`validate({jobId:0})` probing → `N` provisioning) is a genuinely good catch — exactly the kind of cross-layer footgun that's invisible until it bites at runtime.

ahmedhamouda78:
Good catches, both documented in c6e22f4.

The killed-delivery gap is real and you described it exactly: `failed` is only ever written from the handler's `catch`, so a Lambda timeout or OOM kill never reaches it, the record sits at `processing`, and `waitUntilComplete` reports Timeout instead of the dead job it is. I wrote that up as a known limitation next to the existing Timeout guidance, including where to look instead, since SQS still redelivers and the message still lands in the DLQ after `maxRetries`. Detecting it inside AsyncJob is a bigger change than this PR should carry, so I left it as follow-up rather than bolting it on here.

On `receiveCount`, you are right that it is SQS's `ApproximateReceiveCount` and can move without a matching attempt, and `attempt` / `attempts` inherit that. Noted under Handler Context. Worth flagging one thing I hit while checking: the mock's counter is exact, so a test asserting an exact attempt number passes locally and can still be wrong in AWS. That is in the note too.

osama-rizk (dismissed):

.changeset/lucky-pears-observe.md
```diff
@@ -0,0 +1,21 @@
+---
+"@aws-blocks/bb-async-job": minor
```
soberm: Why a minor bump? Is there a breaking change in this PR?

ahmedhamouda78: No breaking change; all the new API changes are opt-in not forced. I changed it to patch.

osama-rizk (approved):

soberm (approved):

PR #282, opened 2026-07-30
fix(bb-distributed-table): secure-by-default durability & encryption, sourced from stack BlocksDefaults

osama-rizk:
## Summary

Fixes the bug-bash finding **"DDB tables ship with PITR OFF, DeletionProtection OFF, SSE-KMS OFF"** (project [aws-amplify/141](https://github.com/orgs/aws-amplify/projects/141/views/1?pane=issue&itemId=195525463)).

Every production DynamoDB table emitted by `bb-distributed-table` was created with only `tableName`, keys, `billingMode`, and TTL, so CloudFormation defaulted to `DeletionProtection: false` (a stray `cdk destroy`/console delete is unrecoverable), `PointInTimeRecovery: DISABLED` (no restore from bad writes), and `SSESpecification: null` (an AWS-owned key, no CloudTrail auditability).

This makes production tables **secure by default** — and does it by building on the **stack-wide `BlocksDefaults`** mechanism from #302 rather than reinventing per-block `sandboxMode` logic.

## Built on `BlocksDefaults` (#302)

Durability posture (removal policy, deletion protection, continuous backups) is resolved from the stack's `BlocksDefaults` — the shared posture object every Building Block reads via `scope.defaults`, seeded from `BlocksPresets.production` / `BlocksPresets.sandbox`. The rule everywhere is `perBlockOption ?? scope.defaults.field`, with each field read independently.

- **Small core addition (`@aws-blocks/core`, minor):** this PR adds **`pointInTimeRecovery: boolean`** to `BlocksDefaults` (and the presets — `true` in `production`, `false` in `sandbox`), the DynamoDB continuous-backup posture. `removalPolicy` and `deletionProtection` already shipped in #302. > The property **name** wants API-BR sign-off before release (noted in the changeset).
- The block **no longer reads the `sandboxMode` context for durability** — that posture now flows through `defaults`. (`sandboxMode` is still read only by the GSI custom resource's drop-and-recreate fast path, an operational concern.)

## Public options on `DistributedTableOptions` (all overrides over `defaults`)

```ts
interface DistributedTableOptions<T, K, Indexes> {
  // ...existing: schema, key, indexes, ttl, table, logger, readValidation
  protection?: 'disposable' | 'retained' | 'locked';                    // default: from stack defaults
  pointInTimeRecovery?: boolean;                                        // default: defaults.pointInTimeRecovery
  pointInTimeRecoveryDays?: number;                                     // 1–35, default 35 (PITR window)
  encryption?: 'aws-managed' | 'customer-managed' | ExternalKmsKeyRef;  // default: 'aws-managed'
}
```

- **`protection`** is one knob spanning removal policy + deletion protection, so the contradictory `deletionProtection: true` + `removalPolicy: 'destroy'` state (which wedges stack deletion in `DELETE_FAILED`) can't be expressed: `'disposable'` → DESTROY + unprotected; `'retained'` → RETAIN + unprotected; `'locked'` → RETAIN + protected. **When omitted, removal policy and deletion protection come from `defaults`** (`BlocksPresets.production` ≈ `'locked'`, `sandbox` ≈ `'disposable'`).
- **`pointInTimeRecovery`** overrides `defaults.pointInTimeRecovery` for one table; **`pointInTimeRecoveryDays`** (1–35) narrows the recovery window to trim backup-storage cost (maps to `recoveryPeriodInDays`).
- **`encryption`** also accepts `DistributedTable.fromKmsKey(arn)` — a branded `ExternalKmsKeyRef` — so several tables can **share one existing CMK** instead of each `'customer-managed'` table minting a dedicated key. Resolved via `Key.fromKeyArn` in the CDK layer; the CDK `IKey` type never leaks into the runtime-agnostic public types. SSE isn't part of `BlocksDefaults` (a per-block concern), defaulting to the AWS-managed `aws/dynamodb` key.
- **Synth-time `addWarningV2` warnings** for: an unrecognized `protection`/`encryption` value (falls back to the stack default), an out-of-range `pointInTimeRecoveryDays`, and durability/encryption options passed alongside `fromExisting()` (which ignores them).

`fromExisting()` tables are otherwise untouched — the customer owns their config.

## Why gate at the construct (via `defaults`) rather than the stack aspect

The natural alternative — enable deletion protection unconditionally and let the `SandboxDisableDeletionProtection` stack mixin relax it in sandbox — **doesn't work**: that mixin duck-types on a `deletionProtection` *instance* property, which the DynamoDB L2 `Table` doesn't expose (constructor prop only; pinned by `core/src/cdk/mixins.test.ts`). Reading `this.defaults.deletionProtection` in the constructor is exactly why #302's construct-level mechanism exists (the mixin is now `@deprecated`). See `DESIGN.md` D-DT-9.

## Validation

- `npm run build:force` ✅ · `npm run lint` ✅ · `npm run lint:deps` ✅ · `npm run check:api` ✅ (up-to-date)
- `@aws-blocks/core`: **673/673** · `bb-distributed-table`: **137/137** · full workspace `npm test`: green
- `@aws-blocks/blocks` conditional-export parity holds
- CDK synth tests cover: the `protection` matrix, PITR window (incl. out-of-range fallback), PITR follows `defaults.pointInTimeRecovery` and per-block override, `fromKmsKey` provisions zero new keys, and durability adopting the stack `defaults` (sandbox vs production preset)

Rebased on latest `main` (includes #302); merges cleanly.

## Changeset

- `@aws-blocks/core` — **minor** (new `BlocksDefaults.pointInTimeRecovery`)
- `@aws-blocks/bb-distributed-table` — **patch** (secure defaults + `protection`/`pointInTimeRecovery(Days)`/`encryption`/`fromKmsKey`)
- `@aws-blocks/blocks` — **minor** (umbrella re-export of the expanded `BlocksDefaults` surface)

## Open for maintainer / bar-raiser

- **Rollout/comms:** the default-posture change means existing prod tables gain PITR + deletion protection + `Retain` on their next deploy (in-place updates, no replacement). A future `cdk destroy` then refuses until protection is relaxed, and the retained table must be removed/imported before re-deploying the same app. How many prod consumers + comms beyond the changelog is a maintainer call.
- **API-BR:** the `BlocksDefaults.pointInTimeRecovery` property name (new `@aws-blocks/core/cdk` surface).

## Checklist
- [x] build / lint / lint:deps / unit / parity green locally
- [x] Changeset added (core + bb-distributed-table + blocks)
- [x] README + DESIGN + JSDoc updated in this PR
- [x] Conditional-export parity holds; runtime methods keep `synthGuard` stubs
- [x] New behavior ships tests (synth + real-deploy e2e)
- [x] Breaking-ish default change flagged above for maintainer review


packages/bb-distributed-table/src/index.cdk.ts
```diff
@@ -71,6 +71,28 @@ export class DistributedTable<T = any> extends Scope {
...
+		// way so `sandbox:destroy` stays a one-command teardown and throwaway
+		// stacks don't accrue backup cost. An explicit option always wins.
+		//
+		// NOTE: deletion protection is gated here (not left to the stack-level
+		// SandboxDisableDeletionProtection mixin) because that mixin duck-types
+		// on a `deletionProtection` instance property, which the DynamoDB L2
+		// Table does not expose — so it can't relax it after the fact.
+		const pitrEnabled = config.pointInTimeRecovery ?? !isSandbox;
+		const deletionProtection = config.deletionProtection ?? !isSandbox;
+		const removalPolicy = config.removalPolicy === 'destroy'
```
soberm: [minor] Both of these fall through to the environment default on any unrecognized string (typo, wrong casing, whatever), since `options` is typed `any` here. For `encryption` that's harmless, worst case you silently get the safer default. For `removalPolicy` it's not: a typo'd `retain` in sandbox falls through to `DESTROY`, the opposite of what someone typed. Might be worth throwing (or at least an `Annotations` warning) on an unrecognized value instead of treating it like `undefined`. Same pattern already exists in bb-kv-store so this isn't new, just flagging it since these are literally the durability knobs this PR is about.

packages/bb-distributed-table/src/types.ts
```diff
@@ -44,6 +44,52 @@ export interface DistributedTableOptions<
...
+	 * - `'retain'`: keep the table (orphan it) — the safe default for
+	 *   production so stack teardown never destroys data.
+	 * - `'destroy'`: delete the table with the stack.
+	 *
+	 * Defaults to **`'retain'` on production deploys** and **`'destroy'` in
+	 * sandbox mode**. Set explicitly to override.
+	 */
+	removalPolicy?: 'retain' | 'destroy';
 	/** Wrap an existing table instead of creating one. */
 	table?: ExternalTableRef;
```
soberm, line 176: [minor] Now that there are four durability/encryption options, might be worth wiring up an `Annotations.of(this).addWarningV2(...)` synth warning (same pattern as pipeline-construct.ts / hosting_construct.ts) for when someone passes e.g. `deletionProtection` or `encryption` alongside `table`. Right now it's a silent no-op, documented in README/DESIGN.md but not surfaced anywhere at synth time. Easy to imagine someone setting `deletionProtection: true` on what they think is a fresh table, not realizing fromExisting means the option just gets ignored.

packages/bb-distributed-table/src/index.cdk.test.ts
```diff
@@ -57,6 +65,156 @@ test('CDK: default DistributedTable provisions a DynamoDB table', () => {
...
+		encryption: 'customer-managed',
+	});
+	const template = Template.fromStack(stack);
+	template.resourceCountIs('AWS::KMS::Key', 1);
+	template.hasResourceProperties('AWS::DynamoDB::Table', {
+		SSESpecification: { SSEEnabled: true, SSEType: 'KMS' },
+	});
+});
+
+test('CDK: fromExisting does not emit durability props (customer owns the table)', () => {
```
soberm: [minor] This is identical to the test right below it (`fromExisting does NOT provision a table (regression)`), same setup, same single assertion on `resourceCountIs('AWS::DynamoDB::Table', 0)`. The name promises a check that durability props aren't emitted, but since no table resource exists at all there's nothing to check props on, it's really just re-testing that fromExisting doesn't provision a table. If the goal is to prove the new durability options are ignored when combined with `table`, it'd be more useful to pass `pointInTimeRecovery: true, deletionProtection: true, encryption: 'customer-managed'` here and also assert `resourceCountIs('AWS::KMS::Key', 0)`, that'd actually exercise the ignore behavior instead of duplicating the test below it.

packages/bb-distributed-table/src/index.cdk.ts
```diff
@@ -71,6 +71,28 @@ export class DistributedTable<T = any> extends Scope {
...
 			isNumericField(fieldName) ? AttributeType.NUMBER : AttributeType.STRING;
 
+		// Secure-by-default durability & encryption.
+		// Production tables enable PITR (continuous backups), deletion
+		// protection, and are retained on stack delete so a stray `cdk destroy`
+		// can't wipe customer data unrecoverably. Sandboxes default the opposite
+		// way so `sandbox:destroy` stays a one-command teardown and throwaway
+		// stacks don't accrue backup cost. An explicit option always wins.
+		//
+		// NOTE: deletion protection is gated here (not left to the stack-level
```
soberm: [nit] This comment block (through line 84) is close to verbatim from DESIGN.md's D-DT-9 writeup on why deletion protection is gated in the constructor. The bullet you added to DESIGN.md's CDK infra list a few lines down just says "(see D-DT-9)" instead of repeating itself, might be worth doing the same here, e.g. `// Gated here rather than the stack-level mixin, see DESIGN.md D-DT-9.` One source of truth for the rationale instead of two copies to keep in sync.

.changeset/bb-distributed-table-secure-defaults.md
```diff
@@ -0,0 +1,21 @@
...
+- **Deletion protection** enabled
+- **SSE-KMS** with the AWS-managed `aws/dynamodb` key (auditable, no per-key charge)
+- **`RemovalPolicy.RETAIN`** (stack deletion orphans rather than destroys the table)
+
+In **sandbox mode** (`--context sandboxMode=true`) PITR and deletion protection stay off and the removal policy is `DESTROY`, so throwaway stacks remain cheap and `sandbox:destroy` still tears down in one command. SSE-KMS is on in both.
+
+Every default is overridable per table via new options — `pointInTimeRecovery`, `deletionProtection`, `encryption` (`'aws-managed' | 'customer-managed'`), and `removalPolicy` (`'retain' | 'destroy'`). Explicit values always win over the environment default. Tables bound via `fromExisting()` are unaffected.
+
+> **Behavior change on next production deploy of an existing app:** the table will gain PITR, deletion protection, an SSE-KMS specification, and a `Retain` deletion policy. These are in-place updates (no table replacement). Because deletion protection becomes enabled, a future `cdk destroy` of a prod stack will refuse to delete the table until you disable protection or pass `deletionProtection: false`.
```
soberm: [major] Agreed with your own callout here, this is the one part of the PR I'd want an explicit maintainer/bar-raiser sign-off on before merge, since it changes behavior for every existing prod table on its next deploy with no opt-in required. The mechanics look right, PITR/DeletionProtection/SSE/RemovalPolicy are all no-interruption updates per the CFN docs so there's no replacement risk. My question is more about rollout: do we know how many internal apps currently consume DistributedTable in prod, and is there a comms plan beyond this changeset note? Most consumers won't read it until after deletion protection is already on and they hit a cdk destroy refusal.

packages/bb-distributed-table/README.md
```diff
@@ -40,7 +40,11 @@ const table = new DistributedTable(scope, id, options)
 | `key` | `TableKeyConfig<T>` | Yes | Primary key configuration: `{ partitionKey, sortKey? }`. Field names must exist in the schema. |
 | `indexes` | `Record<string, TableKeyConfig<T>>` | No | Global secondary index definitions. |
 | `ttl` | `keyof T & string` | No | Enable DynamoDB TTL on the specified attribute. The field should contain a Unix epoch timestamp in seconds. |
-| `table` | `ExternalTableRef` | No | Wrap an existing DynamoDB table instead of creating one. |
+| `pointInTimeRecovery` | `boolean` | No | Enable Point-in-Time Recovery (continuous backups, 35-day window). **Defaults to `true` in production, `false` in sandbox.** |
+| `deletionProtection` | `boolean` | No | Block table deletion until turned off. **Defaults to `true` in production, `false` in sandbox.** |
+| `encryption` | `'aws-managed' \| 'customer-managed'` | No | At-rest encryption key type. `'aws-managed'` (default) uses the `aws/dynamodb` KMS key (auditable, no key charge); `'customer-managed'` provisions a dedicated CMK. |
```
soberm: [minor] Worth a one-line callout that `customer-managed` provisions a brand-new dedicated CMK per table (confirmed by the `resourceCountIs('AWS::KMS::Key', 1)` test), not a shared one. An app with a dozen tables all opting into customer-managed encryption ends up with a dozen KMS keys at about a dollar a month each plus request charges, and there's currently no way to pass in an existing key so multiple tables could share one. Not necessarily asking for that as a feature, just flagging it so it's not a surprise on the bill.

soberm (dismissed):

packages/bb-distributed-table/src/index.cdk.ts
osama-rizk: Good catch — agreed the danger is real for `removalPolicy` specifically. Added an `addWarningV2` at synth for an unrecognized `removalPolicy` or `encryption` string instead of treating it as `undefined`. Left `bb-kv-store` as-is since it's out of scope here. (d13f97b→pushed)

packages/bb-distributed-table/src/types.ts
osama-rizk: Done — `fromExisting` now emits an `addWarningV2` listing any durability/encryption options passed alongside it, so a `deletionProtection: true` on what looks like a fresh table is no longer a silent no-op. Followed the `hosting_construct.ts` pattern you referenced.

packages/bb-distributed-table/src/index.cdk.test.ts
osama-rizk: You're right, it was a straight duplicate. Rewrote it to pass `pointInTimeRecovery`/`deletionProtection: true` + `encryption: 'customer-managed'` and assert both `DynamoDB::Table` and `KMS::Key` are `0` — so it now proves the options are ignored rather than re-testing the regression below.

packages/bb-distributed-table/README.md
osama-rizk: Added a callout: `customer-managed` provisions a dedicated CMK per table (~$1/mo each + requests), no shared-key option today, prefer `aws-managed` unless a table needs its own rotation/key-policy control. Not adding bring-your-own-key here since you weren't asking for it.

packages/bb-distributed-table/src/index.cdk.ts
osama-rizk: I'd lean toward keeping the rationale inline here — the mixin gotcha is exactly what the next person editing this constructor needs in front of them, and a pointer to D-DT-9 is easy to skip past. Happy to collapse it to `// see DESIGN.md D-DT-9` if you'd prefer a single source of truth.

.changeset/bb-distributed-table-secure-defaults.md
osama-rizk: Agreed this needs a maintainer/bar-raiser call rather than just my sign-off. On mechanics we're aligned — all four are in-place updates, no table replacement.

The rollout question (prod-consumer count + comms beyond the changeset) I can't answer from this repo, so flagging it for a maintainer. The one sharp edge is that deletion protection turns on silently and the first thing a consumer notices is a `cdk destroy` refusal. Options to soften that: (a) ship behind an opt-in default for one release with a synth warning, then flip; or (b) keep the flip but pair it with a heads-up to known consumers. Deferring on which — happy to implement either.

packages/bb-distributed-table/src/types.ts
```diff
@@ -44,6 +44,52 @@ export interface DistributedTableOptions<
...
+	 *
+	 * - `'aws-managed'` (default): SSE with the AWS-managed `aws/dynamodb` KMS
+	 *   key. Auditable via CloudTrail with no per-key monthly charge.
+	 * - `'customer-managed'`: provisions a dedicated customer-managed KMS key
+	 *   (CMK) for this table, giving you full control over rotation and key
+	 *   policy. Incurs standard KMS key + request charges.
+	 *
+	 * DynamoDB is always encrypted at rest; this only selects the key type.
+	 */
+	encryption?: 'aws-managed' | 'customer-managed';
```
pranavosu: Could we have a fromExisting-style reference for customer managed keys? Since a key for each table seems expensive

packages/bb-distributed-table/src/types.ts
```diff
@@ -44,6 +44,52 @@ export interface DistributedTableOptions<
...
+	 * stack is deleted.
+	 *
+	 * - `'retain'`: keep the table (orphan it) — the safe default for
+	 *   production so stack teardown never destroys data.
+	 * - `'destroy'`: delete the table with the stack.
+	 *
+	 * Defaults to **`'retain'` on production deploys** and **`'destroy'` in
+	 * sandbox mode**. Set explicitly to override.
+	 */
+	removalPolicy?: 'retain' | 'destroy';
```
pranavosu: deletionProtection and removalPolicy are two knobs for one customer intent ("can this table be destroyed?"), and they can encode an potentially invalid state (true + DESTROY → wedged CloudFormation). Can we combine these into one concept?

packages/bb-distributed-table/README.md
```diff
@@ -40,7 +40,11 @@ const table = new DistributedTable(scope, id, options)
 | `key` | `TableKeyConfig<T>` | Yes | Primary key configuration: `{ partitionKey, sortKey? }`. Field names must exist in the schema. |
 | `indexes` | `Record<string, TableKeyConfig<T>>` | No | Global secondary index definitions. |
 | `ttl` | `keyof T & string` | No | Enable DynamoDB TTL on the specified attribute. The field should contain a Unix epoch timestamp in seconds. |
-| `table` | `ExternalTableRef` | No | Wrap an existing DynamoDB table instead of creating one. |
+| `pointInTimeRecovery` | `boolean` | No | Enable Point-in-Time Recovery (continuous backups, 35-day window). **Defaults to `true` in production, `false` in sandbox.** |
```
pranavosu: This has a cost too right? Like the encrytion option.

packages/bb-distributed-table/README.md
```diff
@@ -283,6 +287,37 @@ const legacy = new DistributedTable(scope, 'legacy', {
...
 - **Durability:** 99.999999999% (11 nines) across 3 AZs
 
+## Durability & Security defaults
+
+Production tables are **secure by default**. On a normal (`npm run deploy`) deploy, every table provisioned by this block ships with:
+
+- **Point-in-Time Recovery** enabled — restore to any second in the last 35 days.
+- **Deletion protection** enabled — a stray `cdk destroy` or console delete is refused until you turn protection off.
+- **SSE-KMS** with the AWS-managed `aws/dynamodb` key — encryption-at-rest that's auditable via CloudTrail, at no per-key charge.
+- **`RemovalPolicy.RETAIN`** — deleting the stack orphans (keeps) the table rather than destroying your data.
```
pranavosu: This also causes future deploys of the same app fails with table-already-exists, could be a rough edge to explain.

packages/bb-distributed-table/src/types.ts
```diff
@@ -44,6 +44,52 @@ export interface DistributedTableOptions<
 	 * ```
 	 */
 	ttl?: keyof T & string;
+	/**
+	 * Enable DynamoDB Point-in-Time Recovery (continuous backups).
+	 *
+	 * PITR lets you restore the table to any second within the retention
+	 * window (35 days), protecting against accidental writes/deletes and
```
pranavosu: How can the customer set this window?

packages/bb-distributed-table/src/index.cdk.test.ts
```diff
@@ -57,6 +65,163 @@ test('CDK: default DistributedTable provisions a DynamoDB table', () => {
 	template.resourceCountIs('AWS::DynamoDB::Table', 1);
 });
 
+// ── Secure-by-default durability & encryption (prod) ────────────────────────
+// Regression for the bug bash finding: prod DDB tables shipped with
```
pranavosu, line 63: Validation here is synth-only, which is thorough for the template shape but doesn't touch the risky path i.e what CloudFormation does when PITR, deletion protection, and an SSESpecification all appear at once on an already-deployed table, alongside this block's GSI custom resource issuing its own UpdateTable.

packages/bb-distributed-table/src/types.ts
osama-rizk: Good call — added `DistributedTable.fromKmsKey(arn)` for exactly this. It returns a branded `ExternalKmsKeyRef` you pass as a third value of the existing `encryption` option:

```ts
const key = DistributedTable.fromKmsKey("arn:aws:kms:...:key/abcd");
new DistributedTable(scope, "orders", { schema, key: { partitionKey: "id" }, encryption: key });
new DistributedTable(scope, "events", { schema, key: { partitionKey: "id" }, encryption: key });
```

So `encryption` is now `'aws-managed' | 'customer-managed' | ExternalKmsKeyRef` — several tables share one CMK (and one monthly charge) instead of each minting a dedicated key. I folded it into `encryption` rather than a separate `encryptionKey` field so there's no contradictory state to encode, and kept the CDK `IKey` type out of the public surface (branded ARN-string ref, same pattern as `fromExisting`). Added CDK tests asserting a shared ref provisions zero new `KMS::Key` resources. (5ab3cb0)

packages/bb-distributed-table/src/types.ts
osama-rizk: Agreed — merged them into a single `protection` enum so the contradictory state is unrepresentable:

```ts
protection?: 'disposable' | 'retained' | 'locked'
// disposable → DESTROY + no deletion protection   (sandbox default)
// retained   → RETAIN  + no deletion protection
// locked     → RETAIN  + deletion protection        (prod default)
```

`deletionProtection: true` + `removalPolicy: 'destroy'` (the DELETE_FAILED wedge you flagged) can no longer be expressed. Kept the middle `'retained'` state so you can orphan-on-delete without locking. This replaces the two separate options entirely. (e8bec55)

packages/bb-distributed-table/README.md
osama-rizk: Yes — added a cost callout. PITR bills for continuous-backup storage per GB-month, so it is not free on a large table. Documented in the README (option row + a note in the durability section), and added `pointInTimeRecoveryDays` (1–35) so you can narrow the recovery window to trim that cost. (e8bec55)

packages/bb-distributed-table/README.md
osama-rizk: Good call — documented it. Because the prod default is now RETAIN (`protection: 'locked'/'retained'`) and the table name is derived deterministically from the block id, deleting the stack orphans the table and redeploying the same app then fails with `Table already exists` until you delete/import the orphan. Added a "gotcha" note to the README durability section and to the changeset behavior-change callout, with the escape (`protection: 'disposable'` for tables you expect to recreate). (e8bec55)

packages/bb-distributed-table/src/types.ts
osama-rizk: It was fixed at 35 days before — now configurable. Added `pointInTimeRecoveryDays?: number` (1–35, default 35), mapping to `recoveryPeriodInDays` on the PITR spec (supported in the pinned CDK 2.257). Out-of-range values warn at synth and fall back to the default rather than failing the deploy; ignored when PITR is off. Updated the JSDoc that previously implied a fixed window. (e8bec55)

packages/bb-distributed-table/src/index.cdk.test.ts
osama-rizk: Right — added real-deploy coverage. `test-apps/comprehensive` now provisions a `secure-items` table carrying PITR (+ a narrowed 7-day window) alongside a GSI, so the GSI custom resource's UpdateTable and the durability config provision together on a live table — exactly the interaction synth can't exercise. It runs in the E2E Sandbox + E2E Production CI jobs.

The strongest signal is structural: if that CloudFormation interaction collided, the deploy step fails before any test runs — so a green deploy disproves it. The e2e then asserts the deployed table is functional (read/write + a GSI query). I kept it API-only rather than asserting via a DescribeTable SDK call — an earlier revision did that and crashed the CI runner (a transitive `@aws-sdk/client-dynamodb` version skew threw at import time); the exact on-table values stay asserted at synth level. (d7b6c20)

packages/bb-distributed-table/src/index.cdk.ts
```diff
@@ -71,6 +99,87 @@ export class DistributedTable<T = any> extends Scope {
...
+			&& !isKmsKeyRef
+		) {
+			Annotations.of(this).addWarningV2(
+				'@aws-blocks/bb-distributed-table:UnknownEncryption',
+				`Unrecognized encryption '${String(config.encryption)}' (expected 'aws-managed', ` +
+					`'customer-managed', or DistributedTable.fromKmsKey(arn)) — falling back to 'aws-managed'.`,
+			);
+		}
+
+		const pitrEnabled = config.pointInTimeRecovery ?? !isSandbox;
```
soberm: [minor] This check is string-only (`tryGetContext('sandboxMode') === 'true'`, defined at line 82), but core's own detectors accept a string OR a boolean, `stack-metadata.ts:22-23` and `blocks-backend.ts:99-100` both do `=== 'true' || === true`. If something ever sets `sandboxMode` as a boolean `true` in context, this construct would classify the stack as production and turn on PITR + deletion protection + RETAIN while the stack tag says sandbox.

Not new to this PR though, `bb-data`, `bb-file-bucket`, `bb-distributed-data`, and `bb-knowledge-base` all use the same string-only check, so core's two files are actually the outliers here. Wouldn't block this PR on it, but might be worth a follow-up ticket to pull a shared `isSandboxContext()` helper into `@aws-blocks/core/cdk` so the sites can't drift.

packages/bb-distributed-table/DESIGN.md
```diff
@@ -100,7 +100,21 @@ A generic `ValidationException` is exactly the kind of catch-all bucket worth av
 
 **Mock/AWS parity:** the mock checks serialized byte length client-side and throws `ItemTooLarge` directly. On AWS, DynamoDB raises a generic `ValidationException` for an oversized item; the runtime narrows on the size-specific message (`size has exceeded`) and re-maps only that case to `ItemTooLarge`. Other `ValidationException` causes (malformed expressions, type mismatches) propagate unchanged. Both layers therefore surface the same `error.name`, and the shared message lives in `errors.ts` (`DistributedTableMessages.itemTooLarge`) so the two stay byte-for-byte aligned.
 
-### D-DT-9: `readValidation` — `off | coerce | strict`, defaulting to `coerce`
+### D-DT-9: Secure-by-default durability & encryption, gated on sandbox mode
```
soberm: [nit] Renumbering the durability decision into the D-DT-9 slot and bumping readValidation to D-DT-10 breaks `docs/DECISIONS.md`. D-015 references `packages/bb-distributed-table/DESIGN.md` D-DT-9 twice (in its Context and again under References) as the pointer to the readValidation mechanics, both of those now land on this durability section instead.

Simplest fix without touching DECISIONS.md: give this decision the next free id (D-DT-11) instead of taking over D-DT-9. Otherwise, flag it for a maintainer to update D-015's two references to D-DT-10.

packages/bb-distributed-table/src/index.cdk.ts
```diff
@@ -41,6 +55,20 @@ export class DistributedTable<T = any> extends Scope {
 			// and grant the runtime Lambda read/write + index query access.
 			// We deliberately skip the GSI custom resource — the customer owns the
 			// table's index lifecycle when they bring their own.
+			//
+			// Durability/encryption options don't apply to an existing table (we
+			// never emit a `Table` resource to attach them to). Surface that at
+			// synth so a `protection: 'locked'` on what looks like a fresh
+			// table isn't a silent no-op.
+			const ignoredForExisting = (['pointInTimeRecovery', 'protection', 'encryption'] as const)
```
soberm: [minor] This list is missing `pointInTimeRecoveryDays`. If someone passes `table: fromExisting(...), pointInTimeRecoveryDays: 14` without also setting `pointInTimeRecovery`, it's silently ignored with no warning, exactly the class of bug this block exists to prevent. One-line fix, just add it to the tuple.

packages/bb-distributed-table/src/index.cdk.test.ts
```diff
@@ -57,6 +65,276 @@ test('CDK: default DistributedTable provisions a DynamoDB table', () => {
...
+	const template = Template.fromStack(stack);
+	template.hasResourceProperties('AWS::DynamoDB::Table', {
+		PointInTimeRecoverySpecification: {
+			PointInTimeRecoveryEnabled: true,
+			RecoveryPeriodInDays: 7,
+		},
+	});
+});
+
+test('CDK: an out-of-range pointInTimeRecoveryDays falls back to the default (no window pinned)', () => {
```
soberm: [minor] None of the four new `addWarningV2` calls (`UnknownProtection`, `UnknownEncryption`, `InvalidPitrDays`, `IgnoredOptionsForExistingTable`) actually get asserted anywhere. This test checks the fallback value but not that a warning was emitted, and there's no test at all for an invalid `protection`/`encryption` string. Worth adding an `Annotations.fromStack(stack).hasWarning(...)` check per warning code so a future refactor can't silently drop one.

soberm:
[nit] The "How each review comment was addressed" table says the real-deploy e2e asserts PITR window, SSE-KMS, and an active GSI via `DescribeContinuousBackups`/`DescribeTable`. That's not what's shipped, `distributed-table-secure-defaults.test.ts` says right in its header comment that it's deliberately API-only with no direct AWS SDK client (an earlier revision that did call `DescribeTable` crashed the CI runner on a transitive `@aws-sdk/client-dynamodb` version skew, per the thread 12 reply). Looks like the description just didn't get updated after the follow-up fix commit. Worth a quick edit so it doesn't mislead anyone reading back on this later.

packages/bb-distributed-table/src/index.cdk.ts
osama-rizk: Good spot, and this is now largely moot after the rebase onto #302: durability no longer reads the `sandboxMode` context at all — removal policy, deletion protection, and PITR all come from `this.defaults` (the stack's `BlocksDefaults` preset). So a boolean-`true` `sandboxMode` can no longer flip a table to the production posture; the tag and the durability posture both derive from what the app passes as `defaults`.

The only remaining `sandboxMode` read (line 82) feeds the GSI custom resource's drop-and-recreate fast path — an operational concern, not durability. Agree that a shared `isSandboxContext()` in `@aws-blocks/core/cdk` is worth a follow-up so the string-only sites (bb-data, bb-file-bucket, bb-distributed-data, bb-knowledge-base, and this GSI check) can't drift from core's `=== 'true' || === true` detectors — filing that separately as you suggested, not blocking here. (f6f9d727)

packages/bb-distributed-table/DESIGN.md
osama-rizk: Fixed — thanks for catching the broken `docs/DECISIONS.md` D-015 references. Restored `readValidation` to **D-DT-9** (so D-015's two 'DESIGN.md D-DT-9' pointers resolve correctly again) and gave the durability decision the next free id, **D-DT-10**, reordering the two sections so they stay in numeric order. Also fixed the internal '(see D-DT-9)' pointer in the Infrastructure section → D-DT-10. No change needed in DECISIONS.md. (f6f9d727)

packages/bb-distributed-table/src/index.cdk.ts
osama-rizk: Fixed — added `pointInTimeRecoveryDays` to the `ignoredForExisting` tuple, so `fromExisting(...) + pointInTimeRecoveryDays: 14` (without `pointInTimeRecovery`) now warns at synth instead of being silently dropped. (f6f9d727)

packages/bb-distributed-table/src/index.cdk.test.ts
osama-rizk: Added — four new tests assert each warning actually fires via `Annotations.fromStack(stack).hasWarning('*', Match.stringLikeRegexp(...))`: an invalid `protection` string (`UnknownProtection`), an invalid `encryption` string (`UnknownEncryption`), an out-of-range `pointInTimeRecoveryDays` (`InvalidPitrDays`), and durability options passed alongside `fromExisting()` (`IgnoredOptionsForExistingTable`). So a future refactor that drops a warning now fails the suite. (f6f9d727)

soberm (dismissed):

packages/bb-distributed-table/src/types.ts
```diff
@@ -57,6 +57,85 @@ export interface DistributedTableOptions<
...
 	ttl?: keyof T & string;
+	/**
+	 * Enable DynamoDB Point-in-Time Recovery (continuous backups).
+	 *
+	 * PITR lets you restore the table to any second within the retention
+	 * window (see {@link pointInTimeRecoveryDays}), protecting against
+	 * accidental writes/deletes and logical corruption.
+	 *
+	 * Defaults to **`true` on production deploys** and **`false` in sandbox
+	 * mode** (`--context sandboxMode=true`) to keep throwaway sandboxes cheap.
```
pranavosu:
```suggestion
	  defaults?
```
is this JSDoc still accurate?

packages/bb-distributed-table/src/types.ts
```diff
@@ -57,6 +57,85 @@ export interface DistributedTableOptions<
...
+	 * The recovery window, in days, that Point-in-Time Recovery keeps
+	 * continuous backups for — you can restore to any second within this many
+	 * preceding days.
+	 *
+	 * Accepts **1–35**; defaults to **35** (the maximum) when omitted. Only
+	 * meaningful when PITR is enabled; ignored when `pointInTimeRecovery` is
+	 * `false`. A shorter window reduces backup-storage cost at the expense of
+	 * how far back you can restore.
+	 */
+	pointInTimeRecoveryDays?: number;
```
pranavosu:
```suggestion
 pointInTimeRecovery?: boolean | { retentionDays: number }
```

Let's combine these into one field, because they're coupled together either way.

Same for the `pointInTimeRecovery` field in [stack defaults ](https://github.com/aws-devtools-labs/aws-blocks/pull/282/changes#diff-06c5b30d717f1b39ff90118b7101ba71a4de766ac5c6fcf56693db0a20da64c1R36)

pranavosu (changes requested):

packages/bb-distributed-table/src/types.ts
osama-rizk: You're right, it was stale. After the rebase onto #302 the default no longer comes from a `--context sandboxMode=true` read — it comes from the stack-wide `defaults` (`BlocksPresets`). Rewrote the JSDoc to say so (and fixed the same stale sandbox-mode wording on `protection`, which had the identical issue). (66b724dc)

packages/bb-distributed-table/src/types.ts
osama-rizk: Done — combined into one field:

```ts
pointInTimeRecovery?: boolean | { retentionDays: number }
```

`true` = on with the default 35-day window, `false` = off, `{ retentionDays: n }` = on with an n-day window (1–35). Since the window only means anything when PITR is on, 'days set but PITR off' is now unrepresentable — same reasoning as the `protection` merge. Dropped the separate `pointInTimeRecoveryDays`.

Applied the same shape to `BlocksDefaults.pointInTimeRecovery` in core (`boolean | { retentionDays: number }`); the presets stay plain booleans. One thing worth a look during API-BR: `{ retentionDays }` is a DynamoDB-flavored detail living in the shared, service-agnostic `BlocksDefaults` — I think it's fine since the field is already named `pointInTimeRecovery` and other services ignore it, but flagging it (noted in the core changeset too). (66b724dc)

pranavosu (approved):

soberm (approved):

PR #283, opened 2026-07-30
feat(bb-distributed-table): readValidation modes (off|coerce|strict), default coerce (fixes #1007)

osama-rizk:
## Summary

Fixes [aws-amplify/private-amplify-backend-staging#1007](https://github.com/aws-amplify/private-amplify-backend-staging/issues/1007) — **"DistributedTable.get() skips schema validation on legacy rows"**.

Writes validate against `schema`, but reads (`get`, `getBatch`, `query`, `scan`) returned the **raw stored value** without reconciling it. After a schema change, a row written under the old schema no longer conforms to `T`:
- a **newly added field is absent** from the read — the value silently violates `T`, and a `.default()` is neither applied on read nor persisted on write-back; and
- if the new field is **required with no default**, the `put()` half of the documented read-modify-write cycle throws `ValidationFailed`, stranding the row.

## What changed

Reads reconcile a stored item with the schema via **`readValidation?: 'off' | 'coerce' | 'strict'`**, defaulting to **`'coerce'`**:

| Mode | On read | Non-conforming item |
|---|---|---|
| **`'coerce'`** (default) | applies the schema (defaults filled, types narrowed) and **preserves keys the schema doesn't declare** | returns the **raw** value + warns — **never throws** |
| **`'strict'`** | validates | **throws** `ValidationFailed` |
| **`'off'`** | raw stored value, no validation | as-is |

`'coerce'` is **lossless**: rather than returning the bare validator output (which strips unknown keys — Zod `.strip()` default), it **deep-merges the coerced value over the raw stored item**. Schema output wins per key, while attributes the current schema doesn't declare — fields from an older schema version, columns another writer owns, nested keys — are preserved. Arrays are replaced wholesale (the coerced array wins). So a legacy row both gains the new schema defaults *and* keeps its extra fields, and a read-modify-write never silently deletes data.

A shared `applyReadValidation()` helper in `errors.ts` keeps the mock and AWS runtime identical (`null` passes through in all modes). `ReadValidationMode` is exported from every entry point (and re-exported from `@aws-blocks/blocks`).

Adds one runtime dependency: **`defu`** (`createDefu`) — maintained, prototype-safe, zero-dependency, ESM. Used for the merge because treating arrays as opaque leaves requires custom config in every merge library regardless (defu concatenates arrays by default, which would duplicate elements since the coerced array is derived from the raw one — overridden via one customizer).

> ### Note: default read behavior vs the current preview
> Against `main`, reads return the raw stored value; this PR makes them return the schema-reconciled value by default. Pass `readValidation: 'off'` to keep raw reads. This is a new option plus a read-path default — not a break of any released API (a fresh addition to a preview package). The default is `'coerce'` so the fix applies without requiring an opt-in flag.

### Why `coerce` is the default (not opt-in, not `strict`)
The bug is that the **default** read violates `T` — a fix only users who find a flag can enable isn't a fix. Backed by how comparable typed data layers behave:

- **Schema-on-read peers:** Mongoose fills defaults/casts on hydrate; ElectroDB returns schema-shaped items. Both coerce, neither throws. DynamoDB-Toolbox `format()` and `zod-firebase` throw, but both ship an escape hatch — none hard-throw with no opt-out.
- **Rails** type-casts on load; **Postgres** materializes `ADD COLUMN … DEFAULT` at read time — coerce-on-read in all but name.
- **SQL ORMs (Prisma/Drizzle) pass raw through** — but only because the engine enforces schema *on write*, a guarantee DynamoDB lacks, so their behavior isn't transferable.

`coerce` is non-throwing (aligns with the reads-return-data-or-`null` contract), `strict` is an opt-in reject, `off` is the raw/perf/migration escape hatch.

### Caveats (documented in DESIGN D-DT-9)
- **Type coercion is validator-dependent.** Standard Schema only guarantees `validate()` *checks*, not *transforms*. Zod fills defaults/coerces; a check-only Valibot/ArkType schema returns its input unchanged, so `'coerce'` won't fill defaults for those. It never fabricates a required value.
- **Transforms that remove a key.** A schema whose transform intentionally renames/removes a key would have the old key resurrected by the merge — such transform-heavy schemas should use `'strict'` or `'off'`.

## Validation

**Tests.** `index.test.ts` + `parity.test.ts` cover all three modes and the preserve semantics: default-`coerce` fills defaults **and** preserves unknown keys (top-level, nested, and across a read-modify-write round-trip), arrays are replaced not duplicated, `off` returns raw, `strict` throws, uncoercible rows return raw without throwing, `null` passthrough in every mode, mock↔AWS parity for each.

```
bb-distributed-table: # tests 117 | # pass 117 | # fail 0
@aws-blocks/blocks conditional-export parity: 41/41
full workspace: 3436/3436
```

**Standalone repro app** (`test-apps/ddb-get-validation-repro/`) — runs against the built mock, no AWS account. Demonstrates each mode, including the read-modify-write that preserves an unknown stored key.

**No regressions:** `npm run build` ✅ · `npm run lint` ✅ · `npm run lint:deps` ✅ (`defu` correctly declared) · API report regenerated (public surface unchanged; no `ae-forgotten-export`).

**Not run — and why:** runtime data-path change with full mock↔AWS parity coverage (AWS path exercised via a fake doc client). No serialization/wire-format change, so no sandbox e2e required.

## Checklist
- [x] build / lint / lint:deps / unit / parity green locally
- [x] Changeset added (covers `bb-distributed-table` + `blocks`)
- [x] README + DESIGN + JSDoc updated; decision recorded in `docs/DECISIONS.md` (D-015)
- [x] Conditional-export parity holds; `ReadValidationMode` exported from all entry points
- [x] New behavior ships tests (mock + parity, all three modes + preserve)
- [x] Default read-path behavior change (raw → reconciled) called out for maintainer review

---

## API surface: `readValidation`

```ts
type ReadValidationMode = 'off' | 'coerce' | 'strict';

interface DistributedTableOptions<T, K, Indexes> {
  // ...existing: schema, key, indexes, ttl, table, logger
  readValidation?: ReadValidationMode;   // default: 'coerce'
}
```

**Why this shape**
- **Three-mode enum, not a boolean** — there are three distinct read behaviors (raw / coerce / throw); a boolean can't carry all three, and both a strict-reject mode and a raw escape hatch are needed.
- **`coerce` default** fixes the bug for everyone (not just flag-setters), stays non-throwing (one bad row can't fail a whole `scan`/`getBatch`; migration reads stay possible), preserves data (deep-merge over raw), and matches schema-on-read precedent.
- **`strict`** for tables where a mismatch should be rejected as corruption.
- **`off`** for hot paths, trusted write-validated data, and reads during a migration.

**Use cases**
1. **Additive schema evolution** — add `currency: z.string().default('USD')`; legacy rows read back with `currency: 'USD'`, keep any older fields, and round-trip. (default `coerce`)
2. **Read-modify-write across schema versions** — `get → mutate → put` works on legacy rows without throwing and without dropping unknown fields.
3. **Bulk reads feeding a UI/API** — `query`/`scan` items arrive in the current shape, no `?? default` guards downstream.
4. **Integrity-sensitive tables** — `readValidation: 'strict'` rejects any drifted/tampered row.
5. **Hot path / migration** — `readValidation: 'off'` skips validation cost and reads rows verbatim.


packages/bb-distributed-table/src/errors.ts
```diff
@@ -143,3 +147,42 @@ export function remapItemTooLarge(err: unknown): unknown {
...
+	const result = schema['~standard'].validate(item);
+	const resolved = result instanceof Promise ? await result : result;
+	if (resolved.issues) {
+		if (mode === 'strict') {
+			throw blocksError(DistributedTableErrors.ValidationFailed, resolved.issues[0]?.message ?? 'stored item failed schema validation on read');
+		}
+		log.warn(
+			`readValidation: stored item failed schema validation, returning the raw value. ${resolved.issues[0]?.message ?? ''}`.trim(),
+			context,
+		);
```
soberm, line 212: [major] The raw-fallback path logs via `log.warn(...)`, but both runtimes construct the default logger at `level: 'error'` (index.aws.ts and index.mock.ts constructors), and `shouldLog` in bb-logger's serializer only emits when `LEVEL_PRIORITY[messageLevel] >= LEVEL_PRIORITY[configuredLevel]` — warn (2) is below error (3), so it's suppressed by default. The README/DESIGN promise that a T-violating row is "returned as-is with a warning" doesn't actually hold unless the caller supplies their own logger at warn or below. Either bump the DT default logger to `warn`, or document that the warning only surfaces with a non-default logger.

packages/bb-distributed-table/src/errors.ts
```diff
@@ -143,3 +147,42 @@ export function remapItemTooLarge(err: unknown): unknown {
...
+		return item;
+	}
+	return resolved.value as T;
```
soberm, line 222: [minor] This returns `resolved.value`, the validator's output, not the raw item. Zod object schemas strip unrecognized keys by default, so a stored row with fields beyond the current schema loses those fields on this coerce path. Combined with the read-modify-write pattern this PR is fixing, a subsequent `put()` persists the stripped item, permanently dropping data the previous raw-read behavior preserved. Worth a caveat in README's "Reads and schema evolution" and DESIGN D-DT-9 that `coerce` can drop unknown keys (validator-dependent), and pointing migration/preservation cases at `readValidation: 'off'`.

soberm:
**packages/blocks/src/index.ts:159** (and index.cdk.ts:38)

[nit] `ReadValidationMode` isn't in this re-export list. A `@aws-blocks/blocks` consumer can pass `readValidation: 'coerce'` via literal inference but can't name the type for their own function signatures/wrappers. Worth adding it alongside the other DT type re-exports here.

packages/bb-distributed-table/src/errors.ts
osama-rizk: Good catch — confirmed. Both runtimes built the default logger at `level: 'error'`, and `shouldLog` gates `warn` (2) < `error` (3), so the raw-fallback warning was silently dropped unless the caller supplied their own warn-level logger. The doc promise was hollow by default.

Fixed in 54e291b: the DT default logger is now `level: 'warn'` in both `index.aws.ts` and `index.mock.ts`. `log.warn` in `applyReadValidation` is the only log the block emits, so this surfaces exactly that message and adds no other noise; callers passing their own logger are unaffected.

packages/bb-distributed-table/src/errors.ts
osama-rizk: Agreed, and verified empirically: a stored row `{id, legacyNote}` read under a schema of just `{id}` comes back `{id}` under `'coerce'` (Zod `.strip()`) but `{id, legacyNote}` under `'off'` — so a subsequent `put()` would persist the stripped row and permanently drop `legacyNote`. Real data-loss vector for RMW.

Fixed in 54e291b:
- README "Reads and schema evolution" now has a ⚠️ callout that `'coerce'` can drop stored keys not in the schema, that RMW then persists the loss, and points preservation/migration cases at `readValidation: 'off'` (or a `.passthrough()` schema).
- DESIGN D-DT-9 gets an "unknown-key stripping caveat" framing it as the one case where the safer default can *lose* info, and why it's an accepted trade-off (adding fields is the common path; removal/staged migrations should use `'off'`).
- Added a test pinning `'coerce'` strips / `'off'` preserves.

osama-rizk:
Thanks for the review @soberm — all three addressed in 54e291b:

- **[major] suppressed warning** — DT default logger bumped `error` → `warn` in both runtimes so the `'coerce'` raw-fallback warning actually emits (it's the only log the block produces, so no added noise).
- **[minor] coerce drops unknown keys** — documented the data-loss-on-RMW risk in README + DESIGN D-DT-9 with mitigations (`readValidation: 'off'` / `.passthrough()`), plus a regression test.
- **[nit] `ReadValidationMode` re-export** — added to `@aws-blocks/blocks` `index.ts` and `index.cdk.ts`; it now appears in `blocks` API.md so umbrella consumers can name the type.

Build / lint / lint:deps green, bb-distributed-table 111/111, blocks parity 41/41, full workspace 3430/3430, API reports regenerated.

soberm (dismissed):

packages/bb-distributed-table/src/index.mock.ts
```diff
@@ -125,23 +127,38 @@ export class DistributedTable<
...
+		// Default level is 'warn' (not 'error') so the readValidation='coerce'
+		// raw-fallback warning actually surfaces — it's the only log the block
+		// emits, so this doesn't add noise. Callers can pass their own logger.
+		this.log = options?.logger ?? new Logger(this, 'logger', { level: 'warn' });
 		this.filePath = join(getMockDataDir(this), 'data.json');
 		this.data = this.loadFromDisk();
 		this.schema = options.schema;
 		this.keyConfig = options.key;
 		this.indexes = (options.indexes ?? {}) as Indexes;
+		this.readValidation = options.readValidation ?? 'coerce';
```
sarayev, line 146: Quick question on the default here (non-blocking — the PR already looks solid): with `readValidation: 'coerce'`, a read runs the stored row through the schema and returns the validator's output, which for Zod's default `.strip()` drops any keys that aren't in the current schema. That means a routine read-modify-write could quietly discard fields that were safely stored before, and because it's the default, people would hit it without opting in. Was `'coerce'` chosen as the default deliberately over the lossless `'off'`, or is it worth making `'off'` the default and treating coercion as an explicit opt-in? Also curious whether the stripping affects all Standard Schema validators or just transform-based ones like Zod — that changes how big the blast radius is.

osama-rizk: Deliberate choice. #1007's bug is that the *default* read returns values violating `T`, so an opt-in fix wouldn't fix the reported default — and coerce-on-read is the norm in comparable layers (Mongoose, Rails, ElectroDB, Postgres `ADD COLUMN DEFAULT`), none default to raw once a schema is declared.

You're right it trades that for a lossy failure in the removal/unknown-key direction. I weighed additive evolution (coerce is strictly safer) as far more common than removal, and the lossy path now has a caveat in README/DESIGN, an `off` / `.passthrough()` escape hatch, and a regression test. If we rank silent data-loss above silent `T`-violation, I'm happy to flip to `off`-default — but I'd rather ship documented, opt-out-able stripping than the original bug.

Blast radius: validator-dependent. Transform/strip validators (Zod by default) drop unknown keys → lossy. Check-only validators (some Valibot/ArkType) return input unchanged → not lossy, but also don't fill defaults, so `coerce` ≈ `off` there. So stripping is scoped to transform-based validators — the same ones where `coerce` adds the most value.

sarayev: Gotcha, would it make sense to add this decision to DECISIONS.MD file, then?

osama-rizk: Good call — done in 06458c1. Added **D-015** to `docs/DECISIONS.md` (the repo-wide ADR log) with the full context/decision/rationale/alternatives/trade-off, and cross-linked it from `bb-distributed-table` DESIGN.md D-DT-9 (which keeps the per-BB mechanics). Follows the same pattern as D-007, where the auth BBs record a cross-cutting default there and link from their DESIGN.md.

sarayev: Thanks for adding the decision, but... 

Coming back to this one, because the more I sit with it the more the concern is specifically **silent, unintended data deletion** — and I think that deserves a bit more weight than "documented trade-off."

The scenario that worries me isn't the developer who removes a field on purpose. It's the one who never touches it at all:

```ts
const order = await table.get('A1');  // stored: {orderId, total, couponCode}
order.total = 50;                     // coerce already stripped couponCode on read
await table.put(order);               // couponCode is now permanently gone from storage
```

They read a row, changed one unrelated field, wrote it back — and a field they didn't even know existed (a legacy column, or one another writer owns) is permanently deleted. No error, no return-value signal, just a `warn` log they'll likely never see. And because `'coerce'` is the default, this is the *out-of-the-box* behavior for a storage primitive.

What makes me hesitate on "documented + opt-out is enough": the person who loses data here is, almost by definition, the one who *didn't* read the caveat — they don't know the extra field is there, so they have no reason to reach for `'off'`. Docs protect the informed user; this footgun mostly hits the uninformed one. And unlike the bug this fixes (a silent `T`-violation, where the data is still there and recoverable), a stripped-then-written-back field is gone for good.

Fully agree the #1007 default needed fixing and that additive evolution is the common case. So rather than re-litigate coerce-vs-off, one option worth considering: **can the default be coerce-but-non-lossy?** i.e. apply the schema's defaults/coercions to satisfy `T`, but *preserve* unknown keys rather than stripping them (merge the coerced result over the raw stored object, or use passthrough semantics by default). That closes #1007 — reads conform to `T` — without silently deleting fields the caller never mentioned. Stripping could then be an explicit opt-in for people who genuinely want it.

Not trying to block — soberm's already approved and the docs/ADR work is great. But "default behavior can permanently delete data with no intent and no hard signal" feels like the kind of thing worth one more look before it ships as the default.

osama-rizk: You're right — documented-opt-out wasn't enough for *silent, unintended, unrecoverable* deletion, and your coerce-but-non-lossy suggestion is the correct resolution. Implemented in 667745b8.

`'coerce'` no longer returns the bare validator output (which strips via Zod's `.strip()` default). It now **deep-merges the coerced output over the raw stored item**: schema output wins per key (defaults filled, types narrowed), while keys the schema doesn't declare — legacy fields, columns another writer owns, nested ones too — are **preserved**. So your exact scenario:

```ts
const order = await table.get('A1');  // {orderId, total, couponCode}  ← couponCode KEPT now
order.total = 50;
await table.put(order);               // couponCode still there, persisted
```

now round-trips losslessly — verified by a test mirroring your snippet (read → mutate unrelated field → put → couponCode survives), plus nested-key and array cases, in both mock and AWS runtime.

Implementation notes:
- Uses **defu** (`createDefu`) — maintained, prototype-safe, zero-dep, ESM. Arrays are the one non-default: the coerced array replaces the raw array wholesale (defu concatenates by default, which would *duplicate* elements since the coerced array is derived from the raw one).
- This closes #1007 (reads conform to `T`) **and** preserves data — they're no longer in tension, so no coerce-vs-off trade-off to accept.
- Residual (documented): a schema whose *transform* deliberately removes a key would have it resurrected by the merge; transform-heavy schemas should use `'strict'`/`'off'`.
- Docs updated (README, DESIGN D-DT-9, DECISIONS D-015) to describe preserve semantics.

Thanks for pushing on this past the first two rounds — the result is strictly better than what I'd have shipped.

soberm (approved):

sarayev:
Re-reviewed — this is exactly the resolution I was hoping for. Deep-merging the coerced output over the raw stored item closes #1007 *and* keeps a read-modify-write from silently deleting fields the caller never touched, so `'coerce'` as the default is now completely defensible. The `defu` choice with arrays-as-leaves (avoiding the concat/duplication trap) and the prototype guard are the right call, and the "reviewer's scenario" RMW test plus the mock↔AWS parity tests lock the behavior in.

Nice catch documenting the one residual — that a transform which intentionally removes a key gets it resurrected by the merge, so transform-heavy schemas should reach for `'strict'`/`'off'`. Agreed that's the right trade and it's clearly called out in DESIGN/DECISIONS.

Thanks for iterating on this past a few rounds — the result is strictly better. LGTM. 👍

sarayev (approved):

PR #285, opened 2026-07-30
fix(test-apps): key OIDC sign-in records by userId instead of module state

### LINKED ISSUE #284: AuthOIDC onSignIn hook stores user in module-level state, invisible across Lambda instances (test-apps/comprehensive)
Split out of the #278 review. Same class of bug as #172, different flow. No test failures have been attributed to it yet, so this is prevention rather than a live break.

### What

In `test-apps/comprehensive/aws-blocks/index.ts` two `AuthOIDC` instances record the signed-in user into a module-level variable from their `onSignIn` hook:

```ts
let lastOidcSignInUser: { userId: string; email: string | null; provider: string } | null = null;
const oidcAuth = new AuthOIDC(scope, 'oidc-auth', {
  providers: oidcProviders,
  onSignIn: async (user) => {
    lastOidcSignInUser = { userId: user.userId, email: user.email, provider: user.provider };
  },
});
```

and the same shape for `lastExtrasSignInUser` on `oidcAuthExtras`. They're read back over separate RPCs, `oidcGetLastSignInUser` and `oidcExtrasGetLastSignInUser`.

### Why it's a problem

The hook fires while serving the OIDC callback; the read is a later, separate request. In a deployed environment those can land on different Lambda instances, so the reader sees its own `null` (assert fails) or a leftover user from an earlier sign-in it happened to handle (wrong-provider/wrong-user assert, or a false pass). This is exactly the pair of failure modes #172 turned out to have. Consumers:

- `test/oidc-auth.test.ts:239`: `oidcGetLastSignInUser()`, asserts `provider === 'google'`
- `test/oidc-auth.test.ts:418`: `oidcExtrasGetLastSignInUser()`, asserts `provider === 'google-extras'`

Note the neighbouring assertion in the same test already does this correctly: `oidcExtrasGetProfile(userId)` reads from the `oidcProfiles` KVStore that `onSignIn` upserts into, and is comment-labelled "the canonical post-sign-in pattern". So half of that hook is already durable and the `lastUser` half is not.

### Suggested fix

Mirror what #278 landed for verification codes:

1. Persist the sign-in record to a KVStore keyed by `userId` instead of a module variable. `oidcAuthExtras` already has `oidcProfiles` and could reuse it.
2. Make `userId` a **required** parameter on `oidcGetLastSignInUser` / `oidcExtrasGetLastSignInUser`, so "give me whoever signed in last" isn't expressible. #278 made this a compile-time guarantee for the code readers and the same trick applies.
3. Have the tests poll for their own `userId` via the shared `test/poll-for-code.ts` helper (or a sibling of it) rather than reading once.

### Notes

- Deliberately left out of #278 to keep that diff scoped to the verification-code path.
- Unlike the code flow, the OIDC tests derive `userId` from the callback response, so the reads can be keyed without new plumbing.
- `test-apps/hosting-ssr`, `hosting-spa` and `hosting-ssr-nuxt` still hold `lastDeliveredCode` in module state too. Lower risk since they're single-user local demos, but worth a look while in here.

ahmedhamouda78:
## What

`AuthOIDC`'s `onSignIn` hooks in the comprehensive test app recorded the signed-in user into module-level variables:

```ts
let lastOidcSignInUser: { userId: string; email: string | null; provider: string } | null = null;
```

The hook runs while the OIDC callback is being served. The e2e reads the result back over a later, separate request. In a deployed environment those two can land on different Lambda instances, so the reader either sees its own `null` or the record of whoever signed in last on whichever instance answered.

The second case is the one that bothers me more than a plain failure. The old assertions only checked `provider`, so a leftover record from the same provider would pass while proving nothing about the sign-in the test had just performed.

Same class as the verification-code flake in #278, different flow. No red builds have been pinned on this one yet, so this is prevention.

## How

Sign-in records go to the existing `oidc-profiles` KVStore under `signin:<instance>:<userId>`, and `userId` is now **required** on `oidcGetLastSignInUser` and `oidcExtrasGetLastSignInUser`. "Give me whoever signed in last" no longer type checks:

```
test/__noarg-probe.ts(3,23): error TS2554: Expected 1 arguments, but got 0.
test/__noarg-probe.ts(4,23): error TS2554: Expected 1 arguments, but got 0.
```

That is a live gate, not just a convention: `test:e2e:local` runs `tsc --noEmit` in its before-hook.

I reused `oidc-profiles` instead of adding a store. It already holds the per-user profile the second instance upserts, which the code comments call the canonical post-sign-in pattern, and no code scans it, only keyed get/put. The keys are also bounded here, unlike the verification codes: the stub IdP returns a fixed user per provider, so each sign-in overwrites its own record rather than piling up. That is why there is no purge method in this one.

The two consumers now resolve their own `userId` from the session via `requireAuth` and poll for that record rather than reading once. The polling lives in `test/poll-for-signin.ts`, a sibling of the `poll-for-code.ts` that #278 adds. Worth folding the two into one generic poller once both have landed; I kept them separate here so this PR does not depend on that one.

## Tests

Two new tests, both with a control run showing they bite. Restoring module-level semantics (a per-instance `Record` holding the last sign-in, which is exactly what the old code did) makes both fail:

```
not ok 1 - sign-in record is keyed per user, not a shared "last sign-in" slot
    unknown userId should read as null, not the most recent sign-in
not ok 2 - each AuthOIDC instance keeps its own sign-in records
    Expected values to be strictly equal:
    - null
# tests 2 / # pass 0 / # fail 2
```

With the fix both pass, along with the rest of the OIDC suite:

```
ok 1 - onSignIn fires on successful sign-in
ok 2 - sign-in record is keyed per user, not a shared "last sign-in" slot
ok 3 - each AuthOIDC instance keeps its own sign-in records
ok 1 - sign-in flow works and onSignIn upserts profile
ok 2 - second sign-in updates profile, does not duplicate
ok 14 - AuthOIDC
```

Also added sign-out cleanup to the `onSignIn fires` test, which previously left its session open.

Ran locally on Node 22: `npm run build`, `npm run test:unit` (0 fail across every workspace), the comprehensive e2e local, `check:api`, `check:exports-consistency`, and all three changeset guards. `verify-coverage` reports no publishable packages changed, so no changeset is needed here.

The e2e shows 372 pass / 2 fail. Both failures are `cli-client.test.ts` shelling out to `npm` and hitting my host's CodeArtifact registry with an expired token, `E401`. They are unrelated to this change and pass in CI.

Fixes #284

### DIFF
--- a/test-apps/comprehensive/aws-blocks/index.ts
+++ b/test-apps/comprehensive/aws-blocks/index.ts
@@ -117,7 +117,37 @@ const authCMfa = new AuthCognito(scope, 'authCMfa', {
 
 // AuthOIDC - OIDC sign-in gate
 // Uses the stub IdP in mock runtime — no real IdP needed for local tests.
-let lastOidcSignInUser: { userId: string; email: string | null; provider: string } | null = null;
+//
+// ── Sign-in records (e2e read-back) ─────────────────────────────────────────
+//
+// `onSignIn` runs while the OIDC callback is being served; the e2e reads the
+// result back over a later, separate request. A module-level variable cannot
+// carry it: in a deployed environment those two requests may be served by
+// different Lambda instances, so the reader either sees its own `null` or the
+// record of whoever signed in last, on either instance. The second case is
+// worse than a failure because it can pass by accident.
+//
+// Persist to the shared KVStore keyed by `userId` and scoped per AuthOIDC
+// instance, and require `userId` on the read, so a test can only ask for the
+// sign-in it is actually waiting on.
+//
+// These keys are bounded, unlike the verification codes: the stub IdP returns a
+// fixed user per provider, so each sign-in overwrites its own record.
+
+const oidcProfiles = new KVStore(scope, 'oidc-profiles');
+
+type SignInRecord = { userId: string; email: string | null; provider: string };
+
+const signInKey = (instance: string, userId: string) => `signin:${instance}:${userId}`;
+
+async function recordSignIn(instance: string, user: SignInRecord): Promise<void> {
+  await oidcProfiles.put(signInKey(instance, user.userId), JSON.stringify(user));
+}
+
+async function readSignIn(instance: string, userId: string): Promise<SignInRecord | null> {
+  const raw = await oidcProfiles.get(signInKey(instance, userId));
+  return raw === null ? null : (JSON.parse(raw) as SignInRecord);
+}
 
 const oidcProviders = [
   stubIdp({ name: 'google', onAuthorize: (req) => req.users[0] }),
@@ -127,16 +157,13 @@ const oidcProviders = [
 const oidcAuth = new AuthOIDC(scope, 'oidc-auth', {
   providers: oidcProviders,
   onSignIn: async (user) => {
-    lastOidcSignInUser = { userId: user.userId, email: user.email, provider: user.provider };
+    await recordSignIn('oidc-auth', { userId: user.userId, email: user.email, provider: user.provider });
   },
 });
 
 // AuthOIDC (second instance) — exercises the onSignIn hook with a profile
 // upsert pattern and bearer-token auth for native clients. Uses custom paths
 // to avoid colliding with the first instance.
-const oidcProfiles = new KVStore(scope, 'oidc-profiles');
-let lastExtrasSignInUser: { userId: string; email: string | null; provider: string } | null = null;
-
 const oidcAuthExtras = new AuthOIDC(scope, 'oidc-auth-extras', {
   providers: [
     stubIdp({ name: 'google-extras', onAuthorize: (req) => req.users[0] }),
@@ -148,7 +175,7 @@ const oidcAuthExtras = new AuthOIDC(scope, 'oidc-auth-extras', {
   // alongside the user, and /aws-blocks/auth/extras/refresh renews tokens.
   allowBearerAuth: true,
   onSignIn: async (user) => {
-    lastExtrasSignInUser = { userId: user.userId, email: user.email, provider: user.provider };
+    await recordSignIn('oidc-auth-extras', { userId: user.userId, email: user.email, provider: user.provider });
     // Upsert profile — the canonical post-sign-in pattern.
     await oidcProfiles.put(`profile:${user.userId}`, JSON.stringify({
       userId: user.userId,
@@ -1193,8 +1220,17 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
     return { success: true };
   },
 
-  async oidcGetLastSignInUser() {
-    return lastOidcSignInUser;
+  /**
+   * Read back the sign-in record `onSignIn` wrote for `userId`, or `null` if
+   * that user has not signed in through this instance.
+   *
+   * `userId` is required. The hook fires on a different request than this read,
+   * so "whoever signed in last" is not a safe question to ask: the answer can
+   * be another user, or another AuthOIDC instance's user, and the test would
+   * pass on the wrong record.
+   */
+  async oidcGetLastSignInUser(userId: string) {
+    return await readSignIn('oidc-auth', userId);
   },
 
   async oidcGetProviders() {
@@ -1210,8 +1246,9 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
     return { userId: user.userId, email: user.email, name: user.name, provider: user.provider, sub: user.sub, iss: user.iss };
   },
 
-  async oidcExtrasGetLastSignInUser() {
-    return lastExtrasSignInUser;
+  /** Read back the sign-in record for `userId`. See `oidcGetLastSignInUser`. */
+  async oidcExtrasGetLastSignInUser(userId: string) {
+    return await readSignIn('oidc-auth-extras', userId);
   },
 
   async oidcExtrasGetProfile(userId: string) {
--- a/test-apps/comprehensive/aws-blocks/index.ts
+++ b/test-apps/comprehensive/aws-blocks/index.ts
@@ -37,11 +37,18 @@ function logCodeLocally(message: string): void {
 // than a missing one: report it as absent so the caller's poller keeps waiting
 // and times out on its own message, rather than failing the test with a JSON
 // syntax error. An empty string is treated the same way.
-function parseStoredRecord<T>(raw: string | null): T | null {
+//
+// Absent is not silent though: a corrupt record means a bad write, which is a
+// real bug and not the race the pollers exist to absorb, so warn with the key
+// that failed. The value and the parser message are deliberately left out — the
+// message quotes the input it choked on, and these records hold verification
+// codes, which must never reach CloudWatch (see `logCodeLocally` above).
+function parseStoredRecord<T>(key: string, raw: string | null): T | null {
   if (!raw) return null;
   try {
     return JSON.parse(raw) as T;
   } catch {
+    console.warn(`[test-app] ignoring unparseable record at "${key}" (${raw.length} bytes) — treating it as absent`);
     return null;
   }
 }
@@ -97,8 +104,8 @@ async function recordDeliveredCode(channel: string, value: DeliveredCode | Deliv
 }
 
 async function readDeliveredCode<T extends DeliveredCode>(channel: string, username: string): Promise<T | null> {
-  const raw = await store.get(codeKey(channel, username));
-  return parseStoredRecord<T>(raw);
+  const key = codeKey(channel, username);
+  return parseStoredRecord<T>(key, await store.get(key));
 }
 
 async function purgeDeliveredCodes(): Promise<number> {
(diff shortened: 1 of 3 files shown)


test-apps/comprehensive/aws-blocks/index.ts
```diff
@@ -159,7 +159,37 @@ const authCMfa = new AuthCognito(scope, 'authCMfa', {
...
+
+const signInKey = (instance: string, userId: string) => `signin:${instance}:${userId}`;
+
+async function recordSignIn(instance: string, user: SignInRecord): Promise<void> {
+  await oidcProfiles.put(signInKey(instance, user.userId), JSON.stringify(user));
+}
+
+async function readSignIn(instance: string, userId: string): Promise<SignInRecord | null> {
+  const raw = await oidcProfiles.get(signInKey(instance, userId));
+  return raw === null ? null : (JSON.parse(raw) as SignInRecord);
```
soberm: `readSignIn` does an unguarded `JSON.parse(raw)`. In practice this is safe because the only writer is `recordSignIn`, which always stores `JSON.stringify(user)`, so the value is well-formed by construction — flagging only as a robustness note, not a defect. If a stray non-JSON value ever landed under a `signin:` key the reader would throw rather than returning a clean failure to the polling test.

Optional: wrap the parse in a try/catch and treat a parse failure the same as a missing record (return null), or leave as-is given the writer fully controls the stored shape. Not required to merge.

test-apps/comprehensive/test/poll-for-signin.ts
```diff
@@ -0,0 +1,84 @@
...
+ * 1. Read once. The write can still be in flight, so the read returns null.
+ * 2. Ask for "the last sign-in" with no key. Another user, or another AuthOIDC
+ *    instance's user, satisfies that instantly and the assertions run against
+ *    the wrong record, which can pass by accident.
+ *
+ * Both are designed out here: `userId` is required, and the record is checked to
+ * belong to that `userId` before it is returned.
+ *
+ * Sibling of `poll-for-code.ts`, which does the same job for verification codes.
+ * Worth folding the two into one generic poller once both have landed.
```
soberm: The header comment notes `poll-for-signin.ts` is a sibling of `poll-for-code.ts` and that the two are "worth folding into one generic poller once both have landed." `poll-for-code.ts` does not exist in this PR's tree (it comes from #278), consistent with keeping this change independent — no action needed, but the cross-reference will dangle until #278 lands.

No change needed now. When #278 lands, revisit consolidating the two pollers into one generic helper.

soberm (dismissed):
Bar-raising review — approve.

The core value here is correctness: the old `onSignIn` hooks wrote the signed-in user to module-level variables that a later, separate read couldn't see across Lambda instances, and because assertions only checked `provider`, a leftover record could pass while proving nothing. This PR fixes both by persisting to the `oidc-profiles` KVStore under `signin:<instance>:<userId>` and making `userId` required on the read. The approach is sound and idiomatic — it reuses the same KVStore the second instance already upserts profiles into rather than adding a store, and the scope is right for a test-app-only change. No blocking or actionable issues; both new tests carry a documented control run showing they bite.

Strengths:
- `recordSignIn`/`readSignIn` (index.ts:143-150) persist to a KVStore keyed by `signin:<instance>:<userId>`, so the read no longer depends on which Lambda instance served the callback — the exact cross-instance defect from #284.
- Making `userId` required on `oidcGetLastSignInUser`/`oidcExtrasGetLastSignInUser` (index.ts:1232, 1250) turns "give me whoever signed in last" into a compile-time error over the old stale-record-prone no-arg reader.
- `pollForSignIn` verifies `seen.userId === userId` before returning (poll-for-signin.ts:59-63) and throws on mismatch, so a backend keying regression surfaces clearly.
- The two new tests (oidc-auth.test.ts:309, :331) assert an unknown userId reads as `null` and each instance answers only for its own user; per the PR body each was confirmed to fail when module-level semantics are restored.
- Reuses `oidc-profiles` KVStore; the `signin:` prefix cleanly namespaces away from `profile:<userId>` keys — no collision.
- The `onSignIn fires` test now signs out at the end (oidc-auth.test.ts:294-298), closing a previously-leaked session.

Two nits below, neither blocking.

osama-rizk (dismissed):

osama-rizk (commented):
Clean, correctly-scoped, and it closes exactly the latent bug I flagged on #278 — with *stronger* enforcement (a required param backed by a compile gate) rather than just a runtime poll. Nothing blocking; findings are all "consider/document."

## What I verified

**The identity linchpin.** The whole fix hinges on one match: the test resolves `userId` via `oidcRequireAuth` and reads back the record `onSignIn` keyed on. Checked both sides on the branch — `oidcRequireAuth` returns `user.userId` from the session, and `recordSignIn` keys on the same `user.userId`. Same identity, so `signInVia → pollOidcSignIn(userId)` resolves the record the hook actually wrote. Per-instance scoping (`signin:oidc-auth:*` vs `signin:oidc-auth-extras:*`) is threaded correctly, which is what makes the "each instance keeps its own records" test meaningful.

**Store reuse is safe.** Reusing `oidc-profiles` for both `signin:*` and `profile:*` keys bites only if something enumerates the store, so I checked: `oidcProfiles` is accessed *only* by keyed `get`/`put` across the whole app file — nothing scans or counts it, and "does not duplicate" reads a specific `profile:<userId>` key rather than counting. So the new keys can't interfere. Your "no code scans it, only keyed get/put" justification checks out against the source.

**The regression genuinely bites** — two independent proofs: the control run (restoring module-level semantics fails both new tests with exactly the false-pass the old suite allowed), and the `tsc --noEmit` gate making the no-arg call a compile error so a future test can't reintroduce it. And the poller throws on `seen.userId !== userId`, so a backend *keying* bug surfaces as a distinct diagnostic rather than a silent wrong-record pass. Nice distinction.

Worth calling out: requiring `userId` at the type level (compile gate) is a stronger fix than #278's runtime-only poll scoping — the unsafe question can't even be written here.

## 1. Poller duplication with `poll-for-code.ts` (already disclosed)

`poll-for-signin.ts` is a near-twin of #278's `poll-for-code.ts`, and you call this out and keep them separate so this PR doesn't depend on #278 landing — defensible sequencing. This is now the third place the poll-until pattern exists (basic-auth, auth-cognito, here). Once both land, the right consolidation is a single generic `pollForRecord<T>(label, read, key, predicate?)` — the sign-in poller's `userId`-match check generalizes cleanly to `predicate`. Worth a tracking note so the three don't drift.

## 2. `signInVia` assumes the `requireAuth` result shape without asserting it

`signInVia` does `return { userId: meCall.result.userId, ... }` after asserting only `status === 200`. If `requireAuth`'s shape ever changed (`userId` → `sub`), `userId` would be silently `undefined`, the poll would key on `signin:<instance>:undefined`, and you'd get a confusing "no record after 15000ms" instead of "userId was undefined." A one-line `assert.ok(meCall.result.userId)` inside `signInVia` localizes that failure. Trivial hardening.

## 3. Parse-failure-as-null is the right call, one caveat

`readSignIn` treats unparseable JSON as `null` so the poller keeps waiting and times out on its own message rather than throwing a raw `SyntaxError` — good instinct for a harness. Edge: if a write ever persisted malformed JSON, this masks it as "never signed in" and you'd debug a phantom race instead of a serialization bug. Can't realistically happen here (writes are `JSON.stringify` of a fixed shape), so fine as-is — just noting the swallowed-corruption vs. clearer-timeout tradeoff is deliberate.

ahmedhamouda78:
Thanks for the review. Both nits are fixed, plus one extra.

`readSignIn` no longer parses blind: missing, empty, or unparseable now returns `null`, the same "no record yet" path the poller already handles, so `pollForSignIn` reports its own timeout instead of a JSON syntax error. Valid JSON behaves exactly as before. While in there I noticed `readDeliveredCode` had the identical shape, so both readers now share one `parseStoredRecord` helper rather than repeating the guard twice.

On the poller comment: `poll-for-code.ts` landed with #278, so the "once both have landed" note no longer said anything. Dropped it and kept the sibling pointer. Left the actual consolidation for a follow-up.

Sorry for the churn, the pushes dismissed your approval. Would appreciate another look when you have a minute.

ahmedhamouda78:
Thanks, both addressed in a34417b.

`signInVia` now asserts the id is a non-empty string before returning it, and prints the offending result when it is not. That failure mode was worse than it looked: `undefined` still builds a valid key, so a broken `requireAuth` response showed up as the poller waiting out its full timeout and reporting "no record" instead of pointing at the real cause.

On masking a corrupt write, you were right that swallowing it was too quiet. It still returns null so the not-found path is unchanged, but it now warns with the key that failed, so real corruption shows up in test output and in CloudWatch. One thing I deliberately left out of the log: the value and the parser message. The message quotes the input it choked on, and these records hold verification codes, which the file is explicit about never letting reach CloudWatch. Key and byte length are enough to find the bad write and carry nothing sensitive, which is also why the warning can fire unconditionally rather than only in local dev.

osama-rizk (approved):

soberm (approved):

PR #286, opened 2026-07-30
feat(core): VPC support

svidgen:
Stacks on #302 (shared hardening defaults). Adds VPC support to Blocks: pass a standard CDK VPC (or let Blocks derive one) and Blocks places compute, wires security groups, and provisions VPC endpoints automatically.

### Usage

```typescript
import * as ec2 from 'aws-cdk-lib/aws-ec2';

const vpc = new ec2.Vpc(app, 'AppVpc', { maxAzs: 2, natGateways: 1 });

export const blocksStack = await BlocksStack.create(app, stackName, {
  backendHandlerPath: join(__dirname, 'index.handler.ts'),
  backendCDKPath: join(__dirname, 'index.ts'),
  defaults: sandboxMode
    ? BlocksPresets.sandbox
    : { ...BlocksPresets.production, vpc: { network: vpc } },   // optional — bring your own; omit to let Blocks derive one when needed
});
```

### Implementation details

**Building Blocks declare their VPC needs at construction.** Each BB extends `BuildingBlockScope` and passes its requirements to the constructor, which registers them on a per-stack registry:

```typescript
export class KVStore extends BuildingBlockScope {
  constructor(scope: ScopeParent, id: string) {
    super(id, { parent: scope, vpc: { gatewayEndpoints: [ec2.GatewayVpcEndpointAwsService.DYNAMODB] } });
  }
}
```

The `vpc` argument is required by the constructor, so a block can't be authored without declaring its needs (`{}` if it needs nothing). At finalization Blocks reads the registry, deduplicates, and provisions endpoints.

**A VPC is derived, not required.** If nothing needs one, none is created. A block that can't function without a VPC declares `requiresVpc: true`; if one is needed and the customer didn't set `defaults.vpc`, Blocks creates a single shared VPC on demand (and emits a notice about the NAT cost). Setting `defaults.vpc = { network }` brings your own.

**Runtime egress is declared as a capability.** A block whose runtime makes outbound internet calls sets `requiresEgress: true`. At synth, Blocks checks whether the runtime's subnets actually provide egress and fails the build on a mismatch (or warns, when a VPC's subnets aren't knowable at synth — e.g. an imported VPC).

### API

```typescript
// On BlocksDefaults (`defaults.vpc`):
vpc?: { network: ec2.IVpc; subnets?: ec2.SubnetSelection; provisionEndpoints?: boolean };

// VpcRequirements — supplied to the BuildingBlockScope constructor via `vpc`:
{ gatewayEndpoints?; interfaceEndpoints?; requiresEgress?: boolean; requiresVpc?: boolean }

// Database cluster placement (CDK-free subnet selection; resolved to CDK types at synth):
new Database(scope, 'main', { subnets: { subnetType: 'isolated' } });
```

### Endpoints

- **S3 (gateway)** and **CloudWatch Logs (interface)** are always provisioned — S3 is free, and Logs is needed for in-VPC log delivery.
- All other endpoints (SSM, SQS, Secrets Manager, Bedrock, etc.) are provisioned only when a block declares them, so an app pays only for what it uses.
- Interface endpoints get a dedicated security group that allows 443 from the Blocks runtime security group only (not the whole VPC CIDR).

### `Database` subnet placement

`Database` accepts an optional `subnets` — a CDK-free subset of `ec2.SubnetSelection`'s common inputs (tier as a string, subnets by id), since the `Database` constructor resolves to a runtime entry that can't import `aws-cdk-lib`. Omit it for the default (prefer isolated, fall back to `private-with-egress`); set it to steer placement for a BYO VPC that lacks an isolated tier or a compliance requirement. The CDK layer resolves it against the VPC.

Where our `subnets` differs from `ec2.SubnetSelection`:

| `ec2.SubnetSelection` | our `subnets` | Comments |
| --- | --- | --- |
| `subnetType?: SubnetType` | `subnetType?: 'isolated' \| 'private-with-egress' \| 'public'` | Only the 3 current tiers; drops deprecated `ISOLATED` / `PRIVATE_WITH_NAT` / `PRIVATE`. |
| `availabilityZones?: string[]` | `availabilityZones?: string[]` | Identical. |
| `subnetGroupName?: string` | `subnetGroupName?: string` | Identical. |
| `onePerAz?: boolean` | `onePerAz?: boolean` | Identical. |
| `subnets?: ISubnet[]` | `subnetIds?: string[]` | By id (CDK-free); rehydrated via `Subnet.fromSubnetId` — loses AZ/route-table metadata. |
| `subnetFilters?: SubnetFilter[]` | *(omitted)* | `SubnetFilter` is a CDK class, not runtime-safe. |

Covers the DB cases we care about — isolated, or specific subnets. We can drop the extra fields later if nobody uses them.

### Alternatives considered

- **A subnet *role* per BB (`runtimeSubnet: SubnetRole`)** — conflated need with tier and rejected valid placements (egress works in `private-with-egress` *or* `public`); modeled as the `requiresEgress` capability instead.
- **A plain opt-in registry (no required declaration)** — a BB could silently forget to declare its needs and fail at deploy; the constructor requirement keeps that forcing while still centralizing dedup.
- **VPC as a required input (throw when absent)** — replaced with lazy derivation; nothing to require if the framework materializes one on demand.
- **Always provisioning SSM** — only `AppSetting`/auth use it, so it's requirement-driven; other apps don't pay for it.
- **Validating provided subnets belong to the VPC** — CDK does no synth-time check either, so it'd be a redundant preflight; the CloudFormation error surfaces at deploy.
- **Per-service endpoint opt-out** (`provisionEndpoints: { exclude }`) — deferred; needs synth-time introspection or an exclusion list of unclear value. For now: `provisionEndpoints: false` + bring your own.
- **Per-compute subnet placement** — deferred to the multi-compute surface (one app-level placement for now).

### Verification

`npm run build`, `lint:deps`, unit tests, and `test:e2e:local` all pass; changesets included.


packages/blocks/README.md
```diff
@@ -181,6 +181,41 @@ Run with `npm run test:e2e`. Write the test first, iterate against mocks until i
 - **`Database` when `DistributedTable` would do** — Aurora costs more and has cold starts; reach for SQL only when you need it.
 - **Curling REST-style paths** — there is no `GET /api/getData`. All calls are JSON-RPC to a single `POST /aws-blocks/api`; use the typed import instead.
 
+## VPC Support
```
svidgen, line 248: FYI: Revision of this section is pending after I see e2e test results.

packages/bb-data/src/infra.ts
```diff
@@ -96,25 +104,29 @@ export function materialize(
   const { minCapacity = DEFAULT_MIN_CAPACITY, maxCapacity = DEFAULT_MAX_CAPACITY, databaseName } = options;
   const envName = name.replace(ENV_NAME_SANITIZE_PATTERN, '_');
 
-  // VPC with isolated subnets only — no NAT gateways needed for Data API path
-  const vpc = new ec2.Vpc(scope, `${name}Vpc`, {
+  // Determine VPC (shared or standalone)
+  const vpc = options.vpcContext?.vpc ?? new ec2.Vpc(scope, `${name}Vpc`, {
```
sarayev: **[blocking]** The Aurora cluster is pinned to `PRIVATE_ISOLATED` (line 163), but the VPC in every docs example (`new ec2.Vpc(app, 'AppVpc', { maxAzs: 2, natGateways: 1 })`) has no isolated tier. So if I follow the README and add a `Database`, synth fails with "no isolated subnet groups in this VPC." Could we fall back to the context's subnet selection (or `PRIVATE_WITH_EGRESS`) when there's no isolated tier, instead of hard-requiring one the docs never create?

packages/bb-data/src/infra.ts
```diff
@@ -96,25 +104,29 @@ export function materialize(
...
     allowAllOutbound: false,
   });
+
+  // Ingress rule differs based on context
   securityGroup.addIngressRule(
-    ec2.Peer.ipv4(vpc.vpcCidrBlock),
+    options.vpcContext
+      ? options.vpcContext.lambdaSecurityGroup
+      : ec2.Peer.ipv4((vpc as ec2.Vpc).vpcCidrBlock),
     ec2.Port.tcp(DEFAULT_POSTGRES_PORT),
```
sarayev: **[minor]** Does this 5432 ingress rule ever actually get used? The cluster runs with `enableDataApi: true` and we reach it over the RDS Data API (HTTPS), not a raw Postgres socket, so nothing seems to open a 5432 connection. If that's right, can we drop it (or add a note on why 5432 is still needed)? As-is "Lambda to Aurora" implies a live path that doesn't exist.

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,143 @@
...
+  // Provision interface endpoints (deduplicated)
+  const provisionedInterface = new Set<string>();
+
+  for (const service of interfaceEndpoints) {
+    const key = service.name;
+    if (provisionedInterface.has(key)) continue;
+    provisionedInterface.add(key);
+
+    const constructId = `VpcIf${key.replace(/[^a-zA-Z0-9]/g, '')}`;
+    new ec2.InterfaceVpcEndpoint(scope, constructId, {
```
sarayev: **[major]** We don't pass `securityGroups` here, so CDK falls back to a default SG that allows 443 from the whole VPC CIDR, broader than `bb-data` which scopes ingress to the Lambda SG. On a bring-your-own-VPC that widens who can reach every interface endpoint. Could we give the endpoints a dedicated SG that only allows 443 from `lambdaSecurityGroup` (available via `getVpcContext(scope)`)?

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,143 @@
...
+  // Provision gateway endpoints (deduplicated)
+  const provisionedGateway = new Set<string>();
+
+  for (const service of gatewayEndpoints) {
+    const key = (service as any).name ?? String(service);
+    if (provisionedGateway.has(key)) continue;
+    provisionedGateway.add(key);
+
+    const constructId = `VpcGw${key.replace(/[^a-zA-Z0-9]/g, '')}`;
+    new ec2.GatewayVpcEndpoint(scope, constructId, { vpc, service });
```
sarayev: **[minor]** None of the endpoints set a `policyDocument`, so they get AWS's default allow-all. IAM still gates the actual calls so this isn't an open door, but as defense-in-depth (especially the gateway endpoints, which have no SG layer) would it be worth scoping the policy to the deploying account? Fine to defer if it's out of scope for v1.

packages/core/src/cdk/vpc-types.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Returned by `BuildingBlockScope.getVpcRequirements()` and collected
+ * during finalization to provision endpoints.
+ */
+export interface VpcRequirements {
+  /** Gateway VPC endpoints this BB needs (e.g., S3, DynamoDB). */
+  gatewayEndpoints?: ec2.GatewayVpcEndpointAwsService[];
+  /** Interface VPC endpoints this BB needs (e.g., SQS, SSM, Secrets Manager). */
+  interfaceEndpoints?: ec2.InterfaceVpcEndpointAwsService[];
+  /** Subnet role for VPC-resident resources (e.g., Aurora needs 'isolated'). */
+  subnetRole?: SubnetRole;
```
sarayev: **[major]** `subnetRole` is defined here, `bb-data` returns `'isolated'`, and `selectSubnets()` implements the mapping, but nothing ever calls `selectSubnets` and `finalizeVpc` never reads `subnetRole`. Right now it's dead config that also reads as load-bearing: someone will change `subnetRole` expecting placement to move and it won't. Wire it through (which would also fix the Aurora subnet issue) or drop it until it's consumed?

packages/core/src/cdk/index.ts
```diff
@@ -45,20 +50,35 @@ export class BlocksStack extends cdk.Stack implements BaseBlocksStack {
...
     // Set globalThis so Building Blocks attach directly to this stack
     (globalThis as any).CURRENT_BLOCKS_STACK = this;
 
-    const infra = setupBlocksInfra(this, props, id);
-    this.handler = infra.handler;
-    this.gateway = infra.gateway;
-    this.apiUrl = infra.apiUrl;
-    this.executionRole = infra.executionRole;
+    // Initialize VPC context before BBs are constructed (so BBs can discover it)
+    if (props.vpc) {
```
sarayev: **[minor]** These two branches are nearly identical, both call `setupBlocksInfra` and assign the same four fields. `BlocksBackend` does it in a single path. Can we mirror that here so they don't drift?
```ts
const vpcContext = props.vpc ? initializeVpc(this, props.vpc) : undefined;
const infra = setupBlocksInfra(this, props, id, vpcContext);
```

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,143 @@
...
+ * Creates the security group, sets VPC context, and returns the VpcContext
+ * that is used for Lambda placement configuration.
+ *
+ * Called during setupBlocksInfra when `vpc` prop is present.
+ * @internal
+ */
+export function initializeVpc(scope: Construct, options: BlocksVpcOptions): VpcContext {
+  const { network: vpc, subnets } = options;
+
+  const resolvedSubnets = subnets ?? { subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS };
```
sarayev: **[minor]** If someone passes a VPC with no `PRIVATE_WITH_EGRESS` subnets (isolated-only, no NAT), Lambda placement fails later with a fairly cryptic CDK error. Worth a quick guard that throws something actionable like "your VPC needs a private-with-egress subnet for Lambda"?

packages/bb-distributed-data/src/index.cdk.ts
```diff
@@ -19,7 +20,13 @@ import { join, resolve } from 'node:path';
 import type { DistributedDatabaseOptions } from './types.js';
 import { LAMBDA_MIGRATIONS_DIR, MIGRATION_LAMBDA_TIMEOUT_MINUTES, ENV_SANITIZE, sanitizeDbRoleName } from './constants.js';
 
-export class DistributedDatabase extends Scope {
+export class DistributedDatabase extends BuildingBlockScope {
+  getVpcRequirements(): VpcRequirements {
```
sarayev: **[major]** This returns `{}` and the comment assumes a NAT-backed egress subnet. If a user selects isolated subnets (the exact compliance case VPC.md recommends a VPC for), it synths clean and then every DSQL call times out at runtime. Since DSQL has no interface endpoint to add, could we at least detect isolated placement and fail fast with a clear message rather than fail silently in prod?

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -97,10 +107,12 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
     assumedBy: new iam.CompositePrincipal(new iam.ServicePrincipal('lambda.amazonaws.com')),
     managedPolicies: [
       iam.ManagedPolicy.fromAwsManagedPolicyName('service-role/AWSLambdaBasicExecutionRole'),
+      // When Lambda is placed in a VPC it needs ENI management permissions
+      ...(vpcContext ? [iam.ManagedPolicy.fromAwsManagedPolicyName('service-role/AWSLambdaVPCAccessExecutionRole')] : []),
     ],
   });
 
-  const handler = new lambda.NodejsFunction(scope, 'Handler', {
+  const handlerProps: any = {
```
sarayev: **[minor]** `const handlerProps: any = { ... }` drops type-checking on the Lambda props, and we ban `any` in customer-facing code. Can we type it as `NodejsFunctionProps` and set the VPC fields conditionally so the shape is still checked? Same for `(service as any).name` in `vpc.ts` and `(vpc as ec2.Vpc).vpcCidrBlock` in `infra.ts`, both properties exist on the interface types so those casts look removable.

packages/core/src/cdk/vpc.test.ts
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, it } from 'node:test';
+import * as assert from 'node:assert/strict';
+import { getVpcContext, setVpcContext } from './vpc.js';
+
+// We can't fully test CDK constructs without a Stack, but we can test
+// the getVpcContext logic with mock constructs.
+
+describe('VPC utilities', () => {
```
sarayev, line 275: **[blocking]** `finalizeVpc` is the core of this PR but nothing tests it. This file only covers the trivial get/set context helpers, and the one e2e app runs with `provisionEndpoints: false` so the provisioning path never executes anywhere. Could we add a `Template.fromStack(...)` test asserting the expected `AWS::EC2::VPCEndpoint` set (plus dedup and the always-on Logs/SSM) for a couple of BBs? Highest-value coverage and no deploy needed.

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -127,7 +139,16 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
       minify: true,
       esbuildArgs: { '--conditions': 'aws-runtime' },
     }),
-  });
+  };
+
+  // Apply VPC placement to the Lambda if VPC context is provided
+  if (vpcContext) {
```
sarayev: **[major]** This VPC placement branch (and the `BlocksStack` one) has no unit coverage, `blocks-backend.test.ts` / `blocks-stack.test.ts` don't reference `vpc` at all. Related: `bb-data`'s `infra.test.ts` only exercises the standalone branch ("PostgreSQL from VPC CIDR") and never passes a `vpcContext`, so the shared-VPC ingress branch (Lambda SG to Aurora) is untested too. Can we add assertions for the `vpcContext` paths?

test-infra/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+## Resources Created
+
+| Resource | Purpose | Cost |
+|----------|---------|------|
+| VPC (2 AZs, 1 NAT) | Network isolation for Lambda | ~$32/mo (NAT) |
+| DynamoDB gateway endpoint | KVStore, DistributedTable | Free |
+| S3 gateway endpoint | FileBucket | Free |
+| SSM interface endpoint | AppSetting, Auth session secrets | ~$7/mo/AZ |
```
sarayev: **[major]** This "Resources Created" table lists SSM, Secrets Manager, and CloudWatch Logs endpoints, but `vpc-test-stack.ts` creates none of them (it's intentionally bare: VPC + 1 NAT). Can we align the table with what the stack actually provisions? Otherwise people go looking for endpoints that aren't there.

packages/blocks/VPC-DESIGN.md
```diff
@@ -0,0 +1,177 @@
...
+
+## BB Endpoint Registration
+
+Each Building Block declares what VPC endpoints it needs via two explicit methods on the `Scope` class. No `instanceof` detection — each BB calls the method matching its endpoint type directly.
+
+### Registration API
+
+```typescript
+// On Scope (core/cdk)
+protected getVpcRequirements()(service: ec2.GatewayVpcEndpointAwsService): void;
```
sarayev: **[major]** This signature is garbled (`getVpcRequirements()(service: ...)`) and the per-BB table below still uses the old `this.getVpcRequirements()(...)` call form, neither matches the shipped `getVpcRequirements(): VpcRequirements` return-object API. Since docs are part of "done" here, can we update this section to the real API?

packages/blocks/VPC.md
```diff
@@ -0,0 +1,103 @@
...
+
+| Resource | Cost | Notes |
+|---|---|---|
+| NAT Gateway | ~$32/month per gateway + $0.045/GB processed | Required for Lambda internet access from private subnets. 1 per AZ recommended for HA. |
+| Interface VPC Endpoints | ~$7.20/month per endpoint per AZ | Auto-provisioned per BB (SQS, Secrets Manager, Bedrock, etc.). Adds up quickly with many BBs. |
+| Gateway VPC Endpoints (S3, DynamoDB) | Free | No additional cost. |
+
+**Example:** A 2-AZ app using KVStore + Database + AsyncJob + AppSetting:
+- 1 NAT Gateway: ~$32/month
+- 4 interface endpoints × 2 AZs × $7.20: ~$57.60/month
```
sarayev: **[minor]** This counts 4 interface endpoints, but `finalizeVpc` always adds CloudWatch Logs on top, so it's really 5 (Secrets Manager, RDS Data, SQS, SSM, CW Logs). That makes the total ~$104/mo rather than ~$90. Worth correcting so the estimate isn't ~16% low.

.github/workflows/pr-checks.yml
```diff
@@ -319,7 +319,15 @@ jobs:
           role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
           aws-region: us-east-1
 
-      - name: E2E Test Sandbox (VPC Smoke)
+      - name: Ensure persistent test VPC
```
sarayev, line 330: **[major]** This step runs `cdk deploy` against the shared, persistent `BlocksTestVpc` stack with no job-level concurrency guard. The `e2e-supabase` job just below has exactly the guard this one needs. Without it, two open PRs both deploy the same stack and CloudFormation fails the second one on an unrelated PR. Can we add:
```yaml
concurrency:
  group: blocks-test-vpc
  cancel-in-progress: false
```

test-apps/vpc-smoke/test/vpc-smoke.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  const json = await res.json() as any;
+  if (json.error) throw new Error(`RPC ${method} error: ${json.error.message || JSON.stringify(json.error)}`);
+  return json.result ?? json.data;
+}
+
+test('VPC Smoke Tests', async (t) => {
+  t.before(async () => {
+    if (ENV === 'local') {
+      console.log('⏭️  VPC smoke tests require deployment — skipping in local mode.');
+      process.exit(0);
```
sarayev: **[nit]** Skipping local runs via `process.exit(0)` works but force-kills the whole runner. `t.skip()` or an early return keeps the exit semantics clean.

test-apps/vpc-smoke/aws-blocks/index.cdk.ts
```diff
@@ -0,0 +1,64 @@
...
+const env = { account: process.env.CDK_DEFAULT_ACCOUNT, region: process.env.CDK_DEFAULT_REGION || 'us-east-1' };
+
+export const blocksStack = await BlocksStack.create(app, stackName, {
+  backendHandlerPath: join(__dirname, 'index.handler.ts'),
+  backendCDKPath: join(__dirname, 'index.ts'),
+  defaults: BlocksPresets.sandbox,
+  vpc: {
+    // Use fromLookup with the stack's own env — CDK will resolve it at synth.
+    // This works because BlocksStack IS a Stack, satisfying CDK's scope requirement.
+    network: ec2.Vpc.fromLookup(new cdk.Stack(app, `${stackName}-vpc-ref`, { env }), 'Vpc', { vpcId }),
```
sarayev, line 59: **[nit]** Creating a throwaway stack just to host `Vpc.fromLookup`, then using the result in a different stack, reads as uncertain (the comment says "BlocksStack IS a Stack"). It's actually fine since `fromLookup` resolves to concrete attributes at synth. A one-line "why a separate lookup stack" would help the next reader.

packages/bb-data/src/infra.ts
svidgen: Ack — good catch, this slipped when I factored out the VPC stack. Fixed: on the shared-VPC path the cluster now lands in the isolated tier only when the VPC actually has one, and falls back to `PRIVATE_WITH_EGRESS` otherwise (via the context's `selectSubnets`), so the documented `natGateways: 1` VPC synths clean. Reachability is unaffected since we're on the Data API, not a socket. Standalone path is unchanged. Fixing the two blockers and retesting before I circle back on the rest of your comments.

packages/core/src/cdk/vpc.test.ts
svidgen: Ack — you're right, this slipped when I factored out the VPC stack and the e2e app ended up on `provisionEndpoints: false`, so the provisioning path never ran anywhere (and the test-infra comment claimed the opposite — my bad). Two fixes: (1) added `Template.fromStack` unit coverage for `finalizeVpc` asserting the endpoint set, gateway/interface dedup, and the always-on Logs/SSM pair; (2) flipped vpc-smoke to `provisionEndpoints: true` so endpoints provision into the per-PR app stack and the real auto-detection path runs e2e. Since AWS only allows one private-DNS interface endpoint per service per VPC, I also added the job-level `concurrency: { group: blocks-test-vpc, cancel-in-progress: false }` gate you flagged below so concurrent PRs queue instead of colliding. Retesting now; will circle back on the other comments after.

svidgen:
Pushed a first pass addressing the two **blocking** items:

- **Aurora isolated-subnet fallback** (`bb-data`) — shared-VPC path now prefers the isolated tier when the VPC has one and falls back to `PRIVATE_WITH_EGRESS` otherwise, so the documented `natGateways: 1` VPC synths clean. Standalone path unchanged.
- **`finalizeVpc` coverage** — added `Template.fromStack` unit tests (endpoint set, gateway/interface dedup, always-on Logs/SSM), flipped vpc-smoke to `provisionEndpoints: true` so the e2e actually exercises the provisioning path, and added the job-level `concurrency` gate (`blocks-test-vpc`) so concurrent PRs queue rather than collide on the shared VPC's private-DNS endpoints.

More changes are coming — I'm not ignoring the rest of the feedback. There's a lot of meaningful review here, so I'm taking it in a few passes: landing fixes incrementally and in isolation so each one is easy to review and any regression is easy to catch as it lands. I'm leaning on the PR's CI (incl. the VPC e2e) as the end-to-end layer between passes. Next pass will pick up the remaining major items (endpoint SG scoping, `subnetRole` wiring, DSQL fail-fast, the untested placement branches) and then the minors/nits + doc corrections.

.github/workflows/pr-checks.yml
svidgen: Needed for other reasons as well, FWIW.

packages/core/src/cdk/vpc.ts
svidgen: Done. Endpoints now get a dedicated SG allowing 443 only from the Lambda SG — and I had to add `open: false` on the endpoint too, because CDK was *also* adding an allow-443-from-the-whole-VPC-CIDR rule even with a custom SG. The new unit test caught that (it initially found 2 ingress rules). So it was a bit worse than it looked; good catch.

packages/core/src/cdk/vpc-types.ts
svidgen: Reworked rather than dropped. `subnetRole` is replaced by `runtimeSubnet` — the subnet role the BB's *parent runtime* (shared Lambda) must be in — which finalizeVpc now actually validates against the Lambda placement and fails synth on mismatch. Own-resource placement (Aurora) stays where it belongs: resolved inline via the now-instructive `selectSubnets`. So the config is consumed and load-bearing now, in both directions.

packages/bb-distributed-data/src/index.cdk.ts
svidgen: Fixed via the runtimeSubnet mechanism above. DSQL now returns `{ runtimeSubnet: 'private-with-egress' }`, so if the Lambda lands in isolated subnets finalizeVpc throws an actionable synth error instead of shipping a deploy that times out on every call.

packages/core/src/cdk/blocks-backend.ts
svidgen: Added VPC placement coverage to blocks-backend.test.ts (Lambda gets VpcConfig + the VPC-access managed policy; no VpcConfig when VPC is omitted) and a shared-VPC path to bb-data infra.test.ts (isolated placement + fallback, and the no-ingress assertion). blocks-stack.test.ts shares the same de-duped setupBlocksInfra path, so it's covered by the backend tests rather than duplicated.

packages/core/src/cdk/blocks-backend.ts
svidgen: Done — all three. handlerProps is typed `NodejsFunctionProps` with VPC fields spread conditionally; `(service as any).name` is just `service.name` (it's public on the gateway service too); and `(vpc as ec2.Vpc).vpcCidrBlock` is gone entirely since the 5432 ingress it fed was removed.

packages/bb-data/src/infra.ts
svidgen: Dropped. The cluster runs Data-API-only (HTTPS via the endpoints), nothing opens 5432, so the rule (and the vpcCidrBlock cast feeding it) is gone. SG now has no ingress; added a test asserting that and updated DESIGN.md + the index.cdk.ts header.

packages/core/src/cdk/index.ts
svidgen: Done — mirrored the BlocksBackend single path: `const vpcContext = props.vpc ? initializeVpc(this, props.vpc) : undefined; const infra = setupBlocksInfra(this, props, id, vpcContext);` (and removed the now-unused ec2 import).

packages/core/src/cdk/vpc.ts
svidgen: Added. initializeVpc now throws an actionable error when it would default to private-with-egress but the VPC has none ('add a PRIVATE_WITH_EGRESS tier, or pass vpc.subnets'). An explicit isolated selection is still honored — the guard only fires on the default path.

test-infra/README.md
svidgen: Fixed — the table now reflects the actual bare stack (VPC + 1 NAT + the three subnet tiers, no endpoints/Aurora) and notes that endpoints are provisioned per-run into the app stack.

packages/blocks/VPC-DESIGN.md
svidgen: Rewritten to the real `getVpcRequirements(): VpcRequirements` return-object API, and the per-BB table now shows the returned objects (including bb-distributed-data's new `runtimeSubnet`).

packages/blocks/VPC.md
svidgen: Corrected to 5 interface endpoints (added CloudWatch Logs) → ~$104/mo.

test-apps/vpc-smoke/test/vpc-smoke.test.ts
svidgen: Done — the suite now t.skip()s in local mode instead of process.exit(0).

test-apps/vpc-smoke/aws-blocks/index.cdk.ts
svidgen: Added a comment explaining the throwaway lookup stack (fromLookup needs a Stack scope and must resolve before BlocksStack.create wires placement; it resolves to concrete attributes at synth, so it's safe to use from the other stack).

.github/workflows/pr-checks.yml
svidgen: Already handled in the first pass — the e2e-sandbox-vpc job got the `concurrency: { group: blocks-test-vpc, cancel-in-progress: false }` guard so concurrent PRs queue instead of colliding on the shared VPC.

packages/core/src/cdk/vpc.ts
svidgen: Deferring this one for now, per your 'fine to defer for v1' note. IAM still gates the actual calls, so it's defense-in-depth rather than a hole. I'd like to do it as its own follow-up (scoped endpoint policies for both interface and gateway endpoints, keyed to the deploying account) rather than fold a half-version into this pass. Tracking it separately — shout if you'd rather it block here.

svidgen:
Second pass — the majors, minors, and nits. Replied inline on each thread; summary here.

**Design note worth surfacing:** `subnetRole` didn't get dropped or blindly wired — the discussion pushed it somewhere better. It's now two distinct concerns:
- **Own-resource placement** (e.g. Aurora): the BB self-resolves via an **instructive** `selectSubnets(scope, role, { fallback })` that verifies the tier exists and throws a **BB-named** error (with an explicit, non-silent fallback) instead of the opaque CDK 'no subnet groups' error.
- **Parent-runtime requirement** (e.g. DSQL): a new `runtimeSubnet` field that `finalizeVpc` **validates** against the Lambda's placement and fails synth on mismatch. It never relocates the Lambda — an unsatisfiable requirement is an error, not a silent move. This is what turns the DSQL-in-isolated footgun into a build error.

**Majors:** endpoint SG scoped to 443-from-Lambda-only (needed `open: false` too — CDK was adding a VPC-CIDR-wide rule even with a custom SG; the new test caught it); `runtimeSubnet` validation + DSQL fail-fast; `any` casts removed; BlocksStack branch de-dup; VPC-DESIGN/VPC.md/test-infra doc fixes; placement-branch tests added.

**Minors/nits:** missing-private-with-egress guard; dropped bb-data's dead 5432 ingress; `t.skip()` over `process.exit(0)`; lookup-stack comment.

**Deferred (called out inline):** endpoint `policyDocument` scoping — defense-in-depth (IAM already gates calls), doing it as a dedicated follow-up rather than a half-version here. Shout if you'd rather it block.

Verified locally: full build clean, 61 core-cdk + 29 bb-data + 157 bb-distributed-data unit tests pass, lint clean, `check:api` clean. CI running now.

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,257 @@
...
+      if (reqs.interfaceEndpoints) {
+        interfaceEndpoints.push(...reqs.interfaceEndpoints);
+      }
+      // Validate the BB's parent-runtime subnet requirement against where the
+      // Lambda is actually placed. We never move the Lambda (that's the
+      // customer's explicit choice) — an unsatisfiable requirement is a synth
+      // error, not a silent relocation. This turns an otherwise-silent runtime
+      // failure (e.g. DSQL in an isolated Lambda: deploys clean, times out on
+      // every call) into an actionable build error.
+      if (reqs.runtimeSubnet && !runtimeSatisfies(lambdaPlacement, reqs.runtimeSubnet)) {
```
Simone319: If I'm reading this correctly, there is only a single `lambdaPlacement` that can satisfy the block's `VpcRequirements.runtimeSubnet` with only one exception `placement === 'private-with-egress' && required === 'isolated'`. Hence I have 2 questions:
1. If a building block need to access internet, looks like `'private-with-egress' | 'public'` both works. But `runtimeSubnet` only allows it to specify one role. Then if the block puts `'private-with-egress'` as requirement, lambda put into the `public` subnet cannot satisfy the requirement, when it should work. Should `runtimeSubnet` be a set instead? 
2. Suppose 2 building blocks requires 2 different role, let's say one `'private-with-egress'` and one `'isolated'`, one compute can't satisfy both right? How can we communicate this to customers to ensure they create the computes accordingly, and use it with the right bb?

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,257 @@
...
+ * deduplicate, and provision endpoints.
+ * Called after all BBs are constructed (alongside finalizeConfigRegistry).
+ * @internal
+ */
+export function finalizeVpc(scope: Construct, options: BlocksVpcOptions): void {
+  const { network: vpc, subnets } = options;
+
+  // The subnet role the shared handler Lambda is actually placed in. Mirrors the
+  // default applied in initializeVpc.
+  const lambdaPlacement: SubnetRole = subnetSelectionRole(subnets) ?? 'private-with-egress';
```
Simone319: With one compute, it makes sense that customers select a single placement for default compute. When multiple compute comes, are we forcing every compute to be put into the subnets? I can see it becomes a limitation for customers who wants their container to be in an isolated subnet, but the async processing lambda to be in `private-with-egress` subnet to make internet calls, or vice versa.

packages/core/src/cdk/index.ts
```diff
@@ -222,3 +238,20 @@ export class Scope extends Construct {
...
+ * to collect gateway/interface endpoint needs and provision them centrally.
+ */
+export abstract class BuildingBlockScope extends Scope {
+  /**
+   * Declare what VPC resources this BB needs when deployed in a VPC.
+   * Called at finalization time ONLY when a VPC is configured.
+   *
+   * Return an empty object `{}` if no VPC-specific resources are needed.
+   */
+  abstract getVpcRequirements(): VpcRequirements;
```
Simone319: nit: Have we considered a centralized registry, similar to `awsSdkIdentifiers` registry? Such that bb does not need to maintain this method. The centralized registry can handle dedup, and `get()` functionality, so that when we need to get a block requirement from `core` for example, we can get it from the registry.

packages/core/src/cdk/vpc.ts
Simone319: One more question on this line, what if the vpc doesn't have `'private-with-egress'`?

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,257 @@
...
+      // CIDR" ingress rule. Our dedicated SG already allows 443 from just the
+      // Blocks Lambda SG; without this, CDK would re-widen access to the entire
+      // VPC — the exact broadening this dedicated SG exists to prevent.
+      open: false,
+    });
+  }
+}
+
+/** The SubnetRole implied by an explicit SubnetSelection, if it maps cleanly. */
+function subnetSelectionRole(selection?: ec2.SubnetSelection): SubnetRole | undefined {
```
Simone319: Reading from `ec2.SubnetSelection` [doc](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.SubnetSelection.html), subnetType is not required. Customers can also just provide a list of subnets, which resolves to `undefined`.

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * Initialize VPC support on a Blocks scope (BlocksStack or BlocksBackend).
+ * Creates the security group, sets VPC context, and returns the VpcContext
+ * that is used for Lambda placement configuration.
+ *
+ * Called during setupBlocksInfra when `vpc` prop is present.
+ * @internal
+ */
+export function initializeVpc(scope: Construct, options: BlocksVpcOptions): VpcContext {
+  const { network: vpc, subnets } = options;
```
Simone319: Maybe I'm missing it somewhere, but are we checking anywhere whether provided subnets in `BlocksVpcOptions` exists in the `vpc`?

packages/core/src/cdk/index.ts
```diff
@@ -45,16 +49,23 @@ export class BlocksStack extends cdk.Stack implements BaseBlocksStack {
   /** Infrastructure defaults for Building Blocks created under this stack. */
   public readonly defaults: BlocksDefaults;
 
+  private _vpcOptions?: BlocksVpcOptions;
```
Simone319: When we enable containers `compute`, vpc becomes required. 

But currently, the UX is a bit fragmented: `vpc` defined in `index.cdk.ts`, while `compute` is defined in `index.ts`, and `compute` requires `vpc` to be created. Any idea how could we ensure the dependency on the type level? Alternative is to throw when no vpc is provided, which I would consider as the last choice.

packages/core/src/cdk/vpc-types.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Internal VPC context propagated through the construct tree.
+ * Set by the CDK-level VPC option. BBs read this to determine their VPC placement.
+ * @internal
+ */
+export interface VpcContext {
+  readonly vpc: ec2.IVpc;
+  readonly lambdaSecurityGroup: ec2.ISecurityGroup;
+  readonly lambdaSubnets: ec2.SubnetSelection;
```
Simone319: Related to the multiple compute sharing some `subnetType` comments, if all the compute falls into the same subnets, the naming should be more generic like `computeSubnets`?

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,257 @@
...
+  if (ctx) {
+    endpointSecurityGroup.addIngressRule(
+      ec2.Peer.securityGroupId(ctx.lambdaSecurityGroup.securityGroupId),
+      ec2.Port.tcp(443),
+      'HTTPS from Blocks Lambda',
+    );
+  }
+
+  // Provision interface endpoints (deduplicated)
+  const provisionedInterface = new Set<string>();
```
Simone319: In an enterprise shared VPC that already has the endpoints to be added, deploy fails on the duplicate. The escape is `provisionEndpoints: false` — which is all-or-nothing. There's no "provision only what's missing" / per-service opt-out.

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,257 @@
...
+    provisionedGateway.add(key);
+
+    const constructId = `VpcGw${key.replace(/[^a-zA-Z0-9]/g, '')}`;
+    new ec2.GatewayVpcEndpoint(scope, constructId, { vpc, service });
+  }
+
+  // Always add CloudWatch Logs (Lambda needs it for log delivery from within VPC)
+  interfaceEndpoints.push(ec2.InterfaceVpcEndpointAwsService.CLOUDWATCH_LOGS);
+  // Always add SSM (auth BBs and AppSetting all use SSM)
+  interfaceEndpoints.push(ec2.InterfaceVpcEndpointAwsService.SSM);
```
Simone319: If those BB needs SSM, they should specify it in their requirement right? Is there any apps that can be created without accessing SSM? In that case customers will be paying for unused endpoint.

packages/core/src/cdk/vpc.ts
```diff
@@ -0,0 +1,257 @@
...
+  for (const service of gatewayEndpoints) {
+    const key = service.name;
+    if (provisionedGateway.has(key)) continue;
+    provisionedGateway.add(key);
+
+    const constructId = `VpcGw${key.replace(/[^a-zA-Z0-9]/g, '')}`;
+    new ec2.GatewayVpcEndpoint(scope, constructId, { vpc, service });
+  }
+
+  // Always add CloudWatch Logs (Lambda needs it for log delivery from within VPC)
```
Simone319: In addition, we also need S3 GW endpoint, for accessing S3 config during start up.

packages/core/src/cdk/index.ts
svidgen: Yeah. I originally went in that direction, actually. But, for the sake of 3rd party blocks, it felt necessary to "demand" VPC that each BB report its requirements. I think this is different than `awsSdkIdentifiers`, IIRC. But, we could revisit both of these together.

packages/core/src/cdk/vpc.ts
svidgen: Good questions ... Perhaps each block should communicate more granular requirements. I.e., `requiresEgress: boolean` and `requiresIngres: boolean`.

svidgen: Switching `runtimeSubnet` to `requiresEgress: boolean` and letting the framework pick which subnets satisfy it (egress works in private-with-egress or public).

packages/core/src/cdk/vpc.ts
svidgen: My assumption is we want a default VPC placement at the app level that can be overridden per-Compute BB, so each compute lands where it needs to.

packages/core/src/cdk/index.ts
svidgen: Leaning this way — BBs pass requirements into the `BuildingBlockScope` constructor so the base registers them centrally. Stays required, so a BB can't silently skip it.

packages/core/src/cdk/vpc.ts
svidgen: `initializeVpc` throws with a fix-it message when the tier's missing.

packages/core/src/cdk/vpc.ts
svidgen: ack. Pulling the role guess out: resolve egress from the selected subnets, and if we can't tell (imported VPC), warn and skip.

packages/core/src/cdk/vpc.ts
svidgen: We don't, and I don't think we should. CDK doesn't check this at synth either (an imported subnet doesn't carry a vpcId), so we'd just be re-predicting the CloudFormation error you'd get anyway. Since you're wiring subnets yourself here, better to let the normal CDK/CFN error through. I'll try to make this clear in docstrings and agent-facing READMEs.

packages/core/src/cdk/vpc-types.ts
svidgen: Will rename `lambdaSubnets` → `computeSubnets` when the per-compute placement change lands.

packages/core/src/cdk/vpc.ts
svidgen: I think we can extend here in the future, but without a preflight check the best I've got is an exclusion list — and honestly that feels a bit awkward. I'm not sure what customers would actually find meaningfully helpful yet. Can we handle it in a follow-up if someone asks for it?

packages/core/src/cdk/vpc.ts
svidgen: ack, fixing — it comes from BB requirements now (only AppSetting and auth actually use SSM), so a KV/DynamoDB app won't get billed for an endpoint it never calls. Leaving CloudWatch Logs on since every in-VPC Lambda needs it.

packages/core/src/cdk/vpc.ts
svidgen: ack

packages/core/src/cdk/index.ts
svidgen: Whoops! I thought I had responded to this comment. 🤔 

My thought is that compute containers will lazily create VPC (or ask for one during instantiation) the same way Aurora does in this PR.

pranavosu (dismissed):

packages/core/src/cdk/vpc.ts
Simone319: Agreed we shouldn't validate this at synth — CDK doesn't either (an imported subnet carries no vpcId), so we'd only be re-predicting the CloudFormation error you'd already get. One follow-up though: you mentioned you'd make this clear in docstrings and agent-facing READMEs, and that doesn't seem to have landed yet. Can we add the caveat — subnets you supply must belong to `network`; not validated at synth, CFN will error if they don't — to the `subnets` docstring (`vpc-types.ts`) and to `VPC.md` as a follow-up action item? No code change needed.

packages/core/src/cdk/vpc-types.ts
Simone319: Fine to leave `lambdaSubnets`/`lambdaSecurityGroup` as-is for now since everything is Lambda today. Let's do the rename to `computeSubnets` once we introduce more compute options, so the generic name lands together with the code that makes it meaningful.

packages/core/src/cdk/vpc.ts
Simone319: Yes, let's handle this in a follow-up. Agreed a bare exclusion list is awkward without a preflight/existing-endpoint check, and it's better to design the escape hatch (provision-only-missing vs. per-service opt-out) around a real customer ask than guess now. `provisionEndpoints: false` as the full opt-out is fine in the meantime.

packages/blocks/VPC.md
```diff
@@ -0,0 +1,105 @@
...
+| Resource | Cost | Notes |
+|---|---|---|
+| NAT Gateway | ~$32/month per gateway + $0.045/GB processed | Required for Lambda internet access from private subnets. 1 per AZ recommended for HA. |
+| Interface VPC Endpoints | ~$7.20/month per endpoint per AZ | Auto-provisioned per BB (SQS, Secrets Manager, Bedrock, etc.). Adds up quickly with many BBs. |
+| Gateway VPC Endpoints (S3, DynamoDB) | Free | No additional cost. |
+
+**Example:** A 2-AZ app using KVStore + Database + AsyncJob + AppSetting:
+- 1 NAT Gateway: ~$32/month
+- 5 interface endpoints × 2 AZs × $7.20: ~$72/month
+  (SQS, Secrets Manager, RDS Data, and the always-on CloudWatch Logs + SSM that
```
Simone319: Stale after the SSM change: SSM is no longer always-on — it's provisioned only from BB requirements (AppSetting + auth); only CloudWatch Logs is unconditional. Please drop SSM from the always-provisioned list here and at line 55, and update the ~$104/mo example since it's now overstated for apps that don't use SSM.

packages/blocks/VPC-DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+| bb-agent | `{ interfaceEndpoints: [BEDROCK_RUNTIME] }` |
+| bb-knowledge-base | `{ interfaceEndpoints: [BEDROCK_RUNTIME] }` |
+| bb-email-client | `{ interfaceEndpoints: [SES] }` |
+| bb-app-setting | `{ interfaceEndpoints: [SSM] }` |
+| bb-realtime | `{ interfaceEndpoints: [APIGATEWAY] }` |
+| bb-auth-cognito | `{ interfaceEndpoints: [SSM] }` |
+| bb-auth-oidc | `{ interfaceEndpoints: [SSM] }` |
+| bb-distributed-data | `{ runtimeSubnet: 'private-with-egress' }` (DSQL over public HTTPS needs Lambda egress) |
+
+> CloudWatch Logs and SSM interface endpoints are always provisioned by
```
Simone319: Stale after the SSM change: 'CloudWatch Logs and SSM interface endpoints are always provisioned … regardless of BB declarations' (106–108) and the 'Always-Added Endpoints … SSM' list (113–115) are no longer accurate. SSM now flows from BB requirements; only CloudWatch Logs is always-on.

packages/blocks/VPC-DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+  defaults: { ...BlocksPresets.production, vpc: { network: sharedVpc, provisionEndpoints: false } },
+});
+```
+
+---
+
+## BB Endpoint Registration
+
+Each Building Block declares what it needs from the VPC by implementing
+`getVpcRequirements()`, which returns a plain `VpcRequirements` object. The
```
Simone319: These sections (70, 77, 90) still document the removed `abstract getVpcRequirements(): VpcRequirements;` API. It's been replaced by the per-stack requirements registry + constructor registration (`vpc-requirements-registry.ts`, `BuildingBlockScope` constructor). Please update to describe the registry model.

packages/blocks/VPC-DESIGN.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+---
+
+## Phased Implementation
+
+### Phase 1: CDK-level VPC (this PR)
+
+- `vpc` prop on `BlocksStack` / `BlocksBackend`
+- `getVpcRequirements()()` / `getVpcRequirements()()` on `Scope`
```
Simone319: Garbled text here: 'getVpcRequirements()() / getVpcRequirements()() on Scope'. Needs a rewrite to match the registry API.

packages/blocks/VPC.md
```diff
@@ -0,0 +1,105 @@
...
+await BlocksStack.create(app, stackName, {
+  backendHandlerPath: join(__dirname, 'index.handler.ts'),
+  backendCDKPath: join(__dirname, 'index.ts'),
+  defaults: { ...BlocksPresets.production, vpc: { network: vpc } },
+});
+```
+
+AWS Blocks handles:
+- **Lambda placement** in private subnets (configurable via `subnets`)
+- **VPC endpoint provisioning** based on which BBs are in scope (DynamoDB, S3, SSM, Secrets Manager, SQS, Bedrock, etc.)
```
Simone319, line 62: Neither VPC.md nor VPC-DESIGN.md mentions the always-on S3 gateway endpoint added in this PR (free, route-table only, used to pull config/secrets/migrations at cold start). Please add it to the endpoints section.

packages/core/src/cdk/vpc.ts
svidgen: Added to the `subnets` docstring and VPC.md.

packages/core/src/cdk/vpc-types.ts
svidgen: Renamed to `computeSubnets`/`computeSecurityGroup` now — internal-only, no API impact.

packages/blocks/VPC.md
svidgen: Fixed — SSM's from AppSetting now, not always-on. Noted the ~$97 case.

packages/blocks/VPC-DESIGN.md
svidgen: Fixed — only CloudWatch Logs is unconditional; SSM's requirement-driven.

packages/blocks/VPC-DESIGN.md
svidgen: Rewritten to the registry/constructor model.

packages/blocks/VPC-DESIGN.md
svidgen: Fixed.

packages/blocks/VPC.md
svidgen: Added to the always-on list in both docs.

pranavosu (dismissed):

Simone319 (dismissed):

Simone319 (approved):

pranavosu (approved):

PR #287, opened 2026-07-31
fix(core): drop redundant --hotswap flag from cdk watch in sandbox

osama-rizk:
## What

Remove the redundant `--hotswap` flag from the `cdk watch` invocation in `npm run sandbox` (`packages/core/src/scripts/sandbox.ts`).

`cdk watch` **already performs hotswap deployments by default** ([aws-cdk docs](https://docs.aws.amazon.com/cdk/v2/guide/cli.html#cli-deploy-watch)), so passing `--hotswap` explicitly is redundant. On some `aws-cdk` CLI versions it also emits a benign duplicate/unknown-option warning that surfaces to users as `[CDK Watch] ...` on stderr.

```diff
   const cdkWatch = spawnCommand("npx", [
-    "cdk", "watch", "--hotswap",
+    // `cdk watch` already hotswaps by default, so passing `--hotswap` is
+    // redundant and emits a duplicate-option warning on some aws-cdk versions.
+    "cdk", "watch",
     `--outputs-file`, `${outDir}/outputs.json`,
```

## Why

Cosmetic dev-experience fix — removes misleading warning output during `npm run sandbox`. No behavior change: hotswap remains the deploy mode for `cdk watch`.

Source: AWS Blocks bug-bash board — [board card](https://github.com/orgs/aws-amplify/projects/141/views/15?pane=issue&itemId=203197131) · [amplify-backend#3249](https://github.com/aws-amplify/amplify-backend/issues/3249) (P3 / DX).

## Testing

- `npm run build` (full monorepo) — passes
- `npm test` in `packages/core` — passes
- `biome lint --changed --since=origin/main` — clean on the changed file

## Changeset

Included (`@aws-blocks/core` patch).

### DIFF
--- a/packages/core/src/scripts/sandbox.ts
+++ b/packages/core/src/scripts/sandbox.ts
@@ -150,8 +150,6 @@ export async function startSandbox(options: SandboxOptions) {
   console.log(`\n   Open http://localhost:${clientPort}\n`);
 
   const cdkWatch = spawnCommand("npx", [
-    // `cdk watch` already hotswaps by default, so passing `--hotswap` is
-    // redundant and emits a duplicate-option warning on some aws-cdk versions.
     "cdk", "watch",
     `--outputs-file`, `${outDir}/outputs.json`,
     `--context`, `projectRoot=${process.cwd()}`,
--- a/packages/core/src/scripts/sandbox.ts
+++ b/packages/core/src/scripts/sandbox.ts
@@ -150,7 +150,9 @@ export async function startSandbox(options: SandboxOptions) {
   console.log(`\n   Open http://localhost:${clientPort}\n`);
 
   const cdkWatch = spawnCommand("npx", [
-    "cdk", "watch", "--hotswap",
+    // `cdk watch` already hotswaps by default, so passing `--hotswap` is
+    // redundant and emits a duplicate-option warning on some aws-cdk versions.
+    "cdk", "watch",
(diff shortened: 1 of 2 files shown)


soberm (dismissed):

packages/core/src/scripts/sandbox.ts
```diff
@@ -150,7 +150,9 @@ export async function startSandbox(options: SandboxOptions) {
   console.log(`\n   Open http://localhost:${clientPort}\n`);
 
   const cdkWatch = spawnCommand("npx", [
-    "cdk", "watch", "--hotswap",
+    // `cdk watch` already hotswaps by default, so passing `--hotswap` is
+    // redundant and emits a duplicate-option warning on some aws-cdk versions.
```
bobbor: this comment is confusing. it documents a change, not the code.
IMHO the comment should be removed as specifying the reason for the parameter-removal in the PR/changeset is enough.

bobbor (changes requested):

packages/core/src/scripts/sandbox.ts
osama-rizk: Good call — removed. The rationale is already captured in the PR description and the changeset, so the inline comment was redundant. Done in 0a5ae30.

sarayev (approved):

soberm (approved):

bobbor (approved):

PR #295, opened 2026-08-03
chore(deps): bump amplify packages and aws-cdk-lib to latest

sarayev:
## Summary

Bumps all amplify-related dependencies and `aws-cdk-lib` to their latest published versions, plus two follow-on pins that were required to keep the dependency tree self-consistent.

### Amplify packages (`test-apps/amplify-gen2/package.json`)

A scan of all 81 `package.json` files (including `overrides`/`resolutions`) found amplify dependencies in exactly one file:

| Package | Field | Before | After | Jump |
|---|---|---|---|---|
| `aws-amplify` | dependencies | `^6.17.0` | `^6.20.0` | minor |
| `@aws-amplify/backend` | devDependencies | `^1.22.0` | `^1.24.0` | minor |
| `@aws-amplify/backend-cli` | devDependencies | `^1.8.2` | `^1.9.0` | minor |

### aws-cdk-lib (47 entries across 46 files)

Every `aws-cdk-lib` declaration normalized to `^2.263.0` — root, all publishable packages' `peerDependencies`, the `create-blocks-app` templates, the native examples, and the test apps.

| Before | Count | After |
|---|---|---|
| `^2.257.0` | 26 | `^2.263.0` |
| `2.257.0` (exact) | 15 | `^2.263.0` |
| `2.260.0` (exact) | 1 | `^2.263.0` |
| `^2.245.0` | 1 | `^2.263.0` |
| `^2.234.1` | 1 | `^2.263.0` |
| `>=2.100.0` | 1 | `^2.263.0` |

### Follow-on pins (required — see below)

| Package | Before | After | Why |
|---|---|---|---|
| `aws-cdk` (CLI) | `2.1122.0` | `2.1135.0` | schema v54 support |
| `@aws-sdk/s3-request-presigner` | `3.1046.0` | `3.1101.0` | align with `client-s3` |

**No major version jumps.** Every bump stays within its current major.

## ⚠️ Please review: two things worth a decision

**1. 18 exact pins became caret ranges.** The requested target value was `^2.263.0`, so entries previously pinned exactly (`2.257.0`, `2.260.0`) and one open range (`>=2.100.0` in `bb-distributed-data`) are now `^2.263.0`. Those exact pins may have been deliberate for reproducibility. Happy to restore the original operators and use `2.263.0` where it was pinned before.

**2. `aws-cdk-lib` is a `peerDependency` of ~24 published packages.** Raising the floor to `^2.263.0` means consumers on older `aws-cdk-lib` will see peer warnings. If the intent is only to move the *dev/test* toolchain, the peer ranges should probably stay lower.

No changeset is included, so **CI's "Require changeset" check will fail** as-is. `amplify-gen2` is in the changeset `ignore` list, but the `aws-cdk-lib` peer-range edits touch publishable packages, which the coverage guard requires changesets for (plus an umbrella entry). I held off because the right bump type is a judgement call — a `patch` on each touched package plus `@aws-blocks/blocks` would satisfy the guard.

## Two real breakages found and fixed

Neither was cosmetic; both broke a clean build/test run.

**`getSignedUrl` type conflict.** The amplify bump floats `@aws-sdk/client-s3` to `3.1101.0` while `@aws-sdk/s3-request-presigner` stayed at `3.1046.0`. Each ships its own copy of the smithy middleware types, so passing an `S3Client` into `getSignedUrl` failed:

```
packages/bb-file-bucket/src/index.aws.ts(127,23): error TS2345:
  Argument of type 'S3Client' is not assignable to parameter of type
  'Client<any, ServiceInputTypes, MetadataBearer, any>'.
  Types have separate declarations of a private property 'handlers'.
```

Fixed by aligning the presigner to `3.1101.0`.

**CDK cloud assembly schema mismatch.** `aws-cdk-lib@2.263.0` emits schema `54.0.0`, but the CLI resolved at `2.1122.0` reads only up to `53.x.x`, so `cdk synth` failed and took out all 4 pipeline synth tests:

```
This CDK CLI is not compatible with the CDK library used by your application.
(Cloud assembly schema version mismatch: Maximum schema version supported is
53.x.x, but found 54.0.0. You need at least CLI version 2.1135.0...)
```

The declared range (`^2.1117.0`) already allowed `2.1135.0`; only the lockfile held it back, so `npm update aws-cdk` was enough.

## Verification

All checks run on Node 22 (`.nvmrc`), matching CI.

| Check | Result |
|---|---|
| `npm ci` (lockfile integrity, from scratch) | pass |
| `npm run clean && npm run build:packages` | pass — 0 TS errors |
| `npm run test:unit` (full suite) | pass — `# fail 0` |
| `test-apps/pipeline` synth tests | 16/16 pass |

> **Note on the build evidence:** an earlier run of this branch appeared green only because `tsc --build` reused cached `.tsbuildinfo` state. Every result above is from an explicit `npm run clean` first. On pristine `main` a clean build passes, and the two failures above reproduce reliably without these fixes — so they are genuinely introduced by the bumps, not pre-existing.

### Lockfile diff

`package-lock.json` shows a large diff, but it is safe:

- **0 downgrades** (verified programmatically across all 3,937 lock entries)
- **694 of 696 removals** are *nested* duplicates collapsing into existing top-level entries (npm dedup)
- Only 2 top-level removals (`@aws-cdk/cloud-assembly-api`, `at-least-node`)

### Pre-existing issues (not from this change)

Typecheck errors in `test-apps/comprehensive` and the unbuilt-workspace `@aws-blocks/blocks/cdk` resolution error were confirmed byte-identical on pristine `main` (matching md5), so they are unrelated.


sarayev:
## CI status update

`Require changeset` now passes. `SwiftFormat` still fails, and the evidence says it is **pre-existing on `main`, not caused by this PR**.

### 1. Changesets added

Added `.changeset/aws-cdk-lib-peer-range-2-263.md` with a **`patch`** entry for the 16 publishable packages the coverage guard flagged, plus `@aws-blocks/blocks`:

`bb-app-setting`, `bb-async-job`, `bb-auth-basic`, `bb-cron-job`, `bb-dashboard`, `bb-distributed-data`, `bb-distributed-table`, `bb-email-client`, `bb-file-bucket`, `bb-knowledge-base`, `bb-kv-store`, `bb-logger`, `bb-metrics`, `bb-realtime`, `bb-tracer`, `pipeline` — and `@aws-blocks/blocks`.

Notes on the choices:
- **16, not 24.** 23 publishable packages had `aws-cdk-lib` peer-range edits, but 7 (`auth-common`, `bb-agent`, `bb-auth-cognito`, `bb-auth-oidc`, `bb-data`, `core`, `hosting`) were already covered by changesets pending on this branch. The guard now reports *"All 24 changed package(s) are covered"*.
- **Umbrella entry included.** `@aws-blocks/blocks` re-exports 15 of the 16, and caret ranges keep sibling patches in range, so `changeset version` would leave the umbrella behind while its packed content moves (#212).
- **`patch` is correct** under the repo's pre-1.0 convention: only the declared peer range and resolved lockfile move — no API change. `create-blocks-app` templates need no entry (they publish as `blocks-app`, not `@aws-blocks/*`).

All four guards plus their unit tests pass locally:

```
✓ 12 changeset(s) are structurally valid.
✓ All 24 changed package(s) are covered by changesets.
✓ No major version bumps declared in changesets.
✓ No changeset here releases a package @aws-blocks/blocks re-exports.
# tests 31 / # pass 31 / # fail 0   (changeset-guard.test.mjs)
```

### 2. SwiftFormat verdict: pre-existing, NOT caused by this PR

- **This PR changes 0 `.swift` files.** All 9 flagged files are byte-identical to `origin/main` (verified with `git diff origin/main --quiet` per file).
- **Identical failure on an unrelated dependency-only PR.** The dependabot PR "bump next from 16.2.9 to 16.2.11" (2026-07-24, run `30123332053`) failed with the *same rule* and a **byte-identical list of 9 files**. Every `native-sdk-swift.yml` run since 2026-07-16 has failed.
- **Why a JS bump wakes a Swift check:** the workflow triggers on `paths: native/swift/**`. This PR edits `native/swift/Demo/typescript-demo/package.json` — an npm manifest that happens to live under that prefix — so the job runs and lints pre-existing Swift source.
- **Root cause:** every error is `(wrapIfStatementBodies)`, a rule `native/swift/.swiftformat` never mentions (neither `--enable` nor `--disable`), so it inherits the tool default. The workflow installs the formatter unpinned via `brew install swiftformat`, so a new upstream release that enables this rule by default turns the check red with no repo change.

**Not fixing it here, deliberately.** A fix would reformat 9 Swift files across `BlocksCodegen`, `BlocksRuntime`, the demo app, a plugin, and a test target — none related to a dependency bump, and it would bury the actual change. It also would not stop recurring.

Suggested separate fix (either one):
- add `--disable wrapIfStatementBodies` to `native/swift/.swiftformat` (matches how ~60 other rules are already explicitly managed), or
- pin the formatter version in the workflow and run the reformat as its own PR.

sarayev:
## SwiftFormat resolved; one new non-required failure to flag

### SwiftFormat: gone (no longer runs on this PR)

PR #296 (the `.swiftformat` fix) was **closed without merging**, so the fix is not on `main` and rebasing would not have helped. Instead I removed this PR's only `native/swift/**` touch: `native/swift/Demo/typescript-demo/package.json` is back to `aws-cdk-lib` `2.260.0`, byte-identical to `main`.

```diff
--- a/native/swift/Demo/typescript-demo/package.json
+++ b/native/swift/Demo/typescript-demo/package.json
-    "aws-cdk-lib": "^2.263.0",
+    "aws-cdk-lib": "2.260.0",
```

That drops the PR out of the `paths: native/swift/**` filter, so the SwiftFormat job no longer runs here — same as any other JS-only PR. **No Swift source was touched, and #296 was not merged.**

This costs nothing: that demo is a **standalone npm project with its own lockfile**, not a root workspace, so it was never part of the monorepo's resolution. Its `package.json` and its own `package-lock.json` are both back at `2.260.0` and mutually consistent. All 46 other `aws-cdk-lib` entries remain at `^2.263.0`.

Re-verified after the change: clean `npm run build:packages` passes with 0 TS errors, and all four changeset guards pass (*"All 24 changed package(s) are covered"*).

### `dependency-review-pr`: new failure, and it is not something this PR can fix

```
package-lock.json » brace-expansion@5.0.8 – DoS via unbounded intermediate arrays,
bypassing the CVE-2026-14257 mitigation (high severity)
GHSA-rgw5-rvv9-x895
```

Three findings worth your attention:

**1. The identical code passed this check three times yesterday.** Runs at 13:01, 13:41 and 13:54 on Aug 3 all succeeded. The advisory was published **2026-08-03T16:35Z** — after those runs. Nothing about the dependency graph changed; the vulnerability database did.

**2. This PR *reduces* the number of vulnerable copies.** Counting every `brace-expansion` in the tree against the advisory's ranges:

| | vulnerable copies |
|---|---|
| `main` | **9** |
| this PR | **7** |

The bump upgrades two copies to the patched `5.0.9` (`@aws-amplify/backend-deployer`, `@aws-amplify/sandbox`). It gets flagged only because dependency-review reports *newly introduced* versions — `aws-cdk-lib`'s bundled copy moves `5.0.5` → `5.0.8`, and both are vulnerable, but only the new one is reported. `main`'s 9 pre-existing copies (including `1.1.15` and three at `2.1.1`) are silently grandfathered.

**3. It cannot be fixed from here.** The flagged copy is `node_modules/aws-cdk-lib/node_modules/brace-expansion` with **`inBundle: true`** — shipped inside the `aws-cdk-lib` tarball. I tested `overrides: { "brace-expansion": "^5.0.9" }` and confirmed the bundled copy stays at `5.0.8`; npm cannot rewrite bundled dependencies. The patched `brace-expansion@5.0.9` was published 2026-07-30, one day before `aws-cdk-lib@2.263.0` (2026-07-31), which still bundles `5.0.8`. **The real fix must come from upstream `aws-cdk-lib`.** `2.263.0` is the latest release, so there is no newer version to move to.

`dependency-review-pr` is **not** in branch protection's required set (`Build, Unit Tests, E2E Local`, `E2E Amplify Interop (Required)`, `E2E Production`, `E2E Sandbox`, `Native SDK E2E (Required)`, `Require changeset`), so it does not block merge. I did not add a suppression, since silencing a live high-severity advisory is your call, not mine.

If you would rather not carry the finding at all, the option is to drop `aws-cdk-lib` from this PR and keep it amplify-only — happy to do that.

PR #298, opened 2026-08-04
fix(bb-auth-basic,bb-logger): report bbName/bbVersion in telemetry

ahmedhamouda78:
## Problem

`AuthBasic` and `Logger` are both listed in the umbrella's `aws-blocks.vendorize` map, so `scripts/generate-bb-names.mjs` has always generated them into `OFFICIAL_BB_NAMES`:

```ts
// packages/core/src/common/official-bb-names.generated.ts
'AuthBasic',
'Logger',
```

But neither passed `bbMeta` to `super()` — both called `super(id, { parent: scope })` — and `Scope` only records a block in its registry when `bbName` is set:

```ts
// packages/core/src/common/index.ts
if (options?.bbName) { ... Scope._bbRegistry.set(options.bbName, ...) }
```

So both entries were **inert**. `Scope.getRegisteredBlocks()` could never name either block, and `product.buildingBlocks` silently under-reported two of the most widely used blocks. Nothing failed — a block that omits `bbMeta` still constructs fine and every existing test still passes — which is why this went unnoticed.

## Fix

Each package now carries the standard prebuild and passes the generated constants through, matching `bb-auth-cognito` and the other 16 blocks:

```ts
super(id, { parent: scope, bbName: BB_NAME, bbVersion: BB_VERSION });
```

`BB_NAME`/`BB_VERSION` come from `src/version.ts`, regenerated on every build by `generate-version.mjs AuthBasic` / `Logger` — the name from the script argument, the version from the package's own `package.json`, so it tracks release bumps on its own. `version.ts` stays gitignored, as it is for every other block.

**Per-package shape:**

| Package | Files | Notes |
|---|---|---|
| `bb-auth-basic` | `package.json`, `src/index.ts` | Single-class block, so one `super()` call — mirrors `bb-auth-cognito`. |
| `bb-logger` | `package.json`, `src/index.aws.ts` | Has the aws/mock/cdk split, but `index.mock.ts` is `export * from './index.aws.js'`, so the one runtime change covers **both** the `default` and `aws-runtime` conditions. |

`bb-logger/src/index.cdk.ts` is **deliberately left alone**: no block passes `bbMeta` from its CDK entry point (0 of 18 checked), because telemetry is reported by the runtime class, not the synth-time construct. The browser `Logger` doesn't extend `Scope`, so it has nothing to report.

### Name consistency across the three coupled artifacts — verified

```
bb-auth-basic | vendorize: ["AuthBasic"] | BB_NAME: AuthBasic | in OFFICIAL_BB_NAMES: true | 0.1.4==0.1.4
bb-logger     | vendorize: ["Logger"]    | BB_NAME: Logger    | in OFFICIAL_BB_NAMES: true | 0.1.2==0.1.2
```

No `BB_NAME_OVERRIDES` entry needed — both class names already equal their runtime names (that map exists only for `AuthOIDC`→`AuthOidc`).

**No `@aws-blocks/core` bump:** both names were already in the vendorize map, so the generated `OFFICIAL_BB_NAMES` file is byte-identical. Confirmed — `verify-coverage` sees exactly 2 changed packages.

## Expected consequence, called out explicitly

`Logger` is composed internally by most other blocks (`new Logger(this, 'logger', { level: 'error' })`), so it will now be reported for effectively every application and `totalCount` rises accordingly. That is the intended reading of the existing `Logger` entry in the vendorize map, not a new disclosure: `getRegisteredBlocks()` still exposes only names already on the official list, and customer-chosen block names remain counted-but-unnamed. Flagging it since it changes telemetry shape/volume repo-wide.

## Versioning

Both bumps are `patch`, not `minor`. In this repo `minor` is the pre-1.0 **breaking** channel (`scripts/changeset-guard.ts`), and registering telemetry changes no public API and breaks no caller. `@aws-blocks/blocks` takes a `patch` because it re-exports both blocks and sibling releases stay inside its caret range (#273, #212).

## Tests

Telemetry-registration tests added to each package's existing `index.test.ts`, using real instances (no mocks):

- `BB_NAME` matches the vendorize / `OFFICIAL_BB_NAMES` spelling
- `BB_VERSION` equals the package version
- an instance carries `bbName`/`bbVersion`
- via public `Scope.getRegisteredBlocks()` — which only names a block whose `bbName` is in `OFFICIAL_BB_NAMES` — the block is reported as **official** rather than dropped into `customBlocksCount`

That last assertion is what ties the three artifacts together, so a future rename on any one side fails here. `bb-logger` additionally pins the mock/aws class identity and that `child()` doesn't register twice.

**Control runs** (proving the tests are load-bearing) — reverting the `bbMeta` and rebuilding:

```
bb-logger      → 3 of 6 fail  (57 tests: 54 pass / 3 fail)
bb-auth-basic  → 2 of 4 fail  (13 tests: 11 pass / 2 fail)
```

The remaining new tests correctly still pass — they pin the generated version module, independent of `super()`.

## Verification (local)

```
npm run build (full)                    BUILD_EXIT=0, 0 × "error TS"
npm test -w packages/bb-logger          57 / 57 pass / 0 fail
npm test -w packages/bb-auth-basic      13 / 13 pass / 0 fail
npm run test:unit (all workspaces)      UNIT_EXIT=0, every "# fail 0", 0 × "not ok"
npm run check:api                       EXIT=0 "All API reports are up-to-date" — no API.md drift
npm run check:exports-consistency       "All aws-blocks export maps are consistent ✓"
tsc --noEmit (test-apps/comprehensive)  TSC_EXIT=0
npm run test:e2e:local                  E2E_EXIT=0
    comprehensive  380 / 379 pass / 0 fail / 1 skipped
    vendorize        8 /   8 pass / 0 fail
cdk synth                               SYNTH_EXIT=0
changeset-guard.test.mjs                31 / 31 pass; all 4 guards ✓
```

`check:api` needed no regeneration: `bbName`/`bbVersion` are inherited readonly props already declared on `Scope`, and `BB_NAME`/`BB_VERSION` aren't re-exported, so the public surface is unchanged.

Found while adding the same telemetry parity to `AuthBearerJwt` in #297. Independent of that PR — this branches off `main` and touches different packages.


osama-rizk (approved):

Simone319:
Look like bb-metric and bb-tracer are also vendorized blocks with no telemetry. Any plan to add them as well? Including any other missing telemetry bb?

Simone319 (approved):

osama-rizk:
Great catch — thank you @Simone319 , you're right, and it's broader than those two. 

I audited every entry in the umbrella's `vendorize` map against whether its runtime class actually passes `bbMeta`. Three vendorized blocks are still inert on `main`:

- `bb-metrics` (`Metrics`)
- `bb-tracer` (`Tracer`)
- `bb-distributed-data` (`DistributedDatabase`)

One deliberate exclusion: `bb-dashboard` (`Dashboard`) is vendorized as well, but its runtime class (`index.aws.ts` / `index.mock.ts`) doesn't extend `Scope` — only the synth-time CDK construct does, and CDK entry points don't report telemetry. So it has nothing to register, same as the browser `Logger` here.

I'll keep this PR scoped to `AuthBasic`/`Logger` and address the other three in a dedicated follow-up (same wiring: `prebuild` → `BB_NAME`/`BB_VERSION` → `super()`, plus registration tests). I'll link it here shortly.

osama-rizk:
Follow-up raised: #313 — adds telemetry registration for `Metrics`, `Tracer`, and `DistributedDatabase` (and documents why `Dashboard` is excluded).

PR #302, opened 2026-08-04
feat(core): stack-wide BlocksDefaults + presets (Infrastructure Options, step 1); adopt in bb-kv-store

osama-rizk:
## Summary

Introduces a single, explicit `defaults` object passed to `BlocksStack.create()` / `BlocksBackend.create()`, chosen once at the app entry point and read by every stateful Building Block. It replaces the per-block `sandboxMode` logic and the `RemovalPolicies` / `SandboxDisableDeletionProtection` mixin/aspect dance for the removal-policy / deletion-protection knobs.

The remaining infra knobs (log retention, API throttling, access logging, PITR, secrets backend, VPC, compute) fit the same `defaults` object and land in follow-up PRs, each with its own API review.

> Supersedes this PR's original experimental `hardening` prop.

## Approach

Two-tier resolution, highest precedence first:

```
per-block option  >  scope.defaults
```

i.e. `options.field ?? scope.defaults.field` — no hidden fallback below the stack default; the app always picks a posture explicitly, starting from a preset:

```ts
await BlocksStack.create(app, stackName, {
  backendHandlerPath: join(__dirname, 'index.handler.ts'),
  backendCDKPath: join(__dirname, 'index.ts'),
  defaults: sandboxMode ? BlocksPresets.sandbox : BlocksPresets.production,
});

// override a single field
defaults: { ...BlocksPresets.production, deletionProtection: false }
```

`defaults` is **anchored on the owning `BlocksStack`/`BlocksBackend`** and resolved by walking up the construct tree (the same way `handler` and `executionRole` resolve), so several backends in one stack each keep their own posture without clobbering each other.

## Changes

**`@aws-blocks/core` (minor):**
- New `blocks-defaults.ts`: the `BlocksDefaults` interface (`removalPolicy: RemovalPolicy`, `deletionProtection: boolean`) and `BlocksPresets.sandbox` / `BlocksPresets.production`.
- `Scope` gains a `defaults` getter (tree-walk to the owning backend); `BlocksStack`/`BlocksBackend` expose `defaults` and take it as a **required** prop.
- Removed the experimental `hardening` prop and its `resolve*` helpers.

**`@aws-blocks/blocks` (minor):** re-exports the new `/cdk` surface.

**Building Block adopters (patch):** `bb-kv-store`, `bb-data`, `bb-distributed-data`, `bb-distributed-table`, and `bb-knowledge-base` now take their removal policy + deletion protection from `defaults` instead of reading the `sandboxMode` context themselves. `bb-kv-store` also exposes `deletionProtection?` as a per-block override (matching `removalPolicy`). `bb-distributed-table` reads `defaults` directly for now — a richer per-block `protection` override lands with #282.

**Templates + test-apps:** every `create()` call site now passes `defaults`; the `RemovalPolicies`/`SandboxDisableDeletionProtection` mixin lines are gone. E2E apps set `defaults: BlocksPresets.sandbox` (disposable), customer templates use `sandboxMode ? sandbox : production`.

## Breaking change

`BlocksStack.create` / `BlocksBackend.create` now **require** a `defaults` field. Migration is one line: pass `BlocksPresets.sandbox` / `BlocksPresets.production` (typically the `sandboxMode ? …` ternary).

## Testing

- Core: `BlocksPresets` posture; backend-anchored resolution (two backends in one stack keep separate defaults; a nested block resolves its owning backend's defaults via the tree-walk).
- Adopters: CDK synth tests that each BB's table/cluster/bucket takes sandbox (DESTROY + protection off) vs production (RETAIN + protection on) from `defaults`, and that per-block options override.
- **Teardown verified via synth:** synthesizing `test-apps/comprehensive` with `defaults: BlocksPresets.sandbox` yields **DeletionPolicy=Delete + DeletionProtection=false** across all 29 DynamoDB tables, the RDS cluster, the DSQL cluster, and all 11 S3 buckets — i.e. the whole stack is disposable through the mechanism, no mixin. (Deploy+destroy is exercised by the E2E Sandbox/Production CI jobs.)
- `npm run build` · `npm run lint` · `npm test` (0 failures) · `npm run lint:deps` all green; API reports regenerated.

## Open questions for reviewers

- **Version bump** — a required prop is a hard compile break; kept at `minor` to match convention but flagged Breaking. Bump to `major` if the policy is semver-strict.
- **`removalPolicy` + `deletionProtection` as two fields** can encode a contradiction; #282 merged the equivalent DDB knobs into one `protection` enum. Worth deciding whether `defaults` should follow suit before more blocks adopt.

## Related (secure-defaults bug bash)
- #163 / #326 — core REST API throttling + access logging (future `defaults` adopters)
- #282 — bb-distributed-table durability (`protection`, PITR, encryption)
- #304 — bb-file-bucket enforceSSL + access logging

## Follow-ups (not in this PR)
`defaults.logRetention` · `defaults.throttling` (re-adopting bb-realtime's WS throttle/access-logs) · `defaults.accessLogging` · `defaults.pointInTimeRecovery` · `defaults.secretsStore` · `vpc` · `compute`; a shared `scope.resolveDefault(key, override)` helper once the per-property fallback repeats; reconciling bb-distributed-table's `protection` (#282) with `defaults`.

### DIFF
--- a/.changeset/blocks-infrastructure-defaults.md
+++ b/.changeset/blocks-infrastructure-defaults.md
@@ -1,0 +1,20 @@
+---
+"@aws-blocks/core": minor
+"@aws-blocks/blocks": minor
+"@aws-blocks/bb-kv-store": patch
+"@aws-blocks/bb-data": patch
+"@aws-blocks/bb-distributed-data": patch
+"@aws-blocks/bb-distributed-table": patch
+"@aws-blocks/bb-knowledge-base": patch
+"@aws-blocks/create-blocks-app": patch
+---
+
+Add infrastructure `defaults` chosen once at the app entry point, replacing the per-block `sandboxMode` logic and the `RemovalPolicies`/`SandboxDisableDeletionProtection` mixin dance for removal-policy and deletion-protection.
+
+`@aws-blocks/core/cdk` now exports `BlocksDefaults` and the `BlocksPresets.sandbox` / `BlocksPresets.production` starting points. `BlocksStack.create` / `BlocksBackend.create` take a required `defaults` prop; start from a preset and override individual fields with a spread. `defaults` is anchored on the owning `BlocksStack`/`BlocksBackend` (resolved by walking up the construct tree, like `handler`/`executionRole`), so multiple backends in one stack each keep their own posture. Building Blocks read the resolved values via `scope.defaults`, and a per-block option always wins (`option ?? scope.defaults.field`).
+
+Adopted across the stateful Building Blocks: `bb-kv-store`, `bb-data`, `bb-distributed-data`, `bb-distributed-table`, and `bb-knowledge-base` now take their removal policy and deletion protection from `defaults` instead of reading the `sandboxMode` context themselves. (`bb-distributed-table` reads `defaults` directly for now; a richer per-block `protection` override lands with #282.)
+
+The `create-blocks-app` scaffolding templates are updated to pass `defaults: sandboxMode ? BlocksPresets.sandbox : BlocksPresets.production` (replacing the `RemovalPolicies`/`SandboxDisableDeletionProtection` mixin), so newly-generated apps satisfy the required prop.
+
+**Breaking:** `BlocksStack.create` / `BlocksBackend.create` now require a `defaults` field — pass `BlocksPresets.sandbox` or `BlocksPresets.production` (typically `sandboxMode ? BlocksPresets.sandbox : BlocksPresets.production`). The previously-shipped experimental `hardening` prop and its `resolve*` helpers are removed; log-retention, API throttling, access-logging and point-in-time-recovery move into `defaults` in follow-up, per-feature changes.
--- a/packages/bb-distributed-data/src/index.cdk.ts
+++ b/packages/bb-distributed-data/src/index.cdk.ts
@@ -24,19 +24,26 @@ export class DistributedDatabase extends Scope {
     super(id, { parent: scope });
 
     const stack = cdk.Stack.of(this);
-    const isSandbox = stack.node.tryGetContext('sandboxMode') === 'true';
     const envName = this.fullId.replace(ENV_SANITIZE, '_');
     const region = stack.region;
     const dbRole = sanitizeDbRoleName(this.fullId);
 
-    // DSQL Cluster — respect explicit removalPolicy; default to DESTROY in sandbox, RETAIN otherwise.
-    const shouldDestroy = options?.removalPolicy === 'destroy' || (!options?.removalPolicy && isSandbox);
-    const removalPolicy = shouldDestroy ? cdk.RemovalPolicy.DESTROY : cdk.RemovalPolicy.RETAIN;
+    // Removal policy and deletion protection are resolved independently from the
+    // stack-wide `defaults` (per-block `removalPolicy` option wins for that field).
+    // Reading `defaults.deletionProtection` directly — rather than deriving it
+    // from `removalPolicy` — keeps every adopting block consistent: the same
+    // `defaults` object yields the same posture no matter which block reads it.
+    const removalPolicy =
+      options?.removalPolicy === 'destroy'
+        ? cdk.RemovalPolicy.DESTROY
+        : options?.removalPolicy === 'retain'
+          ? cdk.RemovalPolicy.RETAIN
+          : this.defaults.removalPolicy;
 
     const cluster = new cdk.CfnResource(stack, `${this.fullId}DsqlCluster`, {
       type: 'AWS::DSQL::Cluster',
       properties: {
-        DeletionProtectionEnabled: removalPolicy !== cdk.RemovalPolicy.DESTROY,
+        DeletionProtectionEnabled: this.defaults.deletionProtection,
       },
     });
 
--- a/packages/bb-kv-store/src/index.cdk.ts
+++ b/packages/bb-kv-store/src/index.cdk.ts
@@ -32,19 +32,24 @@ export class KVStore extends Scope {
 			// and grant the runtime Lambda read/write access.
 			this.table = Table.fromTableName(this, 'table', options.table.tableName);
 		} else {
+			// Resolve durability from the per-block option (a `'destroy'|'retain'`
+			// string, normalized to a CDK RemovalPolicy) falling back to the
+			// stack-wide `defaults`. The stack `defaults` replace the old
+			// `RemovalPolicies.of(stack).destroy()` + `SandboxDisableDeletionProtection`
+			// mixin dance — the sandbox posture now flows in through the chosen preset.
+			const removalPolicy =
+				options?.removalPolicy === 'destroy'
+					? RemovalPolicy.DESTROY
+					: options?.removalPolicy === 'retain'
+						? RemovalPolicy.RETAIN
+						: this.defaults.removalPolicy;
+
 			this.table = new Table(this, 'table', {
 				tableName: this.fullId.substring(0, 255),
 				partitionKey: { name: 'pk', type: AttributeType.STRING },
 				billingMode: BillingMode.PAY_PER_REQUEST,
-				// Default: CDK's RETAIN (undefined here). Customers opt into teardown
-				// via `{ removalPolicy: 'destroy' }`. Templates that apply
-				// `RemovalPolicies.of(stack).destroy()` under `sandboxMode` will
-				// override this at the stack layer regardless.
-				removalPolicy: options?.removalPolicy === 'destroy'
-					? RemovalPolicy.DESTROY
-					: options?.removalPolicy === 'retain'
-						? RemovalPolicy.RETAIN
-						: undefined,
+				removalPolicy,
+				deletionProtection: options?.deletionProtection ?? this.defaults.deletionProtection,
 				// Opt-in: enabling TTL on an already-deployed table is a live table
 				// update, so it must never happen implicitly.
 				timeToLiveAttribute: options?.ttl ? TTL_ATTRIBUTE : undefined,
--- a/packages/core/src/cdk/blocks-backend.ts
+++ b/packages/core/src/cdk/blocks-backend.ts
@@ -12,6 +12,7 @@ import { DEFAULT_NODE_RUNTIME } from './node-version.js';
 import { blocksNodejsBundling } from './bundling.js';
 import { addBlocksStackMetadata } from './stack-metadata.js';
 import { finalizeConfigRegistry, registerConfig } from './config-registry.js';
+import type { BlocksDefaults } from './blocks-defaults.js';
 import { BLOCKS_NAMESPACE, BLOCKS_RPC_PREFIX } from '../constants.js';
 import { registerBuiltinRoutes } from '../builtin-routes.js';
 
@@ -46,10 +47,29 @@ export function assertCdkConditionActive(): void {
 export interface BlocksBackendProps {
   backendHandlerPath: string;
   backendCDKPath: string;
+  /**
+   * Stack-wide infrastructure defaults applied to every Building Block (removal
+   * policy, deletion protection, …). See {@link BlocksDefaults}. Start from
+   * `BlocksPresets.sandbox` or `BlocksPresets.production` and override
+   * individual fields as needed. A per-block option always wins over the
+   * corresponding stack default.
+   */
+  defaults: BlocksDefaults;
 }
 
 /** Shared infra setup — creates Lambda + API Gateway on the given scope. */
 export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id?: string) {
+  // Fail fast with an actionable message at the create() call site if `defaults`
+  // is missing (e.g. a plain-JS caller, `as any`, or a dynamically-built props
+  // object) — otherwise the first Building Block to read `scope.defaults` throws
+  // a cryptic `Cannot read properties of undefined (reading 'removalPolicy')`.
+  if (!props.defaults) {
+    throw new Error(
+      'BlocksStack/BlocksBackend requires a `defaults` field. Pass a posture from ' +
+      '`@aws-blocks/core/cdk` — typically `defaults: sandboxMode ? BlocksPresets.sandbox : BlocksPresets.production`.',
+    );
+  }
+
   // ── Shared execution role ──────────────────────────────────────────────
   // A single IAM role that every Building Block grants to. Provisioned here so
   // it exists before the backend module is imported (Building Blocks reach it
@@ -203,6 +223,8 @@ export class BlocksBackend extends Construct {
   public readonly backendHandlerPath: string;
   /** Shared IAM role assumed by all Blocks compute. Building Blocks grant to this role. */
   public readonly executionRole: iam.IRole;
+  /** Infrastructure defaults for Building Blocks created under this backend. */
+  public readonly defaults: BlocksDefaults;
 
   /**
    * The fullId used by child Scopes to compute their env var names,
@@ -242,6 +264,11 @@ export class BlocksBackend extends Construct {
 
     // Expose self to Building Blocks at CDK time
     (globalThis as any).CURRENT_BLOCKS_STACK = this;
+
+    // Store defaults on the backend (not the stack) so several BlocksBackends
+    // in one stack each keep their own posture; Building Blocks resolve them by
+    // walking up to their owning backend (see Scope.defaults).
+    this.defaults = props.defaults;
 
     const infra = setupBlocksInfra(this, props, id);
     this.handler = infra.handler;
--- a/packages/core/src/cdk/blocks-defaults.ts
+++ b/packages/core/src/cdk/blocks-defaults.ts
@@ -1,0 +1,49 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { RemovalPolicy } from 'aws-cdk-lib';
+
+/**
+ * Default values for every Amazon-authored Building Block created within a
+ * `BlocksStack` or `BlocksBackend`. A per-block option always overrides the
+ * corresponding default (`option ?? scope.defaults.field`).
+ *
+ * Start from {@link BlocksPresets} and override individual fields as needed.
+ */
+export interface BlocksDefaults {
+	/**
+	 * What happens to a stateful resource (table, bucket, user pool) when the
+	 * stack is deleted, via CDK/CloudFormation: `RETAIN` keeps the resource (and
+	 * its data) behind; `DESTROY` tears it down. Applies only to CDK/CloudFormation
+	 * deletions.
+	 */
+	removalPolicy: RemovalPolicy;
+
+	/**
+	 * Whether to block deletion of a stateful resource. Unlike `removalPolicy`,
+	 * this guards against deletion through **any** path — CDK/CloudFormation as
+	 * well as the CLI, API, and AWS console — until it is turned off.
+	 */
+	deletionProtection: boolean;
+}
+
+/**
+ * Prepared starting points for {@link BlocksDefaults}. Pick one and override
+ * individual fields with a spread:
+ *
+ * ```ts
+ * defaults: { ...BlocksPresets.production, deletionProtection: false }
+ * ```
+ */
+export const BlocksPresets = {
+	/** Disposable development stacks: tear down cleanly, no delete guard. */
+	sandbox: {
+		removalPolicy: RemovalPolicy.DESTROY,
+		deletionProtection: false,
+	},
+	/** Durable, protected posture for permanent deployments. */
+	production: {
+		removalPolicy: RemovalPolicy.RETAIN,
+		deletionProtection: true,
+	},
+} satisfies Record<string, BlocksDefaults>;
--- a/packages/core/src/cdk/index.ts
+++ b/packages/core/src/cdk/index.ts
@@ -15,12 +15,17 @@ import {
 import { setupBlocksInfra, BlocksBackend, assertCdkConditionActive } from './blocks-backend.js';
 import { addBlocksStackMetadata } from './stack-metadata.js';
 import { finalizeConfigRegistry } from './config-registry.js';
+import { type BlocksDefaults, BlocksPresets } from './blocks-defaults.js';
 
 export { BlocksBackend, type BlocksBackendProps } from './blocks-backend.js';
 export { DEFAULT_NODE_RUNTIME } from './node-version.js';
 export { blocksNodejsBundling } from './bundling.js';
 export { SandboxDisableDeletionProtection } from './mixins.js';
 export { registerConfig, finalizeConfigRegistry } from './config-registry.js';
+export {
+  type BlocksDefaults,
+  BlocksPresets,
+} from './blocks-defaults.js';
 export { synthGuard } from './synth-guard.js';
 export type { ScopeOptions } from '../index.js';
 export { ApiError, isBlocksError, hasAuthError, DEFAULT_API_ERROR_NAME } from '../errors.js';
@@ -33,11 +38,14 @@ export class BlocksStack extends cdk.Stack implements BaseBlocksStack {
   public readonly backendHandlerPath: string;
   /** Shared IAM role assumed by all Blocks compute. Building Blocks grant to this role. */
   public readonly executionRole: cdk.aws_iam.IRole;
+  /** Infrastructure defaults for Building Blocks created under this stack. */
+  public readonly defaults: BlocksDefaults;
 
   private constructor(scope: Construct, id: string, props: BlocksStackProps) {
     super(scope, id, props);
     this.id = id;
     this.backendHandlerPath = props.backendHandlerPath;
+    this.defaults = props.defaults;
 
     // Set globalThis so Building Blocks attach directly to this stack
     (globalThis as any).CURRENT_BLOCKS_STACK = this;
@@ -131,6 +139,41 @@ export class Scope extends Construct {
 
   get fullId(): string {
     return computeScopeFullId(this);
+  }
+
+  /**
+   * The stack-wide infrastructure {@link BlocksDefaults} registered by
+   * `BlocksStack.create` / `BlocksBackend.create`. Read these in a Building
+   * Block's CDK constructor to resolve a durability value, letting a per-block
+   * option override:
+   *
+   * ```ts
+   * const removalPolicy = options?.removalPolicy ?? this.defaults.removalPolicy;
+   * ```
+   */
+  get defaults(): BlocksDefaults {
+    // Resolve the same way as handler/executionRole: walk up to the owning
+    // BlocksStack/BlocksBackend and read its defaults, so several backends in
+    // one stack each keep their own posture. Falls back to the ambient stack,
+    // then to the production preset when none was registered.
+    let current: Construct = this;
+    while (current.node.scope) {
+        current = current.node.scope as Construct;
+        if (current instanceof BlocksStack || current instanceof BlocksBackend) {
+            return current.defaults;
+        }
+    }
+    const ambient = ((globalThis as any).CURRENT_BLOCKS_STACK as { defaults?: BlocksDefaults } | undefined)?.defaults;
+    if (ambient) return ambient;
+    // No owning BlocksStack/BlocksBackend in the tree and none ambient — this is
+    // usually a deliberate test stub, but could be a real misconfiguration (a
+    // block built outside any Blocks backend). Fall back to the safe production
+    // posture, and log so it's debuggable if it fires unexpectedly.
+    console.warn(
+      `[Blocks] Scope "${this.id}" resolved infrastructure defaults with no owning ` +
+      'BlocksStack/BlocksBackend in scope; falling back to BlocksPresets.production.',
+    );
+    return BlocksPresets.production;
   }
 
   protected buildUserAgentChain(): [string, string][] {
--- a/packages/core/src/common/index.ts
+++ b/packages/core/src/common/index.ts
@@ -3,6 +3,7 @@
 
 import type { StackProps } from 'aws-cdk-lib';
 import type { Construct } from 'constructs';
+import type { BlocksDefaults } from '../cdk/blocks-defaults.js';
 import { CORE_VERSION } from '../version.js';
 import { OFFICIAL_BB_NAMES } from './official-bb-names.generated.js';
 export { OFFICIAL_BB_NAMES } from './official-bb-names.generated.js';
@@ -325,6 +326,14 @@ export function computeScopeFullId(scope: { id: string; parent?: any }) {
 export interface BlocksStackProps extends StackProps {
   backendHandlerPath: string;
   backendCDKPath: string;
+  /**
+   * Stack-wide infrastructure defaults applied to every Building Block (removal
+   * policy, deletion protection, …). Start from `BlocksPresets.sandbox` or
+   * `BlocksPresets.production` and override individual fields as needed. Any
+   * field a block also exposes as a per-block option is overridden by that
+   * option. See `BlocksDefaults` in `@aws-blocks/core/cdk`.
+   */
+  defaults: BlocksDefaults;
 }
 
 export class BlocksStack {
--- a/packages/create-blocks-app/templates/auth-cognito/aws-blocks/index.cdk.ts
+++ b/packages/create-blocks-app/templates/auth-cognito/aws-blocks/index.cdk.ts
@@ -1,7 +1,6 @@
 import * as cdk from 'aws-cdk-lib';
-import { RemovalPolicies, Mixins } from 'aws-cdk-lib';
 
-import { Hosting, BlocksStack, SandboxDisableDeletionProtection } from '@aws-blocks/blocks/cdk';
+import { Hosting, BlocksStack, BlocksPresets } from '@aws-blocks/blocks/cdk';
 import { fileURLToPath } from 'node:url';
 import { dirname, join } from 'node:path';
 import { getStackName } from '@aws-blocks/blocks/scripts';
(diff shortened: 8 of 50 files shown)


packages/bb-realtime/src/index.cdk.ts
```diff
@@ -91,12 +92,42 @@ function getOrCreateSharedInfra(stack: cdk.Stack, handler: cdk.aws_lambda.IFunct
...
+				requestId: '$context.requestId',
+				requestTime: '$context.requestTime',
+				eventType: '$context.eventType',
+				routeKey: '$context.routeKey',
+				connectionId: '$context.connectionId',
+				status: '$context.status',
+				ip: '$context.identity.sourceIp',
+				integrationError: '$context.integrationErrorMessage',
+			}),
+		};
```
soberm: [major] Worth confirming before this merges: access logging here is wired by setting `accessLogSettings` straight on the raw `CfnStage`, but nothing in this PR (or anywhere else in the repo, near as I can tell — grepped for `CfnAccount`, `cloudWatchRole`, `AWS::ApiGateway::Account`, and any log-group resource policy, all zero hits) grants API Gateway permission to actually write to CloudWatch. A first deploy in an account that's never set the account-level CloudWatch Logs role could fail right at `CreateStage` (the classic "CloudWatch Logs role ARN must be set in account settings to enable logging" error), or in some cases just silently produce no logs. The synth tests here only check template shape, so neither failure mode would show up in CI.

I'm not fully confident this actually bites, since your deploy accounts might already have the role from other API Gateway usage, and PR #163 has this exact same gap for the central REST API's access logging. But since this is bb-realtime's first time turning on access logs, I'd deploy to a clean sandbox and confirm logs actually land before merging. If it does need a fix, it probably belongs in `setupBlocksInfra` (a resource policy on the log group, or the account role, provisioned once) rather than per-adopter — every future BB that wants access logs will hit the same wall otherwise.

Smaller thing on the same block: the comment above ("API Gateway v2 has no L2 helper for stage access logs") isn't quite accurate for the aws-cdk-lib version this repo's pinned to (2.257.0) — `WebSocketStageProps` takes `accessLogSettings` directly (`{ destination: new LogGroupLogDestination(logGroup), format }`), so you could skip the `stage.node.defaultChild as CfnStage` reach-around. Produces the same template either way, just a cleaner path.

packages/bb-realtime/src/index.cdk.ts
```diff
@@ -91,12 +92,42 @@ function getOrCreateSharedInfra(stack: cdk.Stack, handler: cdk.aws_lambda.IFunct
...
+	// Apply the stack-wide hardening defaults (overridable per stack via the
+	// `hardening` prop on BlocksStack/BlocksBackend). The WebSocket stage was
+	// previously created with no throttling and no access logs — a single
+	// client could drive unbounded message volume with zero observability.
+	const throttle = resolveApiThrottle(stack);
 	const stage = new WebSocketStage(stack, 'BlocksRtStage', {
 		webSocketApi: wsApi,
 		stageName: 'rt',
 		autoDeploy: true,
+		throttle: { rateLimit: throttle.rateLimit, burstLimit: throttle.burstLimit },
```
soberm: [minor] This 100/200 default traces back to #163, which sized it for the central REST API (100 req/s). Applied here to a WebSocket stage's `DefaultRouteSettings`, it's throttling messages/sec across the whole connection instead, which is a different unit and might be tight for something chatty like live cursors or presence with more than a few concurrent users. Not saying the number's wrong, just seems worth a deliberate check rather than inheriting it by default. It's overridable via `hardening.apiThrottle` either way — might be worth a callout in DESIGN.md so people hitting the ceiling know the knob exists.

soberm:
[minor] `packages/bb-realtime/DESIGN.md` — the "Shared Per Stack" table under "Infrastructure (CDK)" still only lists the Connections Table, WebSocket API, and WebSocket Stage. Seems worth adding a row for the new `BlocksRtAccessLogs` log group and a note that the stage now applies stack-wide throttle/access-log/retention defaults via `@aws-blocks/core/cdk`'s `resolve*` helpers (overridable through the `hardening` prop). The "Before you open a PR" checklist calls for README/DESIGN/JSDoc updates to land in the same PR as the infra change, so this one looks like it just slipped through.

packages/core/src/common/index.ts
```diff
@@ -325,6 +325,14 @@ export function computeScopeFullId(scope: { id: string; parent?: any }) {
...
   backendHandlerPath: string;
   backendCDKPath: string;
+  /**
+   * Stack-wide infrastructure-hardening defaults applied to every Building
+   * Block (log retention, API throttling, DynamoDB point-in-time recovery, …).
+   * Any field a block also exposes as a per-block option is overridden by that
+   * option. Omit to accept the framework's secure defaults. See
+   * `HardeningDefaults` in `@aws-blocks/core/cdk`.
+   */
+  hardening?: import('../cdk/hardening-defaults.js').HardeningDefaults;
```
svidgen: Instead of calling this "hardening", can we call just call it `options` and have it handle the differences between sandbox and production deployments as well?

```ts
new BlocksStack(scope, id, {
  options: BlocksBackendOptions.DeriveFromDeploymentType
}
```

or

```ts
new BlocksStack(scope, id, {
  options: BlocksBackendOptions.Sandbox
}
```

or 

```ts
new BlocksStack(scope, id, {
  options: BlocksBackendOptions.Production
}
```

Something like that.

Then on hover customers can see the options and docstrings indicating when to use what.

What do you think? cc @pranavosu

svidgen (changes requested):

osama-rizk:
**Related secure-defaults work (bug bash)**

This PR is the cross-block *mechanism*; several sibling PRs harden individual blocks and should adopt it as follow-up adopters rather than each hand-rolling its own knob (the goal being one consistent surface + no redundant work):

- #163 (core REST API) — the **original** throttling + access-logging + log-group work; this PR's 100/200 default and 30d/`DESTROY` log-group convention trace back to it. Its throttle/access-logs/retention are exactly `resolveApiThrottle` / `resolveApiAccessLogs` / `resolveLogRetention`.
- #326 (core REST API) — also throttles the core REST API (200/400). **Overlaps directly with #163** — the two are near-duplicates; we should pick one and route it through `resolveApiThrottle` so there's a single default.
- #282 (bb-distributed-table) — DDB secure-by-default: PITR, `protection`, encryption. Intersects on **PITR / sandbox-gating** → will route through `resolvePointInTimeRecovery`.
- #304 (bb-file-bucket) — S3 enforceSSL + access logging. Intersects on **log retention** (`logRetentionDays` 90d vs this PR's `resolveLogRetention` 30d) → retention knob should converge here.

Proposed order: land this mechanism first, then each block adopts the shared resolvers so throttling / access-logs / retention / PITR are configured and named one way across the framework.

.changeset/blocks-infrastructure-defaults.md
```diff
@@ -0,0 +1,10 @@
+---
+"@aws-blocks/core": minor
```
svidgen, line 2: `@aws-blocks/blocks` would also need to be "minor".

packages/bb-kv-store/src/index.cdk.ts
```diff
@@ -32,19 +32,24 @@ export class KVStore extends Scope {
...
-				// via `{ removalPolicy: 'destroy' }`. Templates that apply
-				// `RemovalPolicies.of(stack).destroy()` under `sandboxMode` will
-				// override this at the stack layer regardless.
-				removalPolicy: options?.removalPolicy === 'destroy'
-					? RemovalPolicy.DESTROY
-					: options?.removalPolicy === 'retain'
-						? RemovalPolicy.RETAIN
-						: undefined,
+				removalPolicy,
+				deletionProtection: this.defaults.deletionProtection,
```
svidgen: Anywhere we consume a value from `defaults`, it should arguably be visible and overridable on the BB itself. Hence, we should add `deleteProtection` to `options`/`KVStoreOptions` and handle the override chain similarly to `removalPolicy`.

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -0,0 +1,108 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import * as cdk from 'aws-cdk-lib';
+import { RemovalPolicy } from 'aws-cdk-lib';
+import type { Construct } from 'constructs';
+
+const DEFAULTS_KEY = Symbol.for('BLOCKS_INFRA_DEFAULTS');
+
+/**
+ * Stack-wide infrastructure defaults, chosen once at the app entry point and
```
svidgen: "Stack-wide" isn't quite the right scope. If a customer uses `BlocksBackend.create()`, for example, "stake-wide" is wrong. Maybe change the sentence order to clarify. Something like:

> Default values for every Amazon-authored Building Block created within the BlocksStack or BlocksBackend.

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -0,0 +1,108 @@
...
+ * read by every Amazon-authored Building Block at construction time.
+ *
+ * Building Blocks provision AWS resources (DynamoDB tables, S3 buckets, Cognito
+ * user pools, …) that each need the same class of durability decision — what
+ * happens to stateful data on stack delete, and whether accidental deletion is
+ * blocked. Historically each block re-derived this from the `sandboxMode`
+ * context on its own, and the stack-level `RemovalPolicies`/mixin aspects fought
+ * with those per-block choices. This replaces both with a single, explicit
+ * object the app passes to `BlocksStack.create` / `BlocksBackend.create`.
```
svidgen: Too much context for a customer-facing docstring. (This docstring is customer facing, I assume.) It makes sense to have *separate* documentation somewhere (decision log perhaps) explaining the why. In the docs that user/agent will more likely read, keep it short, relevant, and instructive.

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -0,0 +1,108 @@
...
+ * posture explicitly (start from {@link BlocksPresets}), and a block that needs
+ * something different overrides locally.
+ *
+ * @remarks
+ * This is the first step of the Blocks Infrastructure Options design. It covers
+ * the removal-policy / deletion-protection knobs that were previously managed
+ * via CDK mixins/aspects and per-block `sandboxMode` logic. Further knobs
+ * (log retention, API throttling, access logging, point-in-time recovery,
+ * secrets backend, VPC, compute) are added to this object in follow-up,
+ * per-feature changes, each with its own API review.
```
svidgen: Again, if this is customer-facing, this is too much. This belongs more in the PR description and/or another doc in the repo. (If anywhere.)

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -0,0 +1,108 @@
...
+	 * stack is deleted. `RETAIN` keeps the data behind after a `cdk destroy`;
+	 * `DESTROY` tears it down so a sandbox teardown leaves nothing behind.
+	 */
+	removalPolicy: RemovalPolicy;
+
+	/**
+	 * Whether to block accidental deletion of stateful resources. When `true`,
+	 * a stray console/CLI delete of the resource is rejected until protection is
+	 * turned off. Keep this off in sandbox so `sandbox:destroy` can tear the
+	 * stack down in one command.
```
svidgen: Also prevents CDK/Cloudformation from deleting the resource, correct? Worth mentioning that `removalPolicy` applies only to CDK/Cloudformation whereas `deleteProtection` prevents deletion via CDK/Cloudformation **as well as** any via CLI, API, or AWS console. (If I understand the setting correctly.)

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -0,0 +1,108 @@
...
+ * it explicitly, and `BlocksPresets.sandbox` loosens it for disposable
+ * development stacks.
+ */
+export const BlocksPresets = {
+	/** Disposable development stacks: tear down cleanly, no delete guard. */
+	sandbox: {
+		removalPolicy: RemovalPolicy.DESTROY,
+		deletionProtection: false,
+	},
+	/** Durable, protected posture for real deployments. The framework default. */
```
svidgen: 1. Minus "The framework default."

We currently offer *two* "framework defaults" depending on which presets you pick. 😁 

2. "Permanent deployments" instead of "real" deployment would cast less shade on sandbox deployments.

I'm tempted even to suggest changing "production" to "permanent", since the customer might have permanent non-production environments. We can start with "production" though and add/change for GA if needed. cc @pranavosu

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -0,0 +1,108 @@
...
+ * `BlocksBackend.create` (via `setupBlocksInfra`) from the `defaults` prop,
+ * before the app's blocks are constructed, so any block can read it.
+ *
+ * Stored on the stack instance under a `Symbol.for` key — the same
+ * stack-scoped, synth-time storage pattern used by the config registry.
+ */
+export function registerStackBlocksDefaults(scope: Construct, defaults: BlocksDefaults): void {
+	const stack = cdk.Stack.of(scope);
+	(stack as unknown as Record<symbol, unknown>)[DEFAULTS_KEY] = defaults;
+}
```
svidgen: Anchoring to the stack seems wrong. Are we anchoring everything else at the stack level?

If so, I might need to reel in my previous gripe with the "stack-wide" phrasing. If we can avoid anchoring at the stack level that seems ideal though. But, if someone wants to create *N* `BlocksBackend` instances in a single stack, we should *ideally* be able to support it.

svidgen (changes requested):
Looking pretty good. A few small changes I'd like to see.

If I'm misremembering how we do or don't anchor the scope tree to a stack today, you can skip addressing that for now; but would be good to dive on that separately.

packages/core/src/cdk/blocks-defaults.ts
pranavosu: +1 on both

packages/core/src/common/index.ts
osama-rizk: This comment is outdate will resolve it and will address the newest comment

.changeset/blocks-infrastructure-defaults.md
osama-rizk: Good catch — done. Added `"@aws-blocks/blocks": minor` to the changeset since the umbrella re-exports the new core `/cdk` surface (`BlocksDefaults`, `BlocksPresets`, `registerStackBlocksDefaults`/`getStackBlocksDefaults`). Pushed.

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Agreed — switched to backend-level anchoring. `defaults` now lives on the `BlocksStack`/`BlocksBackend` instance, and `scope.defaults` resolves by walking up the construct tree to the owning backend (the exact pattern `handler` and `executionRole` already use), instead of a single `Symbol` slot on `cdk.Stack.of(scope)`.

Why: with stack anchoring, two `BlocksBackend`s in one stack shared one slot and the second `create()` would clobber the first — a production backend's blocks could silently read a sibling sandbox backend's posture. Backend anchoring lets N backends coexist in one stack, each keeping its own defaults. Removed the stack-scoped `registerStackBlocksDefaults`/`getStackBlocksDefaults` helpers.

Added tests: two backends in one stack keep separate defaults, and a nested block resolves its owning backend's defaults via the tree-walk. Pushed in 85f2e0db.

packages/bb-kv-store/src/index.cdk.ts
osama-rizk: Done — added `deletionProtection?: boolean` to `KVStoreOptions` and resolved it as `options?.deletionProtection ?? this.defaults.deletionProtection`, matching `removalPolicy`. (Kept the CDK name `deletionProtection` rather than `deleteProtection` for consistency with `BlocksDefaults` and the CDK Table prop.) Also refreshed the `removalPolicy` docstring for the new defaults model, and added a test that a per-block `deletionProtection` overrides the resolved default. 2c3abe42

On the broader principle — agreed that anything consumed from `defaults` should be overridable per-block. I kept the fallback inline here (`option ?? this.defaults.field`) rather than abstracting it, since this is the only adopter so far. When the follow-up PRs adopt `defaults` across the remaining stateful BBs, I'll factor the precedence into a small shared helper (e.g. `scope.resolveDefault(key, override)`) so it isn't hand-written per property — the per-resource mapping (and the `'destroy'|'retain'` → `RemovalPolicy` normalization) still has to live in each BB, but the `?? default` rule can be DRY.

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Done — reworded to your suggestion: it now reads *"Default values for every Amazon-authored Building Block created within a `BlocksStack` or `BlocksBackend`."* No more "stack-wide".

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Done — trimmed. The resolution-order/precedence essay is gone; the docstring is now just what the type is plus "start from `BlocksPresets` and override individual fields." The "why" can live in a decision log / the PR description.

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Done — removed. The `FRAMEWORK_HARDENING_DEFAULTS` constant and its explanatory essay are gone entirely (superseded by the required `defaults` + presets model), so there's no longer a long framework-default docstring here.

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Done — clarified. `removalPolicy` now says it applies **only** to CDK/CloudFormation deletions, and `deletionProtection` says it blocks deletion through **any** path — CDK/CloudFormation as well as the CLI, API, and AWS console. Same distinction is mirrored on `KVStoreOptions`.

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Done — dropped "The framework default" (two presets, agreed) and the production preset now reads "Durable, protected posture for **permanent** deployments" instead of "real," so sandbox isn't cast as less real. Kept the name `production` for now; happy to revisit `permanent` for GA.

packages/create-blocks-app/templates/auth-cognito/aws-blocks/index.cdk.ts
```diff
@@ -16,17 +15,11 @@ const projectRoot = app.node.tryGetContext('projectRoot') || process.cwd();
...
-  // Make all resources deletable so sandbox:destroy can clean up the entire stack.
-  // This overrides removal policies and deletion protection (e.g. RDS) for every
-  // resource in the stack, including any you add below.
-  // Remove these lines if you want to manage teardown behavior yourself.
-  RemovalPolicies.of(blocksStack).destroy();
-  Mixins.of(blocksStack).apply(new SandboxDisableDeletionProtection());
-
   // Tell the runtime that cookies need cross-domain attributes (frontend on
   // localhost, API on API Gateway — different registrable domains).
   blocksStack.handler.addEnvironment('BLOCKS_SANDBOX', 'true');}
```
svidgen, line 25: Hmm. Where is this still being used? Should this also be flipping something in the `defaults` instead? (Is addressing this a reasonable scope increase on this PR?)

svidgen (dismissed):

packages/bb-distributed-data/src/index.cdk.ts
```diff
@@ -24,14 +24,19 @@ export class DistributedDatabase extends Scope {
...
-    const removalPolicy = shouldDestroy ? cdk.RemovalPolicy.DESTROY : cdk.RemovalPolicy.RETAIN;
+    // DSQL Cluster removal: the per-block option wins, otherwise the stack-wide
+    // `defaults` (sandbox → DESTROY, production → RETAIN). Deletion protection
+    // follows: on unless the cluster is being destroyed.
+    const removalPolicy =
+      options?.removalPolicy === 'destroy'
+        ? cdk.RemovalPolicy.DESTROY
+        : options?.removalPolicy === 'retain'
+          ? cdk.RemovalPolicy.RETAIN
+          : this.defaults.removalPolicy;
```
soberm, line 41: [blocker] Deletion protection gets resolved two different ways across the five blocks adopting `defaults`, so the same `defaults` object ends up meaning different things depending on which block reads it.

Here and in `bb-data/src/infra.ts:149` (`deletionProtection: removalPolicy !== cdk.RemovalPolicy.DESTROY`), protection is derived from the resolved `removalPolicy` instead of read from `defaults.deletionProtection`. So `defaults: { ...BlocksPresets.production, deletionProtection: false }`, the exact override pattern `BlocksPresets`'s own docstring shows, gets silently dropped in both places: they'll still synth `DeletionProtectionEnabled: true` because `removalPolicy` is still RETAIN.

Flip it around and it's worse. `bb-distributed-table` (`index.cdk.ts:89-90`) and `bb-kv-store` (`index.cdk.ts:52`) read `deletionProtection` independently of `removalPolicy`. So `defaults: { ...BlocksPresets.production, removalPolicy: cdk.RemovalPolicy.DESTROY }` leaves their `deletionProtection` at `true` from the production preset while removalPolicy flips to DESTROY, i.e. DeletionPolicy=Delete + DeletionProtectionEnabled=true, an undeletable table on `cdk destroy`. bb-data and this DSQL cluster wouldn't hit that particular failure since their protection just tracks removalPolicy either way.

Both presets keep the two fields correlated (DESTROY+false, RETAIN+true) so none of this shows up with `BlocksPresets.sandbox`/`.production` unmodified, only once someone overrides a single field, which is the customization path the docstring itself demonstrates.

Given the open question already raised about these two fields, probably worth picking one derivation style before bb-realtime/bb-file-bucket adopt this too: either derive deletionProtection from removalPolicy everywhere, or read `defaults.deletionProtection` independently everywhere, so the same spread produces the same posture no matter which block consumes it.

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -239,6 +250,11 @@ export class BlocksBackend extends Construct {
     // Expose self to Building Blocks at CDK time
     (globalThis as any).CURRENT_BLOCKS_STACK = this;
 
+    // Store defaults on the backend (not the stack) so several BlocksBackends
+    // in one stack each keep their own posture; Building Blocks resolve them by
+    // walking up to their owning backend (see Scope.defaults).
+    this.defaults = props.defaults;
```
soberm, line 271: [minor] Nothing validates that `props.defaults` actually got passed here (same in `BlocksStack`'s constructor). A caller that skips it, plain JS, `as any`, a props object built dynamically, gets a bare `Cannot read properties of undefined (reading 'removalPolicy')` a few frames deep inside whichever Building Block constructs first, instead of a clear error at the `create()` call site. `backendHandlerPath`/`backendCDKPath` have the same gap today so this isn't a new pattern, but might be worth a quick `if (!props.defaults) throw` while touching this, especially since bb-kv-store has a test asserting exactly this principle elsewhere (actionable errors, not cryptic TypeErrors).

packages/core/src/cdk/index.ts
```diff
@@ -132,6 +140,32 @@ export class Scope extends Construct {
...
+    // then to the production preset when none was registered.
+    let current: Construct = this;
+    while (current.node.scope) {
+        current = current.node.scope as Construct;
+        if (current instanceof BlocksStack || current instanceof BlocksBackend) {
+            return current.defaults;
+        }
+    }
+    return ((globalThis as any).CURRENT_BLOCKS_STACK as { defaults?: BlocksDefaults } | undefined)?.defaults
+      ?? BlocksPresets.production;
```
soberm: [nit] Nice that this falls back to the production preset instead of crashing the way `handler`/`executionRole` do when no owning BlocksStack/BlocksBackend is found. Since it's silent though, a real misconfiguration (a block built outside any BlocksStack/BlocksBackend tree) would quietly get production defaults rather than surfacing. Might be worth a log line here so it's debuggable if it ever fires outside the deliberate test-stub pattern.

packages/core/src/cdk/index.ts
```diff
@@ -15,11 +15,16 @@ import {
 import { setupBlocksInfra, BlocksBackend, assertCdkConditionActive } from './blocks-backend.js';
 import { addBlocksStackMetadata } from './stack-metadata.js';
 import { finalizeConfigRegistry } from './config-registry.js';
+import { type BlocksDefaults, BlocksPresets } from './blocks-defaults.js';
 
 export { BlocksBackend, type BlocksBackendProps } from './blocks-backend.js';
 export { DEFAULT_NODE_RUNTIME } from './node-version.js';
 export { SandboxDisableDeletionProtection } from './mixins.js';
```
soberm, line 23: [minor] After this PR, `SandboxDisableDeletionProtection` (and the `RemovalPolicies.of(...).destroy()` pairing) has zero remaining call sites in the repo, everything migrated to `defaults`. It's still a public export though. Worth a `@deprecated` tag pointing at `defaults` so someone reading the public API later doesn't have to guess which mechanism is current.

packages/create-blocks-app/templates/default/aws-blocks/index.cdk.ts
```diff
@@ -16,17 +15,11 @@ const projectRoot = app.node.tryGetContext('projectRoot') || process.cwd();
 const stackName = getStackName({ sandbox: sandboxMode, projectRoot });
 export const blocksStack = await BlocksStack.create(app, stackName, {
   backendHandlerPath: join(__dirname, 'index.handler.ts'),
-  backendCDKPath: join(__dirname, 'index.ts')
+  backendCDKPath: join(__dirname, 'index.ts'),
+  defaults: sandboxMode ? BlocksPresets.sandbox : BlocksPresets.production
```
soberm: [nit] Missing trailing comma here (same in auth-cognito, backend, bare, demo, nextjs, and react templates), while every test-app equivalent has one. Purely cosmetic, prettier would normally add it if these templates are in its glob.

test-apps/pipeline/aws-blocks/index.cdk.ts
```diff
@@ -15,4 +15,5 @@ const stackName = sandboxMode ? 'pipeline-test-sandbox' : 'pipeline-test-prod';
 await BlocksStack.create(app, stackName, {
   backendHandlerPath: join(__dirname, 'index.handler.ts'),
   backendCDKPath: join(__dirname, 'index.ts'),
+  defaults: BlocksPresets.production,
```
soberm: [nit] This hardcodes production for both the sandbox and prod stack name branches, unlike every other test-app here which conditions on `sandboxMode`. Not a problem today, `index.ts` for this app only registers `RawRoute`s (no stateful resources) and the package's only script is a synth test with no deploy/destroy, but if this app ever grows a stateful Building Block it'd be a bit of a landmine. Might be worth aligning to `sandboxMode ? BlocksPresets.sandbox : BlocksPresets.production` for consistency.

packages/core/src/common/index.ts
```diff
@@ -325,6 +325,14 @@ export function computeScopeFullId(scope: { id: string; parent?: any }) {
...
   backendHandlerPath: string;
   backendCDKPath: string;
+  /**
+   * Stack-wide infrastructure defaults applied to every Building Block (removal
+   * policy, deletion protection, …). Start from `BlocksPresets.sandbox` or
+   * `BlocksPresets.production` and override individual fields as needed. Any
+   * field a block also exposes as a per-block option is overridden by that
+   * option. See `BlocksDefaults` in `@aws-blocks/core/cdk`.
+   */
+  defaults: import('../cdk/blocks-defaults.js').BlocksDefaults;
```
soberm: [nit] Could be a plain `import type { BlocksDefaults } from '../cdk/blocks-defaults.js';` at the top instead of the inline dynamic import type. Both erase at runtime the same way, the top-level form just reads a bit more conventionally.

.changeset/blocks-infrastructure-defaults.md
```diff
@@ -0,0 +1,17 @@
...
+"@aws-blocks/blocks": minor
+"@aws-blocks/bb-kv-store": patch
+"@aws-blocks/bb-data": patch
+"@aws-blocks/bb-distributed-data": patch
+"@aws-blocks/bb-distributed-table": patch
+"@aws-blocks/bb-knowledge-base": patch
```
soberm, line 8: [question] create-blocks-app's templates changed too (7 of them) but it's not in this list. Looks like it doesn't declare `@aws-blocks/core` as a dependency so this is probably fine, just want to confirm that's deliberate and not a miss.

soberm (changes requested):

packages/bb-distributed-data/src/index.cdk.ts
osama-rizk: Fixed — picked **independent** resolution everywhere. `bb-data` and `bb-distributed-data` now read `this.defaults.deletionProtection` directly instead of deriving it from `removalPolicy`, matching `bb-kv-store`/`bb-distributed-table`. So the same `defaults` spread produces the same posture in every block, and `{ ...BlocksPresets.production, deletionProtection: false }` is now honored across all four. Chose independent (over derive-everywhere) because the design + presets treat `deletionProtection` as a first-class field with an independent-override docstring, and two blocks already read it that way.

On the `DESTROY + protected` (undeletable) combo you flagged: that's now uniformly representable and remains the already-flagged open question — it only arises from an explicit contradictory override; neither preset produces it. It'll be removed for good if we consolidate the two fields into a single `protection` enum (the #282 direction). Added a dsql test proving removalPolicy and deletionProtection resolve independently. 83d75b98

packages/core/src/cdk/blocks-backend.ts
osama-rizk: Done — `setupBlocksInfra` (called by both `BlocksStack` and `BlocksBackend`) now throws early with an actionable message if `defaults` is missing, instead of the cryptic `Cannot read properties of undefined` a few frames deep. 83d75b98

packages/core/src/cdk/index.ts
osama-rizk: Done — the `scope.defaults` fallback now `console.warn`s when no owning BlocksStack/BlocksBackend (and no ambient stack) is found, before returning the production preset, so a real misconfiguration is debuggable rather than silent. 83d75b98

packages/core/src/cdk/index.ts
osama-rizk: Done — added an `@deprecated` tag on `SandboxDisableDeletionProtection` pointing at the `defaults` prop, noting it's retained only for backward compat. 83d75b98

packages/create-blocks-app/templates/default/aws-blocks/index.cdk.ts
osama-rizk: Done — added the trailing comma on the `defaults` line in all 7 templates (default, auth-cognito, backend, bare, demo, nextjs, react). 83d75b98

test-apps/pipeline/aws-blocks/index.cdk.ts
osama-rizk: Done — aligned `test-apps/pipeline` to `defaults: sandboxMode ? BlocksPresets.sandbox : BlocksPresets.production`, matching the other test-apps. 83d75b98

packages/core/src/common/index.ts
osama-rizk: Done — switched to a top-level `import type { BlocksDefaults } from '../cdk/blocks-defaults.js';` and referenced it directly, instead of the inline dynamic-import type. 83d75b98

.changeset/blocks-infrastructure-defaults.md
osama-rizk: Good catch — it was a miss, not deliberate. Although `create-blocks-app` has no `@aws-blocks/core` dependency, it's published and its shipped **templates** changed (they now require the `defaults` API), so it needs to ship. Added `"@aws-blocks/create-blocks-app": patch` to the changeset. 87a27859

packages/create-blocks-app/templates/auth-cognito/aws-blocks/index.cdk.ts
osama-rizk: Good question. `BLOCKS_SANDBOX` is a **runtime** flag (it tells the deployed runtime to use cross-domain cookie attributes when the frontend is on localhost and the API on API Gateway), so it's orthogonal to `defaults`, which only governs CDK infra durability — folding it in would conflate runtime behavior with infra posture. My change here only removed the mixin lines and preserved this pre-existing block.

That said, grepping the framework source I don't find a current consumer of the `BLOCKS_SANDBOX` env var, so it may actually be stale. Auditing/removing it is a separate concern from this PR's `defaults` mechanism, so I'd prefer to leave it as-is here and track a cleanup separately rather than grow this PR's scope. Happy to file that follow-up.

osama-rizk: Correction to my earlier reply — I mis-searched (I grepped framework `src`, which filtered out the actual consumer): `BLOCKS_SANDBOX` is **not** stale. It's a live template pattern — the `index.cdk.ts` sets `handler.addEnvironment('BLOCKS_SANDBOX', 'true')` in sandbox, and the app's `index.ts` reads it as `crossDomain: process.env.BLOCKS_SANDBOX === 'true'` to flip the auth BB's cookie to cross-domain attributes (frontend on localhost, API on API Gateway).

So it shouldn't move into `defaults`: `defaults` is CDK synth-time **infra durability** (removal policy / deletion protection) and never reaches the runtime, whereas `crossDomain` is a **runtime auth-cookie** behavior — different layer.

The worthwhile cleanup is to stop every template hand-rolling this: have the framework expose a runtime sandbox signal (set once in `setupBlocksInfra` from the `sandboxMode` context) and let the auth BBs default `crossDomain` from it, then drop the ~11 template lines. That changes auth BB defaults, so it wants its own API review and is orthogonal to this PR's `defaults` mechanism. **We'll address it in a follow-up PR** — leaving `BLOCKS_SANDBOX` as-is here.

soberm (dismissed):

Simone319 (approved):

soberm (approved):

PR #304, opened 2026-08-04
fix(bb-file-bucket): harden default bucket security (enforceSSL, opt-in access logs, dev CORS)

pjkroker:
## Problem

A security review (external scanner, rule categories **R6/R8**) produced **4 reports across 3 apps** flagging insecure **default** properties on the S3 bucket provisioned by the `FileBucket` building block. The originating task listed three concerns tersely:

> Default bucket props: versioning OFF (no recovery from overwrite/delete), no server access logs, wildcard CORS for PUT.

The task text was thin, so this description records what I actually found in the code — including one finding whose real root cause was **not** where it first appeared.

### Finding 1 — Versioning OFF by default
`index.cdk.ts` set `versioned: options?.versioned ?? false`. With versioning off there is no recovery from an accidental overwrite or delete.

### Finding 2 — No server access logs (and no transport-encryption enforcement)
The provisioned `s3.Bucket` set `blockPublicAccess: BLOCK_ALL` and `encryption: S3_MANAGED`, but configured **no** server access logging and did **not** set `enforceSSL` — so non-HTTPS requests were not denied at the bucket policy level.

### Finding 3 — "Wildcard CORS for PUT" (root cause was not the deployed bucket)
Investigated thoroughly. The **deployed** S3 bucket emits **no default CORS at all** — CORS comes only from user-supplied `options.corsRules`, and no template or test-app passes any. Core's API CORS (`packages/core/src/cors.ts`) is default-**deny**, never wildcard.

The only wildcard-origin + PUT in FileBucket's own code is the **local dev file-server** (`src/file-server.ts`). It reflected the request `Origin` (falling back to literal `*`), advertised `Access-Control-Allow-Methods: GET, PUT, OPTIONS`, **and** set `Access-Control-Allow-Credentials: true`. Reflecting an arbitrary origin *with credentials* is the textbook permissive-CORS pattern a scanner flags. Because this one shared file is compiled into every app, it plausibly explains "**4 reports across 3 apps**" from a single code location.

## Changes

**`packages/bb-file-bucket/src/index.cdk.ts`**
- **`enforceSSL: true`** is now set unconditionally on the file bucket (and the log bucket). This emits a bucket policy denying any request where `aws:SecureTransport` is `false`. All SDK and presigned-URL traffic is already HTTPS, so it's transparent — a zero-risk, high-value default that matches the hosting/inventory buckets which already enforce it.
- New **opt-in** `accessLogging` provisions a dedicated, private **log bucket** (`BLOCK_ALL`, `S3_MANAGED`, `enforceSSL`, an `ExpireAccessLogs` lifecycle rule) and wires the file bucket's `serverAccessLogsBucket` / `serverAccessLogsPrefix: 'access-logs/'`. Retention is `logRetentionDays` (default **90**). The log bucket follows the file bucket's existing removal/auto-delete rules rather than an unconditional DESTROY.
- The `fromExisting` (external bucket) path returns early and is untouched.

**`packages/bb-file-bucket/src/types.ts`**
- Added `accessLogging?: boolean` and `logRetentionDays?: number` to `FileBucketOptions`, documented as **CDK-only** (ignored by the mock and browser runtimes).

**`packages/bb-file-bucket/src/file-server.ts`** (finding 3)
- Removed the `Access-Control-Allow-Credentials: true` header. Presigned-URL auth is carried in a query-string token, not a cookie, so credentials were never needed; removing it neutralizes the reflected-origin-with-credentials pattern. Origin reflection + `GET, PUT, OPTIONS` are retained so local cross-port dev still works, with a comment noting this is localhost-only dev tooling that never deploys.

**Docs & metadata**
- `README.md`: documented the two new options, that `enforceSSL` is always on, that **production browser uploads require an explicit-origin `corsRules` entry** (never `['*']` with a mutating method — this doc gap is what let the confusion arise), and versioning-for-recovery guidance.
- `DESIGN.md`: three new decisions — **D-FB-9** (versioning stays opt-in), **D-FB-10** (enforceSSL always on; access logging opt-in), **D-FB-11** (dev-CORS hardening; deployed bucket emits no default CORS) — plus mock-difference rows.
- Regenerated `API.md`; added a `patch` changeset covering `@aws-blocks/bb-file-bucket` and the re-exporting umbrella `@aws-blocks/blocks`.

## Decision: versioning left opt-in (not flipped to on-by-default)

Finding 1's "secure default" would be `versioned: true`, but I deliberately did **not** flip it, for three reasons (recorded in **D-FB-9**):
1. **Cost/behavior change** — every overwrite is retained (ongoing storage cost) and the mock's versioned code path would activate.
2. **Type/runtime mismatch** — the version-aware option types (`GetOptionsFor`/`DeleteOptionsFor`/`GetUrlOptionsFor`) key off the `O extends { versioned: true }` literal, so a caller omitting the flag would get non-versioned option types while the bucket is versioned at runtime — a silent disagreement.
3. **Minimal-breaking-change stance** — a default flip belongs in its own minor/major with an explicit `versioned: false` opt-out and a migration note, not bundled with R6/R8 hardening.

The finding is closed via documentation + the one-line opt-in. **Flag for maintainer:** if the security posture requires versioning on by default, say so and I'll do it as a separate change.

## ⚠️ Scope note on finding 3

Because the deployed bucket never emitted default CORS, this PR satisfies the scanner (by hardening the dev server) but does **not** add a construct-level guard against a user passing `corsRules: [{ allowedOrigins: ['*'], allowedMethods: ['PUT'] }]`. That's documented against rather than enforced. Happy to add a synth-time warning (via CDK `Annotations`) if reviewers want belt-and-suspenders.

## Validation

- **Replicated first (red → green):** added failing synth tests (`enforceSSL by default`, `accessLogging provisions a log bucket + LoggingConfiguration`) and a dev-server test (`no Allow-Credentials on a reflected origin`); confirmed they fail against `main`, then implemented the fixes to make them pass.
- **`npm run test --workspace @aws-blocks/bb-file-bucket`** → **94/94 pass**.
- Lint clean; full `bb-file-bucket` + `blocks` build OK.
- Synth spot-check with `accessLogging: true`: 2 S3 buckets + 2 bucket policies (enforceSSL), main bucket `LoggingConfiguration` → log bucket with `access-logs/` prefix.
- No live AWS deploy — the CloudFormation-template assertions are the intended proxy (there is no e2e that asserts destroy-time/logging behavior).

## Notes for reviewers

- Branched from fresh `origin/main`. This is **unrelated** to PR #168 (`fix/bb-file-bucket-removal-policy`, the `autoDeleteObjects` fix) — deliberately kept separate.
- All three findings are one issue with independent, additive, non-breaking fixes → single `patch` changeset.

## Checklist

- [x] PR description included
- [x] Relevant documentation is changed or added

*By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.*

### DIFF
--- a/.changeset/bb-file-bucket-secure-defaults.md
+++ b/.changeset/bb-file-bucket-secure-defaults.md
@@ -1,0 +1,11 @@
+---
+"@aws-blocks/bb-file-bucket": patch
+"@aws-blocks/blocks": patch
+---
+
+Harden `FileBucket` default security posture (addresses security-review findings R6/R8):
+
+- **Transport encryption always on** — the provisioned S3 bucket now sets `enforceSSL: true`, denying any non-HTTPS request via a bucket policy. All SDK and presigned-URL traffic is already HTTPS, so this is transparent.
+- **Opt-in server access logging** — new `accessLogging?: boolean` option (default `false`) provisions a dedicated, private log bucket (public access blocked, SSE-S3, `enforceSSL`, log expiry) and wires the file bucket's server access logs to it under the `access-logs/` prefix. Retention is controlled by `logRetentionDays?: number` (default 90). Both options are CDK-only and ignored by the mock and browser runtimes.
+- **Dev file-server CORS hardening** — the local dev file-server no longer sends `Access-Control-Allow-Credentials: true` alongside a reflected/wildcard origin. Presigned-URL auth is a query-string token (not a cookie), so credentials were never needed; this removes the permissive-CORS pattern flagged by security scanners. Deployed buckets emit no default CORS and are unaffected.
+- **Docs** — README now documents that production browser uploads require an explicit-origin `corsRules` entry (never `['*']` with a mutating method), and that versioning (off by default) should be enabled for overwrite/delete recovery.
--- a/packages/bb-file-bucket/API.md
+++ b/packages/bb-file-bucket/API.md
@@ -76,10 +76,12 @@ export const FileBucketErrors: {
 
 // @public (undocumented)
 export interface FileBucketOptions {
+    accessLogging?: boolean;
     bucket?: ExternalBucketRef;
     corsRules?: CorsRule[];
     lifecycleRules?: LifecycleRule[];
     logger?: ChildLogger;
+    logRetentionDays?: number;
     removalPolicy?: 'destroy' | 'retain';
     versioned?: boolean;
 }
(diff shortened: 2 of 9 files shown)


soberm:
Closing this in favor of https://github.com/aws-devtools-labs/aws-blocks/pull/487

PR #315, opened 2026-08-06
fix(core): route 504 CORS through allowlist + cap Access-Control-Max-Age at 7200

soberm:
## What

The 504 timeout path in `createLambdaHandler` built its CORS headers by hand: it reflected the request `Origin` verbatim (falling back to `'*'`) and always sent `Access-Control-Allow-Credentials: true`. That reflected an unvalidated origin back to the caller with credentials enabled, and `'*'` + credentials is outright invalid, so a timed-out request either leaked a permissive CORS grant or produced a response the browser rejects. The timeout path now goes through `buildCorsHeaders(origin)` — the same allowlist-validated, fail-closed helper the normal response path uses — so an origin that is not on the allowlist gets no CORS grant at all.

Also lowers `Access-Control-Max-Age` from `86400` to `7200` on the OPTIONS preflight response and in the dev server. 86400 is misleading: browsers cap preflight caching well below that (Chromium 7200s, Safari/WebKit lower still), so the advertised value never applied and made the preflight cache lifetime look longer than it really is. 7200 is the largest value actually honoured by the strictest mainstream engine.

Found during the bug bash (P3 report on the CORS preflight Max-Age value); the 504 origin reflection was spotted while confirming the Max-Age code path.

## Changes

- `packages/core/src/lambda-handler.ts` — 504 path uses `buildCorsHeaders(origin)`; OPTIONS `Access-Control-Max-Age` 86400 → 7200.
- `packages/core/src/scripts/dev-server.ts` — `Access-Control-Max-Age` 86400 → 7200.
- `packages/core/src/cors.test.ts` — preflight Max-Age expectation updated to 7200.
- `packages/core/src/lambda-handler.test.ts` — the old test asserting reflective 504 behaviour is replaced by two allowlist-aware tests (allowed origin gets the grant, disallowed origin gets none).

## Testing

`npx tsx --test src/cors.test.ts src/lambda-handler.test.ts` in `packages/core`: **81/81 pass, 0 fail** (21 suites).

The 2 pre-existing `hosting.ts` build errors in this package are unrelated to this change and are left untouched.

### DIFF
--- a/packages/core/src/cors.ts
+++ b/packages/core/src/cors.ts
@@ -2,6 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /**
+ * `Access-Control-Max-Age` for preflight responses, in seconds.
+ *
+ * Chromium caps the preflight cache at 7200s and silently clamps anything
+ * higher, so a larger value buys nothing while widening the window in which a
+ * stale per-origin grant can be served. Shared by the Lambda handler and the
+ * local dev server so the two can't drift.
+ */
+export const CORS_MAX_AGE = '7200';
+
+/**
  * Parse a comma-separated CORS origin string into anchored RegExp patterns.
  *
  * Each entry is treated as a regex pattern:
@@ -73,6 +83,25 @@ export function isOriginAllowed(origin: string): boolean {
 }
 
 /**
+ * Distinct origins already warned about, so a caller retrying — or a bot
+ * spraying bogus `Origin` values — can't amplify one log line per request now
+ * that this helper runs on every response path.
+ */
+const warnedOrigins = new Set<string>();
+
+/** Cap on {@link warnedOrigins} so an untrusted input can't grow it unbounded. */
+const WARNED_ORIGINS_LIMIT = 100;
+
+function warnDisallowedOriginOnce(origin: string): void {
+  if (warnedOrigins.has(origin)) return;
+  if (warnedOrigins.size < WARNED_ORIGINS_LIMIT) warnedOrigins.add(origin);
+  const example = 'CORS_ALLOWED_ORIGINS=https://myapp\\.com,^https?://(localhost|127\\.0\\.0\\.1)(:\\d+)?$';
+  console.warn(
+    `[CORS] Origin "${origin}" is not allowed. Set the CORS_ALLOWED_ORIGINS environment variable to allow this origin. Example: ${example}`
+  );
+}
+
+/**
  * Build the CORS response headers for a request origin.
  *
  * Only reflects the origin when it matches the configured allowlist. When no
@@ -80,18 +109,21 @@ export function isOriginAllowed(origin: string): boolean {
  * `Access-Control-Allow-Origin` / `Access-Control-Allow-Credentials` headers
  * are emitted, so a disallowed origin is never reflected back.
  *
+ * `Vary: Origin` is always emitted, including on the not-allowed path: the
+ * response headers depend on the request `Origin`, so any shared cache (CDN,
+ * forward proxy) must key on it or it can serve one origin's grant — or one
+ * origin's *absence* of a grant — to a different origin.
+ *
  * @param origin - The `Origin` header value from the request (may be empty)
- * @returns The CORS headers to merge into the response (empty when not allowed)
+ * @returns The CORS headers to merge into the response
  */
 export function buildCorsHeaders(origin: string): Record<string, string> {
-  const headers: Record<string, string> = {};
+  const headers: Record<string, string> = { Vary: 'Origin' };
   if (isOriginAllowed(origin)) {
     headers['Access-Control-Allow-Origin'] = origin;
     headers['Access-Control-Allow-Credentials'] = 'true';
   } else if (origin) {
-    console.warn(
-      `[CORS] Origin "${origin}" is not allowed. Set the CORS_ALLOWED_ORIGINS environment variable to allow this origin. Example: CORS_ALLOWED_ORIGINS=https://myapp\\.com,^https?://(localhost|127\\.0\\.0\\.1)(:\\d+)?$`
-    );
+    warnDisallowedOriginOnce(origin);
   }
   return headers;
 }
@@ -102,14 +134,15 @@ export function buildCorsHeaders(origin: string): Record<string, string> {
 export function corsRejection(): { statusCode: number; headers: Record<string, string>; body: string } {
   return {
     statusCode: 403,
-    headers: { 'Content-Type': 'application/json' },
+    headers: { 'Content-Type': 'application/json', Vary: 'Origin' },
     body: JSON.stringify({ error: 'Forbidden: cross-origin request rejected' }),
   };
 }
 
 /**
- * Reset the lazy CORS pattern cache. **For testing only.**
+ * Reset the lazy CORS pattern cache and the warned-origin set. **For testing only.**
  */
 export function _resetCorsPatterns(): void {
   _corsPatterns = undefined;
+  warnedOrigins.clear();
 }
--- a/packages/core/src/cors.ts
+++ b/packages/core/src/cors.ts
@@ -73,6 +73,30 @@ export function isOriginAllowed(origin: string): boolean {
 }
 
 /**
+ * Build the CORS response headers for a request origin.
+ *
+ * Only reflects the origin when it matches the configured allowlist. When no
+ * allowlist is configured, or the origin is configured-but-not-allowed, no
+ * `Access-Control-Allow-Origin` / `Access-Control-Allow-Credentials` headers
+ * are emitted, so a disallowed origin is never reflected back.
+ *
+ * @param origin - The `Origin` header value from the request (may be empty)
+ * @returns The CORS headers to merge into the response (empty when not allowed)
+ */
+export function buildCorsHeaders(origin: string): Record<string, string> {
+  const headers: Record<string, string> = {};
+  if (isOriginAllowed(origin)) {
+    headers['Access-Control-Allow-Origin'] = origin;
+    headers['Access-Control-Allow-Credentials'] = 'true';
+  } else if (origin) {
+    console.warn(
+      `[CORS] Origin "${origin}" is not allowed. Set the CORS_ALLOWED_ORIGINS environment variable to allow this origin. Example: CORS_ALLOWED_ORIGINS=https://myapp\\.com,^https?://(localhost|127\\.0\\.0\\.1)(:\\d+)?$`
+    );
+  }
+  return headers;
+}
+
+/**
  * Build a 403 Forbidden response for cross-origin requests from disallowed origins.
  */
 export function corsRejection(): { statusCode: number; headers: Record<string, string>; body: string } {
--- a/packages/core/src/lambda-handler.ts
+++ b/packages/core/src/lambda-handler.ts
@@ -15,7 +15,7 @@ import {
   errorResponseFromCatch,
   methodNotFoundResponse,
 } from './rpc.js';
-import { getCorsPatterns, isOriginAllowed, corsRejection } from './cors.js';
+import { getCorsPatterns, isOriginAllowed, corsRejection, buildCorsHeaders } from './cors.js';
 
 export { parseCorsPatterns, _resetCorsPatterns } from './cors.js';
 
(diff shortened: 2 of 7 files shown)


Simone319 (dismissed):

osama-rizk (commented):
Reviewed the CORS 504 + Max-Age fix. This is a clean, correct security fix and I'd merge it about as-is. I traced the control flow rather than trusting the summary: the `'*'` → `''` fallback change is exactly right (an empty origin can never match the allowlist, so `buildCorsHeaders('')` returns `{}` — fail-closed), and I confirmed *why* the disallowed-origin-on-504 path is even reachable: the 403 CORS gate at line 447 is guarded by `patterns` being truthy, so with **no allowlist configured** a disallowed origin skips the 403 and lands on the timeout path — which is precisely the hole this closes. Extracting `buildCorsHeaders` into `cors.ts` also *removes* a duplicate (it was a private copy in the handler) and makes the 504, 404, OPTIONS, and RPC paths share one fail-closed helper. Good change.

The three new 504 tests cover the cases that matter (allowed grant / no-allowlist omission / 403-precedence), env cleanup is handled per-test. My comments are mostly at the systemic altitude and none block: one pre-existing cache-safety gap that this PR walks right past (and is the natural place to fix, since it's the same class of bug — unsafe reflection), plus a couple of smaller coverage/robustness notes.

I didn't re-derive the allowlist regex logic — it's unchanged by this PR and already tested. The higher-leverage question here was "does the reflected-origin model interact badly with the newly-advertised cache lifetime," which is finding #1.

packages/core/src/lambda-handler.ts
```diff
@@ -472,7 +456,7 @@ function createHandler(backend: any) {
           ...corsHeaders,
           'Access-Control-Allow-Methods': 'GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS',
           'Access-Control-Allow-Headers': 'Content-Type, Authorization',
-          'Access-Control-Max-Age': '86400',
+          'Access-Control-Max-Age': '7200',
```
osama-rizk: **Cache-safety gap, adjacent to the exact bug this PR fixes — worth addressing here.** This OPTIONS preflight response now advertises `Access-Control-Max-Age: 7200`, explicitly telling caches to store it for 2 hours. But its `Access-Control-Allow-Origin` is a **reflected** value (`buildCorsHeaders` echoes the request origin when allowed), and the response carries **no `Vary: Origin`** and **no `Cache-Control`** (unlike the RPC path just below at ~496, which sets `no-store`). I grepped the whole package — `Vary` appears nowhere.

Failure scenario: the allowlist matches multiple origins (it merges `CORS_ALLOWED_ORIGINS` + `CORS_HOSTING_ORIGINS`, each multi-pattern). Origin A sends a preflight → cached with `Access-Control-Allow-Origin: A` for 7200s → any shared cache keyed on method+path (a CDN, or a corporate forward proxy) can then serve that A-specific grant to origin B's preflight. B's browser sees an ACAO that doesn't match its origin and blocks — or, depending on cache/proxy behavior, the reverse leak. This is the same class of defect as the 504 reflection you're fixing (reflecting an origin without the controls that make reflection safe), so this PR is the right place to raise it.

Minimal fix: add `'Vary': 'Origin'` to the CORS header set (ideally inside `buildCorsHeaders` so every reflected-origin response gets it uniformly). That's the standard mitigation for per-origin reflected ACAO. If preflights genuinely shouldn't be shared-cached at all, add `Cache-Control` too — but `Vary: Origin` is the load-bearing one given you're keeping a 2h Max-Age.

packages/core/src/cors.ts
```diff
@@ -72,6 +72,30 @@ export function isOriginAllowed(origin: string): boolean {
...
+ *
+ * @param origin - The `Origin` header value from the request (may be empty)
+ * @returns The CORS headers to merge into the response (empty when not allowed)
+ */
+export function buildCorsHeaders(origin: string): Record<string, string> {
+  const headers: Record<string, string> = {};
+  if (isOriginAllowed(origin)) {
+    headers['Access-Control-Allow-Origin'] = origin;
+    headers['Access-Control-Allow-Credentials'] = 'true';
+  } else if (origin) {
```
osama-rizk, line 125: Now that `buildCorsHeaders` is the shared entry point called on **every** request path (504, 404, OPTIONS, RPC, raw-route), this `console.warn` on the disallowed-origin branch fires once per request. Previously the reflective 504 path didn't warn; now a disallowed origin that retries — or any bot spraying the endpoint with a bogus `Origin` — produces one warn line per request. On Lambda that's CloudWatch Logs volume (cost + noise); it's a mild log-amplification vector an external caller can trigger at will.

Not a blocker, but worth a beat: consider warning only once per distinct origin (a small `Set` of already-warned origins), or dropping it to a debug-level log, or moving it to the single 403-rejection site rather than the per-header-build helper. The message is genuinely useful for a developer misconfiguring their allowlist — the concern is only its per-request cardinality now that the helper is on the hot path for all responses.

packages/core/src/scripts/dev-server.ts
```diff
@@ -859,7 +859,7 @@ export async function startDevServer(options: DevServerOptions) {
     res.setHeader('Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, PATCH, HEAD, OPTIONS');
     res.setHeader('Access-Control-Allow-Headers', 'Content-Type');
     res.setHeader('Access-Control-Allow-Credentials', 'true');
-    res.setHeader('Access-Control-Max-Age', '86400');
+    res.setHeader('Access-Control-Max-Age', '7200');
```
osama-rizk: Test-coverage note (minor, dev-only): this Max-Age change — and the dev server's CORS behavior generally — has no test; `startDevServer`'s header block isn't exercised anywhere in `packages/core/src/**/*.test.ts`. Low stakes because it's a local dev tool, not shipped runtime, and the value simply mirrors the Lambda path. Flagging only for symmetry: the Lambda `7200` is pinned by `cors.test.ts`, so if someone later reverts one they won't get a failing test on the other. Not worth a full dev-server harness for this alone — a shared constant (`CORS_MAX_AGE = '7200'`) imported by both sites would be the cheaper way to keep them from drifting.

packages/core/src/lambda-handler.ts
osama-rizk: Minor durability nit (separate from the cache comment): `7200` is a magic number in three places now (here, dev-server, and the `cors.test.ts` expectation), and the *reason* it's 7200 — the strictest mainstream browser cap, per the PR description — lives only in the PR body, not the code. A future dev sees `86400` in an MDN example and 'fixes' this back up. A named constant with a one-line comment (`// Chromium caps preflight cache at 7200s; higher values are silently clamped`) captures the rationale where it'll actually be read and removes the duplication. Cheap insurance against a well-meaning regression.

soberm: Good catch, and you're right that this is the natural place for it — same root cause as the 504 path, just with the cache as the amplifier instead of the response itself. Fixed in 6b22f4c.

I put `Vary: 'Origin'` inside `buildCorsHeaders` as you suggested, so every reflected-origin response gets it uniformly rather than me trying to remember each call site. One thing I did differently: it's set unconditionally, including on the *disallowed* and empty-origin paths, not just when we reflect. The absence of an `Access-Control-Allow-Origin` header is itself origin-dependent, so a cache that stores the header-less variant under a non-Vary key can pin a "no grant" response and starve an origin that *should* be allowed. Emitting `Vary` on both branches means there's no variant of this response that a shared cache can key incorrectly. That's why `buildCorsHeaders` no longer returns `{}` — the two tests that asserted `deepStrictEqual(headers, {})` now assert `{ Vary: 'Origin' }`, which I think reads more honestly anyway since "no CORS grant" and "no headers at all" were being conflated.

I added it to `corsRejection()` too — the 403 is origin-dependent by definition.

I left `Cache-Control` off deliberately. `Vary: Origin` makes the 7200s cache correct rather than unsafe, and a private per-origin preflight cache is the thing Max-Age exists for; adding `no-store` would negate the Max-Age we just spent effort tuning. If we later decide preflights shouldn't be shared-cached at all, that's a clean follow-up, but it'd be an argument for dropping Max-Age entirely rather than carrying both.

Verified end-to-end against a real handler invocation and a real HTTP server rather than just the unit tests:

```
OPTIONS, allowed origin   -> Vary: Origin, ACAO: https://good.example.com, Max-Age: 7200
disallowed origin (403)   -> Vary: Origin, no ACAO
504 timeout, allowed      -> Vary: Origin, ACAO present
dev server, localhost     -> ACAO=http://localhost:5173  Vary=Origin  Max-Age=7200
dev server, evil.com      -> ACAO=http://localhost:3000  Vary=Origin  Max-Age=7200
```

New tests pin `Vary` on all three branches (allowed / disallowed / no origin) and on the OPTIONS response.

packages/core/src/cors.ts
soberm: Fair point, and the per-request cardinality is a real change in behavior that I introduced by moving this onto the hot path — worth fixing rather than deferring. Fixed in 6b22f4c.

Went with your first suggestion: a `Set` of already-warned origins, so it warns once per *distinct* origin. I kept it at warn level rather than dropping to debug because, as you say, the message is genuinely the thing that unblocks a developer who's misconfigured their allowlist — at once-per-origin the volume is no longer the problem, so there's no reason to make it harder to find.

I didn't move it to the 403 site, because that would lose the case that actually motivated this PR: when **no** allowlist is configured, the 403 gate at line 447 is skipped entirely (`patterns` is falsy), so a disallowed origin never reaches the rejection path — that's the hole the 504 fix closes. The warning is arguably most valuable in exactly that configuration, since "I set no allowlist and everything is silently denied" is the confusing state. Keeping it in the helper means it still fires there.

One addition beyond what you asked for: the `Set` is capped at 100 entries. An unbounded set keyed on attacker-controlled input just converts a log-volume vector into a memory-growth vector across a warm Lambda's lifetime — spraying distinct random origins would grow it forever. Past the cap it still warns but stops recording, so the worst case degrades back to the current behavior for novel origins instead of retaining them. `_resetCorsPatterns()` clears the set as well, so test isolation holds.

Measured it — 200 identical requests from one bogus origin through the real handler:

```
200 requests from one bogus origin -> 1 warn line(s)
```

Two tests cover it: one asserting repeat-origin warns once while a distinct origin still warns, and one asserting allowed and empty origins never warn.

packages/core/src/scripts/dev-server.ts
soberm: Agreed on both the diagnosis and the remedy you proposed — the drift risk is the real issue here, not the missing dev-server harness. Fixed in 6b22f4c.

Took the shared-constant route you suggested: `CORS_MAX_AGE = '7200'` now lives in `cors.ts` and is imported by both the Lambda handler and the dev server, so there's exactly one value to change and no way to revert one site without the other following.

I also went slightly further on the testability point, because a constant alone still wouldn't have caught a dev-server-only regression: I extracted the dev server's header block into an exported `buildDevCorsHeaders(requestOrigin)` and had the request handler apply it. That's a small refactor, not a harness — no server startup, no pidfiles, no process spawning — but it makes the dev server's actual header set assertable. `dev-server-cors.test.ts` now pins that its Max-Age *is* `CORS_MAX_AGE` (so the two sites are tied together by a test, not just by convention) and that it sets `Vary: Origin`, plus reflection and non-reflection cases.

So the symmetry gap you flagged is closed from both directions now: shared constant prevents the drift, and the dev-server assertion catches it if someone edits `buildDevCorsHeaders` directly.

Confirmed against a real HTTP server on an ephemeral port, not just the pure function:

```
origin=http://localhost:5173 -> ACAO=http://localhost:5173 Vary=Origin MaxAge=7200
origin=https://evil.com      -> ACAO=http://localhost:3000 Vary=Origin MaxAge=7200
```

packages/core/src/lambda-handler.ts
soberm: This is the comment that convinced me — "a future dev sees `86400` in an MDN example and 'fixes' this back up" is exactly the failure mode, and the rationale living only in the PR body is how it happens. Fixed in 6b22f4c.

Used essentially your wording for the comment, since it captures the mechanism (silent clamping) rather than just the number:

```ts
/**
 * `Access-Control-Max-Age` for preflight responses, in seconds.
 *
 * Chromium caps the preflight cache at 7200s and silently clamps anything
 * higher, so a larger value buys nothing while widening the window in which a
 * stale per-origin grant can be served. Shared by the Lambda handler and the
 * local dev server so the two can't drift.
 */
export const CORS_MAX_AGE = '7200';
```

I added the second clause about the staleness window because it ties this to your cache comment above — post-`Vary` the 7200 is safe rather than dangerous, but a longer window is still a longer window in which a per-origin grant sits in a cache, so "higher buys nothing" and "higher costs something" are both true. Wanted the next reader to see both halves and not just conclude the cap is a browser quirk they can ignore.

All three duplicate sites now reference the constant: the Lambda handler, the dev server, and `cors.test.ts`. The test still pins the literal in one place (`assert.strictEqual(CORS_MAX_AGE, '7200')`) so a silent change to the constant is caught — but the OPTIONS assertion compares against `CORS_MAX_AGE` rather than a second hardcoded `'7200'`, so there's one intentional tripwire instead of three copies drifting.

osama-rizk (approved):

Simone319 (approved):

PR #316, opened 2026-08-06
feat(bb-kv-store): add opt-in TTL support and use it to expire AuthCognito sessions

soberm:
## Problem

Two related gaps, one of which is a confirmed bug-bash finding.

**1. `sessions-table-no-TTL` (bug bash, medium)** — `AuthCognito`'s sessions DynamoDB table has no TTL, and every `SessionRecord` stores a **live Cognito refresh token**. Nothing removes a record unless the user explicitly signs out, so:

- the table grows without bound, and
- long-lived credentials are retained at rest indefinitely.

A session abandoned mid-flight (cookie dropped, browser wiped, device lost) leaves its refresh token sitting in the table forever. The code even carried a comment in `getCurrentUser` crediting a "TTL cleanup" path that did not exist.

**2. `KVStore` had no TTL surface at all** — so there was nothing for `AuthCognito` to opt into. Anything written to a `KVStore` lived forever; callers wanting bounded retention had to reach for `DistributedTable` (which has had `ttl` since day one) or hand-roll a cleanup job.

**Issue #, if available:** n/a (bug-bash finding: `sessions-table-no-TTL`)

### `@aws-blocks/bb-kv-store` (patch) — opt-in TTL, purely additive

Two independent opt-ins, both default off:

| Layer | API | Effect |
|---|---|---|
| Construct | `new KVStore(scope, id, { ttl: true })` | Sets `timeToLiveAttribute: 'ttl'` on the table |
| Write | `put(key, value, { ttlSeconds })` or `{ expiresAt }` | Writes a numeric epoch-seconds `ttl` attribute |

- **Backward compatible.** The `ttl` attribute is written *only* when asked for, so existing `put(k, v)` / `put(k, v, { ifNotExists })` calls produce byte-identical items. The third `put` parameter widens from `ConditionalWriteOptions` to `PutOptions`, which **extends** it — every existing call site still type-checks.
- **`ttl` defaults to `false`** because enabling TTL on a table that already exists is a CloudFormation update to the live table, so it must never happen implicitly. Mirrors `bb-distributed-table`'s `ttl` option; the attribute name is fixed to `ttl` (a `KVStore` item is always `{ pk, value }` with the payload opaque inside `value`, so there is no customer attribute to point at — this also matches the `timeToLiveAttribute: 'ttl'` convention already used in `hosting_construct.ts`).
- **Reads filter expired items in every runtime.** DynamoDB's reaper is asynchronous (up to 48 h), so an expired item can still be physically present. `get` returns `null` and `scan` skips it as soon as the expiry passes — otherwise an expired session would keep authenticating for up to two days.
- **The mock emulates expiry** (persists the expiry, prunes lazily on `get`/`scan`), keeping `src/parity.test.ts` parity real rather than approximate. Items without an expiry stay bare strings on disk, so existing `.bb-data` stores load unchanged.
- **One shared resolver** (`src/ttl.ts`) is used by both runtimes so they cannot drift on the value written or on when an item counts as expired. It rejects mutually-exclusive options, non-positive durations, and values large enough to be epoch **milliseconds** — DynamoDB would happily accept milliseconds as a year-5138 expiry, silently defeating the whole feature.

### `@aws-blocks/bb-auth-cognito` (patch) — consume it

- Sessions `KVStore` is now created with `{ ttl: true }`.
- Every session write (create **and** refresh) stamps `now + sessionTtlSeconds` — the same value that already bounds the session cookie's `Max-Age` (`DEFAULT_SESSION_TTL_SECONDS` = 400 d). Refresh slides the expiry forward, matching how the cookie is re-issued, so an active session is never reaped out from under a user.
- Removed the stale "TTL cleanup" comment.

> **This is a retention control, not an authorization one.** Validity is still decided by the access token's `exp` plus a `REFRESH_TOKEN_AUTH` round-trip on every request. A record surviving to its TTL grants nothing; this just reaps abandoned rows and bounds how long a refresh token sits at rest.

### `@aws-blocks/blocks` (patch)

Re-exports the new type as `KVPutOptions`, matching the existing `DTPutOptions` / `FBPutOptions` convention.

All three bumps are **patch**: these packages are pre-1.0, where a minor bump signals a breaking change. Nothing here is breaking — every addition is optional and defaults to the previous behaviour.

## Validation

**331 unit tests pass, 0 failures** (Node 22.22.3, per `.nvmrc`), from a forced clean rebuild (`dist/` + `tsconfig.tsbuildinfo` deleted first, so nothing was skipped by incremental compilation):

| Package | Result |
|---|---|
| `bb-kv-store` | **91/91 pass** (was 43 — +48 new) |
| `bb-auth-cognito` | **240/240 pass** (was 226 — +14 new) |
| `bb-auth-basic`, `bb-auth-oidc`, `bb-distributed-table` | 13/13, 112/112, 117/117 — other `KVStore` consumers, unaffected |

New coverage:

- **`bb-kv-store/src/ttl.test.ts`** (new) — option resolution (units, rounding, `Date` vs epoch seconds), every rejection path (both options set, non-positive, `NaN`/`Infinity`, non-numeric, invalid `Date`, epoch-milliseconds guard); mock expiry (`get`/`scan` filtering, disk pruning, expiry surviving a restart, re-put clearing an expiry, TTL + conditional writes, invalid options writing nothing); **AWS-runtime wire format** — drives the real `KVStore` from `index.aws.ts` against a real `DynamoDBDocumentClient`, intercepting only the network boundary via SDK middleware, asserting `ttl` is an integer in the right window, is absent when not requested, composes with `ConditionExpression`, and that `get`/`scan` filter unreaped expired items.
- **`bb-kv-store/src/parity.test.ts`** — a mock↔AWS parity block: neither runtime writes `ttl` unless asked, both hide expired items on `get` and `scan`, both still return future-dated items, and both reject the same invalid options with the same `error.name`.
- **`bb-kv-store/src/index.cdk.test.ts`** — `TimeToLiveSpecification` absent by default and with `{ ttl: false }`; present as `{ AttributeName: 'ttl', Enabled: true }` with `{ ttl: true }`; composes with `removalPolicy`; `fromExisting` still provisions nothing.
- **`bb-auth-cognito/src/sessions.test.ts`** — a session write stamps an integer `ttl` inside the configured lifetime; refresh slides it forward (verified by rewinding the persisted value and re-writing); an expired record is no longer returned; omitting `ttlSeconds` writes no expiry; a non-positive `ttlSeconds` means "no expiry" rather than instant deletion; `deleteSession` / `deleteByUsername` still work on TTL-stamped records.
- **`bb-auth-cognito/src/index.cdk.test.ts`** — a **real-CDK probe**. That file resolves `@aws-blocks/bb-kv-store` to its *mock* entry point, so the nested table never appears in its templates. The new test synths in a child process started with `--conditions=cdk` (the resolution a customer's `cdk synth` uses) and asserts the sessions table really carries `{ AttributeName: 'ttl', Enabled: true }` — i.e. it proves the actual security fix end-to-end, not just the wiring.

Also verified: `changeset-guard` `validate-structure` / `verify-coverage` / `block-major` all pass; `npm run lint:deps` clean (311 files); `api-extractor` regenerated and stable, so `check:api` passes; `biome lint` introduces no new findings.

Pre-existing and unrelated (identical with these changes stashed): `packages/hosting` and `packages/bb-knowledge-base` fail to build locally because `@aws-sdk/client-cloudfront-keyvaluestore` and `@aws-sdk/client-bedrock-agent` are declared in `package.json` but missing from this machine's `node_modules`; that also causes the 1 `packages/blocks` conditional-export test failure. `npm ci` in CI installs both.

## Checklist

- [x] PR description included

---

### DIFF
--- a/.changeset/kv-store-ttl-and-session-expiry.md
+++ b/.changeset/kv-store-ttl-and-session-expiry.md
@@ -1,0 +1,11 @@
+---
+"@aws-blocks/bb-kv-store": minor
+"@aws-blocks/bb-auth-cognito": minor
+"@aws-blocks/blocks": patch
+---
+
+Add opt-in TTL support to `KVStore` and use it to expire `AuthCognito` session records.
+
+`KVStore` gains a `ttl` construct option that enables DynamoDB Time-to-Live on the table, plus per-write expiry via `put(key, value, { ttlSeconds })` or `{ expiresAt }`. Both default to off, so existing tables and every existing `put()` call are unaffected. Because DynamoDB deletes expired items asynchronously, `get` and `scan` also filter expired items on read in every runtime, and the local mock emulates the same expiry semantics.
+
+`AuthCognito` now enables TTL on its sessions table and stamps each session write with `now + sessionTtlSeconds`. Session records store live Cognito refresh tokens, so without an expiry the table grew without bound and retained those credentials at rest indefinitely; abandoned sessions are now reaped automatically. Authorization is unchanged — validity is still decided by token revalidation on every request.
--- a/.changeset/kv-store-ttl-and-session-expiry.md
+++ b/.changeset/kv-store-ttl-and-session-expiry.md
@@ -6,6 +6,11 @@
 
 Add opt-in TTL support to `KVStore` and use it to expire `AuthCognito` session records.
 
-`KVStore` gains a `ttl` construct option that enables DynamoDB Time-to-Live on the table, plus per-write expiry via `put(key, value, { ttlSeconds })` or `{ expiresAt }`. Both default to off, so existing tables and every existing `put()` call are unaffected. Because DynamoDB deletes expired items asynchronously, `get` and `scan` also filter expired items on read in every runtime, and the local mock emulates the same expiry semantics.
+`KVStore` gains a `ttl` construct option that enables DynamoDB Time-to-Live on the table, plus per-write expiry via `put(key, value, { ttlSeconds })` or `{ expiresAt }`. Both default to off, so existing tables and every existing `put()` call are unaffected. Because DynamoDB deletes expired items asynchronously, `get` and `scan` also filter expired items on read in every runtime, and the local mock emulates the same expiry semantics. Maintenance sweeps that need to act on rows the reaper has not collected yet can opt out with `scan({ includeExpired: true })`.
 
 `AuthCognito` now enables TTL on its sessions table and stamps each session write with `now + sessionTtlSeconds`. Session records store live Cognito refresh tokens, so without an expiry the table grew without bound and retained those credentials at rest indefinitely; abandoned sessions are now reaped automatically. Authorization is unchanged — validity is still decided by token revalidation on every request.
+
+Two things to know before upgrading:
+
+- **Your next `cdk deploy` enables TTL on the existing sessions table.** Even though this is a patch, `AuthCognito` now passes `{ ttl: true }`, so the deploy issues a one-time `UpdateTimeToLive` against the live table. That is an online, non-disruptive DynamoDB operation — no downtime, no data loss — but it is a mutation of a live resource, so it shouldn't surprise anyone diffing a patch upgrade.
+- **Only sessions written after the upgrade expire.** A missing `ttl` attribute means "never expires" (matching DynamoDB), and existing rows are not backfilled, so sessions created before the upgrade keep their refresh tokens at rest indefinitely. Backfilling would need a one-time migration writing `ttl` onto every existing row. To close the retention gap immediately, revoke the pre-existing sessions instead — the revoke sweep deletes rows regardless of expiry state.
--- a/.changeset/kv-store-ttl-and-session-expiry.md
+++ b/.changeset/kv-store-ttl-and-session-expiry.md
@@ -1,6 +1,6 @@
 ---
-"@aws-blocks/bb-kv-store": minor
-"@aws-blocks/bb-auth-cognito": minor
+"@aws-blocks/bb-kv-store": patch
+"@aws-blocks/bb-auth-cognito": patch
 "@aws-blocks/blocks": patch
 ---
 
--- a/packages/bb-auth-cognito/src/index.cdk.ts
+++ b/packages/bb-auth-cognito/src/index.cdk.ts
@@ -283,8 +283,11 @@ export class AuthCognito<const O extends AuthCognitoOptions = AuthCognitoOptions
 		new AppSetting(this, 'session-secret', { secret: true });
 
 		// 5. Session store (KVStore). Propagate `removalPolicy` so retain-mode
-		// customers don't lose live sessions on stack delete.
-		this.sessions = new KVStore(this, 'sessions', { removalPolicy: opts.removalPolicy });
+		// customers don't lose live sessions on stack delete. TTL is enabled so
+		// session records — which hold live Cognito refresh tokens — expire with
+		// the session instead of accumulating forever; the runtime stamps each
+		// write with `now + sessionTtlSeconds`.
+		this.sessions = new KVStore(this, 'sessions', { removalPolicy: opts.removalPolicy, ttl: true });
 
 		// 6. Env vars + IAM
 		const fn = this.handler as lambda.Function;
--- a/packages/bb-auth-cognito/src/sessions.ts
+++ b/packages/bb-auth-cognito/src/sessions.ts
@@ -217,15 +217,34 @@ export interface SessionRecord {
  */
 export class SessionStore {
 	private kv: KVStore<SessionRecord>;
+	private ttlSeconds?: number;
 
-	constructor(scope: ScopeParent, id = 'sessions') {
+	/**
+	 * @param ttlSeconds - When set, every write stamps the record with a DynamoDB
+	 * TTL of `now + ttlSeconds` so abandoned sessions are reaped instead of
+	 * retaining Cognito refresh tokens at rest forever. Callers pass the session
+	 * lifetime, keeping the record's ceiling aligned with the session cookie's
+	 * `Max-Age`. Authorization itself is still decided by token revalidation on
+	 * every request — this only bounds retention.
+	 */
+	constructor(scope: ScopeParent, id = 'sessions', ttlSeconds?: number) {
 		this.kv = new KVStore<SessionRecord>(scope, id);
+		this.ttlSeconds = ttlSeconds !== undefined && ttlSeconds > 0 ? ttlSeconds : undefined;
+	}
+
+	/**
+	 * Expiry options for a write, or `undefined` when no TTL is configured.
+	 * Recomputed per write so refreshing a session slides its expiry forward,
+	 * matching how the session cookie's `Max-Age` is re-issued on refresh.
+	 */
+	private writeOptions(): { ttlSeconds: number } | undefined {
+		return this.ttlSeconds === undefined ? undefined : { ttlSeconds: this.ttlSeconds };
 	}
 
 	/** Insert a new session record; returns the generated session ID. */
 	async createSession(record: SessionRecord): Promise<string> {
 		const sessionId = crypto.randomBytes(24).toString('base64url');
-		await this.kv.put(sessionId, record);
+		await this.kv.put(sessionId, record, this.writeOptions());
 		return sessionId;
 	}
 
@@ -242,7 +261,7 @@ export class SessionStore {
 	async updateSession(sessionId: string, update: Partial<SessionRecord>): Promise<void> {
 		const existing = await this.kv.get(sessionId);
 		if (!existing) return;
-		await this.kv.put(sessionId, { ...existing, ...update });
+		await this.kv.put(sessionId, { ...existing, ...update }, this.writeOptions());
 	}
 
 	/**
--- a/packages/bb-auth-cognito/src/sessions.ts
+++ b/packages/bb-auth-cognito/src/sessions.ts
@@ -272,10 +272,15 @@ export class SessionStore {
 	 * `SessionRecord` is deliberately minimal (tokens only — see its docs), so
 	 * the owning username is recovered by decoding each session's ID token
 	 * rather than read from a denormalized field.
+	 *
+	 * The sweep scans with `includeExpired` because DynamoDB's reaper runs up to
+	 * ~48 h after expiry. A row past its TTL but still on disk holds a live
+	 * refresh token, and a deliberate revoke is the one moment retention matters
+	 * most — so it is deleted rather than skipped as already-invisible.
 	 */
 	async deleteByUsername(username: string): Promise<number> {
 		const toDelete: string[] = [];
-		for await (const { key, value } of this.kv.scan()) {
+		for await (const { key, value } of this.kv.scan({ includeExpired: true })) {
 			if (decodeIdToken(value.idToken).username === username) toDelete.push(key);
 		}
 		for (const id of toDelete) await this.kv.delete(id);
--- a/packages/bb-kv-store/DESIGN.md
+++ b/packages/bb-kv-store/DESIGN.md
@@ -25,9 +25,13 @@ TTL is opt-in at the construct (`{ ttl: true }`) and per write (`put(k, v, { ttl
 
 The attribute name is fixed to `ttl` rather than configurable like `DistributedTable`'s `ttl?: keyof T`. A `DistributedTable` item's attributes are the customer's schema fields, so TTL points at one of them; a `KVStore` item is always `{ pk, value }` with the customer payload opaque inside `value`, so there is no customer attribute to name. Fixing it also matches the `timeToLiveAttribute: 'ttl'` convention already used elsewhere in the repo.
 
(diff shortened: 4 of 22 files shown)


packages/bb-auth-cognito/src/sessions.ts
```diff
@@ -242,7 +261,7 @@ export class SessionStore {
 	async updateSession(sessionId: string, update: Partial<SessionRecord>): Promise<void> {
 		const existing = await this.kv.get(sessionId);
 		if (!existing) return;
-		await this.kv.put(sessionId, { ...existing, ...update });
+		await this.kv.put(sessionId, { ...existing, ...update }, this.writeOptions());
```
osama-rizk, line 264: `deleteByUsername` (just below, ~line 278) drives off `this.kv.scan()`, and `scan()` now filters expired items. So a session that is past its TTL but **not yet reaped** (the async reaper runs up to ~48 h late) is invisible to the revoke sweep and never gets deleted.

Failure scenario: an admin calls `revokeUserSessions` to force a user off. A recently-expired session row for that user is skipped by the scan → its **live Cognito refresh token stays at rest for up to 48 h**, and the returned deleted-count under-reports. Authorization is unaffected (`lookupSession` already returns null for it), so this is a retention leak rather than an auth hole — but it undercuts the exact goal of this PR at the one moment retention matters most: a deliberate revoke.

Consider having `deleteByUsername` iterate raw rows (a scan path that does *not* filter on expiry) so a revoke physically deletes everything still present.

Also: the existing `deleteSession and deleteByUsername still work on TTL-stamped records` test only exercises *live* records, so this expired-but-unreaped case is currently uncovered — worth a test that rewrites a row's `ttl` into the past and asserts `deleteByUsername` still deletes it.

osama-rizk (commented):
Reviewed the TTL work end-to-end. This is high-quality: the shared `ttl.ts` resolver keeping both runtimes from drifting, the epoch-**milliseconds** guard (rejecting a `Date.now()` that DynamoDB would silently accept as a year-5138 "never expires"), read-side filtering for the async reaper, and the backward-compatible on-disk shape are all things that are easy to get wrong and you got right.

Special call-out for `synthUnderCdkConditions` in `index.cdk.test.ts`: spawning a real `--conditions=cdk` child synth to assert the sessions table actually carries `{ AttributeName: 'ttl', Enabled: true }` is exactly the right paranoia. The normal test file resolves nested blocks to their mocks, so a naive test would pass while the real template shipped no TTL. Testing what the customer deploys rather than what the harness sees is worth keeping as a pattern.

I checked for duplication (none — the resolver lives only in this package) and for the same fix in sibling data blocks; that surfaced one systemic item on `bb-distributed-table`, filed inline on DESIGN.md. My other comments are at the deployment/scope altitude — the code itself is correct and well-tested; the open questions are about rollout reality and matching the changeset's claims to what actually ships. Nothing here blocks; a couple are worth a sentence in the changeset.

packages/bb-kv-store/DESIGN.md
```diff
@@ -14,10 +14,21 @@ Creates a single DynamoDB table:
...
+- **Time-to-Live:** Disabled by default; `{ ttl: true }` sets `timeToLiveAttribute: 'ttl'`
 - **Permissions:** `grantReadWriteData` to the parent scope's handler automatically
 
 No sort key, no GSIs. This is intentional — `KVStore` is the simple case. Customers needing sort keys or secondary indexes should use `DistributedTable`.
 
+## Expiry (TTL)
+
+TTL is opt-in at the construct (`{ ttl: true }`) and per write (`put(k, v, { ttlSeconds })` or `{ expiresAt }`), because switching TTL on for a table that already exists is a CloudFormation update to the live table.
+
+The attribute name is fixed to `ttl` rather than configurable like `DistributedTable`'s `ttl?: keyof T`. A `DistributedTable` item's attributes are the customer's schema fields, so TTL points at one of them; a `KVStore` item is always `{ pk, value }` with the customer payload opaque inside `value`, so there is no customer attribute to name. Fixing it also matches the `timeToLiveAttribute: 'ttl'` convention already used elsewhere in the repo.
```
osama-rizk, line 26: Sibling-block consistency: this same DESIGN.md (line ~30) argues that because DynamoDB's reaper is async, **both runtimes must filter expired items on read** — otherwise an expired session keeps authenticating for up to 48 h. That reasoning is sound and is the correctness backbone of this PR.

But `bb-distributed-table` — which this doc references as the block with configurable `ttl?: keyof T` — sets `timeToLiveAttribute` and does **no read-side filtering**: `get()` returns `reconcileRead(result.Item)` with no expiry check, and `query`/`scan`/`getBatch` likewise, and its mock doesn't emulate the reaper at all. Its own docstring even uses a `sessions` table with `ttl: 'expiresAt'` as the example. So a `DistributedTable` consumer using TTL for sessions has exactly the bug this PR fixes for `KVStore`: expired items keep reading as live for up to 48 h.

Not asking you to fix DistributedTable in this PR — but it should get a tracking issue, and it'd be worth a line here noting the read-filter guarantee is KVStore-only today so nobody assumes DistributedTable's `ttl` is equally safe.

packages/bb-auth-cognito/src/index.cdk.ts
```diff
@@ -283,8 +283,11 @@ export class AuthCognito<const O extends AuthCognitoOptions = AuthCognitoOptions
 		new AppSetting(this, 'session-secret', { secret: true });
 
 		// 5. Session store (KVStore). Propagate `removalPolicy` so retain-mode
-		// customers don't lose live sessions on stack delete.
-		this.sessions = new KVStore(this, 'sessions', { removalPolicy: opts.removalPolicy });
+		// customers don't lose live sessions on stack delete. TTL is enabled so
+		// session records — which hold live Cognito refresh tokens — expire with
+		// the session instead of accumulating forever; the runtime stamps each
+		// write with `now + sessionTtlSeconds`.
+		this.sessions = new KVStore(this, 'sessions', { removalPolicy: opts.removalPolicy, ttl: true });
```
osama-rizk, line 290: The KVStore design deliberately makes `ttl` opt-in because "enabling TTL on a table that already exists is a CloudFormation update to the live table, so it must never happen implicitly." This line then does exactly that implicitly for every existing AuthCognito consumer — their next `cdk deploy` after a *patch* bump issues an `UpdateTimeToLive` on the live sessions table.

I checked myself here: enabling TTL is an online, non-disruptive operation (no downtime, no data loss), so this is safe and I'm not asking to revert it. But "safe" and "silent in a patch" are different claims — an operator diffing a patch upgrade shouldn't be surprised by a live-table mutation. Worth one line in the changeset's notes calling out that upgrading triggers a one-time TTL-enable on the existing sessions table.

.changeset/kv-store-ttl-and-session-expiry.md
```diff
@@ -0,0 +1,11 @@
...
+"@aws-blocks/bb-kv-store": patch
+"@aws-blocks/bb-auth-cognito": patch
+"@aws-blocks/blocks": patch
+---
+
+Add opt-in TTL support to `KVStore` and use it to expire `AuthCognito` session records.
+
+`KVStore` gains a `ttl` construct option that enables DynamoDB Time-to-Live on the table, plus per-write expiry via `put(key, value, { ttlSeconds })` or `{ expiresAt }`. Both default to off, so existing tables and every existing `put()` call are unaffected. Because DynamoDB deletes expired items asynchronously, `get` and `scan` also filter expired items on read in every runtime, and the local mock emulates the same expiry semantics.
+
+`AuthCognito` now enables TTL on its sessions table and stamps each session write with `now + sessionTtlSeconds`. Session records store live Cognito refresh tokens, so without an expiry the table grew without bound and retained those credentials at rest indefinitely; abandoned sessions are now reaped automatically. Authorization is unchanged — validity is still decided by token revalidation on every request.
```
osama-rizk, line 11: `isExpired` correctly treats a missing/non-numeric `ttl` as never-expired (mirroring DynamoDB). The consequence: every session row written **before** this change has no `ttl` attribute and is therefore **never reaped** — only sessions created after the upgrade get stamped.

So "the table grew without bound ... abandoned sessions are now reaped automatically" is only true going forward. For any already-deployed app, the pre-existing pile of abandoned sessions — the ones the `sessions-table-no-TTL` bug bash was actually worried about — keeps its refresh tokens at rest indefinitely. That's a defensible scope call (backfill would need a one-time migration writing `ttl` onto existing rows), but the changeset should say so plainly rather than implying the leak is fully closed. Suggest: "Applies to sessions written after upgrade; pre-existing rows are not backfilled."

packages/bb-kv-store/src/ttl.ts
```diff
@@ -0,0 +1,80 @@
...
+		}
+		return nowEpochSeconds() + Math.ceil(ttlSeconds);
+	}
+
+	if (expiresAt instanceof Date) {
+		const ms = expiresAt.getTime();
+		if (!Number.isFinite(ms)) invalidTtl('put: `expiresAt` is an Invalid Date');
+		return Math.floor(ms / 1000);
+	}
+	if (typeof expiresAt !== 'number' || !Number.isFinite(expiresAt) || expiresAt <= 0) {
```
osama-rizk: Minor boundary inconsistency (non-blocking): this rejects `expiresAt <= 0`, yet a test documents `expiresAt` in the past as valid ("caller asked for immediate expiry", using `now - 60`). So `now - 60` is accepted as immediate-expiry but `0` — epoch 1970, equally unambiguously in the past — throws. If past-dates-mean-immediate-expiry is the intended contract, the floor should just be `Number.isFinite` plus the milliseconds guard, not `> 0`. Trivial; flagging only because the boundary isn't principled as written.

packages/bb-auth-cognito/src/sessions.ts
soberm: Good catch — this was a real gap, and the failure scenario you sketched is exactly right.

Fixed in 85ce0c3. `scan` now takes `{ includeExpired }` in both runtimes, and `deleteByUsername` uses it, so a revoke physically deletes every row still present rather than skipping ones the reaper has not collected yet. It is opt-in rather than the default because a read answering "is this still valid?" must never see an expired item — only maintenance sweeps want the unfiltered view.

Also added the test you described (`deleteByUsername physically deletes a session past its ttl that the reaper has not collected`): it rewrites a row's `ttl` into the past, asserts the row is still on disk, then asserts the revoke deletes it and counts it. I checked it is not vacuous — reverting `deleteByUsername` to the plain `scan()` makes it fail.

One wrinkle worth mentioning: the assertion order matters, because the mock's `get` prunes an expired row as a side effect. Calling `lookupSession` first would have deleted the row and left the test asserting nothing, so the test reopens the store instead.

packages/bb-kv-store/src/ttl.ts
soberm: Agreed, the boundary was not principled as written.

Fixed in 85ce0c3: dropped the `> 0` floor on `expiresAt`. It is an instant rather than a duration, so any past timestamp — epoch 0 included — now reads as "expire immediately", which is the contract the existing test already documented. The check is now just `Number.isFinite` plus the unchanged milliseconds guard.

`ttlSeconds` keeps its positive-only rule, since a non-positive duration really is meaningless there rather than just being in the past.

Covered by a new case asserting `0`, `-1` and `new Date(0)` all resolve rather than throw, and noted in DESIGN.md so the asymmetry between the two options is deliberate and written down.

packages/bb-kv-store/DESIGN.md
soberm: Added a line in 85ce0c3, wording it as a KVStore-only guarantee so nobody reads across to the sibling block:

> This read-side filtering is a `KVStore` guarantee only. `DistributedTable` also supports TTL (`ttl?: keyof T`) but sets `timeToLiveAttribute` without filtering reads, so an expired item there keeps reading as live until the reaper collects it — up to 48 hours.

Leaving the `DistributedTable` fix itself out of this PR as you suggested. Its docstring using a `sessions` table with `ttl: 'expiresAt'` as the example makes it worse than a latent issue, since that is the one use case where reading an expired row as live actually matters — worth a tracking issue.

Also documented the new `scan({ includeExpired: true })` escape hatch in the same section, since it is the other half of the read-path story.

packages/bb-auth-cognito/src/index.cdk.ts
soberm: That is a fair distinction — "safe" and "silent in a patch" are not the same claim, and thanks for checking the online/non-disruptive part yourself.

Called it out in the changeset in 85ce0c3:

> **Your next `cdk deploy` enables TTL on the existing sessions table.** Even though this is a patch, `AuthCognito` now passes `{ ttl: true }`, so the deploy issues a one-time `UpdateTimeToLive` against the live table. That is an online, non-disruptive DynamoDB operation — no downtime, no data loss — but it is a mutation of a live resource, so it should not surprise anyone diffing a patch upgrade.

Kept the behaviour as-is, since making it opt-in would mean the retention fix does nothing until every consumer opts in, which defeats the point of the PR.

.changeset/kv-store-ttl-and-session-expiry.md
soberm: You are right, the wording overclaimed. Reworded in 85ce0c3, close to your suggestion:

> **Only sessions written after the upgrade expire.** A missing `ttl` attribute means "never expires" (matching DynamoDB), and existing rows are not backfilled, so sessions created before the upgrade keep their refresh tokens at rest indefinitely. Backfilling would need a one-time migration writing `ttl` onto every existing row.

Kept it as a scope call rather than adding a migration here, but pointed at revoking the pre-existing sessions as the way to close the gap immediately — which is only a real answer because of the `includeExpired` fix on the other thread. Before that, revoke would have skipped exactly the rows an operator was trying to clear.

osama-rizk (approved):

Simone319 (approved):

PR #319, opened 2026-08-06
fix(ci): resolve peer dep conflicts in dependency update workflow

svidgen:
## Problem

The **Weekly Dependency Update** workflow ([failing run](https://github.com/aws-devtools-labs/aws-blocks/actions/runs/30686913674/job/91334440065)) fails at the 'Install and build' step with `ERESOLVE` errors.

**Root cause:** `ncu -u --root --workspaces` bumps `aws-cdk-lib` to latest (e.g. 2.263.0) in the root, but `@guide/bb-queue` declared an exact-pinned peer dependency on `aws-cdk-lib: "2.257.0"`. npm strict peer dep resolution (enforced by `.npmrc`) rejects the mismatch.

## Fix

1. **Relax the peer dep** in `test-apps/extending-blocks-guide/packages/bb-queue/package.json`: change `"aws-cdk-lib": "2.257.0"` → `"^2.257.0"`. CDK is semver-compatible within major version 2, and all other workspace packages already use caret ranges.

2. **Add `--legacy-peer-deps`** to the workflow's `npm install` step as a safety net. Since `ncu` aggressively bumps all deps to latest, transient peer conflicts may arise that don't reflect real incompatibilities. The build and e2e steps still validate correctness.

## Verification

- `npm install` succeeds locally after the peer dep change (no ERESOLVE)
- Only `package.json` peer dep ranges and the workflow file are modified — no source code changes

### DIFF
--- a/.github/workflows/dependency-update.yml
+++ b/.github/workflows/dependency-update.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Install and build
         run: |
-          npm install
+          npm install --legacy-peer-deps
           npm run build
 
       - name: Verify with local e2e
--- a/test-apps/extending-blocks-guide/packages/bb-queue/package.json
+++ b/test-apps/extending-blocks-guide/packages/bb-queue/package.json
@@ -31,7 +31,7 @@
   },
   "peerDependencies": {
     "@aws-blocks/core": "*",
-    "aws-cdk-lib": "2.257.0",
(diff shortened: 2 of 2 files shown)


osama-rizk (approved):

svidgen:
Superseded by #324 which covers this fix plus related dependency management improvements (template version stamping, consistent caret ranges).

PR #324, opened 2026-08-06
fix: dependency version management — stamp versions, fix peer dep pins

svidgen:
## Summary

Fixes two dependency version management issues:

### 1. Weekly dependency update workflow failing

The [dependency update workflow](https://github.com/aws-devtools-labs/aws-blocks/actions/runs/30686913674/job/91334440065) was failing because `ncu` bumped root `aws-cdk-lib` to latest, but `@guide/bb-queue` (a workspace test-app) had an exact peer dep pin on `"2.257.0"` → ERESOLVE. Fixed by changing to caret range (`"^2.257.0"`), consistent with every other package in the repo. Same fix applied to the backend template.

### 2. Template wildcard `*` deps → stamped at scaffold time

Templates ship with `"@aws-blocks/blocks": "*"` in the repo (resolves to workspace version during development). But when customers scaffold from npm, that `*` ended up in their package.json — accepting any version including future breaking majors.

**How it works now:**
1. `@aws-blocks/blocks` is listed as a devDependency in create-blocks-app (changesets auto-bumps it whenever blocks publishes)
2. At scaffold time, the CLI reads that devDep and stamps it into the customer's template
3. If the devDep is missing, it throws with a clear error + GitHub issue URL (it's a packaging bug)

### Scenarios

| Scenario | @aws-blocks/blocks resolution |
|----------|------|
| Customer scaffolds from npm | Caret range from create-blocks-app's devDep (e.g., `^0.2.x`) ✅ |
| Developer scaffolds from local monorepo | `file:` path via monorepo detection (existing behavior) ✅ |

### Changes

- `packages/create-blocks-app/src/index.ts` — Read devDep version, stamp into template wildcard
- `packages/create-blocks-app/src/cli.test.ts` — Verify dep is stamped (not `*`)
- `packages/create-blocks-app/package.json` — Add `@aws-blocks/blocks` devDep (changesets coupling)
- `packages/create-blocks-app/templates/backend/package.json` — `aws-cdk-lib` exact → caret
- `test-apps/extending-blocks-guide/packages/bb-queue/package.json` — `aws-cdk-lib` peerDep exact → caret


pranavosu (dismissed):

packages/create-blocks-app/src/index.ts
```diff
@@ -343,21 +343,17 @@ async function createFreshProject(targetDir: string, templateName: string, skipI
...
+  // replace with the current published version (caret range) so customers
+  // get predictable installs.
+  const cliPkgPath = join(__dirname, '../package.json');
+  const cliPkg = JSON.parse(await readFile(cliPkgPath, 'utf-8'));
+  const blocksVersion = cliPkg.version;
+  for (const section of ['dependencies', 'devDependencies'] as const) {
+    for (const [dep, ver] of Object.entries((pkg[section] ?? {}) as Record<string, string>)) {
+      if (dep.startsWith('@aws-blocks/') && ver === '*') {
+        pkg[section][dep] = `^${blocksVersion}`;
+      }
```
svidgen: Looks off to me actually. Digging into this more. Do not merge.

pranavosu (approved):

ikenyal (changes requested):
The implementation looks correct in a published-CLI layout: I copied the built package outside the monorepo, scaffolded the default and backend templates, and confirmed that @aws-blocks/blocks is stamped as ^0.2.5 (and the backend template uses aws-cdk-lib: ^2.257.0).

However, the added regression test only runs from the monorepo. In that layout isLocalMonorepo is always true, so it exercises the file: rewrite; the ^|file: assertion permits both outcomes and does not verify the new published-package branch. Since stamping the non-wildcard published version is the central behavioral change, please add automated coverage that exercises the non-monorepo path (or isolate that resolution so it can be tested directly).

PR #325, opened 2026-08-07
fix(create-blocks-app): prevent 'aws' prefix in generated resource names

svidgen:
## Problem

A user reported that AWS Resource Groups reject names starting with "AWS" as reserved. When scaffolding a project with a name like `aws-my-app`, the generated `stackId` starts with `aws-`, which flows into CloudFormation stack names and resource names, causing deployment failures.

## Fix

`generateStackId()` now strips a leading `aws-` prefix (case-insensitive) when it appears as a distinct prefix (followed by hyphen or end of string). Names like `awesome-app` where "aws" is part of a longer word are preserved.

**Before:** `aws-my-project` → stackId `aws-my-project-a1b2c3` (❌ deployment fails)
**After:** `aws-my-project` → stackId `my-project-a1b2c3` (✅ deploys fine)

## Testing

- Added 3 tests verifying:
  - Fresh scaffold with `aws-` prefixed name produces a stackId without `aws` prefix
  - Existing project integration with `aws-` prefixed package name produces a stackId without `aws` prefix
  - Names where "aws" is part of a word (e.g., `awesome-app`) are not affected
- All 54 existing tests continue to pass

## Changeset

Patch on `@aws-blocks/create-blocks-app`

### DIFF
--- a/.changeset/fix-aws-prefix-resource-names.md
+++ b/.changeset/fix-aws-prefix-resource-names.md
@@ -1,0 +1,9 @@
+---
+"@aws-blocks/create-blocks-app": patch
+---
+
+Prevent generated stackId from starting with "aws"
+
+AWS reserves the "aws" prefix for resource names in many services (Resource Groups, IAM, etc.), so a project scaffolded with a name like `aws-my-app` would produce a stackId starting with "aws-", causing CloudFormation deployments to fail.
+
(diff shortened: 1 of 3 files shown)


svidgen:
Pursuing https://github.com/aws-devtools-labs/aws-blocks/pull/328 instead.

PR #326, opened 2026-08-07
feat(core): throttle the Blocks API Gateway stage by default

soberm:
## Finding (bug bash)

The repo sets **no API Gateway throttle anywhere**, so every Blocks deployment silently inherits the AWS **account-level default of 10,000 rps / 5,000 burst**.

The core Blocks REST API (`packages/core/src/cdk/blocks-backend.ts`) is internet-facing — authentication runs *in-Lambda*, not at the gateway — so an abused or hot endpoint can amplify Lambda, DynamoDB and Bedrock cost with no ceiling.

## Change

- `RestApi` `deployOptions` now sets `throttlingRateLimit` / `throttlingBurstLimit`, defaulting to **200 rps / 400 burst**. That is ample headroom for typical application traffic while capping cost amplification from abuse.
- New optional `throttling?: { rateLimit: number; burstLimit: number }` prop on `BlocksBackendProps` lets high-traffic deployments raise (or lower) the limit.
- `cachingEnabled: false` is unchanged.

Adding an optional prop is non-breaking → **patch** changeset for `@aws-blocks/core`.

## Scope

Intentionally limited to the shipped core Blocks REST API. `packages/hosting` (`cdn_construct.ts`) and `bb-realtime` are deliberately **not** touched here.

## Testing

Two CDK-assertion tests added to `packages/core/src/cdk/blocks-backend.test.ts`:

1. default synth carries `ThrottlingRateLimit: 200` / `ThrottlingBurstLimit: 400` on `AWS::ApiGateway::Stage`
2. an explicit `throttling` prop overrides it (2500 / 5000)

Full `@aws-blocks/core` suite green:

```
# tests 651
# suites 171
# pass 651
# fail 0
```

### DIFF
--- a/.changeset/apigw-default-throttle.md
+++ b/.changeset/apigw-default-throttle.md
@@ -1,0 +1,5 @@
+---
+"@aws-blocks/core": patch
+---
+
+Apply a default request throttle (200 rps / 400 burst) to the Blocks API Gateway stage instead of inheriting the AWS account default of 10,000 rps. Override it with the new optional `throttling` prop on `BlocksBackendProps` for high-traffic deployments.
--- a/.changeset/apigw-default-throttle.md
+++ b/.changeset/apigw-default-throttle.md
@@ -2,4 +2,4 @@
 "@aws-blocks/core": patch
 ---
 
-Apply a default request throttle (200 rps / 400 burst) to the Blocks API Gateway stage instead of inheriting the AWS account default of 10,000 rps. Override it with the new optional `throttling` prop on `BlocksBackendProps` for high-traffic deployments.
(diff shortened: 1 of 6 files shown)


osama-rizk:
Thanks for raising this, Michael — the cost-amplification concern on the internet-facing core API is spot on.

Heads-up on an overlap: #302 (`feat/shared-hardening-defaults`) sets up a **cross-block hardening mechanism** — a stack-level registry with `resolveApiThrottle` (plus access-logs / log-retention / PITR resolvers) so throttling is configured and named *one* way across every block. It ships the mechanism + bb-realtime as first adopter, and explicitly lists the core + SSR REST APIs as follow-up adopters — so #326 is really that follow-up.

Easiest path is probably to land #302's mechanism first, then rework this to adopt it. Thanks again for catching the gap. cc @svidgen @pranavosu

soberm:
Closing this PR as it overlaps with another PR.

PR #328, opened 2026-08-07
fix(core): reject stackId starting with "aws" at build time with clear guidance

svidgen:
## Problem

A stack named `aws-my-app` passes CDK synth but fails at deploy time. AWS Blocks creates `AWS::ResourceGroups::Group` resources named `{stackId}-resources` and `{stackId}-settings`. The AWS Resource Groups API rejects any group name starting with "aws" (case-insensitive) — it's a reserved prefix.

### The deploy-time error users see WITHOUT this fix

CDK synth passes. `cdk deploy` starts, creates most resources, then hits the Resource Group:

```
aws-my-app |  1/12 | 2:15:32 PM | CREATE_FAILED      | AWS::ResourceGroups::Group | StackResources
Resource handler returned message: "1 validation error detected: Value 'aws-my-app-resources'
at 'name' failed to satisfy constraint: Member must satisfy regular expression pattern:
^(?!(?i)aws)..." (Service: ResourceGroups, Status Code: 400,
Request ID: a1b2c3d4-..., HandlerErrorCode: InvalidRequest)

 ❌  aws-my-app failed: Error: The stack named aws-my-app failed creation, it may
 need to be manually deleted from the AWS console: ROLLBACK_COMPLETE:
 StackResources (AWS::ResourceGroups::Group) Resource handler returned message...
```

The user then has to manually delete the rolled-back stack from the AWS console before trying again.

### The build-time error users see WITH this fix

CDK synth fails immediately — no deployment attempted, nothing to clean up:

```
Error: Invalid stackId "aws-my-app": names beginning with "aws" (case-insensitive) are reserved
by AWS and cannot be used for CloudFormation stacks or resource names. Rename your stackId in
.blocks/config.json (e.g., "my-app").
```

## Changes

- `assertStackNameNotReserved()` in `packages/core/src/cdk/blocks-backend.ts` — validation function with docstring documenting the exact CFN error it prevents
- Called from `setupBlocksInfra()` (shared by both `BlocksStack` and `BlocksBackend`) — single call site
- Removed redundant calls from `BlocksStack.create()` and `BlocksBackend.create()`
- Tests exercise the function directly (case variations, positive cases) + one integration test confirming `BlocksStack.create()` wires the check

Replaces PR #325 which would have instead *sanitized* the prefix at scaffold time in a potentially confusing way. I'm happy to consider that in the future. But, simply rejecting early is safer.

### DIFF
--- a/.changeset/reject-aws-prefix-stack-name.md
+++ b/.changeset/reject-aws-prefix-stack-name.md
@@ -2,4 +2,4 @@
 "@aws-blocks/core": patch
 ---
 
-Reject stack names starting with "aws" (case-insensitive) at CDK synth time with a clear, actionable error message. AWS reserves this prefix for resource names across Resource Groups, IAM, and other services — a stack name starting with "aws" causes a cryptic CloudFormation deployment failure. The new build-time validation catches this early and tells users exactly what to change (the `stackId` in `.blocks/config.json`).
+Reject stack names starting with "aws" (case-insensitive) at CDK synth time with a clear, actionable error message. AWS reserves this prefix for Resource Group names — deploying a stack named "aws-..." fails when CloudFormation creates the stack's Resource Groups. The new build-time check catches this immediately and directs users to rename the stackId in `.blocks/config.json`.
--- a/.changeset/reject-aws-prefix-stack-name.md
+++ b/.changeset/reject-aws-prefix-stack-name.md
@@ -1,0 +1,5 @@
+---
+"@aws-blocks/core": patch
+---
+
+Reject stack names starting with "aws" (case-insensitive) at CDK synth time with a clear, actionable error message. AWS reserves this prefix for resource names across Resource Groups, IAM, and other services — a stack name starting with "aws" causes a cryptic CloudFormation deployment failure. The new build-time validation catches this early and tells users exactly what to change (the `stackId` in `.blocks/config.json`).
--- a/packages/core/src/cdk/blocks-backend.ts
+++ b/packages/core/src/cdk/blocks-backend.ts
@@ -41,6 +41,24 @@ export function assertCdkConditionActive(): void {
   }
 }
 
+/**
+ * Validate that a stack/backend name does not start with "aws" (case-insensitive).
+ *
+ * Without this check, deployment fails at the AWS::ResourceGroups::Group resource with:
+ *   Resource handler returned message: "Group name must not start with 'AWS'"
+ *   (HandlerErrorCode: InvalidRequest)
+ * That error is unhelpful because it doesn't indicate which config to fix.
+ */
+export function assertStackNameNotReserved(name: string): void {
+  if (/^aws/i.test(name)) {
+    throw new Error(
+      `Invalid stackId "${name}": names beginning with "aws" (case-insensitive) are reserved ` +
+      `by AWS and cannot be used for CloudFormation stacks or resource names. ` +
+      `Rename your stackId in .blocks/config.json (e.g., remove the "aws-" prefix).`,
+    );
+  }
+}
+
 export interface BlocksBackendProps {
   backendHandlerPath: string;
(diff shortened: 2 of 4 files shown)


pranavosu (approved):

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -41,13 +41,47 @@ export function assertCdkConditionActive(): void {
...
+/**
+ * Shared infra setup — creates Lambda, API Gateway, and Resource Groups on the given scope.
+ * Called by both BlocksStack and BlocksBackend; centralizes all infrastructure creation
+ * including the aws-prefix validation (which guards the Resource Groups naming constraint).
+ */
 export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id?: string) {
+  // Validate before creating any constructs. The Resource Groups created below
+  // derive their name from the stack/backend id; the 'aws' prefix is reserved.
+  const stackName = id ?? cdk.Stack.of(scope).stackName;
+  assertStackNameNotReserved(stackName);
```
sarayev, line 83: **[Major]** I think this validates the wrong string in one of the two call paths.

Here we check `id ?? cdk.Stack.of(scope).stackName`, but the group name is actually assembled further down (~line 139) as:

```ts
let rootStack = cdk.Stack.of(scope);
while (rootStack.nestedStackParent) rootStack = rootStack.nestedStackParent;
const groupPrefix = (id && id !== rootStack.stackName) ? `${rootStack.stackName}-${id}` : rootStack.stackName;
// → `${groupPrefix}-resources` / `${groupPrefix}-settings`
```

For the common `BlocksStack.create('aws-my-app', …)` case this is fine — `id === rootStack.stackName`, so we're checking the same string that gets used.

The gap is the `BlocksBackend` path: `id` is usually a harmless construct id (e.g. `"Blocks"`), so the check passes — but if the enclosing root stack is named `aws-*`, the group becomes `aws-…-Blocks-resources` and deploy still fails with the exact error we're trying to prevent. Nested stacks have the same shape, since the check uses `cdk.Stack.of(scope)` while the name walks up to `rootStack`.

Running the assertion against the final `groupPrefix` (or `rootStack.stackName`) right where the groups are created would cover both paths.

.changeset/reject-aws-prefix-stack-name.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aws-blocks/core": patch
+---
+
+Reject stack names starting with "aws" (case-insensitive) at CDK synth time with a clear, actionable error message. AWS reserves this prefix for Resource Group names — deploying a stack named "aws-..." fails when CloudFormation creates the stack's Resource Groups. The new build-time check catches this immediately and directs users to rename the stackId in `.blocks/config.json`.
```
sarayev, line 5: **[Minor]** Small accuracy nit on the naming description: the docstring and changeset say the groups are named `{stackId}-resources`, but the real name comes from `groupPrefix`, which can be `{rootStack.stackName}-{id}-resources`. Worth aligning the wording with the actual logic so future readers (and the changelog) aren't misled.

packages/core/src/cdk/blocks-stack.test.ts
```diff
@@ -104,3 +104,42 @@ describe('assertCdkConditionActive', () => {
     }
   });
 });
+
+describe('assertStackNameNotReserved', () => {
+  // Validation is centralized in assertStackNameNotReserved() and called from both
```
sarayev, line 109: **[Minor]** The comment here notes the check fires from both `BlocksStack.create()` and `BlocksBackend.create()`, but only the `BlocksStack` path has a wiring test. A parallel `BlocksBackend.create(scope, 'aws-test', …)` test would lock in the second call path — and it'd likely surface the naming mismatch in the Major comment above, which is a nice bonus.

PR #329, opened 2026-08-07
[core + bb-lambda-compute] Add Compute abstraction and a Lambda compute package

Simone319:
## Problem

Multi-compute needs a first-class *compute* abstraction — a type that owns a runtime (a Lambda function + its API Gateway today, other runtimes later) — that the framework can build against. This is the foundational, non-breaking step: introduce the abstraction and add a Building Block for the Lambda compute, without changing the default app's behavior or exposing anything to customers.

## Changes

- **New abstract `Compute` in core (internal-only).** Added behind a new entry point `@aws-blocks/core/cdk/internal` (added to `package.json` `exports`) so these types are reachable by framework/test code but stay off the public `@aws-blocks/core` / `@aws-blocks/core/cdk` surface (api-extractor confirms they're absent from `API.md`). `Compute` is an abstract `Scope` subclass: it declares `namespaces` (routing seam, currently unpopulated) and `setEnv()`, and resolves its owning `BlocksStack`/`BlocksBackend` on construction to derive its runtime identity (backend entry + `BLOCKS_STACK_NAME`).
- **New `@aws-blocks/bb-lambda-compute` package** holding the concrete `LambdaCompute` (a `NodejsFunction` fronted by its own API Gateway, assuming the shared `executionRole`). Its `entry` and `BLOCKS_STACK_NAME` are **derived from the owning stack/backend**, never caller-supplied, so every compute runs the same backend and agrees on the resource-name namespace.
- **Conditional-export layers.** The backend module that will declare `new LambdaCompute(...)` is imported both at CDK synth and at runtime, so the package ships the standard layers: the `cdk` layer provisions the infrastructure (the only layer that touches `aws-cdk-lib`); the `aws-runtime`, local-dev (`default`/`types`), and `browser` layers are **inert handles** — they construct so the import succeeds and `{ compute }` references resolve, but provision nothing and `setEnv` is a no-op (config is injected at synth). This keeps `aws-cdk-lib` out of the runtime and browser bundles. `LambdaCompute` is **not re-exported from `@aws-blocks/blocks`**, so customers cannot reach it yet.
- **Consolidated the owner tree-walk on `Scope`.** A single `root` field (resolved once in the constructor) now backs `handler`, `executionRole`, and the `backendHandlerPath` / `backendStackName` accessors, replacing repeated copies of the walk.
- **Excluded `bb-lambda-compute` from customer-facing surfaces** consistent with it being internal infrastructure: the Building Block catalog (`scripts/sync-catalog.mjs` + `scripts/gen-block-docs.mjs` `EXCLUDED` sets) and the vendorize map check.

**Non-breaking:** pure additive dead code — nothing in the default path constructs a `LambdaCompute`, and customers cannot reach the compute types. App behavior is unchanged.

> **Note:** `@aws-blocks/bb-lambda-compute` is a new package but is not published until a release cuts it (changeset-driven).

## Validation

- `npm run build:force` (full monorepo) green.
- `@aws-blocks/core`: 654/654 pass, including internal `LambdaCompute` tests (function + gateway shape, shared-role assumption, owner-derived `BLOCKS_STACK_NAME`, distinct construct paths, and construction failure with no Blocks owner).
- `@aws-blocks/bb-lambda-compute`: 6/6 pass (uses an inline `StubBlocksStack` — no on-disk fixtures).
- `@aws-blocks/blocks`: 42/42 pass, including conditional-export parity (now auto-discovers bb-lambda-compute via its `aws-runtime` condition) and the vendorize-map check (bb-lambda-compute excluded as internal infrastructure).
- `npm run lint`, `npm run lint:deps`, `npm run check:api`, `npm run sync-docs:check` all clean.

## Checklist

- [x] PR description included

---

### DIFF
--- a/packages/bb-lambda-compute/src/index.cdk.ts
+++ b/packages/bb-lambda-compute/src/index.cdk.ts
@@ -1,0 +1,79 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { ScopeParent } from '@aws-blocks/core';
+import { DEFAULT_NODE_RUNTIME } from '@aws-blocks/core/cdk';
+import { BLOCKS_NAMESPACE, Compute } from '@aws-blocks/core/cdk/internal';
+import * as cdk from 'aws-cdk-lib';
+import * as apigateway from 'aws-cdk-lib/aws-apigateway';
+import * as lambda from 'aws-cdk-lib/aws-lambda-nodejs';
+import type { LambdaComputeProps } from './types.js';
+
+export type { LambdaComputeProps } from './types.js';
+
+/**
+ * A Lambda-backed {@link Compute}: a `NodejsFunction` fronted by its own API
+ * Gateway REST API. The compute *owns* these resources.
+ *
+ * The function assumes the shared execution role (`this.executionRole`), so
+ * Building Block grants reach it via that role. The handler entry and
+ * `BLOCKS_STACK_NAME` are **derived from the owning BlocksStack/BlocksBackend** —
+ * never caller-supplied — so every compute in an app runs the same backend and
+ * agrees on the runtime resource-name namespace.
+ *
+ * @internal Not exported from the package's public entry point. Customers
+ * cannot instantiate a compute until the customer-facing surface exists.
+ */
+export class LambdaCompute extends Compute {
+	/** The Lambda function backing this compute. */
+	readonly fn: lambda.NodejsFunction;
+	/** The API Gateway REST API fronting {@link fn}. */
+	readonly apiGateway: apigateway.RestApi;
+
+	constructor(scope: ScopeParent, id: string, _options?: LambdaComputeProps) {
+		super(id, { parent: scope });
+
+		// Entry + BLOCKS_STACK_NAME are derived from the owning stack/backend
+		// (resolved by Compute) — never caller-supplied — so every compute in an
+		// app runs the same backend and agrees on the resource-name namespace the
+		// runtime rebuilds from BLOCKS_STACK_NAME.
+		this.fn = new lambda.NodejsFunction(this, 'Handler', {
+			entry: this.backendHandlerPath,
+			runtime: DEFAULT_NODE_RUNTIME,
+			handler: 'handler',
+			role: this.executionRole,
+			memorySize: 2048,
+			timeout: cdk.Duration.seconds(60 * 15),
+			environment: {
+				NODE_ENV: 'production',
+				BLOCKS_STACK_NAME: this.backendStackName,
+			},
+			bundling: {
+				minify: true,
+				esbuildArgs: { '--conditions': 'aws-runtime' },
+			},
+		});
+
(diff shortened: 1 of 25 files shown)


packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -0,0 +1,79 @@
...
+			memorySize: 2048,
+			timeout: cdk.Duration.seconds(60 * 15),
+			environment: {
+				NODE_ENV: 'production',
+				BLOCKS_STACK_NAME: this.backendStackName,
+			},
+			bundling: {
+				minify: true,
+				esbuildArgs: { '--conditions': 'aws-runtime' },
+			},
```
svidgen, line 54: This might a good time to let some of these values become configurable at construction time.

Most critical:
1. `memorySize`
2. `timeout`

Less critical, but probably desirable at some point:
1. `environment`
2. `reservedConcurrentExecutions`
3. `provisionedConcurrentExecutions`

More dubious, but maybe for lower level tinkering/troubleshooting?
1. `bundling/minify`
2. `runtime`

(Reasonable as-is for fast-follow as well.)

packages/core/src/cdk/index.ts
svidgen, line 1: Just want to sanity check: Does the naming of anything change (destructively) before/after this change?

packages/blocks/src/vendorize-map.test.ts
```diff
@@ -46,8 +46,10 @@ describe('aws-blocks.vendorize map', () => {
     const missing: string[] = [];
 
     for (const dep of deps) {
-      // Skip infrastructure packages (not user-facing BBs)
-      if (dep.endsWith('/core') || dep.endsWith('/auth-common')) continue;
+      // Skip infrastructure packages (not user-facing BBs). bb-lambda-compute
+      // is the internal default compute — CDK-only and not customer-instantiable
+      // — so it has no public symbol to vendorize.
+      if (dep.endsWith('/core') || dep.endsWith('/auth-common') || dep.endsWith('/bb-lambda-compute')) continue;
```
svidgen, line 52: We don't let customers vendorize `LambdaCompute`?

svidgen (approved):
🚢

pranavosu (approved):

packages/bb-lambda-compute/src/index.cdk.ts
Simone319: Definitely, we should allow customers to configure the computes. Same for container. Will expose the configuration options in a followup PR.

packages/core/src/cdk/index.ts
Simone319: Not in this PR. The idea is that nothing about multi-compute is exposed to customers and the behavior remains the same until everything is ready. Only the final PR will modify the customer surface and expose the compute option to customers.

packages/blocks/src/vendorize-map.test.ts
Simone319: Now because multi-compute is under development, we don't verdorize it, we exclude it from the blocks catalog. When it's production ready, it should be vendorized, exported from blocks, and added to the blocks catalog.

PR #338, opened 2026-08-11
feat(bb-agent): compute-agnostic client streaming API (createChat)

pjkroker:
## Problem

The Agent BB's client streaming API is a three-step channel dance — `getChannel` → `subscribe` → await `established` → `stream` — with three problems it forces on every frontend developer:

1. **Ceremony.** Displaying one streamed answer requires understanding channels, subscriptions, and their lifecycle.
2. **A subscribe-before-send race.** The agent emits chunks the moment `stream()` is called, so subscribing *after* sending silently drops early chunks. The types don't prevent it — it's a runtime rule enforced only by documentation.
3. **A transport-coupled, leaky surface.** `stream()` returns `{ channelId, channel, complete }` and `resume(channelId, …)` takes a leading channel id — Realtime pub/sub artifacts that leak into call sites, so the streaming mechanism can't change without rewriting app code.

Design docs: [Doc 106 — Client Streaming API Redesign (createChat)](https://chorus.aws.dev/doc/tdC7sED3ryoh/Agent-BB--106--Client-Streaming-API-Redesign-createChat) and the runtime-strategy parent [Doc 104 — Runtime Options](https://chorus.aws.dev/doc/eFBzXXKnPi3j/Agent-BB--104---Runtime-Options-AgentCore-vs-Lambda-vs-Gener), whose section 3.3 names this redesign as the client-side enabler of the AgentCore-keeping-Realtime direction.

**Issue #, if available:** GA — [Agent BB][Docs] Client API Redesign

## Changes

Adds **`createChat`** — a compute/transport-agnostic client API — alongside the existing `useChat` (kept and `@deprecated`, non-breaking). Client-side only: the server methods (`stream`/`getChannel`/`resume`) are unchanged; the new transport is built on top of them.

- **`createChat({ transport, api, on* })`** (`packages/bb-agent/src/index.chat.ts`) — the common case is one call, `chat.sendMessage('Hello')`. `sendMessage` **fuses subscribe + run**, so the subscribe-before-send race is impossible by construction. HITL resume folds into the same call: `sendMessage({ interruptResponses })` (no separate `respondToInterrupt`). Also exposes `newConversation` / `loadConversation` / `getMessages` / `getConversationId` / `isLoading` / `destroy`.
- **Flexible primitives** `run()` (produce) and `subscribe()` (consume) — the easy default is sugar over these; they unlock fan-out, observer-only attach, and decoupled produce/consume.
- **The transport seam** `ChatTransport` (`packages/bb-agent/src/transport.ts`) — `run` + `subscribe`, the ONLY place a runtime is named. `realtimeTransport(io)` is the Lambda + Realtime implementation (bridges the callback Realtime channel into an `AsyncIterable` `ChunkStream`); a future runtime supplies a different transport with no client-code change.
- **Docs** — README gains a `createChat (recommended)` section (React example + copy-paste `realtimeTransport` wiring); `useChat` is relabeled deprecated; DESIGN.md gains a compute-agnostic client-streaming section. Changeset bumps `@aws-blocks/bb-agent` (minor) and `@aws-blocks/blocks` (patch, docs-sync gate).

Review-round fixes also included: tear down a subscription if `unsubscribe()` races before `io.subscribe` resolves (pre-resolution leak) and close the queue on attach failure; clear `loading` + fire `onError` + drop the empty assistant placeholder when attach/run rejects (so a failed turn can't wedge the chat); reuse the last channel id when resuming an inference-only interrupted turn; keep `run()` from mutating the `sendMessage` path's channel state.

## Validation

- **Unit tests** (`packages/bb-agent/src/index.test.ts`): `createChat` streams text and clears loading on `done`; interrupt fires `onInterrupt` and resume via `sendMessage({ interruptResponses })` continues the turn; a failing attach/run clears loading, fires `onError`, and leaves no empty assistant bubble. These run in the `Build, Unit Tests, E2E Local` lane.
- **Typecheck:** `tsc --build` clean for the changed `bb-agent` sources.
- **CI note:** the required unit/local lanes and the changeset gate are green. The remaining red `E2E Sandbox` / `E2E Production` / `Telemetry E2E` lanes fail at `cdk deploy` with `ServiceLimitExceeded: RolesPerAccount 1500` on the shared E2E account — an account IAM-role-quota exhaustion that stops the stack before any test runs, unrelated to this client-only diff. Re-run once role headroom is reclaimed.

## Checklist

- [x] PR description included

---


packages/bb-agent/src/index.chat.ts
```diff
@@ -0,0 +1,369 @@
...
+		},
+
+		subscribe(opts?: { channelId?: string; observer?: boolean }): ChunkStream {
+			const channelId = opts?.channelId ?? conversationId;
+			if (!channelId)
+				throw new Error('subscribe() needs a channelId — start a conversation first or pass one explicitly.');
+			return transport.subscribe(channelId, opts?.observer ? { observer: true } : undefined);
+		},
+
+		newConversation() {
```
soberm, line 364: [major] `newConversation()` mid-turn leaves `loading` stuck `true`, wedging all future sends.

`newConversation()` calls `activeStream.unsubscribe()`, which closes the `ChunkQueue`. That ends the background `for await` in `startTurn` **normally** (no `done`/`error`/`interrupt` terminal chunk), and that loop's `finally` only re-runs `unsubscribe()` + nulls `activeStream` — it never calls `setLoading(false)`. Since `createChat` is a reused factory instance (unlike a per-turn `useChat`), after calling `newConversation()` while a turn is in flight the `if (loading) return;` guard in `sendMessage` (line 250) silently discards **every** subsequent send.

`destroy()` (line 362) has the same shape but is terminal, so it matters less.

Fix: call `setLoading(false)` in `newConversation()` (and reset `assistantId`/`assistantText`), or have the background consumer's `finally` clear loading whenever the stream ends without a terminal chunk. Please add a test: start a turn, call `newConversation()`, then assert a following `sendMessage` is not dropped — the current tests only cover terminal-chunk completion, so this path is untested.

packages/bb-agent/src/index.chat.ts
```diff
@@ -0,0 +1,369 @@
...
+			// inference-only turn through the same entry point that started it.
+			const channelId = id ?? crypto.randomUUID();
+			if (typeof input === 'string') {
+				return transport.run({ channelId, conversationId: id, message: input });
+			}
+			return transport.run({ channelId, conversationId: id, interruptResponses: input.interruptResponses });
+		},
+
+		subscribe(opts?: { channelId?: string; observer?: boolean }): ChunkStream {
+			const channelId = opts?.channelId ?? conversationId;
```
soberm, line 358: [minor] The `subscribe()` primitive can't reach an inference-only channel.

With no `channelId` it falls back to `conversationId`, but for inference-only chats the live channel is the private `lastChannelId` (a random UUID), not `conversationId` (which stays `null`). So `subscribe()` here either throws (`null`) or, if a conversation later exists, targets the wrong id. Either expose the resolved current channel or document that `subscribe()` on an inference-only chat requires an explicit `channelId`.

packages/bb-agent/src/index.chat.ts
```diff
@@ -0,0 +1,369 @@
...
+				// over a still-open subscription and never see a fresh confirmation.
+				stream.unsubscribe();
+				if (activeStream === stream) activeStream = null;
+			}
+		})();
+	}
+
+	return {
+		async sendMessage(input: SendInput) {
+			if (loading) return;
```
soberm, line 286: [minor] `sendMessage` silently drops concurrent sends. `if (loading) return;` discards the call with no signal to the caller, so a UI can't tell the send was ignored. Consider returning a boolean, or at least documenting the drop in the JSDoc.

packages/bb-agent/src/transport.ts
```diff
@@ -0,0 +1,199 @@
...
+				.then((sub) => {
+					if (unsubscribed) {
+						// unsubscribe() was called during the attach window — detach now.
+						sub.unsubscribe();
+						return;
+					}
+					unsub = sub.unsubscribe;
+					return sub.established;
+				})
+				.catch((err) => {
```
soberm, line 180: [minor] Possible unhandled rejection via the public `subscribe()` primitive. This `.catch` closes the queue and rethrows, rejecting the `established` promise. `createChat.startTurn` awaits `established` inside a try/catch, so the easy path is safe — but a power user calling the public `subscribe()` primitive directly and iterating without awaiting/catching `established` gets an unhandled promise rejection. Add a one-line JSDoc note on `ChunkStream.established` / the primitive that `established` must be awaited or `.catch`-ed.

packages/bb-agent/src/transport.ts
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/**
+ * A minimal single-consumer push queue that adapts a callback source (the
+ * Realtime channel's `subscribe(handler)`) into an `AsyncIterable`. Chunks
+ * pushed before the consumer pulls are buffered in `queue`; pulls that arrive
+ * first park in `pending`. `close()` ends iteration once the buffer drains.
+ */
+class ChunkQueue implements AsyncIterable<AgentStreamChunk> {
+	private queue: AgentStreamChunk[] = [];
```
soberm, line 83: [nit] `ChunkQueue.queue` is an unbounded buffer. It's single-consumer and bounded in practice (the consumer always drains), but a fast producer against a stalled consumer would grow it without limit. Fine to leave as-is; noting for completeness.

packages/bb-agent/src/index.chat.ts
sharonyajain: Good catch — fixed in 353cd19e. Root-caused exactly as you describe: an unsubscribe-driven end (newConversation()/destroy() mid-turn) closes the queue with no terminal chunk, so the background consumer's `finally` never cleared `loading`. Rather than patch newConversation() alone, the `finally` now clears `loading` unconditionally, covering every non-terminal stream end. Added the regression test you asked for: start a turn, call newConversation(), assert the next sendMessage reaches the transport (was dropped before the fix).

packages/bb-agent/src/index.chat.ts
sharonyajain: Documented in 353cd19e: added a JSDoc note on the subscribe() primitive that an inference-only chat's live channel is internal (a random UUID, not the null conversationId), so callers must pass an explicit channelId (the id run() returns) rather than rely on the current-channel fallback.

packages/bb-agent/src/index.chat.ts
sharonyajain: Documented in 353cd19e: the sendMessage JSDoc now states the call is a no-op while a turn is in flight and points at isLoading() to check first. Kept the void return to match the design doc's surface rather than widening it to a boolean.

packages/bb-agent/src/transport.ts
sharonyajain: Documented in 353cd19e: added a note on ChunkStream.established that it must be awaited or .catch-ed — createChat does this internally, but a caller using the subscribe() primitive directly must, or an attach failure surfaces as an unhandled rejection.

packages/bb-agent/src/transport.ts
sharonyajain: Agreed, leaving as-is per your note, single-consumer and drained in practice; the terminal-chunk close (or unsubscribe) bounds it.

sharonyajain:
Changes:

**1. `any` in public API types (blocking) — fixed (37e1140d).**
Removed `any` from the customer-facing createChat surface: `ChatMessage.metadata` and `getConversation(...).metadata` are now `Record<string, JSONValue>`; the interrupt `reason` is `unknown` on both `getPendingInterrupts` and `onInterrupt`. Approval-message metadata is built JSONValue-clean (undefined fields omitted, `input` coerced). Now consistent with the README's own `reason?: unknown` guidance.

**2. README resume snippet not runnable (blocking) — fixed (37e1140d).**
The `realtimeTransport` resume wiring (README + `transport.ts` JSDoc) now states `responses` is `InterruptResponse[]` (optional `approved`) and that you should either type `agentResume` to accept `InterruptResponse[]` (pass-through) or map `approved: r.approved ?? false`. Copy-paste now typechecks either way.

**3. `onError` erases typed errors — fixed (37e1140d).**
`onError` is now `(error: string, cause?: unknown)`. The attach/run reject paths pass the original `Error` as `cause`, so a call site can `isBlocksError(cause, ...)`. (The stream `error` chunk carries only a message string, so `cause` is `undefined` there — documented.)

packages/bb-agent/src/index.chat.ts
```diff
@@ -0,0 +1,415 @@
...
+			conversationId = id;
+			const { messages: history } = await api.getConversation(id);
+			messages = history
+				.filter((m) => m.role === 'user' || m.role === 'assistant' || m.role === 'approval')
+				.map((m) => ({
+					id: nextId(),
+					role: m.role as 'user' | 'assistant' | 'approval',
+					content: m.content,
+					// Incoming metadata is Record<string, unknown> (any backend shape); coerce
+					// to the JSONValue ChatMessage renders. Undefined stays undefined.
```
pranavosu: Why coerce at all? can metadata be JSONValue from the source

sharonyajain: Dropped the coerce in 6be6f15f. `ChatConversationApi.getConversation` metadata is now `Record<string, JSONValue>` (JSONValue from the source), and `loadConversation` assigns it straight through. The reason it was `unknown`+coerce before: the Agent BB's own `Message.metadata` is `MessageMetadata` — a closed interface, so ts won't assign it to a JSONValue index type without help. Rather than widen the shared `MessageMetadata`/`TokenUsage` types repo-wide, the app now maps its backend metadata at the RPC boundary (the `agentGetConversation` wiring), which is where that shape adaptation belongs. Also exported `JSONValue` from `@aws-blocks/bb-agent/client` so customers can type it.

soberm (approved):

sharonyajain:
Done in `04cb39e5` — went with your first shape (typed metadata per variant).

`createChat`'s `ChatMessage` is now a discriminated union on `role`:

```ts
export interface ApprovalMetadata {
  approved?: boolean; trust?: boolean; toolName?: string; input?: JSONValue;
}
export type ChatMessage =
  | { id: string; role: 'user' | 'assistant'; content: string; metadata?: Record<string, JSONValue> }
  | { id: string; role: 'approval'; content: string; metadata?: ApprovalMetadata };
```

So a consumer writes `if (m.role === 'approval') { m.metadata?.approved }` with **no cast** — that's the "customer doesn't type unnecessary code" goal. Also removed the `role as ...` cast that was in `loadConversation`: it now builds the union per role and projects stored JSON metadata into `ApprovalMetadata` by reading the known keys (verified at runtime — unknown keys are dropped). README createChat example updated to show the cast-free approval render.

Scope note: I applied this to **createChat only** (the surface this PR owns). The deprecated `useChat` hook's `ChatMessage` in `index.hooks.ts` is a shipped public type — reshaping it into a union is a breaking change wider than this PR, so I left it. Happy to do that as a follow-up if you want the deprecated surface migrated too. The line-339 anchor is a `useChat` consumer, so its cast lives on that deprecated path — flag if you'd rather I pull the useChat redesign into this PR.

packages/bb-agent/src/index.chat.ts
```diff
@@ -0,0 +1,448 @@
...
+	| { id: string; role: 'approval'; content: string; metadata?: ApprovalMetadata };
+
+/** Conversation CRUD — plain request/response RPC to the backend, the same across every runtime. */
+export interface ChatConversationApi {
+	/** Create a new conversation and return its id. Called lazily on the first turn of a fresh chat. */
+	createConversation(): Promise<{ conversationId: string }>;
+	/** Load a conversation's message history for rendering. */
+	getConversation(
+		id: string,
+	): Promise<{ messages: { role: string; content: string; metadata?: Record<string, JSONValue> }[] }>;
```
pranavosu:
```suggestion
	): Promise<{ messages: { role: string; content: string; metadata?: unknown }[] }>;
```

I think this Record type makes the customer DX worse. Users [should not have to map objects](https://github.com/aws-devtools-labs/aws-blocks/pull/338/changes#diff-5caee74d0a160d7ec2336664c5d44af7652e98f52b9740ab44deb862c44a82afR42) every time they use this function. We might have to introduce a narrow or a cast in our code but that's preferable over the user needing to create an adapter.

pranavosu (changes requested):

packages/bb-agent/src/index.chat.ts
sharonyajain: Done in `32c3293b` — took the suggestion. `getConversation`'s `metadata` is now `unknown`, so a customer passes their backend metadata **straight through** with no adapter:

```ts
getConversation: (id) => api.agentGetConversation(id)   // was a .map(...) mapping block
```

Both the test-app and the e2e-test wirings collapsed to that one-liner, and the `toJSONMetadata` / `toMetadataRecord` helpers are deleted — that's the repeated customer boilerplate gone.

The narrow moved into our code, as you said it should: `loadConversation` runs the value through an `asJSONRecord()` (plain object → clean JSON record; anything else → `undefined`) before rendering, then projects the approval fields into `ApprovalMetadata`. Verified at runtime — raw metadata narrows correctly and a non-object metadata is dropped safely rather than crashing. No cast on the public surface.

packages/bb-agent/src/index.chat.ts
```diff
@@ -0,0 +1,468 @@
...
+	/**
+	 * Called when the agent encounters an error. `error` is the message; `cause` is the
+	 * original error object when the failure was a rejected attach/run (undefined for a
+	 * stream `error` chunk, which carries only a message), so a call site can
+	 * `isBlocksError(cause, ...)` to branch on error type.
+	 */
+	onError?: (error: string, cause?: unknown) => void;
+	/** Called when the agent pauses for human approval. Continue with `sendMessage({ interruptResponses })`.
+	 *  Each interrupt's `interruptId` is the same field you pass back in `InterruptResponse` — no remap. */
+	onInterrupt?: (interrupts: Array<{ interruptId: string; name: string; reason?: unknown }>) => void;
```
pranavosu: why do `onInterrupt` and [getPendingInterrupts](https://github.com/aws-devtools-labs/aws-blocks/pull/338/changes#diff-5ca8d457f3e28c07ccba61a9ae77fe06b5ecd26f8d5cdd99dbf58d4c4cb582c5R68) have an `interruptId` vs `id` ? can we consolidate it to one or the other?

PR #341, opened 2026-08-12
[core + 2 more] Make LambdaCompute the default compute

Simone319:
> **Stacked on #329.** Base branch is `zimzha/multi-compute-a2-compute-block`, so this diff shows only the A3 changes on top of the `Compute` abstraction + `bb-lambda-compute` package. GitHub re-targets this PR to `main` automatically when #329 merges.

## Problem

With the `Compute` abstraction and `LambdaCompute` in place (#329), the framework needs to actually resolve which compute a block runs on and make a stack's default compute own the function + API Gateway — the step that turns the abstraction into the app's real backend.

## Changes

- **`Scope.compute` resolution.** A `compute` getter on `Scope` resolves the compute a block runs on: the nearest `_compute` assigned on the block or an ancestor scope, else the owning stack/backend's default compute. For any app that never assigns a compute, this always resolves to the default — a no-op refactor.
- **Default-compute injection.** Core exposes a `DefaultComputeFactory` type (internal path) and takes a factory as a **required argument** to `create()`, calling it to build the default without importing a concrete compute class. The umbrella `@aws-blocks/blocks` — the one package that depends on both core and `bb-lambda-compute` — supplies `LambdaCompute` as that factory through a plain `import`, wrapping `BlocksStack`/`BlocksBackend.create()` so the customer call site is unchanged. The factory is a positional argument, not a prop, so customers can't set it.
- **Ownership inversion.** `LambdaCompute` now **owns** the Lambda function + API Gateway that back a stack's `handler` / `gateway` / `apiUrl`; `setupBlocksInfra` no longer creates them, and those stack accessors become getters that delegate to the default compute (typed via `LambdaShapedCompute`). The default compute is created in `create()` before the backend import, so a block reading `this.compute` in its constructor resolves it.

**Logical-ID note:** because the function + API Gateway now live under the default compute's construct path (`.../DefaultCompute/...`), their CloudFormation logical IDs change (a one-time replace of the function + API Gateway, and the API URL changes). These resources are internal, not a customer-facing contract, and there is no production usage yet, so this needs no action beyond redeploy.

## Alternatives considered

**Self-registration via a side-effect import (rejected).** An earlier iteration had core keep a module-level factory slot with `registerDefaultComputeFactory(fn)`, and `@aws-blocks/bb-lambda-compute/register` fill it via a side-effect `import` that `@aws-blocks/blocks` pulled in. Rejected in favor of the injected `create()` argument because:

- **The dependency edge becomes real and visible.** `@aws-blocks/blocks` → `@aws-blocks/bb-lambda-compute` is now a normal `import` the module graph, bundler, tree-shaker, and `lint:deps` all see — not a load-bearing side-effect import that a "remove unused import" cleanup would silently break, surfacing far away in `create()`.
- **No process-global mutable state.** The factory travels as a function argument scoped to one `create()` call, instead of a last-writer-wins singleton shared across every stack in a synth (which the multi-stage/pipeline synths — see the ESM cache-busting tests — are sensitive to).
- **The contract is compiler-enforced.** A required argument means a bare-`@aws-blocks/core` caller who omits it gets a type error, not a runtime throw at synth.
- **Not actually the house idiom.** `registerClientMiddleware` looks similar but is mechanically different — an *instance* method a block calls from its constructor, registering a *string* into a *collector array* consumed later by codegen. The compute `register` was a novel "overwrite a singleton with an executable closure, invoked synchronously" pattern that merely borrowed the name; reusing the name for different semantics is a readability cost, not a consistency win.

The tradeoff accepted: the umbrella now wraps `BlocksStack`/`BlocksBackend.create()` (preserving the instance types) instead of re-exporting the classes verbatim. That wrapper must track core's `create()` signature by hand, but it's local, boring, and testable — versus an invisible import-order contract.

## Validation

- `npm run build:force` (full monorepo) green.
- `@aws-blocks/core`: 654/654 pass.
- `@aws-blocks/bb-lambda-compute`: 11/11 pass (6 CDK-shape + 5 `Scope.compute` resolution).
- `@aws-blocks/blocks`: 42/42 pass (conditional-export parity + vendorize-map).
- `npm run lint`, `npm run lint:deps`, `npm run check:api`, `npm run sync-docs:check` all clean.

## Checklist

- [x] PR description included

---


packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -54,6 +58,14 @@ export class LambdaCompute extends Compute {
 			},
 		});
 
+		// In sandbox mode, allow localhost origins so the local dev frontend can
+		// reach the deployed Lambda API via CORS.
+		const isSandbox =
+			this.node.tryGetContext('sandboxMode') === 'true' || this.node.tryGetContext('sandboxMode') === true;
+		if (isSandbox) {
+			this.fn.addEnvironment('CORS_ALLOWED_ORIGINS', '^https?://(localhost|127\\.0\\.0\\.1)(:\\d+)?$');
+		}
```
svidgen: This should probably build on `defaults` mechanism: https://github.com/aws-devtools-labs/aws-blocks/pull/302

Please make sure between you and @osama-rizk, depending on which PR merges first, we ultimately avoid sandbox detection in the BB.

packages/core/src/cdk/index.ts
```diff
@@ -41,21 +62,25 @@ export class BlocksStack extends cdk.Stack implements BaseBlocksStack {
...
-    const infra = setupBlocksInfra(this, props, id);
-    this.handler = infra.handler;
-    this.gateway = infra.gateway;
-    this.apiUrl = infra.apiUrl;
+    const infra = setupBlocksInfra(this, id);
     this.executionRole = infra.executionRole;
   }
 
-  static async create(scope: Construct, id: string, props: BlocksStackProps) {
+  static async create(scope: Construct, id: string, props: BlocksStackProps, defaultComputeFactory: DefaultComputeFactory) {
```
svidgen: If I'm reading this right, this similarly should be part of `props.defaults` after https://github.com/aws-devtools-labs/aws-blocks/pull/302

But, I'm actually slightly lost. This is the `BlocksStack.create()` method, which appears to be adding a *required* fourth positional parameter. I would expect that to impact all of the templates and be a breaking change. But, I don't see those changed. Where is my disconnect? 😅

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -237,21 +206,22 @@ export class BlocksBackend extends Construct {
...
-    // Override BLOCKS_STACK_NAME to include the parent stack name so runtime
-    // resource lookups (DynamoDB table names) match the CDK-time fullId
-    // and are unique per deployment.
-    this.handler.addEnvironment('BLOCKS_STACK_NAME', this.fullId);
+    // The default compute (and thus handler/gateway) is created in create(),
+    // after construction — it derives BLOCKS_STACK_NAME from this.fullId.
   }
 
-  static async create(scope: Construct, id: string, props: BlocksBackendProps) {
+  static async create(scope: Construct, id: string, props: BlocksBackendProps, defaultComputeFactory: DefaultComputeFactory) {
```
svidgen: I have the same confusion here as I do for `BlocksStack.create()` with the new required parameter.

svidgen (commented):
I might need to spend more time with this one to fully grok it. I've left questions for now. Don't block on me there. More importantly, are you coordinating with @osama-rizk on the new `defaults` option for `BlocksBackendProps`?

packages/bb-lambda-compute/src/index.cdk.ts
Simone319: I will move it to the defaults as it's already merged.

packages/core/src/cdk/index.ts
Simone319: `bb-lambda-compute` depends on core, so core can't import `LambdaCompute` back — that'd be a circular dependency. The 4th param is how we break it: the umbrella (`@aws-blocks/blocks`), which depends on both, [injects the factory.](https://github.com/aws-devtools-labs/aws-blocks/blob/45c39c56b654ae943c2257cfd2a77d0cf83df7be/packages/blocks/src/index.cdk.ts#L31-L40) Templates call the umbrella's 3-arg `create()`, so nothing there changes.

`Alternatives considered` section in the PR description talks about the decision making process.

svidgen (approved):

packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -40,6 +45,11 @@ export class LambdaCompute extends Compute {
 		this.fn = new lambda.NodejsFunction(this, 'Handler', {
 			entry: this.backendHandlerPath,
 			runtime: DEFAULT_NODE_RUNTIME,
+			// Default to arm64 (Graviton) — ~20% cheaper at equal performance, and
+			// transparent for the framework's own pure-JS bundles. `architecture` is
+			// internal for now (see types.ts); a customer override arrives with the
+			// public compute-configuration surface.
+			architecture: options?.architecture ?? Architecture.ARM_64,
```
osama-rizk, line 52: The `timeout` just below (line 56) hardcodes `cdk.Duration.seconds(60 * 15)`. Before this migration, `setupBlocksInfra` created this same function using the exported `SHARED_HANDLER_TIMEOUT_SECONDS` constant. That constant is still the documented contract other resources size against: `bb-async-job` sets its SQS visibility timeout to `SHARED_HANDLER_TIMEOUT_SECONDS + maxBatchingWindowSeconds` (`bb-async-job/src/index.cdk.ts:98`), with a comment that it must cover "the handler's full budget" or SQS will redeliver a message that's still being processed.

They're two independent `900`s now. If anyone changes the constant (say to 300), async-job's visibility timeout drops while the real function timeout stays 900 → SQS redelivers mid-processing → duplicate job execution. Latent today (both equal 900), but the invariant is no longer enforced by the code. Suggest importing and using `SHARED_HANDLER_TIMEOUT_SECONDS` here instead of the literal.

packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -48,12 +58,24 @@ export class LambdaCompute extends Compute {
...
 				esbuildArgs: { '--conditions': 'aws-runtime' },
-			},
+			}),
 		});
 
+		// Allowed CORS origins come from the stack's `defaults` (e.g. the sandbox
+		// preset allows localhost so a local dev frontend can reach the deployed
+		// API). Comma-joined to match how the runtime `getCorsPatterns()` parses
+		// CORS_ALLOWED_ORIGINS.
+		const allowedOrigins = this.defaults.allowedOrigins;
```
osama-rizk, line 74: Now that CORS origins are driven by `defaults.allowedOrigins` rather than the `sandboxMode` context flag, the doc comment in `core/src/cors.ts:44` is stale — it still reads "`CORS_ALLOWED_ORIGINS` — set as Lambda env var by blocks-backend in sandbox mode." It's now set here in `LambdaCompute` from the resolved defaults, and isn't gated on sandbox mode. Worth updating that comment so the next debugger isn't sent to the wrong place. (`cors.ts` isn't in this diff, hence the note here.)

.changeset/lambda-compute-arm64.md
```diff
@@ -0,0 +1,12 @@
...
+"@aws-blocks/blocks": patch
+---
+
+`LambdaCompute`: default the function to **arm64 (AWS Graviton)**.
+
+The compute's Lambda now defaults to `Architecture.ARM_64` instead of CDK's `x86_64` default. arm64 Lambda is ~20% cheaper per GB-second than x86_64 at equivalent performance. Because `LambdaCompute` is the default compute for every `@aws-blocks/blocks` app, the shared handler runs on Graviton out of the box.
+
+The framework's own Building Blocks are pure-JavaScript esbuild bundles with no architecture-specific native code, so the switch is transparent for them. A backend `entry` bundle is the customer's own handler plus whatever they import, though — so an app that bundles an **x86-only native dependency** into its backend should be aware of the default. There is no per-app override yet; a customer-facing way to pin the architecture (`architecture` on `LambdaComputeProps` is present but internal for now) will be exposed alongside the public compute-configuration surface.
+
+> **Behavior change on next deploy:** the handler function's architecture is `arm64` (an in-place update on an existing function).
```
osama-rizk, line 12: This says the arm64 switch is "an in-place update on an existing function," but the `core-compute-resolution` changeset says the function's CloudFormation logical ID changes (it moves under the `DefaultCompute/...` construct path), which means CloudFormation **replaces** the function rather than updating it in place. Landed together, it's a replacement, not an in-place update. Worth reconciling the two redeploy notes so the combined story is accurate.

osama-rizk (commented):
Solid change overall — the ownership inversion (compute owns the function + gateway) is clean, test coverage is strong (compute resolution, umbrella wiring, CORS-from-defaults, arm64 default + override), and the changesets are unusually thorough. The CORS single-source-of-truth (`defaults.allowedOrigins` instead of a separate `sandboxMode` context flag) is a real improvement; just note it's a behavior change — deploying with `--context sandboxMode=true` but `production` defaults no longer enables localhost CORS. I also checked the arm64 default against first-party Building Blocks and it's safe: the only hashing block uses `bcryptjs` (pure JS), so the shared handler has no x86-only native code.

One thing to fix outside the code: the PR description states the default-compute factory is "a positional argument, not a prop, so customers can't set it," and the "Alternatives considered" section argues for a required `create()` argument. The shipped code passes it as a **prop** on `CoreBlocksStackProps` (see `de1a67a` / `9021ff3`). Reviewers reading the description will have the wrong mental model of the mechanism, and the "customers can't set it" guarantee now rests on the prop being `@internal` and absent from the public `BlocksStackProps`, not on positionality. Worth updating the body to match.

Inline: one systemic risk (handler timeout no longer tied to `SHARED_HANDLER_TIMEOUT_SECONDS`) plus two doc nits (stale CORS comment, arm64 changeset's "in-place" wording).

packages/bb-lambda-compute/src/index.cdk.ts
Simone319: I will create following PR to make timeout configurable, where would retire the  `SHARED_HANDLER_TIMEOUT_SECONDS` bridge between async and compute. But really good observation, will remember to use the actual compute timeout instead of relying on the constant in the next PR.

packages/bb-lambda-compute/src/index.cdk.ts
Simone319: The comment still held true. The `allowedOrigins` are set as `CORS_ALLOWED_ORIGINS` env var in the line below.

.changeset/lambda-compute-arm64.md
Simone319: It's because the [lambda-compute-arm64.md](https://github.com/aws-devtools-labs/aws-blocks/pull/341/files/9021ff3e03ef892126b2462af8a85b14170de1aa#diff-713fa4eccc1e95fb825e9f44356d7765c629d4f4167173750aaa4d3b392cac25) is created as a separate PR and merged to this branch.

osama-rizk (approved):

PR #351, opened 2026-08-13
fix(core): shim import.meta.* in the CJS Lambda bundle so import.meta.url survives bundling

osama-rizk:
## Problem

The backend Lambda handler is bundled by CDK's `NodejsFunction`, which emits **CommonJS**. In a CJS bundle `import.meta` is empty, so any bundled code doing `fileURLToPath(import.meta.url)` compiles to `fileURLToPath(undefined)` and throws `ERR_INVALID_ARG_TYPE` at Lambda module load — every request 502s. esbuild only emits a **warning** (`empty-import-meta`), so the broken bundle builds, `cdk synth` succeeds, deploy goes through, and the failure only surfaces at runtime in AWS.

## How users hit this (real report)

One of three issues in a customer bug report against a fresh Aurora scaffold — repro: https://github.com/skazantsev/aws-blocks-aurora-repro (`REPORT.md`, issue #3). The trigger is idiomatic app code: an Aurora `Database` with migrations needs an absolute `migrationsPath`, the scaffold is ESM (no `__dirname`), so the natural way to compute it is `fileURLToPath(import.meta.url)`. That module is **dual-loaded** — real ESM at synth (works) but bundled to CJS at runtime (throws). It stays invisible until deploy because `npm run dev` runs real ESM against the PGlite mock.

## Fix — auto-shim `import.meta.*` (not fail the build)

`blocksNodejsBundling()` in `@aws-blocks/core/cdk` wraps a `NodejsFunction` bundling config and, for CJS output, shims `import.meta.url` / `import.meta.dirname` / `import.meta.filename` to their CommonJS equivalents (`pathToFileURL(__filename)`, `__dirname`, `__filename`) via an esbuild `--define` + `banner`:

```ts
bundling: blocksNodejsBundling({ minify: true, esbuildArgs: { '--conditions': 'aws-runtime' } }),
```

Applied to the backend handler (`core`) and both migration lambdas (`bb-data`, `bb-distributed-data`). ESM output (`OutputFormat.ESM`) is left untouched — `import.meta` works there natively.

### Why shim instead of erroring at synth (changed from the first revision)

The first revision promoted `empty-import-meta` to a hard esbuild **error**. Per @soberm's review, that fires on the *syntactic presence* of `import.meta` anywhere in the bundled graph — including dead code in a transitively-bundled dependency — with no escape hatch, so it could hard-fail a currently-green `cdk synth` for reasons unrelated to the app. I researched how the ecosystem handles this:

- **esbuild** doesn't shim by default (warns + empties `import.meta`), but its blessed fixes are ESM output *or* a `--define` replacement — and **defining `import.meta` also suppresses the warning**.
- **Rollup** shims `import.meta.url` **by default** (that's the origin of the "portable ESM/CJS fallback" pattern shipped in many packages).

So shimming is the ecosystem-standard "make it work" approach, and it resolves both problems at once: it fixes the actual crash **and** removes the transitive-dependency false-positive blast radius (no build failure at all). Note: inside a bundle these resolve to the bundled output location, not your source tree — correct for synth-time values (e.g. `migrationsPath`) and dead interop fallbacks, but runtime code should not read source-relative files via `import.meta.url` (the READMEs say so).

## Tests

- `bundling.test.ts`: helper contract (CJS injects the `--define` + banner; ESM untouched; caller banner preserved) **plus** a real esbuild `build()` of a fixture that uses top-level `import.meta.url`, then `require()`s the output and asserts it loads without throwing and resolves a real path — i.e. proves the shim actually prevents the load crash.
- `blocks-backend.test.ts`: a handler using `import.meta.url` now bundles successfully through `BlocksBackend.create()`.

## Verification

- `npm run build` ✅ · `npm run lint` ✅ · `npm run lint:deps` ✅ · `verify-umbrella` ✅
- core **640/640** · `bb-data` **366/366** · `bb-distributed-data` **157/157**
- Changesets: `@aws-blocks/core`, `@aws-blocks/blocks`, `@aws-blocks/bb-data`, `@aws-blocks/bb-distributed-data` (all patch). Umbrella bumped because it re-exports `@aws-blocks/core/cdk` (incl. `blocksNodejsBundling`).

## Notes

- `test:e2e:local` not run — this change only affects synth-time CDK bundling of deployed handlers, which the local dev-server e2e doesn't exercise; the real esbuild bundle-and-require path is covered directly by `bundling.test.ts`.

Fixes (issue #3 of 3): https://github.com/aws-amplify/private-amplify-backend-staging/issues/1040
Repro: https://github.com/skazantsev/aws-blocks-aurora-repro


packages/core/src/cdk/blocks-backend.ts
```diff
@@ -89,7 +69,17 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
...
+        '--conditions': 'aws-runtime',
+        // NodejsFunction bundles the handler to CJS. In a CJS bundle `import.meta`
+        // is empty, so any bundled code that does `fileURLToPath(import.meta.url)`
+        // (a customer handler, a Building Block's aws-runtime code, or a dependency)
+        // becomes `fileURLToPath(undefined)` and throws at Lambda load. esbuild only
+        // *warns* about this ("empty-import-meta"), so a broken bundle deploys and
+        // fails on first invocation. Promote the warning to an error so `cdk synth`
+        // fails loudly at build time, pointing at the exact offending file/line.
+        '--log-override': 'empty-import-meta=error',
+      },
```
Simone319: Look like a common mistake that a human/agent could make. Shall we also document it in the doc somewhere to not use `fileURLToPath(import.meta.url)`?

osama-rizk: Good call — done. I added a callout in the `## Migrations` section of both `bb-data` and `bb-distributed-data` READMEs (the exact spot an author computes a path, and where an app-building agent looks — the docs live in `node_modules/@aws-blocks/*/README.md`):

- Use `migrationsPath` as a path **relative to the project root** (`'./aws-blocks/migrations'`) — the sanctioned pattern, resolved from the dir you run `cdk`/`npm run deploy` in.
- **Don't** derive it with `fileURLToPath(import.meta.url)`: the backend module is ESM locally but bundled to **CommonJS** in Lambda, where `import.meta` is empty → `fileURLToPath(undefined)` throws at load.
- If bundled runtime code genuinely needs a directory, use `__dirname` (esbuild shims it correctly in the CJS bundle).

Between the build-time guard (now fails `cdk synth` and names the file:line) and this doc note, both the human and agent paths are covered. Commit: 63ca5c7c.

Simone319 (dismissed):

packages/core/src/cdk/bundling.ts
```diff
@@ -0,0 +1,41 @@
...
+ *   bundling: blocksNodejsBundling({ minify: true, esbuildArgs: { '--conditions': 'aws-runtime' } }),
+ * });
+ */
+export function blocksNodejsBundling(options: BundlingOptions = {}): BundlingOptions {
+  return {
+    ...options,
+    esbuildArgs: {
+      ...options.esbuildArgs,
+      // Applied last so the guard can't be relaxed by a caller-supplied --log-override.
+      '--log-override': 'empty-import-meta=error',
```
soberm: [major] Worth getting explicit maintainer sign-off on the blast radius here before merging. esbuild's `empty-import-meta` check fires on the mere syntactic presence of `import.meta` anywhere in the whole bundled graph (entry plus every transitively bundled dependency), not on whether that code path is actually reached at runtime. I checked this directly: a `typeof require === 'undefined'` branch that's provably dead once bundled to CJS, a plain `import.meta?.url` reference, and an unused export inside a transitive dependency all still trip the error.

That "portable ESM/CJS fallback" pattern shows up in a lot of npm packages. Now that this helper is wired into all three `NodejsFunction`s (backend handler plus both migration lambdas), any app whose bundle transitively pulls in one of those packages, even in code that never executes, gets a hard synth failure on upgrade, with no relation to the bug being fixed and no runtime impact today. There's also no escape hatch, `BlocksBackendProps` and the migration options don't expose a way to relax this if it fires on something unrelated, so the only recourse is pinning back the dependency or this package.

Shipping this as a patch implies safe/non-breaking, but it can break a currently-green `cdk synth` for reasons that have nothing to do with the app's own code. Not saying fail-loud is wrong given how nasty the underlying bug is (silent 502s), just flagging that the discussion so far covers the customer's-own-handler case and not this transitive-dependency false-positive exposure.

soberm (commented):
Went through the latest push (63ca5c7). Good news first, a couple of things I would've flagged on the earlier revision are already fixed here.

The `assert.rejects` on the CJS bundle guard test now uses a real validator function tied to the fixture name, plus a positive control that bundles the clean handler through the same guarded config, so the rejection is actually attributable to the `import.meta.url` usage now rather than "something threw." Matching on the fixture name instead of the esbuild diagnostic text is the right call, not just a style choice: CDK's local bundling `exec()` inherits stdio for the child process instead of capturing it, so esbuild's `empty-import-meta` message never actually makes it into the thrown `Error`, it only prints to the terminal. The entry path does end up in the command-echo message though, which is what the regex is keying off. [nit] It's still technically "a failure occurred that mentions this fixture" rather than "specifically the import.meta bug" (an unrelated syntax error in that exact file would also satisfy it), but given the constraint above that's about as tight as this can get without a lot more machinery.

Also good: extracting `blocksNodejsBundling()` and wiring it into both migration-lambda `NodejsFunction`s (bb-data, bb-distributed-data) closes the gap where those two previously had zero esbuild protection. Applying the `--log-override` last so a caller's own `esbuildArgs` can't relax it is a nice defensive touch.

Two things I'd still want addressed before merge:

**[major] Changeset coverage for the umbrella package.** Neither `.changeset/handler-import-meta-cjs-guard.md` nor the new `.changeset/migration-lambda-import-meta-guard.md` bumps `@aws-blocks/blocks`. `packages/blocks/src/index.cdk.ts` re-exports `@aws-blocks/core/cdk` wholesale (so `BlocksBackend` and now `blocksNodejsBundling` flow through it), and `packages/blocks/package.json` lists `core`, `bb-data`, and `bb-distributed-data` as dependencies, all three of which now have pending changesets in this PR. That's exactly the case `verify-umbrella` in `scripts/changeset-guard.ts` exists to catch. It'll probably pass CI right now regardless, since there's an unrelated pending changeset on main (`pglite-initdb-retry.md`) that happens to already bump the umbrella, but that's coincidental and timing dependent. If that one gets released before this merges and nothing else pending covers the umbrella at that point, this exact PR would fail the check. Every other changeset in this repo pairs a re-exported sibling bump with the umbrella bump, so I'd rather this PR add `"@aws-blocks/blocks": patch` to one of the two files here than rely on someone else's changeset to cover it.

**[major] Blast radius of the warn-to-error promotion.** Left the details as an inline comment on `bundling.ts` since it has a specific line, but it's really a question about the design as a whole now that it applies to all three bundled Lambdas.

Otherwise this is a solid, well-tested fix for a genuinely nasty bug (silent 502s in production from something that only ever warns at build time). The root-cause writeup in the PR description is thorough, and the fixture's `moduleDir` reference to stop esbuild from tree-shaking away the only `import.meta.url` usage is a nice bit of test rigor.

packages/core/src/cdk/bundling.ts
osama-rizk: Agreed, and this changed my mind — I pivoted the fix from "fail at synth" to "auto-shim" in 940aa4ad.

You're right that `empty-import-meta` fires on the syntactic presence of `import.meta` across the whole bundled graph (dead code and transitive deps included), with no escape hatch. I looked at how the ecosystem handles `import.meta.url` → CJS: esbuild doesn't shim by default but blesses a `--define` replacement (which *also* suppresses the warning), and Rollup shims it by default (the source of the "portable ESM/CJS fallback" you mentioned).

So `blocksNodejsBundling()` now shims `import.meta.url`/`.dirname`/`.filename` to `pathToFileURL(__filename)` / `__dirname` / `__filename` via `--define` + `banner` for CJS output (ESM is left untouched). That:

- fixes the actual crash (the repro's handler now loads and deploys), and
- **removes this blast radius entirely** — a transitive dep that merely contains `import.meta`, even in unreachable code, no longer fails the build; there's nothing to relax and no escape hatch to add.

`bundling.test.ts` proves it end to end: it bundles a fixture that uses top-level `import.meta.url` and `require()`s the output, asserting it loads without throwing and resolves a real path. The one honest caveat (documented in the BB READMEs): inside a bundle these resolve to the bundled output location, not your source tree, so runtime code still shouldn't read source-relative files via `import.meta.url`.

osama-rizk:
@soberm thanks for the thorough pass — both `[major]`s addressed:

**Umbrella changeset.** Fixed in 940aa4ad — added `"@aws-blocks/blocks": patch` to `.changeset/handler-import-meta-cjs-guard.md`. You're right that relying on the unrelated `pglite-initdb-retry` bump was coincidental/timing-dependent. `verify-umbrella` now passes on this PR's changesets alone:
> ✓ 3 re-exported package(s) released, and @aws-blocks/blocks is bumped alongside them.

**Blast radius.** Replied on the `bundling.ts` thread — short version: I pivoted from warn→error to **auto-shimming** `import.meta.*` (esbuild `--define` + banner, the approach esbuild blesses and Rollup does by default), which fixes the crash *and* eliminates the transitive-dependency false-positive you flagged. No escape hatch needed because there's no longer a build failure to escape.

**On your `[nit]` re: the old test** — moot now; the test asserts the shimmed bundle loads and resolves a real path rather than matching an error message.

Thanks also for confirming the earlier test/`blocksNodejsBundling` improvements.

packages/core/src/cdk/bundling.test.ts
```diff
@@ -0,0 +1,90 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { test, describe } from 'node:test';
+import assert from 'node:assert';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { createRequire } from 'node:module';
+import { build } from 'esbuild';
```
soberm, line 11: The new test imports esbuild (import { build } from 'esbuild') but @aws-blocks/core does not declare esbuild in dependencies, devDependencies, or peerDependencies — it currently resolves only as a transitive dependency of aws-cdk-lib's aws-lambda-nodejs. AGENTS.md §11 makes npm run lint:deps a blocking gate for undeclared imports, and relying on a transitive dep is fragile (an aws-cdk-lib change could drop or hoist it differently).

osama-rizk: Good catch — fixed in 724dc658. Declared `esbuild` as a devDependency of `@aws-blocks/core` (`^0.27.0`, matching `@aws-blocks/hosting` and the resolved `0.27.7`), so `bundling.test.ts`'s JS-API import no longer relies on aws-cdk-lib's transitive hoisting.

For the record, `lint:deps` didn't flag it because `scripts/lint-deps.ts` excludes `.test.ts` (EXCLUDED_SUFFIXES), so it's not a gate failure today — but your point about the transitive-only reliance being fragile stands, and an explicit devDep is the right fix. `lint:deps` still green after the change.

soberm (approved):

Simone319 (approved):

PR #353, opened 2026-08-13
fix(bb-async-job): batch 10 SQS messages per invocation with partial batch failures

soberm:
## Problem

`bb-async-job` configured its SQS Lambda event source with `batchSize: 1`, and offered no way to configure the batching window.

**Issue #, if available:**

## Changes

- `packages/bb-async-job`: `SqsEventSource` default `batchSize` changed from `1` to `10`.
- Added `maxBatchingWindowSeconds` option (default `5`), mapped to the event source `maxBatchingWindow`.
- Partial batch failure reporting (`ReportBatchItemFailures`) is always enabled — it is an internal constant, not an option. It must stay on for any `batchSize > 1`, since without it SQS deletes every message in a batch as soon as the invocation returns, so one failing record would silently discard the rest.
- `batchSize` and `maxBatchingWindowSeconds` remain overridable by callers; an explicitly passed value takes precedence over the default.
- The `AsyncJob` constructor now range-checks both options and throws `AsyncJobErrors.InvalidOption` (`InvalidOptionException`) naming the construct, the option and the offending value. `batchSize` must be an integer 1–10 with no batching window, or 1–10000 once `maxBatchingWindowSeconds > 0`; `maxBatchingWindowSeconds` must be an integer 0–300. These are the limits AWS enforces when the event source mapping is created, so the failure now happens at synth instead of mid-deployment as an opaque CloudFormation error.
- Updated `README.md`, `DESIGN.md` (the Infrastructure (CDK) event source mapping entry now documents the real mapping — always-on partial batch reporting, the `maxBatchingWindowSeconds` default of `5`, and the ranges the constructor enforces) and regenerated `API.md` (`npm run update:api`).

- `packages/bb-agent`: opts out of the new defaults on its interactive path — the internal `AsyncJob` is now constructed with `batchSize: 1` and `maxBatchingWindowSeconds: 0` at both the runtime (`src/agent.ts`) and CDK (`src/index.cdk.ts`) sites. `stream()` submits one job per agent turn (plus a second on HITL resume) and the caller is blocked on that job starting, so a 5s batching window would land straight on a human-facing path; `batchSize: 1` also keeps one failing turn from sharing a batch with others, which matters because that handler is not idempotent. Both sites carry identical options so the runtime and CDK-synth paths provision the same event source mapping. A sweep of the remaining consumers found none that need the opt-out.

No runtime handler changes: `handleEventSourceRecords` in `packages/core/src/lambda-handler.ts` already iterates `event.Records` and returns `{ batchItemFailures }`.

## Validation

- `packages/bb-async-job/src/index.cdk.test.ts` — 12 CDK synth tests. Event source mapping: default `BatchSize: 10`, `FunctionResponseTypes: ['ReportBatchItemFailures']`, `MaximumBatchingWindowInSeconds: 5`, explicit `batchSize` override (partial batch reporting still on), batching window override, and window `0`. Option guards: `batchSize` `0` rejected, `batchSize` `11` rejected with no batching window, `batchSize` `500` accepted with a window (synthesizes `BatchSize: 500`), `batchSize` `10001` rejected even with a window, `maxBatchingWindowSeconds` `301` rejected, and a valid `batchSize: 10` / window `300` combination still synthesizes.
- `packages/core/src/lambda-handler.test.ts` — 3 runtime partial-batch tests over 10-record SQS batches: mixed results report only the failed `messageId`s, all-success returns `{}`, all-fail reports all 10.
- `test-apps/comprehensive/test/async-job.test.ts` — the E2E result polls now run on a wall-clock deadline instead of an attempt count. The original loops spent a fixed 100 iterations of fetch+100ms sleep, so the real budget scaled inversely with environment latency: fast Sandbox reads (~66ms) collapsed it to ~16s against 15-22s of real handler latency, failing 4 tests, while Production passed only accidentally because its slower ~120ms fetches inflated the same 100 attempts to ~22s. `pollForResult` now loops while `Date.now() < deadline` on a deployed-aware budget — `isDeployed ? 60s : 15s`, mirroring the `BLOCKS_TEST_ENV` detection in `async-job-status.test.ts` — and the `delaySeconds` tests get `delaySeconds * 1000 + budget` so the defer window is fully covered. Assertions and test intent are unchanged, including the `immediate === null` checks; no production defaults were touched.
- `packages/bb-agent/src/index.cdk.test.ts` — 3 new CDK synth tests pin the opt-out in the synthesized template: `BatchSize: 1`, `MaximumBatchingWindowInSeconds: 0`, and partial batch reporting still on. Confirmed non-vacuous: with the opt-out removed they fail with the inherited `BatchSize: 10` / window `5`. Run under `--conditions=cdk`, otherwise the internal BBs resolve to mocks and synthesize nothing.
- Builds clean; `bb-async-job` 54/54 tests pass; `bb-agent` 81/81 tests pass (78 existing + 3 new); `core` 667/667 tests pass; comprehensive local E2E 386 pass / 0 fail (1 skipped), including all 9 AsyncJob tests.

## Checklist

- [x] PR description included

---


packages/bb-async-job/src/index.cdk.ts
```diff
@@ -38,7 +84,9 @@ export class AsyncJob<T = unknown> extends Scope {
 		super(id, { parent: scope });
 
 		const maxRetries = options.maxRetries ?? 3;
-		const batchSize = options.batchSize ?? 1;
+		const batchSize = options.batchSize ?? 10;
```
sarayev, line 81: Flipping the default to `batchSize: 10` with a 5s batching window changes behavior for everyone already using `AsyncJob`, so I want to make sure we're doing it deliberately. Two things stand out:

- Latency: every job now waits up to 5s to fill a batch before the Lambda even starts. For something like a 'send welcome email on signup' job, that added delay is a surprising default.
- Duplicate processing: per-record failures are isolated cleanly, but if the Lambda times out or OOMs partway through a batch, all 10 records redeliver — including the ones that already succeeded. At `batchSize: 1` that couldn't happen. You clearly know this (it's exactly why `bb-agent` opts out, and the comment there calls out the non-idempotent handler), but the new default hands that same edge to every other consumer.

Would it be safer to keep the default at `1` and let people opt into batching explicitly? And regardless of the default, could we add a line to the README/DESIGN noting that `batchSize > 1` assumes an idempotent handler? Right now that requirement is implicit.

.changeset/async-job-batch-sqs-messages.md
```diff
@@ -0,0 +1,36 @@
+---
+"@aws-blocks/bb-async-job": patch
```
sarayev, line 3: Small one: the summary line reads `feat(...)` but the bump is `patch`. Since this changes default runtime behavior (batching + up to 5s of added latency), should it be a `minor` bump so consumers pinned on `~0.1.x` don't inherit the new behavior without noticing? Totally fine to defer given we're pre-1.0 — just flagging the feat/patch mismatch in case it's unintentional.

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -38,7 +84,9 @@ export class AsyncJob<T = unknown> extends Scope {
...
+		const maxBatchingWindowSeconds = options.maxBatchingWindowSeconds ?? 5;
+		validateEventSourceOptions(this.fullId, batchSize, maxBatchingWindowSeconds);
```
sarayev, line 83: The range checks only run in the CDK layer — `index.mock.ts` and `index.aws.ts` don't look at `batchSize` or the window at all. So if someone sets, say, `batchSize: 500` with no window, local `npm run dev` runs happily and they only hit `InvalidOptionException` at synth. That's probably an acceptable trade (synth is fast and the message is clear), but was it a conscious choice? If it's cheap to validate in a shared options path, local dev could fail fast too and match the CDK behavior.

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -30,6 +30,52 @@ export type {
 	WaitUntilCompleteOptions,
 } from './types.js';
 
+/** SQS event source limits: a batch may exceed 10 only when a batching window is set. */
+const MAX_BATCH_SIZE_WITHOUT_WINDOW = 10;
+const MAX_BATCH_SIZE_WITH_WINDOW = 10000;
+const MAX_BATCHING_WINDOW_SECONDS = 300;
+
+function blocksError(name: string, message: string): Error {
```
sarayev: Heads up — there's already an identical `blocksError(name, message)` in this same package at `status.ts` (~line 85), same implementation (sets `err.name`, no `cause`). Could we export that one and reuse it here instead of defining a second copy? Two identical error factories in one package tend to drift apart over time.

packages/bb-async-job/src/index.cdk.test.ts
```diff
@@ -0,0 +1,187 @@
...
+				handler: async () => {},
+				batchSize: 10001,
+				maxBatchingWindowSeconds: 30,
+			}),
+		/batchSize/,
+		/got: 10001/
+	);
+});
+
+test('CDK guard: maxBatchingWindowSeconds 301 is rejected', () => {
```
sarayev, line 172: The guard tests cover `batchSize` 0 / 11 / 10001 and window 301, which is great — could we also add a negative window (`-1`) and a non-integer case (`2.5` or `NaN`)? The code guards both `Number.isInteger` and `< 0`, so a couple more cases would pin those branches and document the intent.

test-apps/comprehensive/test/async-job.test.ts
```diff
@@ -6,6 +6,49 @@ import assert from 'node:assert';
...
+  } while (Date.now() < deadline);
+  return result;
+}
+
+/**
+ * Budget for a delayed job: the delay itself plus the full standard budget. Note the
+ * delaySeconds tests only assert non-execution at t~=0, not continuously throughout the
+ * delay window.
+ */
+function delayedPollBudgetMs(delaySeconds: number): number {
```
sarayev, line 48: A couple of small ones, take or leave: the delay `2` is repeated independently in `asyncJobSubmitDelayed(..., 2)` and `delayedPollBudgetMs(2)` — a shared `const DELAY_SECONDS = 2` would keep them in sync if either changes. And the CDK test stub sets `globalThis.CURRENT_BLOCKS_STACK` without resetting it between tests; harmless while they run sequentially, just a little shared-global grit.

sarayev (commented):
Overall this looks solid — turning on `reportBatchItemFailures` alongside a batch size > 1 is exactly the right fix, and the synth-time range checks plus the `bb-agent` opt-out are thoughtful. A few things I'd love your take on before it goes in; most are minor. (Posting as comments, not blocking.)

### Additional note — whole-batch failure path (core/src/lambda-handler.ts ~637–694, not changed here)
Related to the batch-size bump: the record handlers run inside `Promise.allSettled`, so per-record failures are isolated — but there's no try/catch around the surrounding logic (handler-map lookup, the loop that builds `batchItemFailures`). If anything throws there, the whole batch fails, and raising the default batch size to 10 makes that blast radius bigger. Might be worth a follow-up, or an explicit 'we're comfortable with this' note.

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -68,7 +120,17 @@ export class AsyncJob<T = unknown> extends Scope {
...
-		this.handler.addEventSource(new SqsEventSource(this.queue, { batchSize }));
+		// Partial batch responses MUST stay on for any batchSize > 1: without them a
+		// single failing record makes SQS treat the whole batch as handled and delete
+		// every message in it (silent loss). The runtime handler already returns
+		// `{ batchItemFailures }`, so this is never configurable.
+		this.handler.addEventSource(
+			new SqsEventSource(this.queue, {
+				batchSize,
+				reportBatchItemFailures: true,
+				maxBatchingWindow: Duration.seconds(maxBatchingWindowSeconds),
```
osama-rizk, line 130: The visibility timeout at `index.cdk.ts:99` is hardcoded to `900` (= the shared Lambda's max timeout) with the comment "must be >= the shared Lambda's timeout." That invariant held at `batchSize: 1` with no window, but adding a batching window changes the message-lifetime math: an SQS message's visibility clock starts when the poller *receives* it — before the batching window elapses and before the handler runs. So worst-case invisibility becomes `maxBatchingWindowSeconds + handler execution time`, not just handler time.

With the default 5s window and a handler that runs near the 900s ceiling, that's ~905s > 900s, so the message becomes visible again mid-invocation and SQS redelivers it → duplicate processing. The validator allows `maxBatchingWindowSeconds` up to 300, which sharpens it: a 300s window leaves only ~600s of real budget before redelivery. AWS's guidance is `visibilityTimeout >= maxBatchingWindow + functionTimeout` (they suggest 6× the function timeout).

bb-agent is safe here (window 0), so the exposure is a future consumer taking the defaults with a long-running, non-idempotent handler. Consider `visibilityTimeout = 900 + maxBatchingWindowSeconds`, or documenting a handler-duration ceiling. (I haven't reproduced this against a live deployment — flagging the mechanism.)

.changeset/async-job-batch-sqs-messages.md
osama-rizk, line 3: This is classified `patch`, but the title is `feat(...)` and the body describes a change to a shared default: every existing consumer that didn't set `batchSize` now gets `batchSize: 10` + a 5s batching window instead of `1` with none — i.e. up to 5s of added delivery latency and batched-invocation semantics they didn't opt into. Conventionally that's at least a `minor`; a downstream consumer on patch-range updates would inherit the behavior change silently. Is `patch` the intended bump here?

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -30,6 +30,52 @@ export type {
 	WaitUntilCompleteOptions,
 } from './types.js';
 
+/** SQS event source limits: a batch may exceed 10 only when a batching window is set. */
+const MAX_BATCH_SIZE_WITHOUT_WINDOW = 10;
+const MAX_BATCH_SIZE_WITH_WINDOW = 10000;
+const MAX_BATCHING_WINDOW_SECONDS = 300;
+
+function blocksError(name: string, message: string): Error {
+	const err = new Error(`${name}: ${message}`);
```
osama-rizk: `blocksError` sets `err.name = name` *and* prefixes the same `name` into the message (`` `${name}: ${message}` ``), and the callers pass a message that already starts with `AsyncJob "${id}": …`. Since Node's `Error.prototype.toString()` returns `name + ': ' + message`, any log or stack trace using the default string reads `InvalidOptionException: InvalidOptionException: AsyncJob "x": …` — the name is doubled.

If the other `AsyncJobErrors` (`Timeout`, `StatusNotTracked`) already follow this same `${name}: ${message}` convention, this is just house style and fine to ignore. If not, dropping the `${name}: ` prefix from the message (keeping `err.name`) avoids the doubling. Which is it?

packages/bb-agent/src/agent.ts
```diff
@@ -167,6 +167,9 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
 
 		this.job = new AsyncJob(this, 'job', {
 			schema: jobPayloadSchema,
+			// Interactive/HITL path: immediate delivery, no batching latency; batchSize 1 also isolates failures for this non-idempotent handler.
+			batchSize: 1,
```
osama-rizk: This opt-out is duplicated verbatim at the CDK site (`index.cdk.ts:159-161`), kept in sync only by matching comments plus the new synth tests. It's a pre-existing dual-runtime/CDK pattern, not something this PR introduced — but each option added here widens the drift surface: edit one site and the runtime and the provisioned event source mapping diverge silently. A shared `const` for the opt-out options, imported by both sites, would make them impossible to desync. Non-blocking — flagging the maintenance hazard now that there are two options to keep aligned.

osama-rizk (commented):
Reviewed the batching change end to end — this is a strong PR. The always-on partial-batch-reporting rationale, the per-message retry-accounting comment at the DLQ, and the explicit non-vacuousness check on the bb-agent synth tests are all exactly what I'd want to see. The E2E polling rework (attempt-count → wall-clock budget) is a real fix, not churn — the old fixed-iteration loop's effective budget scaled *inversely* with environment read speed, which is why fast Sandbox reads were collapsing it.

Four notes below. The one I'd weigh most is the visibility-timeout interaction with the new batching window — a correctness concern for long-running, non-idempotent consumers that take the defaults. The changeset semver and the error-string doubling are questions rather than assertions. None block the code as written for bb-agent, which correctly opts out.

packages/bb-async-job/src/index.cdk.ts
soberm: Deliberate, and I want to keep 10 — but you are right that the idempotency requirement was implicit, so I documented it (README "Batching and handler requirements", plus D-AJ-1a in DESIGN.md).

On keeping the default at 1: batchSize 1 *is* the bug this PR fixes. One Lambda invocation per message makes AsyncJob a ~10x cost amplifier on exactly the bursty/queued workloads it exists for; 10 collapses up to ten messages into one invocation, so invocations, cost and concurrency pressure all drop by up to 10x with no tuning. 10 is also not an arbitrary middle value: it is the maximum SQS allows *without* a batching window (the 11–10000 range requires maxBatchingWindowSeconds > 0), so it is the ceiling of the no-window regime, and the 5s window is what lets batches actually fill.

The usual objection to bigger batches — one bad record reprocesses the batch — does not apply: reportBatchItemFailures is always on and the runtime handler always returns { batchItemFailures }, so a poison message redelivers only itself. What remains is exactly the case you named: a whole-invocation failure (timeout/OOM) redelivers all ten, including records that had already succeeded. That is a widening of a window that already exists — the queue is standard, delivery is at-least-once, and the README already tells handlers to expect more than one delivery — not a new class of hazard. It is now called out explicitly rather than left implicit.

On latency: the cost is bounded by up to 5s before invocation, on a queue whose contract is "async". Anything where 5s is surprising is a latency-sensitive path, and those set maxBatchingWindowSeconds: 0 (keeping batching) or batchSize: 1 — which is what bb-agent does, for the non-idempotent reason you spotted. So the tunable is there for the minority case, while the default optimizes the common one.

Net: 10 is the largest invocation/cost reduction available in the no-window regime, retry semantics are preserved per message, and opting out is one option — strictly better than shipping 1 (the bug) and better than a smaller default, which would save less with no compensating upside.

packages/bb-async-job/src/index.cdk.ts
soberm: Conscious, yes. The values only mean anything to an event source mapping, and the mock/AWS runtime entry points never create one — the mock queue invokes the handler in-process per message, so batchSize and the window have no effect there. Validating them in a shared options path would mean rejecting values in a runtime that has no behavior to reject them for, and it would still not catch the interesting case (someone deploying without ever running local dev). Since synth is the first step of every deploy and the message names the option, the value and the applicable range, I would rather leave the check at the single place where the constraint is real than duplicate it into a code path it does not describe. Happy to revisit if we ever grow a genuine shared options-normalisation layer.

packages/bb-async-job/src/index.cdk.ts
soberm: Good catch — done. status.ts now exports blocksError and index.cdk.ts imports it, so there is one factory in the package again.

packages/bb-async-job/src/index.cdk.test.ts
soberm: Added: negative window (-1), fractional window (2.5), NaN window, plus fractional and NaN batchSize — five new guard tests, so both the Number.isInteger and the < 0 branches are pinned for each option. Also added three assertions on the queue visibility timeout while I was in there (see the reply on osama-rizk's comment).

test-apps/comprehensive/test/async-job.test.ts
soberm: Both done: DELAY_SECONDS = 2 is now shared by asyncJobSubmitDelayed / asyncJobSubmitBatchDelayed and delayedPollBudgetMs, and the CDK test files clear CURRENT_BLOCKS_STACK in an afterEach so the global does not leak out of the file.

packages/bb-async-job/src/index.cdk.ts
soberm: This is the right catch and I fixed it — the visibility timeout is now `900 + maxBatchingWindowSeconds` (LAMBDA_TIMEOUT_SECONDS + window) rather than a flat 900. Your mechanism is exactly right: visibility starts at receive, so worst-case message lifetime is window + handler runtime, which is 905s at the default and 1200s at the 300s maximum — a flat 900 let SQS redeliver a message whose invocation was still running. Documented as D-AJ-1a in DESIGN.md (with AWS's `visibilityTimeout >= maxBatchingWindow + functionTimeout` guidance) and covered by three new synth assertions: 905 at the defaults, 1200 at a 300s window, 900 with the window at 0. Note this only removes a redelivery the block itself caused — at-least-once still applies, so handlers still need to be idempotent, which the README now says out loud.

.changeset/async-job-batch-sqs-messages.md
soberm: Not intended — bumped to `minor` for @aws-blocks/bb-async-job and @aws-blocks/blocks, with the reasoning in the changeset body: an existing consumer that never set batchSize now inherits batched invocations plus up to 5s of added delivery latency, which should not arrive on a patch range. bb-agent stays `patch` since it only sets options on its own job. To be explicit about the default itself: I am keeping 10. batchSize 1 is the defect this fixes (one invocation per message, ~10x cost amplification on the queued workloads AsyncJob is for), 10 is the SQS ceiling for the no-window regime rather than an arbitrary value, reportBatchItemFailures keeps per-message retry semantics intact, and latency-sensitive consumers opt out with batchSize 1 / window 0 the way bb-agent does.

packages/bb-async-job/src/index.cdk.ts
soberm: House style — the same `${name}: ${message}` + `err.name = name` factory is used verbatim by status.ts here, and by bb-dashboard/widgets.ts and bb-app-setting/index.aws.ts, so Timeout and StatusNotTracked already read the same way. Changing it here alone would make bb-async-job internally inconsistent for a cosmetic win, so I have left it and instead deduplicated it: index.cdk.ts now imports the status.ts factory rather than defining its own (sarayev flagged the same duplication). If we want to drop the prefix it should be one sweep across the blocks that share the pattern, which I would rather not fold into this PR.

packages/bb-agent/src/agent.ts
soberm: Agreed, and cheap enough to fix here: added packages/bb-agent/src/job-event-source.ts exporting INTERACTIVE_JOB_EVENT_SOURCE = { batchSize: 1, maxBatchingWindowSeconds: 0 }, spread at both the runtime (agent.ts) and CDK (index.cdk.ts) sites. The rationale comment now lives with the constant instead of being duplicated, and the two sites cannot desync on these options. The synth tests stay as the backstop for the provisioned mapping.

soberm:
Replies posted inline on all nine comments; pushed cc710f4a with the code changes. Summary of what changed:

- **Visibility timeout** (osama-rizk): now `900 + maxBatchingWindowSeconds`, so the batching window is covered — 905s at the defaults, 1200s at a 300s window. New synth assertions + D-AJ-1a in DESIGN.md.
- **Duplicate `blocksError`** (sarayev): `status.ts` exports it, `index.cdk.ts` imports it.
- **Guard tests** (sarayev): added negative / fractional / NaN cases for both `maxBatchingWindowSeconds` and `batchSize`.
- **bb-agent opt-out duplication** (osama-rizk): shared `INTERACTIVE_JOB_EVENT_SOURCE` const spread at both the runtime and CDK sites.
- **Changeset semver** (both): `minor` for `bb-async-job` and `blocks`, with the reasoning in the body.
- **Docs** (sarayev): README now states that `batchSize > 1` assumes an idempotent handler and spells out the whole-invocation redelivery case and the latency trade.
- **Test-app nits** (sarayev): shared `DELAY_SECONDS`, `CURRENT_BLOCKS_STACK` cleared in an `afterEach`.

Kept as-is with reasoning inline: the `${name}: ${message}` doubling (house style shared across blocks — a sweep, not this PR), validation staying in the CDK layer only (the mock/AWS runtimes never create an event source mapping), and the `batchSize: 10` default.

### On the whole-batch failure path in `core/src/lambda-handler.ts` (~637–694)
Taking the explicit "we are comfortable with this" option, and here is why. The unguarded region is the handler-map lookup and the loop that builds `batchItemFailures`. The lookup is inside the per-record async function, so a miss already surfaces as a rejected promise in `allSettled` and lands in `batchItemFailures` — it does not escape. What is left outside is the `globalThis` map read (a `??` with a `new Map()` fallback, cannot throw) and a `forEach` over `results` doing property reads and array pushes on shapes we just constructed. There is no call into user code there, so a throw would mean the runtime itself is broken, and in that state failing the batch — every message redelivered, nothing deleted — is the behavior you want anyway.

Raising the batch size does widen the blast radius of a whole-invocation failure, but the realistic causes are Lambda timeout and OOM, which no try/catch in that function can catch. That is covered by the README note and D-AJ-1a instead. Wrapping the collection loop would also change semantics for non-SQS event sources (Kinesis/DDB streams rely on the throw), so I would rather not do it as a drive-by in a bb-async-job PR — happy to take it as a follow-up in core if you want the belt-and-braces version.

.changeset/async-job-batch-sqs-messages.md
soberm: Correcting myself — I flipped this to `minor` earlier and have reverted it: it stays `patch` for all three packages (4f371485).

The reason is the 0.x convention in this repo. Every package here is pre-1.0, where `minor` is what we use to signal a *breaking* change, not "notable behavior change" — under `^0.x` ranges a minor is the breaking boundary. This change is not breaking: the two options are new and optional, the API is unchanged, and the new default has a first-class opt-out (`batchSize: 1` / `maxBatchingWindowSeconds: 0`). Bumping minor would mis-signal a break and diverge from the all-patch changeset history in this repo.

Your underlying point — that the behavior change should be visible rather than silent — is fair, and that is carried by the changeset body, which spells out the new default, the added latency and how to opt out, so it lands in the CHANGELOG rather than in the version number. The `feat(...)` / patch mismatch you spotted was the real signal; the summary line now matches the intent.

soberm: Following up to correct my earlier reply: I did bump this to `minor`, then reverted it — it is `patch` for all three packages again (4f371485).

Reason: all packages here are pre-1.0, and in this repo `minor` on a 0.x version is the signal for a *breaking* change (it is the breaking boundary for `^0.x` ranges). This change is not breaking — `AsyncJobOptions` only gained optional fields, no existing call signature or type changed, and the new default is fully opt-out-able with `batchSize: 1` / `maxBatchingWindowSeconds: 0`. Shipping it as minor would tell consumers "expect a break" when there is none, and would be the only non-patch entry in the repo history.

The concern about silent inheritance is legitimate, so it is handled where it is actually readable: the changeset body states the old and new defaults, the up-to-5s added delivery latency, the visibility-timeout change and the opt-out, which is what shows up in the CHANGELOG for anyone taking the update.

osama-rizk (dismissed):

packages/bb-async-job/src/index.cdk.ts
soberm: Follow-up on this one (fea26167) — I had cited AWS guidance incorrectly and have corrected it, plus removed the duplicate constant.

What I got wrong: AWS's actual recommendation is `6 x functionTimeout + MaximumBatchingWindowInSeconds`, not `1x`. The 6x is not padding for a single execution — it buffers *invoke-side throttling retries*: if the function is at its concurrency ceiling, the poller keeps retrying the invoke while the message stays invisible, so a 1x timeout can expire before the handler has even run once. That is reachable here, because the event source targets the Lambda shared with API and CronJob traffic (D-AJ-2). So the mechanism you flagged has a second path I was implicitly waving away.

I am still shipping the 1x minimum, but now for a stated reason rather than a wrong citation. At the 900s shared timeout, 6x means a 5400–5700s visibility timeout: a genuinely stuck message would take ~1.5h per attempt to become visible again and ~4.5h to reach the DLQ at the default maxRetries of 3. That trades a rare duplicate delivery for a retry loop slow enough to be operationally useless. The residual exposure is redundant work rather than loss — delivery is at-least-once and handlers must be idempotent regardless, and `receiveCount` in the handler context lets a handler detect the redelivery. D-AJ-1a now spells all of that out, including that it is explicitly *not* AWS's recommended value, and the code comment no longer claims otherwise.

Also fixed while here: `900` was hardcoded a second time in bb-async-job, unlinked from the real timeout in `core/src/cdk/blocks-backend.ts`. `@aws-blocks/core/cdk` now exports `SHARED_HANDLER_TIMEOUT_SECONDS` (used to set the Lambda's timeout) and bb-async-job imports it, so the formula cannot go stale if that timeout ever changes — which would otherwise be the same class of bug as the one this thread is about.

bobbor (commented):
Reviewed the full diff. Verdict: **LGTM, no blocking findings** — leaving suggestions/nits inline.

What I verified:
- Core `handleEventSourceRecords` was already partial-batch correct (`Promise.allSettled` → `{ batchItemFailures }`); this PR activates it CDK-side with `reportBatchItemFailures: true`, which is correctly always-on (mandatory for any `batchSize > 1`).
- Validation guards match AWS's real limits (batchSize 1–10 without window, 1–10000 with; window 0–300, integer-checked) and fail at synth instead of mid-deploy.
- `InvalidOptionException` follows the repo's error-by-name convention (`AsyncJobErrors` as-const + `blocksError`, matched via `isBlocksError`).
- bb-agent opt-out is applied at both the runtime (`agent.ts`) and CDK (`index.cdk.ts`) sites via the shared `INTERACTIVE_JOB_EVENT_SOURCE` constant, so the two paths can't drift; the new CDK tests are non-vacuous (they fail with the opt-out removed).
- Changeset covers all four touched published packages (core, bb-async-job, bb-agent, blocks).
- The E2E poll fix (attempt-count → wall-clock deadline) is sound and removes the environment-latency-dependent budget; a >60s regression would still fail.
- Remaining AsyncJob consumers in-repo (comprehensive test app) have idempotent handlers (KVStore `put`), so inheriting the new defaults is safe.

One ask: the description reports local E2E only — since this changes AWS-path event-source behavior, a sandbox `test:e2e:sandbox` run on this branch would be good to confirm before merge (or point me to it if it already happened).

packages/bb-async-job/src/status.ts
```diff
@@ -82,7 +82,8 @@ export const statusTableOptions = {
 	ttl: 'expiresAt',
 } as const;
 
-function blocksError(name: string, message: string): Error {
+/** Build a typed `AsyncJob` error whose `name` is one of `AsyncJobErrors`. */
+export function blocksError(name: string, message: string): Error {
```
bobbor: **Suggestion (non-blocking):** promoting `blocksError` from a private helper in the status-tracking module to a public export so `index.cdk.ts` can use it works, but couples the CDK layer to the runtime status module. The natural home is `errors.ts` alongside `AsyncJobErrors` — consider moving it there and importing from both `status.ts` and `index.cdk.ts`.

.changeset/async-job-batch-sqs-messages.md
```diff
@@ -0,0 +1,50 @@
+---
+"@aws-blocks/core": patch
```
bobbor, line 2: **Suggestion (non-blocking):** changing the default `batchSize` 1→10 plus a 5s batching window is a behavior change for existing consumers (per-message isolation is gone unless handlers are idempotent, and there's added latency). `patch` is defensible pre-1.0 and the changeset documents the opt-out and the at-least-once/idempotency contract — just flagging it for a conscious maintainer decision on the bump level.

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -47,13 +89,22 @@ export class AsyncJob<T = unknown> extends Scope {
...
 
-		// Visibility timeout must be >= the shared Lambda's timeout (900s)
-		const visibilityTimeout = 900;
+		// A message's visibility clock starts when the poller receives it, which is
+		// before the batching window elapses and before the handler runs — so the
+		// worst-case invisibility a message needs is the window plus the handler's
+		// full budget. Anything less lets SQS redeliver a message that is still
+		// being processed. This is the deterministic minimum, not AWS's padded
+		// recommendation of 6x the function timeout plus the window (see D-AJ-1a).
+		const visibilityTimeout = SHARED_HANDLER_TIMEOUT_SECONDS + maxBatchingWindowSeconds;
```
bobbor, line 98: **Nit:** visibility timeout at 1x handler timeout (+ window) rather than AWS's 6x guidance is a documented tradeoff in DESIGN.md — fine as-is. A handler that regularly hits concurrency throttling could see duplicate deliveries slightly more often; the docs cover it, so nothing to change.

osama-rizk (approved):

sarayev (approved):

PR #354, opened 2026-08-13
fix(bb-auth-oidc): reject reserved response params in stub IdP redirect_uri

soberm:
## Problem

`stubIdp()` accepted `redirect_uri` values real IdPs reject, giving false confidence about real-provider compatibility. Google rejects a `redirect_uri` whose query already contains a reserved OAuth/OIDC response param (`scope`, `code`, `state`, `error`, `response_type`, `access_token`, `token_type`, `id_token`, `expires_in`, `session_state`, `iss`) — it appends those itself — with `Invalid redirect_uri: contains reserved response param scope`. The stub accepted them, so client PKCE with `redirectPath: '/auth/spa-callback?scope=...'` passed on the stub and failed on real Google. Found during the OIDC explorer exercise.

**Issue #, if available:** n/a (friction log #14)

## Changes

`packages/bb-auth-oidc/src/engines/stub-idp.ts`:

- Added `RESERVED_RESPONSE_PARAMS` and `redirectUriRejectionReason(uri)`, one validator returning the `error_description` or `null`: existing scheme rule, plus a new rejection of any reserved response param in the query (case-insensitive) using Google's wording.
- Wired into both existing enforcement points — `parseAuthorizeRequest` (GET query and login-form fields) and the token endpoint's re-check — replacing the boolean `isAllowedRedirectUri` call sites (helper retained for the scheme check).

Gaps left open, documented in the validator's doc comment:

- **Exact-match registration: not implemented.** The stub has no registered-`redirect_uri` allowlist (providers are named only; the app derives its callback at runtime), so it still accepts URIs Google would refuse for not matching a pre-registered value, including extra non-reserved query params.
- **Custom schemes** are rejected outright, vs. real IdPs allowing them for registered native clients.
- **PKCE** is required unconditionally, not per-client policy.

Changeset: `@aws-blocks/bb-auth-oidc` `patch`.

## Validation

Three tests added to `describe('stub IdP /authorize')` in `packages/bb-auth-oidc/src/index.test.ts`, using the existing `authorizeContext`/`baseParams` helpers: reserved `scope` rejected (400 + exact error body), reserved `state` spelled `State` rejected (case-insensitivity), and `?tenant=acme` still accepted (302, `code=` and `tenant=acme` preserved). Existing clean-URI and custom-scheme tests unchanged.

In `packages/bb-auth-oidc` on Node 22: `npm run build` (tsc) clean, and `node --test --test-concurrency=1 dist/*.test.js` → **115 pass, 0 fail**.

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (and PR referenced) — no user-facing docs affected; remaining stub-vs-real-IdP gaps documented in code.

---


packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -733,3 +733,57 @@ function isAllowedRedirectUri(uri: string): boolean {
...
+ * - Custom schemes are rejected outright here, whereas real IdPs accept them
+ *   only for registered native/installed-app clients.
+ * - PKCE is required unconditionally by the stub and is not validated against
+ *   per-client registration policy.
+ */
+function redirectUriRejectionReason(uri: string): string | null {
+	if (!isAllowedRedirectUri(uri)) return 'redirect_uri must be HTTPS or loopback HTTP';
+	const { searchParams } = new URL(uri);
+	for (const name of searchParams.keys()) {
+		if (RESERVED_RESPONSE_PARAMS.has(name.toLowerCase())) {
```
osama-rizk: Should this match be case-sensitive? OAuth 2.0 param names and URI query keys are both case-sensitive, so `state` and `State` are distinct keys. Lowercasing here means the stub rejects a `redirect_uri` like `.../callback?State=abc` (capital-S `State` is not a reserved key), which a real IdP would accept — the stub ends up *stricter* than the provider it emulates, producing a callback config that works in prod but fails only against this test double.

Worth confirming Google's actual behavior here. If Google matches case-sensitively, dropping the two `.toLowerCase()` calls (here and in the error message on the next line) restores fidelity.

packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -733,3 +733,57 @@ function isAllowedRedirectUri(uri: string): boolean {
 	return false;
 }
 
+/**
+ * OAuth 2.0 / OIDC response parameters that an authorization server itself
+ * appends to the redirect_uri. Google rejects a registered or requested
+ * redirect_uri whose query already contains any of these, because the IdP
+ * would otherwise have to duplicate or clobber the parameter on the way back.
+ */
+const RESERVED_RESPONSE_PARAMS = new Set([
```
osama-rizk, line 742: This set omits `error_description` and `error_uri`, but the AS appends those too — `redirectWithError()` in this same file sets `error_description` on the redirect. So a `redirect_uri` carrying `?error_description=x` passes this guard (exactly the collision the check exists to prevent) and then gets silently clobbered on any denial. Since the doc comment defines the set as "parameters that an authorization server itself appends," `error_description` (and `error_uri`, per RFC 6749 §4.1.2.1) belong here.

packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -391,9 +390,10 @@ export async function handleToken(providerName: string, ctx: BlocksContext): Pro
 	// Reject custom schemes at the token endpoint as defense in depth — if
 	// a code somehow slipped through with a bad redirect_uri, catch it here
 	// before issuing tokens.
-	if (redirectUri && !isAllowedRedirectUri(redirectUri)) {
+	const tokenRejection = redirectUri ? redirectUriRejectionReason(redirectUri) : null;
```
osama-rizk, line 395: Is this re-check reachable? A code is only minted after `parseAuthorizeRequest` has validated `pending.redirectUri` with this same `redirectUriRejectionReason`, and the token request only reaches here when its `redirectUri` exact-matches that already-validated `pending.redirectUri`. So `tokenRejection` looks like it can never be non-null.

The comment frames it as defense-in-depth, but if no input can trigger it, consider either dropping it in favor of trusting the authorize-time gate, or noting explicitly that it guards a hypothetical future code path that mints codes without going through `/authorize`.

packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -733,3 +733,57 @@ function isAllowedRedirectUri(uri: string): boolean {
...
+ *   otherwise-valid URI that a real IdP would refuse for not exactly matching
+ *   a pre-registered value — including extra non-reserved query params.
+ * - Custom schemes are rejected outright here, whereas real IdPs accept them
+ *   only for registered native/installed-app clients.
+ * - PKCE is required unconditionally by the stub and is not validated against
+ *   per-client registration policy.
+ */
+function redirectUriRejectionReason(uri: string): string | null {
+	if (!isAllowedRedirectUri(uri)) return 'redirect_uri must be HTTPS or loopback HTTP';
+	const { searchParams } = new URL(uri);
```
osama-rizk: Minor: `isAllowedRedirectUri(uri)` on the line above already does `new URL(uri)` internally, and this reparses the same string. Having `isAllowedRedirectUri` return the parsed `URL` (or parsing once here and passing it in) would avoid the second allocation on every authorize/token call. Not blocking.

packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -733,3 +733,57 @@ function isAllowedRedirectUri(uri: string): boolean {
...
+ *   redirect_uri allowlist (providers are declared by name only, and the app
+ *   derives its own callback URL at runtime), so it cannot reject an
+ *   otherwise-valid URI that a real IdP would refuse for not exactly matching
+ *   a pre-registered value — including extra non-reserved query params.
+ * - Custom schemes are rejected outright here, whereas real IdPs accept them
+ *   only for registered native/installed-app clients.
+ * - PKCE is required unconditionally by the stub and is not validated against
+ *   per-client registration policy.
+ */
+function redirectUriRejectionReason(uri: string): string | null {
```
osama-rizk, line 784: The "known remaining permissiveness" list here is genuinely helpful. One gap worth adding to it: a `redirect_uri` containing a fragment (e.g. `.../callback#state=1`) is accepted, since only `searchParams` (the query) is inspected. RFC 6749 §3.1.2 forbids fragments in `redirect_uri` and real IdPs reject them outright — another spot where the stub is more permissive than Google. Either reject fragment-bearing URIs, or list it alongside the other documented gaps so it's a known limitation rather than a silent one.

osama-rizk (commented):
Reviewed the reserved-response-param guard. Nicely done overall — the doc comments (especially the explicit "known remaining permissiveness vs. Google" list) make the fidelity tradeoffs legible, and the three added tests exercise the new path correctly. The findings below are all fidelity/cleanup, none correctness-breaking.

The two I'd weigh most: the **case-sensitivity** of the reserved-param match (a stub being *stricter* than the real IdP causes false failures, which defeats its purpose) and the omission of **`error_description`** from the reserved set. One caveat on the first — I haven't verified Google's exact case-sensitivity behavior myself, so I've left it as a question rather than an assertion.

packages/bb-auth-oidc/src/engines/stub-idp.ts
soberm: Confirmed and fixed in aebc990 — dropped both `.toLowerCase()` calls, so matching is now case-sensitive.

You were right on the spec: OAuth 2.0 param names (RFC 6749 Appendix B) and URI query keys are both case-sensitive, so `State` is a distinct key from the reserved `state` and a real IdP passes it through untouched. The stub being *stricter* than the provider is the worse failure mode, since it rejects a callback config that works in prod.

Notably the existing test was encoding the bug: `rejects redirect_uri whose query carries the reserved `state` param` was passing `?State=abc` (capital S) and asserting a 400. I changed that test to use lowercase `?state=abc`, and added a new test asserting `?State=abc` is **accepted** (302, and `State=abc` survives on the callback) so the fidelity behavior is now pinned.

packages/bb-auth-oidc/src/engines/stub-idp.ts
soberm: Good catch — added both `error_description` and `error_uri` to the set in aebc990.

You are right that this was the exact collision the guard exists to prevent: `redirectWithError()` in this file unconditionally does `target.searchParams.set('error_description', description)`, so a `redirect_uri` carrying `?error_description=x` passed the guard and then got silently clobbered on any denial. `error_uri` is included per RFC 6749 §4.1.2.1 even though this stub never emits it, since the doc comment defines the set as params an authorization server appends. Added a test covering the `error_description` rejection.

packages/bb-auth-oidc/src/engines/stub-idp.ts
soberm: You are right that it is unreachable — I traced it and could not construct an input that trips it, for exactly the reason you gave: a code is only minted after `parseAuthorizeRequest` validated `pending.redirectUri` through this same function, and the `redirectUri !== pending.redirectUri` check immediately above means the token-time value is that already-validated string.

I took the second option in aebc990 — kept the check but replaced the misleading "defense in depth" comment with an accurate one stating it is unreachable via the current `/authorize` flow and guards a future path that mints a code without going through `/authorize`. Happy to just delete it instead if you would rather not carry unreachable code.

packages/bb-auth-oidc/src/engines/stub-idp.ts
soberm: Fixed in aebc990. `redirectUriRejectionReason` now parses once and passes the `URL` into `isAllowedRedirectUri(parsed: URL)`, which drops its own `new URL()` and try/catch. The parse failure case moved up into the caller (returns the same scheme-rejection reason), so behavior is unchanged and there is now exactly one allocation per authorize/token call. `isAllowedRedirectUri` had no other callers, so the signature change is contained to this file.

packages/bb-auth-oidc/src/engines/stub-idp.ts
soberm: Good find — I went with rejecting rather than documenting, in aebc990.

`redirectUriRejectionReason` now returns `Invalid redirect_uri: must not contain a fragment` when `parsed.hash` is non-empty, with a test asserting the 400. RFC 6749 §3.1.2 requires `redirect_uri` to be an absolute URI without a fragment, and real IdPs reject it, so rejecting is the higher-fidelity option and closes the gap rather than recording it. Added it as rule 3 in the doc comment and left the remaining three gaps in the "known permissiveness" list.

osama-rizk (approved):

bobbor (approved):
Reviewed the final state at aebc990. Validator logic is correct — every param the stub appends (`code`, `state`, `error`, `error_description`) is covered by the reserved set, the case-sensitive match direction is right (URLSearchParams percent-decodes keys, so no encoding bypass), and the fragment rejection closes the RFC 6749 §3.1.2 gap. All review findings were addressed in code. Tests cover both rejection and pass-through paths, CI green.

One non-blocking nit: the PR description's Changes section still says the reserved-param match is case-insensitive and mentions three tests — both stale after aebc990 (case-sensitive, six tests). Worth editing before merge since it becomes the merge-commit record.

PR #356, opened 2026-08-13
[bb-kv-store + 10 more] Attach IAM grants to the shared execution role

Simone319:
## Problem

Building Blocks granted IAM permissions directly on the handler `NodejsFunction` (`this.handler.grant*()` / `this.handler.addToRolePolicy(...)`). That couples every block's IAM wiring to the one concrete Lambda function, which blocks the multi-compute model where a block's permissions must land on a shared role that any compute type can assume.

This PR is Issue **B3** of Multi-Compute Phase 1. It depends only on **A1** (the shared `BlocksRole` / `Scope.executionRole`, already merged into `main`) and is intentionally **standalone** — it does not stack on A2/A3.

## Changes

Migrate every data, auth, and agent block's CDK layer from granting on the handler function to granting on the shared Blocks execution role (`this.executionRole`):

- `grant*(this.handler)` → `grant*(this.executionRole)`
- `this.handler.addToRolePolicy(...)` → `this.executionRole.addToPrincipalPolicy(...)`

Blocks migrated: `bb-kv-store`, `bb-distributed-table`, `bb-file-bucket`, `bb-data`, `bb-app-setting`, `bb-knowledge-base`, `bb-email-client`, `bb-auth-cognito`, `bb-auth-oidc`, `bb-distributed-data`, `bb-agent`.

Because the shared role is the same role the handler assumes, effective runtime permissions are **identical** — this is a behavior-preserving refactor.

Additional notes:
- **`bb-distributed-data`**: the DSQL endpoint and region now flow through the config registry (`registerConfig`, loaded into `process.env` at cold start like every other block) instead of direct handler environment variables; the migration Lambda now maps the shared execution role's ARN.
- Removed dead `this.handler as lambda.Function` casts and now-unused `lambda` imports in the auth/data blocks.
- Test stubs updated to expose `executionRole` (a `BlocksRole` the stub handler assumes).

**Not in scope (deliberately):** `handler` is **not** removed here — event/observability blocks (`bb-async-job`, `bb-cron-job`, `bb-realtime`, `bb-logger`, `bb-tracer`, `bb-dashboard`) still use it. `bb-tracer` also does a compute-specific `tracingConfig` poke, so it moves in Track C (C6), not here. `Scope.handler` is removed last, in C9, once all consumers migrate.

## Validation

- `npm run build` ✓
- Unit tests for all affected packages pass (kv-store 95, distributed-table 117, file-bucket 92, data 366, app-setting 59, knowledge-base 129, email-client 20, auth-cognito 241, auth-oidc 112, distributed-data 157, agent 78) ✓
- `npm run lint` — zero blocking errors ✓
- `npm run lint:deps`, `npm run check:api` ✓
- `npm run test:e2e:local` — 8/8 ✓
- Confirmed no data/auth/agent block references `this.handler` for IAM after the change.

## Checklist

- [x] PR description included

### DIFF
--- a/packages/bb-auth-oidc/src/index.cdk.ts
+++ b/packages/bb-auth-oidc/src/index.cdk.ts
@@ -28,13 +28,13 @@
  *   the provider helpers.
  */
 
-import { type ScopeParent } from '@aws-blocks/core';
+import type { ScopeParent } from '@aws-blocks/core';
 import { Scope, registerConfig } from '@aws-blocks/core/cdk';
 import { AppSetting } from '@aws-blocks/bb-app-setting';
 import { KVStore } from '@aws-blocks/bb-kv-store';
 import * as cdk from 'aws-cdk-lib';
 import * as cognito from 'aws-cdk-lib/aws-cognito';
-import type * as lambda from 'aws-cdk-lib/aws-lambda';
+
 import type { IDependable } from 'constructs';
 import type { AuthOIDCOptions, CognitoFederatedProvider, ProviderConfig } from './types.js';
 import {
@@ -175,7 +175,6 @@ export class AuthOIDC<
 			client.node.addDependency(dep);
 		}
 
-		const fn = this.handler as lambda.Function;
 		const envPrefix = `BLOCKS_AUTH_OIDC_COGNITO_${this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')}`;
 		registerConfig(this, `${envPrefix}_POOL_ID`, pool.userPoolId);
 		registerConfig(this, `${envPrefix}_CLIENT_ID`, client.userPoolClientId);
--- a/packages/bb-distributed-data/src/index.cdk.test.ts
+++ b/packages/bb-distributed-data/src/index.cdk.test.ts
@@ -22,12 +22,17 @@ const MIGRATIONS_DIR = '.bb-data/__test_cdk_migrations__';
 function synth(build: (stack: cdk.Stack) => void): Template {
   const app = new cdk.App();
   const stack = new cdk.Stack(app, 'TestStack');
+  const executionRole = new cdk.aws_iam.Role(stack, 'BlocksRole', {
+    assumedBy: new cdk.aws_iam.ServicePrincipal('lambda.amazonaws.com'),
+  });
   const handler = new lambda.Function(stack, 'Handler', {
     runtime: DEFAULT_NODE_RUNTIME,
     handler: 'index.handler',
     code: lambda.Code.fromInline('exports.handler = async () => {};'),
+    role: executionRole,
   });
   (stack as any).handler = handler;
+  (stack as any).executionRole = executionRole;
   (globalThis as any).CURRENT_BLOCKS_STACK = stack;
   try {
     build(stack);
@@ -67,13 +72,18 @@ test('CDK: per-block removalPolicy is independent of deletion protection', () =>
   // RETAINed on stack delete but not deletion-protected.
   const app = new cdk.App();
   const stack = new cdk.Stack(app, 'IndepStack');
+  const executionRole = new cdk.aws_iam.Role(stack, 'BlocksRole', {
+    assumedBy: new cdk.aws_iam.ServicePrincipal('lambda.amazonaws.com'),
+  });
   const handler = new lambda.Function(stack, 'Handler', {
     runtime: DEFAULT_NODE_RUNTIME,
     handler: 'index.handler',
     code: lambda.Code.fromInline('exports.handler = async () => {};'),
+    role: executionRole,
   });
   (stack as any).handler = handler;
   (stack as any).defaults = BlocksPresets.sandbox;
+  (stack as any).executionRole = executionRole;
   (globalThis as any).CURRENT_BLOCKS_STACK = stack;
   try {
     new DistributedDatabase(scope(stack), 'mydsql', { removalPolicy: 'retain' });
@@ -90,15 +100,20 @@ test('CDK: per-block removalPolicy is independent of deletion protection', () =>
 test('CDK: sandbox defaults disable deletion protection', () => {
   const app = new cdk.App();
   const stack = new cdk.Stack(app, 'SandboxStack');
+  const executionRole = new cdk.aws_iam.Role(stack, 'BlocksRole', {
+    assumedBy: new cdk.aws_iam.ServicePrincipal('lambda.amazonaws.com'),
+  });
   const handler = new lambda.Function(stack, 'Handler', {
     runtime: DEFAULT_NODE_RUNTIME,
     handler: 'index.handler',
     code: lambda.Code.fromInline('exports.handler = async () => {};'),
+    role: executionRole,
   });
   (stack as any).handler = handler;
   // The cluster's removal/protection follows the stack-wide defaults (resolved
   // via the globalThis fallback here), not the sandboxMode context.
   (stack as any).defaults = BlocksPresets.sandbox;
+  (stack as any).executionRole = executionRole;
   (globalThis as any).CURRENT_BLOCKS_STACK = stack;
   try {
     new DistributedDatabase(scope(stack), 'mydsql');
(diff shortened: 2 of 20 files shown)


packages/bb-distributed-data/src/index.cdk.ts
```diff
@@ -44,20 +44,22 @@ export class DistributedDatabase extends Scope {
 
     const endpoint = cluster.getAtt('Endpoint').toString();
 
-    // Env vars for runtime
-    this.handler.addEnvironment(`BLOCKS_${envName}_ENDPOINT`, endpoint);
-    this.handler.addEnvironment(`BLOCKS_${envName}_REGION`, region);
+    // Config for runtime — flows through S3 config (loaded into process.env at
+    // cold start) like every other block, instead of a direct env var.
+    registerConfig(this, `BLOCKS_${envName}_ENDPOINT`, endpoint);
```
sarayev, line 56: **[major]** This moves the DSQL endpoint/region from a direct handler env var to `registerConfig`, which only lands in `process.env` after `loadConfigToProcessEnv()` runs at cold start. The runtime reads it eagerly in the constructor (`index.aws.ts:31`, `process.env[...ENDPOINT] ?? ''`), so it now depends on config being loaded before the app module is imported. Was this validated on a real DSQL sandbox deploy? `test:e2e:local` uses the mock, so it won't exercise the S3 → `process.env` path or catch a cold-start ordering issue — and AGENTS.md asks for a sandbox e2e on delivery changes like this.

Separately, the `?? ''` fallback on `index.aws.ts:31` masks a missing endpoint as an empty string instead of failing fast. Could we throw there if it's unset?

packages/bb-kv-store/src/index.cdk.test.ts
```diff
@@ -17,20 +17,25 @@ import { Scope, DEFAULT_NODE_RUNTIME } from '@aws-blocks/core/cdk';
...
     this.id = id;
     (globalThis as any).CURRENT_BLOCKS_STACK = this;
+    this.executionRole = new cdk.aws_iam.Role(this, 'BlocksRole', {
+      assumedBy: new cdk.aws_iam.ServicePrincipal('lambda.amazonaws.com'),
+    });
     this.handler = new cdk.aws_lambda.Function(this, 'StubHandler', {
       runtime: DEFAULT_NODE_RUNTIME,
       handler: 'index.handler',
       code: cdk.aws_lambda.Code.fromInline('exports.handler = async () => {};'),
+      role: this.executionRole,
```
sarayev, line 42: **[minor]** Because the stub handler is created with `role: this.executionRole` (same as prod), a "policy exists on the role" assertion passes whether the grant targets `this.handler` or `this.executionRole` — they resolve to the same role. So nothing here would catch a regression back to `this.handler.addToRolePolicy(...)`. The PR confirmed "no block references `this.handler` for IAM" manually — could we codify that as a test (a source scan, or asserting the handler has no inline policy of its own) so the guarantee doesn't erode later?

packages/bb-distributed-data/src/index.cdk.test.ts
```diff
@@ -110,19 +120,39 @@ test('CDK: handler gets dsql:DbConnect policy (least privilege)', () => {
...
+  (stack as any).handler = handler;
+  (stack as any).executionRole = executionRole;
+  (globalThis as any).CURRENT_BLOCKS_STACK = stack;
+  try {
+    new DistributedDatabase(scope(stack), 'mydsql');
+
+    // Endpoint + region now flow through the config registry (loaded into
+    // process.env at cold start) like every other block — not a direct env var
+    // on the handler.
+    const registry = (stack as any)[Symbol.for('BLOCKS_CONFIG_REGISTRY')] as
```
sarayev, line 171: **[minor]** This reaches into `Symbol.for('BLOCKS_CONFIG_REGISTRY').entries` directly, which is brittle to internal refactors. The auth-cognito test asserts config via `finalizeConfigRegistry(...)` + template inspection — could we match that pattern here rather than poking the registry map?

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -175,7 +175,6 @@ export class AuthOIDC<
 			client.node.addDependency(dep);
 		}
 
-		const fn = this.handler as lambda.Function;
 		const envPrefix = `BLOCKS_AUTH_OIDC_COGNITO_${this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')}`;
```
sarayev, line 178: **[nit]** Removing `const fn = this.handler as ...` leaves AuthOIDC with no execution-role grants at all (I grepped to confirm nothing was dropped). Just confirming that's intended — OIDC relies entirely on its composed AppSetting/KVStore blocks for IAM, unlike AuthCognito which grants directly?

sarayev (commented):
Reviewed the full diff plus the runtime/core code around it. Clean, well-scoped refactor — grants move from `this.handler` to `this.executionRole`, and since the prod handler is created with `role: executionRole` (core `blocks-backend.ts:74`), effective permissions are identical today. `addToPrincipalPolicy` is the right swap, `executionRole` is a concrete mutable `Role` (so no silent no-op), and the DSQL migration-lambda mapping stays consistent (it keeps its own admin role and maps the app's `executionRole` to the DML DB role). A few things worth confirming — left inline.

One extra nit that doesn't map to a changed line:

**[nit]** `Scope.executionRole` is typed `iam.IRole` (core `src/cdk/index.ts:119`). It's a concrete mutable `Role` today, so `addToPrincipalPolicy` applies — but if that getter ever returns an imported/immutable role, every block's grant becomes a silent no-op and permissions quietly vanish. Might be worth an assertion or a doc note on the getter to lock that assumption in.

packages/bb-kv-store/src/index.cdk.test.ts
Simone319: Before launching multi-compute, we plan to remove this.handler as one of the last steps. In which case, if there's still reference to `this.handler`, the code wouldn't compile anymore.

packages/bb-distributed-data/src/index.cdk.test.ts
Simone319: This pattern already exists in at least three other test files:

index.cdk.test.ts 225
index.cdk.test.ts 146
bb-knowledge-base/src/index.cdk.test.ts:158, 256
All reach into `(stack as any)[Symbol.for('BLOCKS_CONFIG_REGISTRY')].entries` directly, because `config-registry.ts` exports only `registerConfig / finalizeConfigRegistry` — there's no public read API for entries. 

I will prefer separating the concern from this PR, and have a standalone task to refactor the tests.

packages/bb-distributed-data/src/index.cdk.ts
Simone319: That ?? '' is only in the constructor, feeding `registerSdkIdentifiers(this.fullId, { clusterEndpoint })`. The actual connection path is `get base()`, which independently re-reads the env and already fails fast:
```
const endpoint = process.env[`BLOCKS_${envName}_ENDPOINT`];
const region   = process.env[`BLOCKS_${envName}_REGION`];
if (!endpoint || !region) {
  throw new Error(`Missing env: BLOCKS_${envName}_ENDPOINT / BLOCKS_${envName}_REGION`);
}
```

Simone319: The ordering is safe by construction — `createLambdaHandler` awaits `loadConfigToProcessEnv()` before importing the backend module, so `process.env[…ENDPOINT]` is populated before both the `index.aws.ts` constructor read and `get base()`. And it's validated on a real deploy: the green **E2E Sandbox** check deploys `test-apps/comprehensive` (which has a live DSQL `DistributedDatabase`) and `dsql.test.ts` runs insert/get/update/delete/transaction against it, exercising the S3-config → `process.env` path end-to-end.

packages/bb-auth-oidc/src/index.cdk.ts
Simone319: Removed because the const is created but never referenced. There is no role grants prior to this code change either. And you're right that the OIDC relies entirely on its composed AppSetting/KVStore blocks for IAM.

sarayev (approved):

osama-rizk (commented):
Reviewed the full migration end to end — clean, behavior-preserving refactor. **Recommend approve.**

I verified the load-bearing claim rather than taking it on faith: the handler is created with `role: executionRole` (core `blocks-backend.ts`), so granting on the handler vs. the role writes to the same role's inline policy — byte-identical synthesized IAM, which is also why the synth tests can't (and needn't) distinguish the two. Every migrated grant is pure-IAM, so none relied on the grantee being a `Function`.

The one non-mechanical change — `bb-distributed-data` moving endpoint/region to the config registry — also checks out on all three hazards: deploy-time tokens round-trip correctly through `jsonData`+`Lazy`; the migration Lambda is unaffected (it keeps its own direct `DSQL_ENDPOINT` env var and its own admin role); and `loadConfigToProcessEnv()` is awaited *before* `backendFactory()` imports the backend, so even the constructor-time `registerSdkIdentifiers` read sees the populated value. Test hygiene is good — the distributed-data test was correctly rewritten (registry assertion, not handler env var), not just patched. Changeset `patch` is right (no API change).

One low-severity note inline.

packages/bb-distributed-data/src/index.cdk.ts
```diff
@@ -51,20 +51,22 @@ export class DistributedDatabase extends Scope {
...
```
osama-rizk, line 56: Behavior-preserving as intended — flagging the one real delta hiding under "like every other block": the DSQL endpoint/region were always-present direct Lambda env vars, and now depend on the cold-start config-registry S3 read succeeding. If that read ever fails, `get base()` throws `Missing env` and queries fail — a slightly wider cold-start failure surface for a block's *core connection info*, and this is the block carrying a deploy-time token that has to round-trip through `BucketDeployment`. It's consistent with the architecture (every `registerConfig` block accepts this trade-off), so not a blocker — just worth a sentence in the description so the added availability coupling is a conscious call rather than incidental.

osama-rizk (approved):

PR #358, opened 2026-08-13
refactor(data-common): PGlite init-retry helper follow-ups from #205 review

sharonyajain:
Follow-up to #205 addressing @sarayev's three review comments based on improvement on the shared PGlite init-retry helper (`data-common`). No behavior change for the happy path or the recovery path shipped in #205 — this is polish + hardening.

### 1. Narrow the WASM-trap classifier (`isPgliteUnreachableTrap`)
Previously matched the bare word `unreachable` anywhere in the message/stack, which could misclassify an unrelated probe failure (an `assertUnreachable` helper, an "unreachable host" message) as a retryable trap. Now matches specific signatures only: `RuntimeError: unreachable`, `wasm trap: unreachable`, and the Emscripten `Aborted()` form.

### 2. Uniform cleanup in `initializePgliteWithRetry`
Retryability is now classified **before** consulting the attempt budget. A non-retryable error is always rethrown untouched (we never close an instance we didn't diagnose as a dead WASM trap), and this fixes the `maxAttempts === 1` edge that previously closed the instance on *any* failure without classifying it.

### 3. Mark the init-retry exports `@internal`
`initializePgliteWithRetry`, `isPgliteUnreachableTrap`, `PgliteLike`, and `PgliteInitRetryOptions` exist only so the two engine packages can share the helper — they were needlessly part of `data-common`'s public API. Marked `@internal` (and documented the `PgliteLike` methods); `API.md` regenerated to match.

### Tests
- Positive coverage for each trap signature (`RuntimeError: unreachable`, `wasm trap:`, `Aborted()`), including stack-only and cause-chain paths.
- New negative test: a bare `unreachable` without a trap signature is **not** retried.
- New `maxAttempts=1` regressions: a non-retryable error does not close the instance; a trap still closes it and does not recreate.


ikenyal (approved):
Reviewed the classifier narrowing, the retry-budget ordering, and the generated API report. The new ordering correctly preserves a non-retryable instance when maxAttempts is 1 while still closing a diagnosed trap before giving up. The focused PGlite helper tests cover message, stack, cause-chain, negative, and exhausted-budget cases; I also ran all 18 of those tests locally. No blocking issues found.

packages/data-common/src/pglite-init.ts
```diff
@@ -45,33 +56,45 @@ export interface PgliteInitRetryOptions {
...
+ *  - V8/Node:            `RuntimeError: unreachable` (optionally `... executed`)
+ *  - other engines:      `wasm trap: unreachable`
+ *  - Emscripten runtime: `Aborted()` (e.g. `Aborted(). Build with -sASSERTIONS ...`)
+ *
+ * Deliberately narrower than a bare `/unreachable/i`: the word "unreachable" is
+ * common (helpers like `assertUnreachable`, "unreachable host" messages), so we
+ * require it next to a trap-context token. This keeps a stray "unreachable" in
+ * an unrelated failure's message or stack from being misclassified as retryable.
+ */
+const PGLITE_INIT_TRAP_RE = /RuntimeError:\s*unreachable\b|wasm trap:\s*unreachable\b|\bAborted\(\)/i;
```
soberm: [major] Wanted to double-check this against a real trap rather than take the doc comment on faith, so I built a minimal WASM module that actually traps on `unreachable` and threw it in Node. The real shape: `.message` comes back as literally `"unreachable"` — no `RuntimeError:` prefix. That prefix only shows up in `.stack`, since V8 prepends `toString()` there. I also checked pglite's own `dist/index.js`: `_pg_initdb()` isn't wrapped in a try/catch on the init path, so that raw shape is exactly what this classifier receives.

So for that specific trap, this regex only matches through the `.stack` fallback below, not `.message` directly. The old bare `/unreachable/i` matched the message directly, so we've quietly added a dependency on `.stack` staying intact for this exact case. Probably fine today, but since this helper exists purely because of a hard-to-repro production trap, I'd rather not have its reliability hinge on an implicit assumption about stack traces surviving.

Cheap addition that doesn't reopen the original false-positive problem (checking `.name` is independent of `.stack`):
```ts
current.name === 'RuntimeError' && /unreachable/i.test(current.message)
```
A plain `Error` (the `assertUnreachable`/"host unreachable" cases in the tests below) will never have `name === 'RuntimeError'`, so it stays just as narrow.

Side note, not a concern: the `Aborted()` alternative here is solid, not a guess. Traced Emscripten's `abort(what)` in `postgres.js` and `abort("")` is a real call site (the generic `__abort_js` import stub), so `'Aborted(). Build with -sASSERTIONS...'` is the actual message shape, not a fabricated test string.

packages/data-common/src/pglite-init.ts
```diff
@@ -45,33 +56,45 @@ export interface PgliteInitRetryOptions {
...
+ * @internal
  */
 export function isPgliteUnreachableTrap(error: unknown): boolean {
   const seen = new Set<unknown>();
   let current: unknown = error;
   while (current != null && !seen.has(current)) {
     seen.add(current);
     if (current instanceof Error) {
-      if (/unreachable/i.test(current.message) || (current.stack != null && /unreachable/i.test(current.stack))) {
+      if (PGLITE_INIT_TRAP_RE.test(current.message) || (current.stack != null && PGLITE_INIT_TRAP_RE.test(current.stack))) {
```
soberm: [major] This is the specific line that only catches the real trap via `.stack` rather than `.message` — see the comment on the regex definition above for the full trace. Suggested tweak here:
```ts
if (
  PGLITE_INIT_TRAP_RE.test(current.message) ||
  (current.name === 'RuntimeError' && /unreachable/i.test(current.message)) ||
  (current.stack != null && PGLITE_INIT_TRAP_RE.test(current.stack))
) {
```

packages/data-common/src/pglite-init.ts
```diff
@@ -45,33 +56,45 @@ export interface PgliteInitRetryOptions {
...
+const PGLITE_INIT_TRAP_RE = /RuntimeError:\s*unreachable\b|wasm trap:\s*unreachable\b|\bAborted\(\)/i;
+
 /**
  * Return true when an error looks like a PGlite WASM `unreachable` init trap.
  *
- * The trap surfaces with an "unreachable" message (often wrapped as
- * `RuntimeError: unreachable` or an `Aborted()` error), sometimes only visible
- * in the stack or a nested `cause`. This walks the message, stack, and cause
- * chain, guarding against cyclic causes.
+ * The trap surfaces as one of the {@link PGLITE_INIT_TRAP_RE} signatures,
```
soberm, line 80: [nit] `{@link PGLITE_INIT_TRAP_RE}` points at a non-exported const. Harmless right now since `docModel` is disabled for this package, but if that ever gets turned on this link won't resolve.

packages/data-common/src/pglite-init.test.ts
```diff
@@ -167,6 +180,37 @@ test('rethrows a non-retryable error immediately without recreating', async () =
...
+          recreated++;
+          return new FakePglite(0);
+        },
+        { backoffMs: 0, maxAttempts: 1 },
+      ),
+    /unreachable/,
+  );
+  assert.strictEqual(initial.closed, true, 'a trapped instance is still closed even with no retries left');
+  assert.strictEqual(recreated, 0, 'no recreate once the attempt budget is exhausted');
+});
```
soberm, line 230: [minor] Both new tests here check attempt 1 in isolation. Might be worth one more covering "traps on attempt 1 (retries into a new instance), then throws something unrelated on attempt 2" — asserting the second instance isn't closed either. It's implied by the per-iteration independence of the catch block, but this reorder is fixing exactly this class of ordering bug, so locking in the multi-attempt case seems like cheap insurance.

packages/data-common/API.md
```diff
@@ -44,28 +44,34 @@ export interface DatabaseEngine {
 // @public
 export const DOLLAR_QUOTE_TAG_RE: RegExp;
 
-// @public
+// Warning: (ae-internal-missing-underscore) The name "initializePgliteWithRetry" should be prefixed with an underscore because the declaration is marked as @internal
```
soberm, line 47: [nit] Flagging in case this looks like a regression at a glance: these `ae-internal-missing-underscore` warnings match the existing `unwrapQuery` entry already in this file, and `check-api-reports.ts` doesn't gate on that token (only `ae-forgotten-export` is in its warning list), so there's nothing to fix here.

soberm (dismissed):
Follow-up to #205 that addresses all three review comments here: narrows the WASM-trap classifier, reorders retryability-classification ahead of the attempt-budget check (fixes the `maxAttempts=1` cleanup issue), and marks the four engine-only exports `@internal` with docs. Traced the reordered cleanup logic against every existing test plus the two new `maxAttempts=1` tests and didn't find any regressions.

A couple of notes inline, one of them (the trap classifier's dependency on `.stack`) is worth a look but not blocking — the two shapes I could actually verify against real WASM/Emscripten behavior both still work correctly today.

Nice to see the negative regex tests (`assertUnreachable`, "host unreachable") locked in explicitly instead of just trusting the narrower regex, and the `@internal` + docs change matches exactly what was asked for without any runtime behavior change (dtsRollup is disabled for this package, so it's a documentation-level signal as intended, not enforcement).

sarayev (commented):
Took another pass focused on correctness/edge cases — overall this cleanly addresses the #205 feedback and the reordered retry logic looks right. A few thoughts inline. One PR-level note: since this narrows the classifier that both `bb-data` and `bb-distributed-data` depend on, it'd be good to confirm CI runs their engine suites here and not just the `data-common` tests. I checked their fixtures against the new regex and they still match, so I'd expect them to pass — just want the consumers actually exercised rather than assumed.

---

### `packages/data-common/src/pglite-init.ts` (line 70)

This is a nice tightening over the old bare-word match. One thing I'd double-check: `\bAborted\(\)` only matches when the parens are *empty*, but Emscripten aborts usually carry a reason — things like `Aborted(Cannot enlarge memory arrays…)` or `Aborted(OOM)`. Since the whole point here is surviving memory pressure, an OOM-style abort feels like exactly the case we'd want to catch, and that form would slip through unless it also happened to include `RuntimeError: unreachable` (which is what saved us in the signature from #205).

I think dropping the closing paren from the pattern covers it without giving anything up, since this only ever runs against the `SELECT 1` probe:
```ts
const PGLITE_INIT_TRAP_RE = /RuntimeError:\s*unreachable\b|wasm trap:\s*unreachable\b|\bAborted\(/i;
```
Might be worth a quick positive test for `Aborted(Cannot enlarge memory arrays)` too — the current `Aborted()` test uses the empty form, and the other `Aborted()` test passes a custom `isRetryable`, so the default classifier never actually sees a reason string.

---

### `packages/data-common/src/pglite-init.ts` (line 130)

Not something this PR introduced, but since we're in here: `?? 3` won't catch a `NaN` (only `null`/`undefined`), so if a caller ever passed `maxAttempts: NaN`, `Math.max(1, NaN)` stays `NaN` and `attempt >= NaN` is always false — which would turn a persistent trap into an unbounded close/recreate loop. Easy to shore up if you feel like it:
```ts
const maxAttempts = Number.isFinite(options.maxAttempts)
  ? Math.max(1, Math.trunc(options.maxAttempts as number))
  : 3;
```
Totally fine to leave for a follow-up — just flagging while it's fresh.

---

### `packages/data-common/src/pglite-init.ts` (line 156)

Small edge case: if `recreate()` rejects (the engines do real work in there — `mkdirSync`, `new PGlite(...)` — so `ENOSPC`/`EACCES`/construction failures are possible), that error propagates out instead of the init-trap error we caught above, so the failure ends up looking like a recreate problem rather than "PGlite kept trapping." Wrapping it to keep the original cause would make debugging nicer:
```ts
try {
  instance = await recreate();
} catch {
  throw new Error('Failed to recreate PGlite after init trap', { cause: error });
}
```
Or, if we'd rather keep it simple, a line in the `@param recreate` doc noting the factory shouldn't throw would do the trick.

---

### `packages/data-common/src/pglite-init.test.ts` (line 214)

The reorder means `onRetry` intentionally *doesn't* fire on the final, give-up attempt — which is the right call. The only thing is, the `[1,2]` test here succeeds on the 3rd attempt, so it doesn't actually pin that behavior: if someone later moved `onRetry` back above the budget check, this test would still pass while `onRetry` wrongly fired on the exhausting attempt. A tiny exhaustion-path test would guard it:
```ts
test('does not invoke onRetry on the final, exhausting attempt', async () => {
  const attempts: number[] = [];
  await assert.rejects(
    () => initializePgliteWithRetry(new FakePglite(99), () => new FakePglite(99), {
      backoffMs: 0, maxAttempts: 2, onRetry: (a) => attempts.push(a),
    }),
    /unreachable/,
  );
  assert.deepStrictEqual(attempts, [1]);
});
```

packages/data-common/src/pglite-init.ts
sharonyajain: Great find, and thanks for verifying against a real WASM module rather than the doc comment. Addressed in 926185f: added a stack-independent clause — current.name === 'RuntimeError' && /unreachable/i.test(current.message), so the real V8/Node trap is caught via .message, not only the .stack fallback. Stays just as narrow (a plain Error never has name === 'RuntimeError'). Added a positive test for a RuntimeError whose message is only "unreachable" with no signature in the stack, plus a negative for a plain Error('unreachable').

packages/data-common/src/pglite-init.test.ts
sharonyajain: Added in 926185f — a non-retryable failure on a later attempt is rethrown without closing that instance (attempt 1 traps → recreate, attempt 2 fails non-retryably → rethrown, second instance not closed). Locks in the per-iteration independence.

sarayev (approved):

soberm (approved):

PR #366, opened 2026-08-14
feat(core): expose backendModulePath; self-heal handler init 

pjkroker:
**Stacked PR 1 of 2** — AgentCore + Realtime migration. `main` → **this (core)** → bb-agent migration (#368).                                                                
                                                                                                                                                                                                               
## Problem                                                                                                                                                                                                     
The Agent BB ran its streaming loop on **Lambda + SQS + Realtime (API Gateway WebSocket)**. The fundamental ceiling is **Lambda's 15-minute max execution** — long-running / multi-step agents were not achievable on AWS. Bedrock AgentCore Runtime (sessions up to 8h) is purpose-built for this.

## Scope of this PR
Additive CDK-**core** changes with no consumer of their own (the loop move lands in #368), so behavior-preserving except the init-retry fix noted below:

1. **`backendModulePath` on `BlocksStack` / `BlocksBackend`** — exposes the app's `backendCDKPath` so a BB that co-bundles the backend at synth can find it via `globalThis.CURRENT_BLOCKS_STACK.backendModulePath`. (Consumed by #368's AgentCore Runtime co-bundle.)
2. **`getConfigLocation()`** — splits the config-bucket creation out of `finalizeConfigRegistry` into an eager, memoized helper returning `{ bucketName, key }`. The handler path is unchanged; a BB whose compute runs **as** the shared role but outside the handler (the AgentCore Runtime) uses it to inject `BLOCKS_CONFIG_BUCKET`/`BLOCKS_CONFIG_KEY` so its `loadConfigToProcessEnv()` loads the same app config the handler does. IAM is unchanged — the config-read grant is on the shared role, which such compute inherits.
3. **`BlocksRole` stays BB-agnostic** — core does **not** trust any compute beyond `lambda`. The role keeps its `CompositePrincipal` so a BB whose compute runs as the shared role adds its own trust principal from **its own** construct (the Agent BB adds `bedrock-agentcore` in #368) — same "each block augments the shared role" pattern as #356.
4. **`fix(core)`: the Lambda handler self-heals a failed init** instead of poisoning the container — see below.   

## Changes
- `packages/core/src/cdk/blocks-backend.ts` — assign `backendModulePath` from `props.backendCDKPath`; reword the `BlocksRole` `CompositePrincipal` comment (no `bedrock-agentcore` trust here — lambda only).
- `packages/core/src/cdk/config-registry.ts` — `getConfigLocation()` (eager memoized bucket); `finalizeConfigRegistry` reuses it and still uploads/wires when a bucket exists even with zero entries.
- `packages/core/src/cdk/index.ts` / `src/index.cdk.ts` — export `backendModulePath` + `getConfigLocation`.
- `packages/core/src/common/index.ts` — declare `backendModulePath` on the shared `BlocksStack` type.
- Tests: `blocks-backend.test.ts` (lambda-only trust + `backendModulePath`); `config-registry.test.ts` (getConfigLocation ↔ finalize, incl. empty-bucket upload + idempotency); `lambda-handler.test.ts` + `config.test.ts` (self-heal regressions).
                                                                                                                                                              
### 3. The handler self-heal fix                                                                                                                                                                               
**Bug (pre-existing):** `createLambdaHandler` memoized its init promise, including on **rejection**. If `initialize()` threw — most often because `loadConfigToProcessEnv()` can't read `blocks-config.json`   
during the brief post-deploy window before it's readable — the rejected promise was reused for the container's entire lifetime, so every later request returned a 500. A *transient* config blip became a      
*permanent* per-container outage: the whole API returns `undefined`, only appearing to "recover" as Lambda cycled in fresh containers.  

**Why that window exists:** the handler doesn't bundle its config — it **fetches `blocks-config.json` from S3 at cold-start** (Lambda env vars have a ~4 KB cap, so BB config is stored in S3). That file is   
produced by *separate* CloudFormation resources, so "the handler can serve" and "its config is readable" are **decoupled and race during a deploy**. Two things can briefly lag right after a fresh deploy:    
1. **The object isn't written yet** — it's uploaded by a CDK `BucketDeployment` custom resource that has *no ordering dependency* on the handler / API Gateway, which can become reachable first.              
2. **The read grant hasn't propagated** — the handler's `s3:GetObject` on the config bucket lives in a freshly-created (overflow) IAM policy, and IAM changes aren't instantaneous.

                                                                       
                                                                                                                                                                                                               
```ts                                                                                                                                                                                                          
// Before — a failed initialize() is cached forever; every later request re-awaits the rejection:                                                                                                              
if (!handler) {                                                                                                                                                                                                
  if (!initPromise) initPromise = initialize();                                                                                                                                                                
  await initPromise;                                                                                                                                                                                           
}                                                                                                                                                                                                              
                                                                                                                                                                                                               
// After — a failed init resets the promise so the next request retries:                                                                                                                                       
if (!handler) {                                                                                                                                                                                                
  if (!initPromise) {                                                                                                                                                                                          
    initPromise = initialize().catch((err) => {                                                                                                                                                                
      initPromise = null; // don't cache the rejection — let the next request re-run initialize()                                                                                                              
      throw err;                                                                                                                                                                                               
    });                                                                                                                                                                                                        
  }                                                                                                                                                                                                            
  await initPromise;                                                                                                                                                                                           
}                                                                                 
```                                                                                                                                                                                                                 
Paired with config.ts no longer caching a not-found result (a 404 is treated as a transient window, not a permanent empty-config poison), the handler self-heals as soon as the config object becomes readable.
                                                                                                                                                                                                               
Behavioral change: a failed init is now retried per-request rather than cached — a strict robustness win, but it does change failure semantics (a genuinely-broken deploy retries instead of failing fast).

## Validation
- Unit: `blocks-backend.test.ts`, `config-registry.test.ts`, `lambda-handler.test.ts`, `config.test.ts`; full core suite green.


## Checklist

- [x] PR description included

### DIFF
--- a/.changeset/core-handler-init-retry.md
+++ b/.changeset/core-handler-init-retry.md
@@ -17,3 +17,10 @@ config loader no longer caching a not-found result (a 404 is treated as transien
 poisoning the cache with `{}`), the handler **self-heals** as soon as config becomes readable.
 This is most likely to surface on large/slow deploys (which widen the post-deploy window), but the
 fragility was general.
+
+**Behavior change:** a failed handler init is now **retried per request** instead of cached. A
+transient config blip self-heals (the win), but a *genuinely* broken deploy (bad IAM, wrong bucket,
+malformed config) now re-runs `initialize()` — an S3 GET + backend import — on **every** invocation
+until it recovers or the container cycles, rather than failing fast. Expect added per-request latency
+and repeated S3/CloudWatch activity under a broken deploy; revisit any alarms that assumed a fast,
+sticky init failure.
--- a/.changeset/core-handler-init-retry.md
+++ b/.changeset/core-handler-init-retry.md
@@ -1,0 +1,19 @@
+---
+"@aws-blocks/core": patch
+---
+
+fix(core): retry handler initialization on failure instead of caching the rejection
+
+`createLambdaHandler` memoized its init promise and, if `initialize()` threw — most often
+because `loadConfigToProcessEnv()` couldn't read `blocks-config.json` during the brief
+post-deploy window before it's readable — cached that **rejected** promise for the container's
+entire lifetime. Every subsequent request then re-awaited the same rejection and returned a 500,
+so a *transient* config blip became a *permanent* handler outage for that container (visible as a
+whole stack of API calls returning `undefined`, only "recovering" as Lambda cycled in fresh
+containers).
+
+A failed `initialize()` now resets the promise so the next invocation retries. Combined with the
+config loader no longer caching a not-found result (a 404 is treated as transient rather than
+poisoning the cache with `{}`), the handler **self-heals** as soon as config becomes readable.
+This is most likely to surface on large/slow deploys (which widen the post-deploy window), but the
+fragility was general.
--- a/packages/core/src/cdk/blocks-backend.ts
+++ b/packages/core/src/cdk/blocks-backend.ts
@@ -106,33 +106,15 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
   // so permissions would quietly vanish. If a bring-your-own-role option is ever
   // added, it must resolve to a mutable role (`{ mutable: true }`).
   const executionRole = new iam.Role(scope, 'BlocksRole', {
-    // CompositePrincipal (rather than a bare ServicePrincipal) so additional
-    // compute types can assume this same shared role as they are introduced
-    // (e.g. ECS tasks via ecs-tasks.amazonaws.com), by adding principals here.
+    // CompositePrincipal (rather than a bare ServicePrincipal) so a Building Block
+    // whose compute runs AS this shared role can add its own trust principal here
+    // (e.g. the Agent BB adds bedrock-agentcore in its CDK construct) — core stays
+    // agnostic and only Lambda is trusted by default.
     assumedBy: new iam.CompositePrincipal(new iam.ServicePrincipal('lambda.amazonaws.com')),
     managedPolicies: [
       iam.ManagedPolicy.fromAwsManagedPolicyName('service-role/AWSLambdaBasicExecutionRole'),
     ],
   });
-
-  // The Agent BB's AgentCore Runtime runs AS this shared role, so its loop inherits every Building
-  // Block's grant (the same way any compute does) rather than needing a bespoke, under-granted role.
-  // Scope that trust to this account/region with aws:SourceAccount + aws:SourceArn — this is AWS's
-  // recommended AgentCore Runtime trust policy (the service sends both keys on AssumeRole), so it
-  // tightens the confused-deputy surface (only AgentCore runtimes in THIS account can assume the
-  // role) without breaking assumption. Added as its own trust statement so the condition applies
-  // only to AgentCore, not to the unconditioned lambda trust.
-  executionRole.assumeRolePolicy?.addStatements(
-    new iam.PolicyStatement({
-      effect: iam.Effect.ALLOW,
-      principals: [new iam.ServicePrincipal('bedrock-agentcore.amazonaws.com')],
-      actions: ['sts:AssumeRole'],
-      conditions: {
-        StringEquals: { 'aws:SourceAccount': cdk.Aws.ACCOUNT_ID },
-        ArnLike: { 'aws:SourceArn': `arn:${cdk.Aws.PARTITION}:bedrock-agentcore:${cdk.Aws.REGION}:${cdk.Aws.ACCOUNT_ID}:*` },
-      },
-    }),
-  );
 
   // ── Resource Groups ───────────────────────────────────────────────────
   let rootStack = cdk.Stack.of(scope);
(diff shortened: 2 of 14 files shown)


packages/core/src/cdk/blocks-backend.ts
```diff
@@ -86,6 +86,25 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
...
+    new iam.PolicyStatement({
+      effect: iam.Effect.ALLOW,
+      principals: [new iam.ServicePrincipal('bedrock-agentcore.amazonaws.com')],
+      actions: ['sts:AssumeRole'],
+      conditions: {
+        StringEquals: { 'aws:SourceAccount': cdk.Aws.ACCOUNT_ID },
+        ArnLike: { 'aws:SourceArn': `arn:${cdk.Aws.PARTITION}:bedrock-agentcore:${cdk.Aws.REGION}:${cdk.Aws.ACCOUNT_ID}:*` },
+      },
+    }),
+  );
```
Simone319: I think the idea is that each block adds the required policy to the `executionRole`. So the BB specific logic does not crowd the backend definition. And if no agent is created, the `executionRole` should not by default allow `bedrock-agentcore.amazonaws.com` to assume. 

Can we moving this piece to the AgentCore constructor?

Simone319 (changes requested):

packages/core/src/cdk/blocks-backend.ts
pjkroker: Done — moved the `bedrock-agentcore` assume-role trust out of core and into AgentCoreRuntime's CDK construct. BlocksRole now only trusts lambda by default — it keeps its CompositePrincipal so a compute-BB adds its own principal from its own construct.

Simone319 (dismissed):

Simone319 (dismissed):

packages/core/src/lambda-handler.ts
```diff
@@ -318,7 +318,18 @@ export function createLambdaHandler(backendFactory: () => Promise<any>) {
...
+      // because loadConfigToProcessEnv() couldn't read blocks-config.json during the brief
+      // post-deploy window before it's readable), a memoized rejected promise would poison this
+      // container for its whole lifetime — every later request re-awaits the same rejection and 500s.
+      // Resetting initPromise lets the next invocation re-run initialize() (config.ts re-fetches,
+      // since it doesn't cache failures), so the handler self-heals once config is available.
+      if (!initPromise) {
+        initPromise = initialize().catch((err) => {
+          initPromise = null;
+          throw err;
+        });
```
svidgen: Should this also contain a few retries? Perhaps with jittered backoff (~200ms, ~400ms, ~800ms, ~1600ms, and maybe ~3200ms before calling it quits).

packages/core/src/cdk/config-registry.ts
```diff
@@ -65,42 +110,37 @@ export function finalizeConfigRegistry(
...
-	});
+	// Nothing to do only if no config was registered AND no bucket was created (via
+	// getConfigLocation). If a co-located BB created the bucket, still upload (even an empty {}) and
+	// wire the handler so that compute's loadConfigToProcessEnv() resolves instead of 404-ing forever.
+	if (registry.entries.size === 0 && !registry.bucket) return;
 
-	const configKey = 'blocks-config.json';
+	// Ensure the bucket exists (a BB may already have created it via getConfigLocation).
+	const { bucketName, key } = getConfigLocation(scope);
+	const configBucket = registry.bucket!;
```
jvh-aws: Can we fix the linter warning? (nit)

.changeset/core-handler-init-retry.md
```diff
@@ -0,0 +1,19 @@
...
+entire lifetime. Every subsequent request then re-awaited the same rejection and returned a 500,
+so a *transient* config blip became a *permanent* handler outage for that container (visible as a
+whole stack of API calls returning `undefined`, only "recovering" as Lambda cycled in fresh
+containers).
+
+A failed `initialize()` now resets the promise so the next invocation retries. Combined with the
+config loader no longer caching a not-found result (a 404 is treated as transient rather than
+poisoning the cache with `{}`), the handler **self-heals** as soon as config becomes readable.
+This is most likely to surface on large/slow deploys (which widen the post-deploy window), but the
+fragility was general.
```
jvh-aws, line 19: Can we callout the behavior change in the initialization more concretely? E.g.

  **Behavior change:** a failed handler init is now **retried per request** instead of
  cached. A transient config blip self-heals (the win), but a *genuinely* broken deploy
  (bad IAM, wrong bucket, malformed config) now re-runs `initialize()` — an S3 GET + backend
  import — on **every** invocation until it recovers or the container cycles, rather than
  failing fast. Expect added per-request latency and repeated S3/CloudWatch activity under a
  broken deploy; revisit any alarms that assumed a fast, sticky init failure.

jvh-aws (commented):
A few minor comments. Otherwise it looks good to me

.changeset/core-handler-init-retry.md
pjkroker: yes, great catch

pranavosu (dismissed):

soberm (dismissed):

soberm (dismissed):

Simone319 (dismissed):

soberm (approved):

Simone319 (approved):

PR #367, opened 2026-08-14
feat(bb-realtime): expose Realtime.publishCallbackUrl() for a co-located compute

pjkroker:
**Stacked PR 2 of 3** — AgentCore + Realtime migration. `main` → core (#366) → **this (`grantPublish`)** → bb-agent migration (#368).         
                                                                                                                                                                                                               
## Problem                                                                                                                                                                                                     
The Agent BB ran its streaming loop on **Lambda + SQS + Realtime (API Gateway WebSocket)**. The fundamental ceiling is **Lambda's 15-minute max execution** — long-running / multi-step agents were not achievable on AWS. Bedrock AgentCore Runtime (sessions up to 8h) is purpose-built for this, so we move to loop there.

That loop runs **as the shared Blocks execution role** but in a container *outside* the Blocks handler. To publish to Realtime it needs the API Gateway WebSocket **callback URL** — and that value isn't discoverable outside the handler (the container doesn't get `BLOCKS_CONFIG_BUCKET`/`KEY`, so it can't load the shared S3 config). Realtime needs a way to hand that URL to a co-located BB.

## Scope of this PR                                                                                                                                                                                            
One additive, **CDK-synth-only** change: a public `Realtime.publishCallbackUrl(): string` that returns the shared WebSocket stage's callback URL. It has **no consumer in this PR** — #368's AgentCore Runtime 
is the first caller — so it's behavior-preserving on its own.                                                                                                                                                  
                                                                                                                                                                                                               
**No IAM is granted.** The only publisher (the AgentCore Runtime) runs *as* the shared execution role, which already holds the publish permissions — `execute-api:ManageConnections` (via `wsApi.grantManageConnections(handler)`) and full read/write on the connections table (`DistributedTable` grants it to the same handler). Running as that role inherits both, so a `publish()` from the container needs no extra grant — only the callback URL, which is a config handoff, not an IAM concern.

## Changes                                                                                                                                                                                                     
**`publishCallbackUrl()` — `packages/bb-realtime/src/index.cdk.ts`**                                                                                                                                           
  Returns the shared API Gateway WebSocket stage `callbackUrl` (to inject as `BLOCKS_RT_CALLBACK_URL`). Adds no IAM statements.                                                                                
                                                                                                                                                                                                               
  **Types — `types.ts`:** `RealtimePublishInfo { publishCallbackUrl(): string }`, re-exported from the entry points. (Removed the abandoned `RealtimePublishGrant` / `RealtimePublisherGrants`.)               
                                                                                                                                                                                                               
  **Conditional-export parity:** `index.ts` (default/mock) and `index.aws.ts` (aws-runtime) carry throwing `publishCallbackUrl` stubs (`/CDK-synth-only/`) so it can't silently no-op off the CDK path.        
                                                                                                                                                                                                               
  **Umbrella:** re-exported `RealtimePublishInfo` from `@aws-blocks/blocks` (`index.cdk.ts` + `index.ts`), matching how other BBs surface their types.                                                         
                                                                                                                                                                                                               
  **Tests / docs:** `index.cdk.test.ts` synth test (asserts it returns the stage callback URL); `index.test.ts` guard test (throws outside CDK). `DESIGN.md` records the CDK-synth-only asymmetry in the "Mock 
vs AWS Behavior Differences" table. `API.md` regenerated; changeset (minor).

## Validation
  - `npm run build` + bb-realtime tests (mock + cdk) + `packages/blocks` conditional-export parity all green. 

## Checklist

- [x] PR description included


packages/bb-realtime/src/types.ts
```diff
@@ -131,6 +132,46 @@ export interface RealtimeServer<T extends NamespaceDefs> {
...
 }
 
+// ── External-publisher grant (CDK) ──────────────────────────────────────────
+
+/**
+ * Runtime config an external (non-Blocks-handler) publisher needs to publish to
+ * Realtime channels. Returned by {@link RealtimePublisherGrants.grantPublish}.
+ *
+ * Publishing from outside the Blocks handler Lambda (e.g. an AgentCore Runtime
+ * container) can't use the in-process SDK-identifier registry or the Lambda's S3
```
Simone319: Why couldn't AgentCore use the in-process SDK-identifier registry?

packages/bb-realtime/src/types.ts
```diff
@@ -131,6 +132,46 @@ export interface RealtimeServer<T extends NamespaceDefs> {
...
+	 * fan-out, plus read/query on the connections table to resolve subscribers. Returns the
+	 * runtime config the grantee must inject into its process env to publish.
+	 *
+	 * CDK-synth only — throws in the runtime/mock build.
+	 *
+	 * @param grantee - the IAM principal to grant (e.g. a Runtime's execution role).
+	 * @returns the `BLOCKS_RT_CALLBACK_URL` the grantee must inject (the connections table name is
+	 *   re-derived in-process by `publish()`, so it isn't returned — see {@link RealtimePublishGrant}).
+	 */
+	grantPublish(grantee: IGrantable): RealtimePublishGrant;
```
Simone319: Do we have any use case to grant to an external principal other than the shared executionRole?

If there is no such use case yet, can we just have wsApi grant to the executionRole [here](https://github.com/aws-devtools-labs/aws-blocks/blob/main/packages/bb-realtime/src/index.cdk.ts#L101), instead of relying on this method? The rest of the blocks are already applying the same practice in: https://github.com/aws-devtools-labs/aws-blocks/pull/356/

Simone319 (commented):
In general, I am struggling to understand the need of the PR, depending on if there is use case for granting to external publisher.

packages/bb-realtime/src/types.ts
pjkroker: > Do we have any use case to grant to an external principal other than the shared executionRole?                                                                                                               
                                                                                                                                                                                                                 
Agreed — there isn't one today, and I checked whether the grant does anything for the current sole caller. It doesn't: the caller passes the shared execution role, which already has postToConnection (grantManageConnections(handler)) and full RW on the connections table (DistributedTable grants it to handler) — and the AgentCore Runtime runs as that same role, so it inherits both. grantPublish(role) just
  re-grants a subset it already holds.                                                                                                                                                                           
                                                                                                                                                                                                                 
  The one place an off-shared-role publisher would show up is a BB-specific compute layer, but since that is not there yet, I don't want to bake a public API around it prematurely. So I'll adopt your
  suggestion: drop grantPublish() and rely on the shared role's existing grants (bb-agent only needs the callback URL, which I'll expose minimally). We can add a real grantPublish later. Thanks for catching this.

packages/bb-realtime/src/index.cdk.ts
```diff
@@ -133,6 +136,22 @@ export class Realtime extends Scope {
...
+	 * A co-located Building Block whose compute runs outside the Blocks handler (e.g. the Agent BB's
+	 * AgentCore Runtime) injects this as `BLOCKS_RT_CALLBACK_URL` so its `publish()` can post to
+	 * subscribers. No IAM grant is exposed here on purpose: such a compute runs AS the shared Blocks
+	 * execution role, which already holds the publish permissions (API Gateway `postToConnection` +
+	 * the connections table, both granted to the handler on the same role). The only thing it can't
+	 * otherwise discover outside the handler is this endpoint.
+	 *
+	 * @returns the API Gateway WebSocket callback URL to inject as `BLOCKS_RT_CALLBACK_URL`.
+	 */
+	publishCallbackUrl(): string {
```
Simone319, line 151: nit, up to your deicision: (sorry it slips through the initial review)

**Suggestion: replace `Realtime.publishCallbackUrl()` with a synth-time config-registry read**

`AgentCoreRuntime` currently obtains the Realtime callback URL by calling `props.realtime.publishCallbackUrl()`. Since `bb-realtime` already registers this value into the shared config registry unconditionally in `Realtime`'s constructor (`registerConfig('BLOCKS_RT_CALLBACK_URL', stage.callbackUrl)`, `index.cdk.ts:105`), we can read it back at synth time instead of exposing a bespoke method for it.

Proposed: add a small synth-time getter to `config-registry.ts` (it already has the private `getRegistry`/`REGISTRY_KEY` plumbing — this just surfaces a read):

```ts
export function peekConfig(scope: Construct, key: string): unknown {
  return getRegistry(cdk.Stack.of(scope)).entries.get(key);
}
```

Then in `agentcore-runtime.cdk.ts`:

```ts
// before
const callbackUrl = props.realtime.publishCallbackUrl();

// after
const callbackUrl = peekConfig(this, 'BLOCKS_RT_CALLBACK_URL');
```

**Why this is cleaner:**
- **Removes a single-purpose method from Realtime's public API surface.** `publishCallbackUrl()` exists solely for this one consumer; a generic registry read keeps Realtime's API focused on its actual responsibilities.
- **Eliminates a latent runtime footgun on Realtime's public API.** `publishCallbackUrl()` is a synth-only method that would throw if called from `aws-runtime` code (it reaches for the CDK-time shared stage that doesn't exist in a live process). Because it sits on the public `Realtime` surface, nothing stops a consumer from mistakenly calling it at runtime and hitting that throw. 
- **Reuses existing machinery.** The value is already in the registry (`index.cdk.ts:105`) — we're consuming what's already published rather than adding a parallel accessor. Forwarding the CDK token through `Map.get()` is mechanically identical to today (tokens resolve once at end-of-synth regardless of path).
- **A general-purpose `peekConfig` is reusable** for any future construct that needs a synth-time value produced by another BB without holding a direct reference to it.

PR #368, opened 2026-08-14
feat(bb-agent): run the streaming loop on AgentCore Runtime (keep Realtime)

pjkroker:
**Stacked PR 2 of 2** — AgentCore + Realtime migration. `main` → core (#366) → **this (bb-agent migration)**.

## Problem
The Agent BB ran its streaming loop on **Lambda + SQS + Realtime (API Gateway WebSocket)**. The fundamental ceiling is **Lambda's 15-minute max execution** — long-running / multi-step agents were not achievable on AWS. Bedrock AgentCore Runtime (sessions up to 8h) is purpose-built for this.


## Scope of this PR
The payoff PR: move the Agent BB's Strands loop to a **Bedrock AgentCore Runtime** while **keeping the Realtime BB as the streaming transport** — chunks reach the browser over Realtime exactly as before (*not* AgentCore's `/ws`). Consumes #366 (`backendModulePath` + `getConfigLocation`).

**The client-facing API does not change** — `stream` / `resume` / `getChannel`, the `chunks` Realtime channel, and the `useChat` subscribe contract are all preserved. Locally the loop still runs in-process against the mock Realtime.

**Loop runs on AgentCore (`agent.aws.ts`, `agentcore-entry.ts`)**                                                                                                                                              
- `stream()` / `resume()` call `InvokeAgentRuntime`, which starts the turn as a **background async task** and returns in seconds (it doesn't hold the connection). The microVM stays alive on AgentCore's own terms — AgentCore polls the container's health endpoint and the SDK reports `HealthyBusy` while a background task is in flight, keeping it running up to the 8h max session — while the loop publishes chunks to Realtime.                                                                                                                                                                                                   
- The internal AsyncJob (SQS → Lambda) loop path is removed.                                                                                                                                                  
                                                                                                                                                                                                               
**Provisioning — self-contained `AgentCoreRuntime` (`agentcore-runtime.cdk.ts`)**
- Synth-time co-bundle of the app backend + the AgentCore entry harness; `Runtime` via `fromCodeAsset` (Node 22 CodeZip, no Docker), discovering the backend via `globalThis.CURRENT_BLOCKS_STACK.backendModulePath` (#366).
- Runs **as the shared `BlocksRole`**, so the loop inherits every Building Block's grants — including Realtime publish (granted to the handler on that same role) and other BBs an agent's *tools* touch (KVStore, tables). To that shared role it adds **only** the AgentCore-specific bits, **once per stack** (a `Symbol.for` guard): the scoped `bedrock-agentcore` assume-role **trust** (added here, not in core), Bedrock model access, and the handler's `InvokeAgentRuntime` permission.
- **Injects the config location** (`BLOCKS_CONFIG_BUCKET`/`BLOCKS_CONFIG_KEY`, via #366's `getConfigLocation`) so the container's `loadConfigToProcessEnv()` loads the **same full app config as the handler** — delivering `BLOCKS_RT_CALLBACK_URL` (registered by the Realtime BB) and any other `registerConfig()` value a tool's BB reads. (This is why no `publishCallbackUrl`/grant helper is needed — the earlier #367 was retired.)
                                                                                                                                                                                                               
**Deps**
- Removes `@aws-blocks/bb-async-job`; bumps `@strands-agents/sdk` → `^1.7.0`; adds `bedrock-agentcore` + `@aws-sdk/client-bedrock-agentcore`.
- `API.md` + `DESIGN.md` updated; changeset (minor).
> **On the new `Compute` abstraction (#329):** the AgentCore runtime is kept self-contained deliberately. `Compute` today models *hosting the whole app's RPC handler* (`LambdaCompute`), and there's no "run a
single BB's workload on a dedicated compute" surface yet — which is what an agent-loop-runner needs. So `AgentCoreRuntime → AgentCoreCompute extends Compute` is a **gated follow-up** (unblocked when that    
surface lands), not something to force now.

## Validation
- `npm run build` + bb-agent tests (78 unit + 4 CDK synth + 1 co-bundle) + conditional-export parity + `lint:deps`, all green.
- **Deployment note:** changes the deployed infra shape (adds an AgentCore Runtime, removes the agent's SQS/AsyncJob resources). Flagged for maintainer review even though the app-author API is unchanged.

 ## Checklist                                                                                                                                                                                                   
                                                                                                                                                                                                                 
  - [x] PR description included                                                                                                                                                                                  


packages/bb-agent/src/agentcore-entry.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Standalone launch: load config, import the developer backend by path, then serve.
+ * Used when the artifact runs this file directly with BB_AGENT_BACKEND_MODULE pointing at the
+ * backend. When an app co-bundles the backend with `serve` (agentcore-bundle.ts), it calls
+ * `serve()` directly instead and this `main()` is not the entry.
+ */
+export async function main(): Promise<void> {
+	// Same cold-start contract as the Lambda handler: pull BB resource identifiers
+	// (table names, bucket names, Realtime callback URL) into process.env before importing
+	// the backend, so BB constructors can resolve them.
+	await loadConfigToProcessEnv();
```
pranavosu, line 116: will this work since we don't add the BLOCKS_CONFIG_BUCKET and BLOCKS_CONFIG_KEY vars to the Runtime [here](https://github.com/aws-devtools-labs/aws-blocks/pull/368/changes#diff-945ae68b96d28c4b4aa58c996ed46833fd41aa8392de3ec8904c1aeda998ed84R160)?

This may cause access failures in some building blocks that rely on env vars. e.g we may not be able to call db queries from tools from the AgentCore runtime.

packages/bb-agent/package.json
```diff
@@ -35,23 +35,29 @@
...
     "@aws-blocks/bb-file-bucket": "^0.1.4",
     "@aws-blocks/bb-logger": "^0.1.4",
     "@aws-blocks/bb-realtime": "^0.1.4",
     "@aws-blocks/core": "^0.2.0",
     "@aws-sdk/client-bedrock": "^3.700.0",
+    "@aws-sdk/client-bedrock-agentcore": "^3.700.0",
     "@opentelemetry/api": "^1.9.0",
-    "@strands-agents/sdk": "~1.3.0",
+    "@strands-agents/sdk": "^1.7.0",
+    "bedrock-agentcore": "^0.4.0",
```
pranavosu, line 59: can we avoid using `bedrock-agentcore` ? it pulls in a lot of dependencies that we might not need

packages/bb-agent/package.json
```diff
@@ -35,23 +35,29 @@
...
     "build": "tsc --build",
-    "test": "node --test dist/index.test.js"
+    "test": "node --test dist/index.test.js && node --conditions=cdk --test dist/index.cdk.test.js && node --test dist/agentcore-bundle.test.js"
   },
   "dependencies": {
-    "@aws-blocks/bb-async-job": "^0.1.4",
     "@aws-blocks/bb-distributed-table": "^0.1.5",
     "@aws-blocks/bb-file-bucket": "^0.1.4",
     "@aws-blocks/bb-logger": "^0.1.4",
     "@aws-blocks/bb-realtime": "^0.1.4",
```
Simone319: The change requires the latest version of `@aws-blocks/bb-realtime` that contains the callbackUrl for the endpoint to inject into the container right? We should enforce it here.

packages/bb-agent/src/agentcore-entry.ts
pjkroker: Good catch — you were right, and it's fixed in the latest push. The AgentCore Runtime is now injected                                                                                                   BLOCKS_CONFIG_BUCKET / BLOCKS_CONFIG_KEY (via a new core helper, getConfigLocation()), so the                                                                                                             container's loadConfigToProcessEnv() loads the same full app config the handler does — a tool                                                                                                           that uses a config-backed Building Block now works in the runtime. No new IAM: the runtime runs as                                                                                                       the shared execution role, which already carries the config-read grant.                                                                                                                                                                                                          Side effect: this also delivers BLOCKS_RT_CALLBACK_URL from the config blob (the Realtime BB                                                                                                              registerConfigs it), so the separate publishCallbackUrl handoff became unnecessary and #367 was                                                                                                           retired. See agentcore-runtime.cdk.ts (env injection) and config-registry.ts (getConfigLocation).

packages/bb-agent/package.json
pjkroker: Can we keep it? `bedrock-agentcore` is the official AgentCore Runtime harness (implements the /ping +                                                                                                        /invocations + async-task contract), and it's imported in exactly one place — `agentcore-entry.ts` —                                                                                                      which is only ever bundled into the AgentCore code asset (the Node CodeZip). So its transitive deps                                                                                                     (Fastify/ws/etc.) don't reach the Lambda handler or the browser/mock bundles; they only affect the                                                                                                      container image. Hand-rolling that serving contract to drop deps would be fragile for marginal gain.

packages/bb-agent/package.json
pjkroker: This one's moot now after a design change. bb-agent no longer depends on a bb-realtime callbackUrl                                                                                                        helper — I retired that approach (#367's publishCallbackUrl).

Simone319 (dismissed):

packages/bb-agent/package.json
pranavosu: It doesn't affect the runtime but npm installs the whole tree for every consumer regardless of which code path imports it. Ideally, we minimize that impact but let's keep this for now and we can revisit later.

svidgen: Is it big enough that we should considering "unlinking" the `Agent` BB from `@aws-blocks/blocks`? We could add a section to that README of "Supported Optional Packages" or similar. "To add them, first `npm install` the listed NPM package."

pjkroker: Agreed, this is a good idea, and I think it's best as a follow-up rather than folding it into this PR (this migration doesn't change the linkage; it just made the deps heavier, which surfaced it). I'll add it to the board.                                                                                                                                                                                         

Worth noting the weight is mostly pre-existing: even before AgentCore, bb-agent was already the heaviest BB by far, e.g. openai,  and Strands/MCP. `AgentCore `added new harness deps (`bedrock-agentcore` + its Fastify/ws stack) on top. So the "optional package" case holds on the Agent BB's inherent weight regardless of this migration.  

And we should look at `bb-knowledge-base` in the same pass — it's the other heavy BB and the natural second member of a "Supported Optional Packages" tier, so it's worth designing the pattern (criteria for optional-vs-bundled, how the umbrella surfaces them, the npm install docs) to cover both rather than one-off-ing the Agent.

pranavosu (dismissed):

packages/bb-agent/src/agent.ts
```diff
@@ -199,17 +205,65 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
...
+	 * Run one agent turn (initial message or HITL resume) and publish its chunks to Realtime.
+	 *
+	 * This is the single execution entry the compute layer invokes: the AgentCore Runtime
+	 * entrypoint (agentcore-entry.ts) calls it as a background async task on AWS, and locally
+	 * {@link dispatchTurn} calls it in-process. Errors are caught and published as an `error`
+	 * chunk (not re-thrown) so the client never hangs and a non-idempotent turn isn't retried.
+	 *
+	 * @param payload - the turn to run (see {@link AgentTurnPayload}).
+	 * @internal Invoked by the compute layer (agentcore-entry / dispatchTurn), not customer API.
+	 */
```
jvh-aws, line 261: On the mock path (`agent.ts:255`), `dispatchTurn` runs `invokeTurn` in-process, so a tool receives the **live** `context` object. On AWS (`agent.aws.ts:69`), `context` is `JSON.stringify`'d into the `InvokeAgentRuntime` payload and re-parsed by Zod in the container.
  
I verified against the real code: a `context` of `{ when: new Date(), tags: new Set(['a']), note: undefined }` reaches a tool handler as a live `Date` + live `Set` locally, but on AWS arrives as `when: "2020-01-01T00:00:00.000Z"` (string), `tags: {}` (silent data loss), and `note` dropped. So `context.when.getTime()` or `context.tags.has(...)` passes every local test and throws/misbehaves after `cdk deploy`.
  
**Ask:** round-trip `context` through `JSON.parse(JSON.stringify(context))` (or validate it against `toolContextSchema`) inside the mock's `dispatchTurn`, so local dev reproduces the wire boundary. This turns a deploy-only failure into a local one.

jvh-aws (commented):
I noticed there's no way to actually stop a running agent - no stop/cancel/abort anywhere. It wasn't a big deal on the old lambda path since it capped out at 15min, but now that the loop runs on agentcore it can go for up to 8h, so a runaway agent could create a lot of bedrock spend. We probably want to include a stop and potentially poll API. Potentially also worth passing a maxTurns cap through AgentConfig into the strands loop so it self-limits?

packages/bb-agent/src/agent.ts
pjkroker: Great catch — fixed. The mock dispatchTurn now round-trips context through JSON.parse(JSON.stringify(...)) before running the loop, so a tool sees the same wire shape locally as on AWS (Date→string,  Set→{}, undefined dropped) — turning that deploy-only failure into a local one.

soberm (dismissed):

packages/bb-agent/src/agentcore-runtime.cdk.ts
```diff
@@ -43,11 +44,19 @@ export interface AgentCoreRuntimeProps {
 	agentcoreAssetPath?: string;
 }
 
-export class AgentCoreRuntime extends Scope {
+export class AgentCoreRuntime extends Compute {
```
Simone319: Why are we making AgentCoreRuntime extend Compute? Computes are now the bbs that are used to run the API/AsyncJob/CronJob workloads. And there are certain contract that all Compute should implement. Even though it doesn't show any sign that AgentCore couldn't fulfill any of the contracts, I would postpone adding it to Compute unless there's a need.

soberm: Agreed — reverted. `AgentCoreRuntime` goes back to `extends Scope` with its existing `getConfigLocation()` self-injection.

Context on why it was there and why it's safe to drop: I originally added the `Compute` extension while chasing the E2E failures (Aurora/DSQL/Realtime "not configured"), on the theory that the container wasn't receiving `BLOCKS_CONFIG_BUCKET/KEY`. Further investigation showed that wasn't the cause: the Agent E2E suite (which is what actually exercises this container) was passing, and the real failure was a Lambda-side config self-heal gap — the handler's post-deploy retry was gated on `initialize()` throwing, but the transient empty-config path returned without throwing, so the request-serving Lambda locked in empty config for the container's lifetime. That fix lives on the base PR (#366, `TransientConfigError` / `isConfigResolved()`), and #368 inherits it.

So there's no demonstrated need to make AgentCore a `Compute` — the container already gets the authoritative config coordinates via `getConfigLocation()`, and its own entry loads them at cold start. You're right that `Compute` is the API/AsyncJob/CronJob-workload abstraction with its own contract, and coupling AgentCore to it without a need overloads that abstraction. Removed.

soberm (dismissed):

Simone319 (dismissed):

PR #371, opened 2026-08-14
fix(bb-knowledge-base): disable installLatestAwsSdk on ingestion custom resource

soberm:
## Problem

The `StartIngestion` `AwsCustomResource` in `bb-knowledge-base` left
`installLatestAwsSdk` at its CDK default of `true`. With that default, the
`AWS679f53fac002430cb0da5b7982bd2287` provider Lambda runs an `npm install` of
the AWS SDK on every invocation before it can call the API. That costs roughly
15-30s of extra cold start and forces the provider up to 512MB of memory.

None of that is needed here: the resource calls
`BedrockAgent.startIngestionJob`, a stable API already bundled in the Lambda
runtime's AWS SDK v3. The default also causes CDK to emit an
`installLatestAwsSdkNotSpecified` warning at synth time.

## Changes

- `packages/bb-knowledge-base/src/index.cdk.ts` — added `installLatestAwsSdk: false`
  to the `new cr.AwsCustomResource(this, 'StartIngestion', { ... })` options, with a
  short comment explaining why the bundled SDK is sufficient.
- `packages/bb-knowledge-base/src/index.cdk.test.ts` — added a regression test
  asserting the synthesized `Custom::AWS` resource renders
  `InstallLatestAwsSdk: false`.
- `.changeset/kb-ingestion-install-latest-sdk.md` — patch bump for
  `@aws-blocks/bb-knowledge-base` and the `@aws-blocks/blocks` umbrella (which
  re-exports `KnowledgeBase`).

Internal construct wiring only — no public API change, so `API.md` is untouched.

Scope note: the `packages/hosting/src/cdn_construct.ts` `DeployInvalidation`
custom resource has the same anti-pattern. It is intentionally left out of this
PR.

## Validation

**Clean build** (`rm -rf dist *.tsbuildinfo && npm run build`, Node 22.22.3): exit 0.

**Unit tests** (`npm test -w packages/bb-knowledge-base`, Node 22.22.3):
130 passed / 0 failed across 15 suites. The `index.cdk.test.js` file alone is
9 passed / 0 failed, including the new assertion.

**Synth evidence.** Synthesizing the construct and dumping `Custom::AWS` shows
the property actually rendering in the CloudFormation template:

```json
=== Custom::AWS logical id: appdocsStartIngestion693D6A58 ===
{
  "InstallLatestAwsSdk": false,
  "Create": {
    "Fn::Join": ["", [
      "{\"service\":\"BedrockAgent\",\"action\":\"startIngestionJob\",\"parameters\":{\"knowledgeBaseId\":\"",
      { "Fn::GetAtt": ["appdocsKB240A965D", "KnowledgeBaseId"] },
      "\",\"dataSourceId\":\"",
      { "Fn::GetAtt": ["appdocsDataSource4AE12091", "DataSourceId"] },
      "\"},\"physicalResourceId\":{\"id\":\"", "..." , "-ingest\"}}"
    ]]
  }
}
```

The `installLatestAwsSdkNotSpecified` synth-time warning is no longer emitted
for this construct (grep of synth stderr returns nothing).

**Non-vacuity (red/green).** With the `installLatestAwsSdk: false` line removed
and the package rebuilt, the new test fails (`# pass 0 / # fail 1`) — before the
change the property is absent from the synthesized template. Restoring the line
returns it to green.

## Checklist

- [x] PR description included
- [x] Relevant documentation is changed or added (and PR referenced) — n/a, no public API or doc surface changed; changeset added

---

### DIFF
--- a/.changeset/kb-ingestion-install-latest-sdk.md
+++ b/.changeset/kb-ingestion-install-latest-sdk.md
@@ -1,0 +1,20 @@
+---
+"@aws-blocks/bb-knowledge-base": patch
+"@aws-blocks/blocks": patch
+---
+
+fix(bb-knowledge-base): disable `installLatestAwsSdk` on the ingestion custom resource
+
+The `StartIngestion` `AwsCustomResource` left `installLatestAwsSdk` at its default
+(`true`), so the provider Lambda ran an `npm install` of the AWS SDK on every
+invocation before it could call the API. That costs roughly 15-30s of extra cold
+start and forces the provider up to 512MB of memory.
+
+Nothing needed it. The resource calls `BedrockAgent.startIngestionJob`, a stable
+API already bundled in the Lambda runtime's AWS SDK v3, so the bundled client is
+sufficient and the install is pure overhead. Setting `installLatestAwsSdk: false`
+also silences CDK's `installLatestAwsSdkNotSpecified` synth-time warning for this
+construct.
+
+Internal construct wiring only — no public API change. The synthesized
+`Custom::AWS` resource now renders `InstallLatestAwsSdk: false`.
--- a/.changeset/kb-ingestion-install-latest-sdk.md
+++ b/.changeset/kb-ingestion-install-latest-sdk.md
@@ -8,7 +8,10 @@ fix(bb-knowledge-base): disable `installLatestAwsSdk` on the ingestion custom re
 The `StartIngestion` `AwsCustomResource` left `installLatestAwsSdk` at its default
 (`true`), so the provider Lambda ran an `npm install` of the AWS SDK on every
 invocation before it could call the API. That costs roughly 15-30s of extra cold
-start and forces the provider up to 512MB of memory.
+start and raises the shared provider Lambda to 512MB of memory, though that
+reduction is only realized once every `AwsCustomResource` in the stack opts out —
+the provider is a stack-level singleton and its `memorySize` is resolved once at
+synth. The per-resource cold-start saving applies regardless.
 
 Nothing needed it. The resource calls `BedrockAgent.startIngestionJob`, a stable
 API already bundled in the Lambda runtime's AWS SDK v3, so the bundled client is
@@ -23,3 +26,9 @@ since the client's initial release, not a recent addition.
 
 Internal construct wiring only — no public API change. The synthesized
 `Custom::AWS` resource now renders `InstallLatestAwsSdk: false`.
+
+Upgrade note: because `InstallLatestAwsSdk` renders as a property on the
+`Custom::AWS` resource while the `physicalResourceId` stays stable, the first
+deploy after upgrading is a CloudFormation *Update* and fires `onUpdate` — so one
+`startIngestionJob` kicks off. This is harmless (ingestion is idempotent and
+fire-and-forget), but expect to see an ingestion start right after upgrading.
--- a/.changeset/kb-ingestion-install-latest-sdk.md
+++ b/.changeset/kb-ingestion-install-latest-sdk.md
@@ -16,5 +16,10 @@ sufficient and the install is pure overhead. Setting `installLatestAwsSdk: false
 also silences CDK's `installLatestAwsSdkNotSpecified` synth-time warning for this
 construct.
 
+The tradeoff being accepted: the provider now uses whichever SDK v3 the Lambda
+runtime bundles at deploy time rather than installing the newest one. That is safe
+here because `startIngestionJob` is a foundational Bedrock Agent operation present
+since the client's initial release, not a recent addition.
+
 Internal construct wiring only — no public API change. The synthesized
 `Custom::AWS` resource now renders `InstallLatestAwsSdk: false`.
--- a/packages/bb-knowledge-base/src/index.cdk.test.ts
+++ b/packages/bb-knowledge-base/src/index.cdk.test.ts
@@ -271,9 +271,8 @@ test('CDK (s3:// source): DATA_SOURCE_ID config is wired to the data source id (
 test('CDK: the ingestion AwsCustomResource opts out of installing the latest AWS SDK (InstallLatestAwsSdk: false)', () => {
   const template = synth();
 
-  // startIngestionJob is a stable BedrockAgent API bundled in the Lambda
-  // runtime's SDK v3 — the provider must NOT npm-install the SDK at invoke time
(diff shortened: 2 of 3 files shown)


cadivus (dismissed):

.changeset/kb-ingestion-install-latest-sdk.md
```diff
@@ -0,0 +1,25 @@
...
+"@aws-blocks/bb-knowledge-base": patch
+"@aws-blocks/blocks": patch
+---
+
+fix(bb-knowledge-base): disable `installLatestAwsSdk` on the ingestion custom resource
+
+The `StartIngestion` `AwsCustomResource` left `installLatestAwsSdk` at its default
+(`true`), so the provider Lambda ran an `npm install` of the AWS SDK on every
+invocation before it could call the API. That costs roughly 15-30s of extra cold
+start and forces the provider up to 512MB of memory.
```
sarayev: **minor** — Small accuracy thing on the memory claim. The `AWS679f…` provider is a stack-level singleton shared by every `AwsCustomResource` in the stack, and its `memorySize` is resolved once at synth. So the 512MB → default drop only actually happens if *all* AwsCustomResources in the deployed stack opt out. Since the hosting CDN's `DeployInvalidation` (which the scope note calls out) still uses the default, any app that composes KnowledgeBase + hosting keeps the shared provider at 512MB and won't see the memory win.

The per-resource cold-start / npm-install saving is real regardless — it's just the memory framing that reads a touch optimistic. Could we soften this to something like: "…raises the shared provider Lambda to 512MB, though that reduction is only realized once every AwsCustomResource in the stack opts out."

packages/bb-knowledge-base/src/index.cdk.ts
```diff
@@ -431,6 +431,9 @@ export class KnowledgeBase extends Scope {
 				},
 				physicalResourceId: cr.PhysicalResourceId.of(stableIngestId),
 			},
+			// Stable BedrockAgent API bundled in the Lambda runtime's SDK v3 — skip the
+			// npm-install at invoke time (avoids ~15-30s cold start + 512MB bump).
+			installLatestAwsSdk: false,
```
sarayev, line 437: **minor** — Worth a one-line heads-up in the changeset here. Because `InstallLatestAwsSdk` renders as a property on the `Custom::AWS` resource and the `physicalResourceId` is stable, adding it is a CloudFormation *Update* for existing stacks — which fires `onUpdate` → one `startIngestionJob` on the next deploy. It's completely harmless (ingestion is idempotent and fire-and-forget), but it'd be good to note it so nobody's surprised to see an ingestion kick off right after upgrading.

packages/bb-knowledge-base/src/index.cdk.test.ts
```diff
@@ -268,6 +268,14 @@ test('CDK (s3:// source): DATA_SOURCE_ID config is wired to the data source id (
   );
 });
 
+test('CDK: the ingestion AwsCustomResource opts out of installing the latest AWS SDK (InstallLatestAwsSdk: false)', () => {
+  const template = synth();
+
+  // StartIngestion is the only Custom::AWS resource in this template, so an
+  // unscoped match is unambiguous.
+  template.hasResourceProperties('Custom::AWS', Match.objectLike({ InstallLatestAwsSdk: false }));
```
sarayev, line 279: **minor** — This is correct today: `BucketDeployment` synthesizes `Custom::CDKBucketDeployment`, not `Custom::AWS`, so the comment holds. The one catch is that `hasResourceProperties` only checks that *some* `Custom::AWS` matches — if a second `AwsCustomResource` with `installLatestAwsSdk: false` ever lands in this construct, this would still pass even if `StartIngestion` itself regressed to `true`.

Dropping a `template.resourceCountIs('Custom::AWS', 1);` right before this line turns that load-bearing comment into an actually-enforced invariant, so the test fails loudly if a second one shows up.

sarayev (commented):
Took a close look and this is a nice, tight change — the rationale holds up. `startIngestionJob` is a long-stable BedrockAgent API and it's bundled in the provider runtime's SDK v3, so opting out of the invoke-time install is the right call. No correctness, security, or logic concerns from me.

Left three minor notes inline: two are just changelog/wording accuracy, one is a small test-hardening suggestion. None are blocking.

One informational aside (no action needed): this assumes `@aws-sdk/client-bedrock-agent` is present in the provider Lambda's bundled SDK — which it is for the nodejs18.x+ AwsCustomResource provider, and the client has been GA since 2023, so we're safe there.

.changeset/kb-ingestion-install-latest-sdk.md
soberm: Good catch — softened. The changeset now reads that the change 'raises the shared provider Lambda to 512MB of memory, though that reduction is only realized once every `AwsCustomResource` in the stack opts out', and explains the stack-level-singleton / synth-time `memorySize` reason, with the per-resource cold-start saving called out as unconditional. I applied the same softening to the inline code comment in `index.cdk.ts`, which had the same optimistic '+ 512MB bump' framing. Fixed in 6f568f6.

packages/bb-knowledge-base/src/index.cdk.ts
soberm: Agreed, worth flagging. Added an 'Upgrade note' paragraph to the changeset: because `InstallLatestAwsSdk` renders as a property while the `physicalResourceId` stays stable, the first deploy after upgrading is a CloudFormation *Update* and fires `onUpdate`, so one `startIngestionJob` kicks off — harmless (idempotent, fire-and-forget), but no longer a surprise. Fixed in 6f568f6.

packages/bb-knowledge-base/src/index.cdk.test.ts
soberm: Right, the comment was load-bearing but unenforced. Added `template.resourceCountIs('Custom::AWS', 1);` immediately before the property assertion so a second `AwsCustomResource` fails the test loudly instead of silently masking a `StartIngestion` regression. Re-verified red/green: with `installLatestAwsSdk: false` removed the test fails, restored it passes (9 pass / 0 fail in `index.cdk.test.js`, 130 pass / 0 fail for the package). Fixed in 6f568f6.

sarayev (approved):

cadivus (approved):

PR #374, opened 2026-08-14
fix(hosting): use nodejs24.x for SSR framework compute and fail loudly on unknown runtimes (#361)

### LINKED ISSUE #361: bug(hosting): Nuxt SSR (and other frameworks) Lambda is pinned to deprecated nodejs20.x (undocumented), compute construct also uses old node versions

Other

### Describe the bug

AWS Blocks Hosting deploys a newly built Nuxt/Nitro SSR bundle using the deprecated Lambda runtime `nodejs20.x`, even when the application and synthesis process use Node.js 24 (or newer versions). This version is not "officially supported" by AWS and may not receive security updates, even though new creation is not blocked yet. https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported. Furthermore, having hidden runtime changes creates hard to recreate bugs between local/production.

The Nitro adapter explicitly emits `runtime: 'nodejs20.x'` for both its `node-server` and `aws-lambda` compute resources. `ComputeConstruct.resolveRuntime()` also has no `nodejs24.x` case and silently falls back to Node.js 20 for unrecognized runtime strings.

In a production synthesis using Node.js 24.18.0, the resulting CloudFormation resource was:

```text
HostingComputedefaultFunction35E53DE1
Runtime: nodejs20.x
```

This is not limited to the developer's local Node version failing to propagate: Hosting itself rebuilds the Nuxt application during synthesis, then assigns the resulting SSR Lambda to Node.js 20.

Node.js 20 was deprecated by AWS Lambda on April 30, 2026. AWS Blocks Hosting's own internal Lambda handlers already use `Runtime.NODEJS_24_X`; only the framework compute produced by the adapters remains pinned to Node.js 20.

### Expected behavior

Hosting should deploy newly built Nuxt/Nitro SSR compute on a supported, configurable Node.js Lambda runtime, with its build and runtime compatibility contract kept aligned. I'm not saying support Node26 now, btut any versions officially supported should be available. This can dynamically read the tsconfig, but just having a default and a parameter to pass like vanilla CDK would be great.

At minimum:

- the Nitro adapter should support Node.js 24 rather than unconditionally emitting `nodejs20.x`;
- the runtime resolver should recognize `nodejs24.x`; and
- unsupported runtime strings should produce a clear error instead of silently falling back to Node.js 20.

Node.js 20 may remain available as an explicit legacy compatibility choice, but should not be the unconditional runtime for new framework builds.

### Reproduction steps

1. Use Node.js 24 and install `@aws-blocks/blocks@0.2.6`, which resolves `@aws-blocks/hosting@0.1.6`.
2. Create a Nuxt application with a production `Hosting` construct using `framework: 'nuxt'`.
3. Synthesize the production CDK application (not sandbox mode).
4. Observe that Hosting runs the Nuxt build, including an `NITRO_PRESET=aws-lambda` build.
5. Inspect the synthesized `AWS::Lambda::Function` for `Hosting/Compute-default`.
6. Observe that its `Runtime` is `nodejs20.x`, and CDK emits a deprecation warning recommending `nodejs24.x`.

### Code snippet

```typescript
new Hosting(blocksStack, 'Hosting', {
  root: join(__dirname, '..'),
  buildCommand: 'pnpm build',
  buildOutputDir: '.output',
  framework: 'nuxt',
  api: blocksStack,
  compute: {
    memorySize: 1024,
    timeout: cdk.Duration.seconds(30),
  },
})
```

### Log output

```plain text
Running build (NITRO_PRESET=aws-lambda): npm run build

WARNING Runtime: Runtime 'nodejs20.x' was deprecated on '2026-04-30'. Creation was disabled on '2027-02-01' and update on '2027-03-03'. Please consider updating to 'nodejs24.x' (CloudFormation Validate)
   .../Hosting/Compute-default/Function/Resource (...) aws-cdk-lib.aws_lambda.CfnFunction
```

### Environment

- AWS Blocks version: `@aws-blocks/blocks@0.2.6`; `@aws-blocks/hosting@0.1.6`
- Node.js version: `24.18.0`
- Package manager (npm/yarn/pnpm): `pnpm 11.21.0`
- OS: `macOS 26.5.2`
- CDK version (if applicable): `aws-cdk-lib 2.262.1`
- Framework: `Nuxt 4.5.1`, `Nitro 2.13.4

### Additional context

The installed package contains both halves of the behavior:

- `packages/hosting/src/adapters/nitro.ts` emits `runtime: 'nodejs20.x'` for both supported Nitro compute shapes.
- `packages/hosting/src/constructs/compute_construct.ts` recognizes Node.js 18, 20, and 22, but not Node.js 24; unknown values fall back to Node.js 20.
- `packages/hosting/src/constructs/node_runtime.ts` already uses `Runtime.NODEJS_24_X` for Hosting-owned handlers while explicitly excluding framework SSR/edge bundles.

AWS Lambda currently supports `nodejs22.x` through April 30, 2027 and `nodejs24.x` through April 30, 2028:

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

sarayev:
## Summary

Fixes #361. AWS Blocks Hosting was deploying SSR framework Lambdas (Nuxt/Nitro, Astro, SvelteKit, Next.js) on the deprecated `nodejs20.x` runtime, and `resolveRuntime()` silently fell back to Node 20 for any runtime it didn't recognize — masking the problem and making a partial fix ineffective.

## Changes

- **New shared constants** (`framework_runtime.ts`): `FRAMEWORK_COMPUTE_RUNTIME = 'nodejs24.x'` for regional SSR compute and `FRAMEWORK_EDGE_COMPUTE_RUNTIME = 'nodejs24.x'` for Lambda@Edge. Adapters reference these instead of hardcoded literals, so future bumps are a one-line change.
- **`resolveRuntime()`** now recognizes `nodejs24.x`, defaults to `nodejs24.x` when no runtime is declared, and **throws `UnsupportedRuntimeError`** (with the list of supported values) instead of silently returning `NODEJS_20_X`.
- Updated adapters: `nitro.ts` (both presets), `astro.ts`, `sveltekit.ts`, `nextjs.ts` (regional/server-function **and** Lambda@Edge paths).
- Corrected a misleading comment in `node_runtime.ts` that implied the adapters detect the framework's compiled Node version (they don't — the runtime is a fixed shared constant).
- Also fixed two hosting-owned handlers (ISR revalidation worker + tag-table seed) that were pinned to Node 20 against the stated `DEFAULT_NODE_RUNTIME` contract.

## Lambda@Edge: also bumped to nodejs24.x

An earlier revision of this PR left the edge path on `nodejs20.x` out of caution. That decision has been revised — the edge path is now bumped too:

1. **Lambda@Edge supports Node 24.** For Node.js, Lambda@Edge draws from the same managed runtime table as regional Lambda: `nodejs24.x` is supported (deprecates 2028-04-30), while `nodejs20.x` is already past its 2026-04-30 deprecation, so it can no longer back newly-created edge functions. Staying on Node 20 was the riskier option.
2. **The OpenNext banner patch was revalidated.** `patchEdgeBundlesForLambdaEdge` rewrites OpenNext's `import * as process from "node:process"` banner. The `TypeError: Cannot assign to read only property 'env' of object '[object Module]'` it works around is **spec-level** behavior — ES Module namespace object bindings are non-writable per ECMA-262 — not a Node 20 quirk. Its replacement (`const process = (await import("node:process")).default;`) relies only on top-level `await import(...)` and the long-stable `node:process` default export, both of which behave identically on Node 20, 22, and 24. There is no Node-version-specific assumption in the patch, so no fallback to `nodejs22.x` was needed.

Code comments in `framework_runtime.ts` and `nextjs.ts` were updated to record this reasoning and to drop the now-inaccurate "Node 20 ESM semantics" framing.

## Testing

- `npm run build` — clean (`tsc --build` + handler bundling), Node 22
- `npm test` — **832/832 pass**, 0 fail (+3 new tests): default runtime resolves to `nodejs24.x`; explicit `nodejs22.x` honored; unrecognized runtime (`python3.12`) throws `UnsupportedRuntimeError`.
- Existing tests asserting `nodejs20.x` (regional and edge) updated to `nodejs24.x`.

Note: `resolveRuntime()` still *accepts* explicit `nodejs20.x`/`nodejs18.x` from a user-supplied manifest — this change only moves the framework-emitted defaults.

### DIFF
--- a/packages/hosting/src/adapters/nextjs.ts
+++ b/packages/hosting/src/adapters/nextjs.ts
@@ -167,7 +167,10 @@ export const nextjsAdapter = (
     patchStreamingWrapperForApiGateway(openNextDir);
 
     // Patch each edge bundle's process import banner for Lambda@Edge
-    // compatibility. See patchEdgeBundleForLambdaEdge.
+    // compatibility. See patchEdgeBundleForLambdaEdge for the mechanics, and
+    // the FRAMEWORK_EDGE_COMPUTE_RUNTIME doc comment in framework_runtime.ts
+    // for why this patch is runtime-independent (ES Module namespace exports
+    // are non-writable per spec — it is not a Node-version quirk).
     patchEdgeBundlesForLambdaEdge(openNextDir);
 
     // Patch the image-optimization bundle for the Next.js 16 image-optimizer
--- a/packages/hosting/src/adapters/nextjs.ts
+++ b/packages/hosting/src/adapters/nextjs.ts
@@ -19,6 +19,10 @@ import fg from 'fast-glob';
 import semver from 'semver';
 import { getPackageInfoSync } from 'local-pkg';
 import { HostingError } from '../hosting_error.js';
+import {
+  FRAMEWORK_COMPUTE_RUNTIME,
+  FRAMEWORK_EDGE_COMPUTE_RUNTIME,
+} from '../framework_runtime.js';
 import type {
   ComputeResource,
   CustomHeader,
@@ -1945,7 +1949,7 @@ const translateOpenNextOutput = (
         handler: fn.handler ?? 'index.handler',
         placement: 'global',
         streaming: false,
-        runtime: 'nodejs20.x',
+        runtime: FRAMEWORK_EDGE_COMPUTE_RUNTIME,
       };
     }
   }
@@ -2076,7 +2080,7 @@ const mapOriginToCompute = (
       handler: origin.handler ?? 'index.handler',
       placement: 'regional',
       streaming: origin.streaming ?? true,
-      runtime: origin.runtime ?? 'nodejs20.x',
+      runtime: origin.runtime ?? FRAMEWORK_COMPUTE_RUNTIME,
       memorySize: origin.memorySize,
       timeout: origin.timeout,
       environment: origin.environment,
@@ -2091,7 +2095,7 @@ const mapOriginToCompute = (
       port: origin.port ?? 3000,
       placement: 'regional',
       streaming: origin.streaming ?? false,
-      runtime: origin.runtime ?? 'nodejs20.x',
+      runtime: origin.runtime ?? FRAMEWORK_COMPUTE_RUNTIME,
       environment: origin.environment,
     };
   }
@@ -2103,7 +2107,7 @@ const mapOriginToCompute = (
       handler: origin.handler ?? 'index.handler',
       placement: 'global',
       streaming: false,
-      runtime: origin.runtime ?? 'nodejs20.x',
+      runtime: origin.runtime ?? FRAMEWORK_EDGE_COMPUTE_RUNTIME,
       environment: origin.environment,
     };
   }
@@ -2115,7 +2119,7 @@ const mapOriginToCompute = (
     handler: 'index.handler',
     placement: 'regional',
     streaming: origin.streaming ?? true,
-    runtime: 'nodejs20.x',
+    runtime: FRAMEWORK_COMPUTE_RUNTIME,
   };
 };
 
--- a/packages/hosting/src/adapters/nextjs.ts
+++ b/packages/hosting/src/adapters/nextjs.ts
@@ -167,7 +167,7 @@ export const nextjsAdapter = (
     patchStreamingWrapperForApiGateway(openNextDir);
 
     // Patch each edge bundle's process import banner for Lambda@Edge
-    // nodejs20.x compatibility. See patchEdgeBundleForLambdaEdge.
+    // compatibility. See patchEdgeBundleForLambdaEdge.
     patchEdgeBundlesForLambdaEdge(openNextDir);
 
     // Patch the image-optimization bundle for the Next.js 16 image-optimizer
@@ -1335,18 +1335,20 @@ export const patchStreamingWrapperForApiGateway = (
 };
 
 /**
- * Patch every OpenNext edge bundle so it runs on Lambda@Edge nodejs20.x.
+ * Patch every OpenNext edge bundle so it runs on Lambda@Edge.
  *
  * OpenNext prepends each edge bundle with `import * as process from "node:process";`
- * (banner injected by createEdgeBundle.ts). Under Node 20 ESM the resulting
- * Module namespace's `env` binding is non-writable. The next-emitted Next.js
+ * (banner injected by createEdgeBundle.ts). Per the ECMAScript spec a Module
+ * namespace object's bindings are non-writable, so the next-emitted Next.js
  * shim then runs `process.env = e.g.process.env` and crashes with:
  *
  * TypeError: Cannot assign to read only property 'env' of object '[object Module]'
  *
  * Replace the namespace import with a default import so `process` is the
  * live mutable singleton (default export of `node:process`). Idempotent —
- * re-running on an already-patched file finds no banner to swap.
+ * re-running on an already-patched file finds no banner to swap. Version-
+ * agnostic: it relies only on top-level `await import(...)` and the
+ * `node:process` default export, both stable across Node 18–24.
  *
  * Affects only edge functions (Lambda@Edge); the regional `default` bundle
  * uses a different banner and is unaffected.
@@ -1401,7 +1403,7 @@ export const patchEdgeBundlesForLambdaEdge = (openNextDir: string): void => {
     });
   }
   process.stderr.write(
-    `\u{1F527} Patched ${total} edge bundle(s) for Lambda@Edge nodejs20.x compatibility.\n`,
+    `\u{1F527} Patched ${total} edge bundle(s) for Lambda@Edge compatibility.\n`,
   );
 };
 
--- a/packages/hosting/src/constructs/compute_construct.ts
+++ b/packages/hosting/src/constructs/compute_construct.ts
@@ -342,16 +342,27 @@ export class ComputeConstruct extends Construct {
   }
 
   private resolveRuntime(runtime?: string): Runtime {
-    if (!runtime || runtime === 'nodejs20.x') {
-      return Runtime.NODEJS_20_X;
+    if (!runtime) {
+      return Runtime.NODEJS_24_X;
+    }
+    if (runtime === 'nodejs24.x') {
+      return Runtime.NODEJS_24_X;
     }
     if (runtime === 'nodejs22.x') {
       return Runtime.NODEJS_22_X;
     }
+    if (runtime === 'nodejs20.x') {
+      return Runtime.NODEJS_20_X;
+    }
     if (runtime === 'nodejs18.x') {
       return Runtime.NODEJS_18_X;
     }
-    return Runtime.NODEJS_20_X;
+    throw new HostingError('UnsupportedRuntimeError', {
+      message: `Unsupported compute runtime '${runtime}'.`,
+      resolution:
+        'Use one of: nodejs24.x, nodejs22.x, nodejs20.x, nodejs18.x. ' +
+        'Omit the runtime to use the default (nodejs24.x).',
+    });
   }
 
   private validateWebAdapterRegion(
--- a/packages/hosting/src/constructs/compute_construct.ts
+++ b/packages/hosting/src/constructs/compute_construct.ts
@@ -212,7 +212,7 @@ export class ComputeConstruct extends Construct {
     if (computeResource.type === 'handler') {
       // Native Lambda handler — no Web Adapter needed
       this.function = new LambdaFunction(this, 'Function', {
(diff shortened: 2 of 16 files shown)


packages/hosting/src/constructs/compute_construct.ts
```diff
@@ -342,16 +342,27 @@ export class ComputeConstruct extends Construct {
...
+    if (runtime === 'nodejs24.x') {
+      return Runtime.NODEJS_24_X;
     }
     if (runtime === 'nodejs22.x') {
       return Runtime.NODEJS_22_X;
     }
+    if (runtime === 'nodejs20.x') {
+      return Runtime.NODEJS_20_X;
+    }
     if (runtime === 'nodejs18.x') {
```
soberm, line 357: Is `nodejs18.x` actually still worth supporting here given it's long past EOL (April 2025)? Since this PR is already tightening the "silently fall back" behavior, might be worth throwing for `nodejs18.x` too and only allowing 20/22/24, or at least flagging as a follow-up. Not blocking though, the PR description says it deliberately keeps accepting user-supplied 20/18, so probably intentional.

packages/hosting/src/constructs/compute_construct.ts
```diff
@@ -342,16 +342,27 @@ export class ComputeConstruct extends Construct {
...
       return Runtime.NODEJS_18_X;
     }
-    return Runtime.NODEJS_20_X;
+    throw new HostingError('UnsupportedRuntimeError', {
```
soberm, line 360: Worth double-checking whether other `HostingError` codes in this package follow a naming convention (do they end in `Error` or not?) before this ships as a public-facing error code users might match on.

packages/hosting/src/adapters/nextjs.ts
```diff
@@ -2076,7 +2082,7 @@ const mapOriginToCompute = (
       handler: origin.handler ?? 'index.handler',
```
soberm, line 2085: Since `resolveRuntime` now throws on anything not in its allow-list, if `origin.runtime` is ever a genuinely unsupported string, it'll surface as a synth-time throw from deep inside `ComputeConstruct` rather than adapter-level validation. Probably fine since fail-loud is the point of this PR, just want to confirm the error at that point still has enough context (bundle/resource name) for a user to know which compute resource was misconfigured.

packages/hosting/src/adapters/nextjs.ts
```diff
@@ -163,7 +167,7 @@ export const nextjsAdapter = (
     patchStreamingWrapperForApiGateway(openNextDir);
 
     // Patch each edge bundle's process import banner for Lambda@Edge
```
soberm, line 169: This comment reads a little vague now that the version-specific framing is gone ("Patch each edge bundle's process import banner for Lambda@Edge compatibility"). Might be worth a one-word pointer back to `framework_runtime.ts`'s doc comment for the "why", since that's where the real reasoning now lives.

packages/hosting/src/framework_runtime.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Single source of truth for the AWS-managed Node.js runtime that executes the
+ * SSR/server bundles the framework adapters emit (Next.js/OpenNext, Nitro,
+ * Astro, SvelteKit). Adapters must reference this constant instead of writing a
+ * `nodejs*.x` literal into the manifest, so a runtime bump is a one-line change.
+ *
+ * Scope: REGIONAL framework compute only. Lambda@Edge compute uses
+ * {@link FRAMEWORK_EDGE_COMPUTE_RUNTIME}, which is tracked separately because
+ * Lambda@Edge can only associate a subset of Lambda's managed runtimes.
+ */
+export const FRAMEWORK_COMPUTE_RUNTIME = 'nodejs24.x';
```
soberm, line 19: Both constants are hardcoded to `nodejs24.x` right now. Given the doc comment says to bump only after re-validating both separately, might be worth a one-liner noting they're currently equal by coincidence, not by design, otherwise a future contributor might assume they always track together and collapse the indirection.

.github/workflows/publish-tag-swift-release.yml
```diff
@@ -20,16 +20,27 @@ jobs:
 
       - name: Extract version from branch name
         id: version
+        env:
+          BRANCH: ${{ github.event.pull_request.head.ref }}
         run: |
-          BRANCH="${{ github.event.pull_request.head.ref }}"
           VERSION="${BRANCH#release/swift-}"
+
+          if ! printf '%s' "$VERSION" | grep -Eq '^[0-9]+\.[0-9]+\.[0-9]+([-+][0-9A-Za-z.-]+)?$'; then
```
soberm, line 28: This hardening (SemVer validation + newline check on the branch-derived version before tagging) is good practice, but it's unrelated to the nodejs24 runtime fix that's the stated point of this PR (fixes #361). Was this meant to ship separately? Might be worth splitting out for a cleaner audit trail, even though the change itself looks solid.

soberm (commented):
## Summary

This PR bumps the SSR framework compute runtime (Nuxt/Nitro, Astro, SvelteKit, Next.js — both regional and Lambda@Edge) from `nodejs20.x` to `nodejs24.x`, introduces shared constants (`FRAMEWORK_COMPUTE_RUNTIME`, `FRAMEWORK_EDGE_COMPUTE_RUNTIME`) so adapters stop hardcoding literals, and makes `resolveRuntime()` throw `UnsupportedRuntimeError` for unrecognized runtimes instead of silently defaulting to Node 20. It also fixes two hosting-owned handlers pinned to Node 20 and bundles in some unrelated hardening of the Swift release-tagging workflow. Tests are updated accordingly with three new tests covering the resolution behavior.

## Recommendation

Comment / minor changes requested — nothing here blocks the merge. The core logic change is correct, well-tested, and the reasoning about Lambda@Edge Node 24 support and the OpenNext patch being version-agnostic is sound and well documented. Two things worth a reply from the author before merge: whether the CI workflow hardening should be split into its own PR, and whether `nodejs18.x` should still silently resolve given the "fail loud" theme here. Everything else is optional polish.

.github/workflows/publish-tag-swift-release.yml
sarayev: Update: rebased this out. The branch was accidentally based on the Swift script-injection fix branch (#340) rather than `main`; I've rebased the #361 commits directly onto `main`, so that commit is no longer an ancestor and the workflow file no longer appears in this PR's diff. The PR is now scoped solely to `packages/hosting/`.

packages/hosting/src/constructs/compute_construct.ts
sarayev: Deliberately kept for now: `resolveRuntime()` still accepts explicit `nodejs18.x`/`nodejs20.x` from a user-supplied manifest so existing/deployed functions on those runtimes don't hard-break at synth — this PR only moves the framework-emitted defaults to Node 24. Tightening the accepted set to 20/22/24 (and eventually 22/24) is a reasonable next step, but it's a behavior change for existing users, so I'd rather handle it as a separate follow-up than fold it in here. Filed as #378.

packages/hosting/src/constructs/compute_construct.ts
sarayev: Checked — the package's `HostingError` codes are consistently suffixed with `Error` (`WafRegionError`, `UnsupportedComputeTypeError`, `UnsupportedNitroPresetError`, `MissingBuildIdError`, …), so `UnsupportedRuntimeError` already matches the convention. Left as-is.

packages/hosting/src/adapters/nextjs.ts
sarayev: Good point — updated. `resolveRuntime` now takes the compute resource name and the thrown error reads e.g. `Compute resource 'serverFn' declares an unsupported runtime 'python3.12'.` Since every adapter funnels through `ComputeConstruct`, this gives resource-level context everywhere without per-adapter wrapping. Added a test assertion guarding the resource name in the message.

packages/hosting/src/adapters/nextjs.ts
sarayev: Updated — the comment now points to `framework_runtime.ts` for the full rationale and notes the patch is runtime-independent (ES module namespace exports are non-writable per spec, not a Node-version quirk).

packages/hosting/src/framework_runtime.ts
sarayev: Added — the doc block now notes the two constants are equal by coincidence, not by design (they're governed by different support tables and must be bumped independently), with an explicit 'do not collapse them into a single constant.'

soberm (approved):

Simone319 (approved):

PR #376, opened 2026-08-14
fix(dart): Lower `build_runner`/`mockito` floors for Flutter 3.41

cadivus:
Fixes the package version problem shown by https://github.com/aws-devtools-labs/aws-blocks/pull/375

### DIFF
--- a/native/dart/packages/blocks_codegen/pubspec.yaml
+++ b/native/dart/packages/blocks_codegen/pubspec.yaml
@@ -14,6 +14,6 @@ dependencies:
   blocks_runtime: ^0.1.2
 
 dev_dependencies:
-  build_runner: ^2.16.0
+  build_runner: ^2.14.0
   lints: ^6.0.0
-  test: ^1.31.0
+  test: ^1.25.0
--- a/native/dart/packages/blocks_runtime_flutter/pubspec.yaml
+++ b/native/dart/packages/blocks_runtime_flutter/pubspec.yaml
@@ -18,8 +18,8 @@ dependencies:
   app_links: ^7.0.0
 
 dev_dependencies:
-  build_runner: ^2.16.0
+  build_runner: ^2.14.0
   flutter_test:
     sdk: flutter
   lints: ^6.0.0
-  mockito: ^5.8.1
+  mockito: ^5.6.0


VarshithaPamisetty (approved):

native/dart/packages/blocks_codegen/pubspec.yaml
```diff
@@ -14,6 +14,6 @@ dependencies:
   blocks_runtime: ^0.1.2
 
 dev_dependencies:
-  build_runner: ^2.16.0
+  build_runner: ^2.14.0
```
sharonyajain, line 17: [Non-blocking] Lowering the caret floors makes `pub get` RESOLVE on the SDK floor, but nothing in this PR demonstrates the codegen (build_runner) and test suite actually COMPILE/RUN against the lower tool versions (build_runner 2.14.0, test 1.25.0, mockito 5.6.0). Because the constraints stay caret (`^`), the normal 'stable' CI job still resolves to the newest versions, so a green stable job proves nothing about the floor. PR #375 explicitly added the SDK-floor job and stated 'the failure is expected at this point'.

```yaml
-  build_runner: ^2.16.0
+  build_runner: ^2.14.0
-  test: ^1.31.0
+  test: ^1.25.0
```

**Why it matters:** cause: a lowered floor only widens the allowed range -> mechanism: if any codegen builder or mock uses an API introduced after 2.14.0 / 1.25.0 / 5.6.0, the package resolves but fails to build/test on the minimum SDK, while stable CI stays green and hides it -> consequence: the '#375 floor failure' this PR claims to fix would still be red (or would break for a contributor pinned to the minimum SDK), i.e. the fix would not actually fix.

**Suggestion:** Confirm PR #375's SDK-floor CI job is green on head cd0f8af before merge. If any lowered dev-dep is consumed via a newer-only API, raise that specific floor to the lowest version that BOTH resolves on the SDK floor AND compiles, rather than the round number.

sharonyajain (approved):

native/dart/packages/blocks_codegen/pubspec.yaml
cadivus: This is wanted here. Otherwise, we are incompatible with the floor version.

Issue #355, opened 2026-08-13
RawRoute registry is module-local — duplicate @aws-blocks/core copies silently 404 every Building Block route


cabcookie:
### Before opening, please confirm
- [x] I have searched for duplicate or closed issues.
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Which package(s) are affected?
`@aws-blocks/core`, `Building Block: auth-oidc`

---

### Describe the bug

`packages/core`'s `dist/raw-route.js` keeps its route table in a **module-local** array:

```js
const routeRegistry = [];
```

`matchRoute()` iterates that array, and `new RawRoute(...)` pushes into it. Both live in the same
module instance — so the table is per *module copy*, not per process.

If a dependency tree ends up with more than one physical copy of `@aws-blocks/core`, the Lambda
bundle contains more than one `raw-route` module. Building Blocks that mount their routes through
*their* copy then register into a table that the dispatcher — which reads *its* copy — never looks
at. Every such route returns `404 {"error":"Not Found"}`.

The failure is completely silent. The dispatcher's 404 branch logs nothing, so CloudWatch shows
only `START` / `END` / `REPORT` with no error in between. Nothing in a template, a synth, or a test
suite sees it: the CDK side is entirely correct — the CloudFront behaviour exists, the API Gateway
route exists, the SSM parameters exist. The break is between build time and run time.

**What makes this specifically a library concern rather than a consumer one:** the same package
already defends against multiple instances elsewhere. Lambda event handlers are stored on
`globalThis.__BLOCKS_LAMBDA_EVENT_HANDLERS__`, so every copy shares one map. The route registry
does not do that. In our deployed bundle we counted 18 occurrences of that `globalThis` key (18
core copies sharing one handler map, working correctly) alongside 2 independent `raw-route`
modules (two registries, one reader). The inconsistency inside one package is the actual defect.

### Expected behavior

`RawRoute` registration and lookup share one registry per process, regardless of how many copies of
`@aws-blocks/core` a bundler includes — the same guarantee `__BLOCKS_LAMBDA_EVENT_HANDLERS__`
already provides for event handlers.

Failing that, a duplicate-instance situation should be **loud** (a warning at registration, or a
404 that logs which registry it searched) rather than an unexplained 404.

### Reproduction steps

1. Create a Blocks app whose dependency tree resolves `@aws-blocks/core` to more than one physical
[...]

PR #377, opened 2026-08-15
fix(core): share the RawRoute registry across duplicate core copies

cabcookie:
Fixes the silent 404 reported in #355.

## The defect

`raw-route.ts` kept its route table in module-local state:

```ts
const routeRegistry: RegisteredRoute[] = [];
let registrationLocked = false;
```

That makes the table per *module copy*, not per process. When a dependency tree resolves `@aws-blocks/core` to more than one physical copy, the Lambda bundle carries more than one `raw-route` module. Building Blocks register their routes through *their* copy; the dispatcher reads *its own*; every route registered through the "wrong" copy answers `404 {"error":"Not Found"}`.

Two things made it hard to find, and both are addressed here:

- **It was completely silent.** The dispatcher's 404 branch logged nothing, so CloudWatch showed only `START` / `END` / `REPORT`.
- **Synth had the same split.** `hosting.ts` builds CloudFront behaviors from `getRegisteredRoutes()`, so an affected route could be missing its behavior too — a 404 before the request ever reaches Lambda.

As #355 notes, this package already defends against exactly this elsewhere: Lambda event handlers live on `globalThis.__BLOCKS_LAMBDA_EVENT_HANDLERS__` and are shared across copies. The route registry was the inconsistency.

## The change

- Registry state (`routes` + `locked` + a copy counter) moves onto a well-known `globalThis` key, `__AWS_BLOCKS_RAW_ROUTE_REGISTRY_V1__`. `locked` travels with `routes` deliberately — splitting them would let one copy lock the shared table while another is still registering.
- A plain string key rather than `Symbol.for()`, on purpose: duplicate copies are diagnosed by counting strings in a deployed bundle, and a symbol is invisible to that. It also matches the existing `__BLOCKS_LAMBDA_EVENT_HANDLERS__` precedent.
- Copies that take part in routing are counted, and a second one warns once.
- An unmatched RawRoute now logs method, path and the size of the table the dispatcher consulted — a `0 routes registered` line is the fingerprint of a split registry. The path is already in the API Gateway access logs, so this adds no new category of data.

## Behavior change (deliberate, please review)

Duplicate-route detection now reaches **across** copies. An app with two core copies *and* two instances of the same Building Block previously split its routes silently — half of them 404ing; it now throws at startup. Such an app was already broken, and this only makes it say so, but it is a visible change and worth a maintainer's judgement. It is called out in the changeset and pinned by a test.

## Tests

The interesting case is not expressible with mocks, so the tests install two genuinely separate copies of the package — two directory trees, each with its `package.json` beside `dist/`, the shape npm produces for a nested `node_modules/@aws-blocks/core` — and replay the production sequence: a Building Block registers through one copy, the dispatcher looks up through the other.

- **Unbundled** — cross-copy registration and dispatch, the shared lock, and the cross-copy duplicate throw.
- **Bundled** — the same probe sent through esbuild with `--conditions=aws-runtime` and `--minify`, the way `NodejsFunction` builds a handler, then executed. This pins that a bundler keeps two copies rather than collapsing them (i.e. that the failure mode reaches the deployed artifact at all), that reaching the registry through `globalThis` survives minification, and that the key stays a greppable literal.
- **Synth** — a route registered by another copy must still produce its CloudFront behavior.
- **Diagnostics** — the 404 branch logs, a matched route does not, and the copy count appears when more than one copy is loaded.

Each of these fails against the previous implementation. Against `main`'s `raw-route.ts`, the two-copy probe reports `0` routes and no match — the reported 404, reproduced.

One note in case it saves a reviewer time: an early draft justified counting copies on first use (rather than at module load) by tree-shaking, on the grounds that `sideEffects` marks this module side-effect-free. That was measured and does not hold — esbuild keeps top-level statements in a module whose exports are used, `--minify` included. The choice stands on its other reason: it counts the copies that actually take part in routing. The comment in `raw-route.ts` says so explicitly, so the claim does not get re-derived later.

## Checks

- `npm run build` ✅
- `npm run lint:deps` ✅ (315 source files)
- `biome lint` on the changed files ✅
- `packages/core` unit tests ✅ — 684/687, and 228/228 across the three files this PR touches
- `npm run test:e2e:local` ⚠️ — did not run cleanly here, for reasons I could trace to this machine rather than to the change

Two notes on the checks that are not green, both of which I believe are pre-existing. Please tell me if you would rather see either of them fixed before this is considered.

**Unit suite.** Without `--test-timeout`, `dist/telemetry/telemetry.test.js` hangs indefinitely inside the full-suite run (`telemetry/client` and `telemetry/--telemetry-file flag`), so `npm test` in `packages/core` never returns at all. That same file passes in **2.4 s / 81 tests** on its own, and in a small group. With `--test-timeout=30000` the suite completes in **44 s** at 684/687. At full parallelism other heavy files intermittently time out too, varying between runs.

**e2e.** Two tests in `test-apps/comprehensive/test/cli-client.test.ts` fail here — `default export fails when CWD has no config.json` and `generateClient with incorrect URL fails` — both after exactly ~12.1 s with an empty `stderr`. `runCli()` gives the subprocess `execFileSync(..., { timeout: 12_000 })`; on timeout `e.status` is null, so `exitCode` falls back to 1 and the first assertion passes, while `e.stderr` is empty and the second fails with `got: ''`. Measured directly on this machine, with this change reverted, that subprocess takes **16–40 s** to start — `npx tsx` startup alone — so the 12 s budget cannot be met here regardless of this PR. The client itself behaves correctly (`fetch failed` is produced, just too late).

Separately: a full `npm run test:e2e:local` leaves the working tree modified — `test:vendorize` adds `vendor/bb-kv-store` to `test-apps/comprehensive/package.json` and to `package-lock.json`, and leaves `vendor/` plus `cdk.out.*` behind, none of which are gitignored.

Happy to open a separate issue for any of these; they are independent of #355 but each costs a contributor time.


packages/core/src/raw-route.ts
```diff
@@ -105,19 +105,102 @@ export interface RegisteredRoute {
...
+/**
+ * Well-known `globalThis` key holding the shared registry state.
+ *
+ * A plain string rather than a `Symbol.for()` key, deliberately: duplicate
+ * copies are diagnosed by counting strings in the deployed bundle, and a
+ * symbol would be invisible to that. It also matches the existing
+ * `__BLOCKS_LAMBDA_EVENT_HANDLERS__` precedent. The `_V1_` suffix lets a
+ * future incompatible shape claim its own key rather than corrupt this one.
+ */
+const REGISTRY_KEY = '__AWS_BLOCKS_RAW_ROUTE_REGISTRY_V1__';
```
Simone319, line 152: nit: Duplicate copies in the wild are usually *different versions*, so a future `_V2_` bump would silently re-split the registry across a mixed tree, with the copy counter reporting `copies: 1` (false negative) — reintroducing this exact bug. Document the key as a permanent contract: evolve the shape additively; any V2 must read/adopt V1's state.

Simone319 (dismissed):

packages/core/src/raw-route.ts
cabcookie: Fixed in a38d437.

The mechanism checks out: two keys means two `copies` counters, so each side increments its own from 0 to 1, both report `copies: 1`, and the warning never fires while the routes split — the original 404 with its own diagnostic switched off.

The key doc now says so outright: the key is permanent, `_V1_` names the shape rather than a version to bump, the shape may only grow, and a `_V2_` key — if it ever proves unavoidable — has to read the V1 state and keep writing it.

I also made `getState()` hold up the other half of that contract, since documenting additive evolution while the code assumes every field is present is only half the fix. A state left by a peer copy without `routes` would throw on the first registration, and one without `copies` would go NaN on the increment and mute the duplicate warning — the same silent-diagnostic failure by a different route. Both are now filled in rather than trusted. `locked` needs no default: absent reads as unlocked, which is what a registry still accepting registrations means.

Two tests cover it (`registry state from a peer copy is adopted, not trusted`); both fail without the guard. `packages/core`: 230/230 across raw-route, hosting and lambda-handler.

Simone319 (dismissed):

soberm (dismissed):

soberm (approved):

Simone319 (approved):

PR #380, opened 2026-08-17
[core] Retarget config registry to shared role + per-stack compute registry

Simone319:
## Problem

The config registry granted `s3:GetObject` on `blocks-config.json` to a single hardcoded handler function and stamped `BLOCKS_CONFIG_BUCKET` / `BLOCKS_CONFIG_KEY` on that one function. That couples config delivery to one Lambda and doesn't generalize to the multi-compute model, where every compute must be able to read the shared config object.

This is Issue **B1** of Multi-Compute Phase 1. It stacks on **A3** (it consumes the `computes` list and default compute introduced there; its self-registration lives in the A2 `Compute` base).

## Changes

- **Retarget `finalizeConfigRegistry`** — signature is now `finalizeConfigRegistry(root, executionRole, computes)`:
  - grants `s3:GetObject` **once to the shared execution role** (so every compute that assumes the role can read the object), instead of to one function's role;
  - stamps `BLOCKS_CONFIG_BUCKET` / `BLOCKS_CONFIG_KEY` on **every compute** via `compute.setEnv(...)`.
  - The previous `ConfigFinalizeRoot` wrapper interface is removed in favor of three explicit parameters.
- **Per-stack compute registry** (`packages/core/src/cdk/compute/compute-registry.ts`, new): `registerCompute` / `getComputes`, with state stored on the owning stack keyed by a `Symbol` and resolved via `cdk.Stack.of()` — mirroring the config registry. This keeps each stack's computes isolated in a multi-stack synth (Amplify Gen2 nested stacks, Pipelines stages); a process-global list would let one stack's finalize see another's computes.
- **`Compute` base self-registers** — `registerCompute(this)` in its constructor, so every compute joins its stack's registry the moment it's constructed (no separate discovery pass).
- **`BlocksStack.computes` / `BlocksBackend.computes`** now read from the registry (`getComputes(this)`); `create()` passes `(root, root.executionRole, root.computes)` to finalize.
- Updated the two CDK test callers (`bb-auth-cognito`, `bb-knowledge-base`) to the new signature.

Behavior-preserving for the default single-compute app: the same config object is written to S3, the same two coordinates reach the runtime, and the same `s3:GetObject` permission is available — now via the shared role.

## Validation

- `npm run build` ✓
- Unit tests: `@aws-blocks/core` 656/656 (incl. a new multi-stack registry-isolation test), `bb-auth-cognito` 241/241, `bb-knowledge-base` 129/129 ✓
- `npm run lint` (zero blocking errors), `npm run lint:deps`, `npm run check:api` ✓
- `npm run test:e2e:local` 8/8 ✓

## Checklist

- [x] PR description included

### DIFF
--- a/.changeset/core-config-registry-shared-role.md
+++ b/.changeset/core-config-registry-shared-role.md
@@ -1,0 +1,26 @@
+---
+"@aws-blocks/core": patch
+---
+
+refactor(core): retarget the config registry to the shared role and every compute
+
+`finalizeConfigRegistry` no longer takes a single handler function. It now takes
+the owning construct plus the shared execution role and the stack's computes
+(`finalizeConfigRegistry(root, executionRole, computes)`) and:
+
+- grants `s3:GetObject` on the config object **once to the shared execution
+  role** instead of to one function's role, so every compute that assumes the
+  role can read it; and
+- stamps `BLOCKS_CONFIG_BUCKET` / `BLOCKS_CONFIG_KEY` on **every compute** via
+  `compute.setEnv(...)` rather than on a single hardcoded handler.
+
+Adds a per-stack compute registry (`registerCompute` / `getComputes`): computes
+self-register on their owning stack in the `Compute` base constructor (state
+keyed on the stack, resolved via `cdk.Stack.of()`, like the config registry), so
+a multi-stack synth keeps each stack's computes isolated. `BlocksStack.computes`
+/ `BlocksBackend.computes` read from it.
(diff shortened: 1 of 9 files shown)


Simone319:
Discarding and recreating against `main` (then re-linking into stack #342) so PR Checks/E2E run — mirroring how A3 (#341) is set up. Same head branch `zimzha/multi-compute-b1-config-registry`, no code change.

PR #385, opened 2026-08-18
feat(core): logRetention + throttling + accessLogging stack defaults (Infrastructure Options)

osama-rizk:
## Summary

Extends the stack-wide `BlocksDefaults` mechanism (#302) with three additive fields and adopts them across Blocks-managed infrastructure. Each field is read independently via `option ?? scope.defaults.field` — a per-block option always wins, and no field is derived from another (a #302 reviewer blocker).

New `BlocksDefaults` fields (and both `BlocksPresets`):

| Field | Type | sandbox | production |
|---|---|---|---|
| `logRetention` | `RetentionDays` | `ONE_WEEK` | `ONE_YEAR` |
| `throttling` | `{ rateLimit; burstLimit }` | `200 / 400` | `200 / 400` |
| `accessLogging` | `boolean` | `false` | `false` (opt-in) |

Also newly exported from `@aws-blocks/core/cdk`: the `BlocksThrottling` type and `ensureApiGatewayAccount()`; `Scope` gains a `handlerLogGroup` getter.

## 🔎 For API-BR review

Please sanity-check the three property **names/shapes** before this is widely adopted:
- `logRetention: RetentionDays` (CDK enum — appears only in the `/cdk` type)
- `throttling: { rateLimit: number; burstLimit: number }` (new `BlocksThrottling` type)
- `accessLogging: boolean`

## Adoption sites

**Log retention** — Blocks-managed log groups now follow `defaults.logRetention` instead of AWS's infinite default:
- Core shared handler Lambda — `setupBlocksInfra` now owns a single log group, exposed as `scope.handlerLogGroup`.
- `bb-logger` — reconfigures that shared group's retention (`retention ?? defaults.logRetention`) instead of creating a colliding `/aws/lambda/<fn>` group.
- `bb-distributed-table` — GSI-manager Lambdas.
- `bb-distributed-data` — DSQL migration Lambda.
- `bb-data` — Aurora migration Lambda.
- `bb-app-setting` — secret-init custom-resource Lambda.

  _(The framework `custom-resources.Provider` Lambdas these BBs wrap still use AWS's default retention — the L2 `Provider` exposes no log-group override.)_

**Throttling** — applied to the core REST API stage and the `bb-realtime` WebSocket stage (on a WS stage the throttle unit is **messages/sec** across the connection; noted in `bb-realtime/DESIGN.md`).

**Access logging** — when enabled, structured JSON access logs to a dedicated CloudWatch log group (retention = `defaults.logRetention`, `RemovalPolicy.DESTROY`) on the core REST API and `bb-realtime` WebSocket stages. The account-level API Gateway CloudWatch Logs role is provisioned **once per stack** via a shared `ensureApiGatewayAccount()` helper (`Symbol.for`-keyed so co-located stages reuse one `AWS::ApiGateway::Account`); the RestApi's auto `cloudWatchRole` is disabled to avoid a duplicate. Each access-logging stage takes an explicit dependency on the account so a clean-account first deploy applies the setting before the stage is created.

## #163 / #326 reconciliation

Both #163 and #326 add throttling to the core REST API directly (near-duplicates). This PR is the single source of truth for API Gateway throttling and **supersedes** them.

## Deferred to a follow-up PR

- **Hosting adoption** (SSR REST API throttling/access logging + compute/revalidation log retention, and the account-role wiring for hosting) is intentionally **out of scope here** and will land in a separate PR.
- Out of scope per the original plan: `secretsStore`, `pointInTimeRecovery`, VPC, compute, sandbox template cleanup.

## Testing

- `npm run build && npm run lint && npm run lint:deps && npm test` all green (3709 pass / 0 fail).
- Synth tests (`aws-cdk-lib/assertions`) cover: preset values; sandbox vs production retention on the handler group; 200/400 throttle on the REST and WS stages; per-block/per-stack overrides winning; access-log group + single `AWS::ApiGateway::Account` present when enabled and absent when disabled; GSI log-group retention.

## Notes for reviewers

- `check:api` produces no diff — api-extractor tracks the mock/types entry, not the `/cdk` entry where these types live.

## Sandbox deploy verification (done)

Deployed `test-apps/comprehensive` to a sandbox (account 444209914584, us-west-2) with `accessLogging` enabled, exercising the core REST API, `bb-realtime` WebSocket, `bb-data`, `bb-distributed-data`/`-table`, and `bb-app-setting`. Verified end-to-end, then torn down (`DELETE_COMPLETE`):

- ✅ Deploy succeeded **with access logging on** — no `CreateStage` failure. Our `BlocksApiGatewayAccount` + CloudWatch role were created and `apigateway get-account` pointed at our stack's role (explicit stage→account dependency ordered it correctly).
- ✅ **Access logs land** — 3 live `POST /aws-blocks/api` requests (HTTP 200) produced structured JSON entries in the `ApiAccessLogs` CloudWatch group (`requestId`/`httpMethod`/`resourcePath`/`status`…).
- ✅ REST stage: `accessLogSettings` → JSON format + `ApiAccessLogs` group; throttle **200/400** confirmed live.
- ✅ WebSocket stage: `AccessLogSettings` → `BlocksRtAccessLogs` group; throttle **200/400**.
- ✅ All **6 log-retention groups** at **7 days (ONE_WEEK, sandbox)**: handler, GSI-manager, GSI-isComplete, DSQL-migration, bb-data-migration, bb-app-setting-secret-init.

Caveat observed: the test account already had an API Gateway CloudWatch role (from a prior hosting deploy), so this wasn't a truly *clean* account — our stack provisioned and re-pointed the account role (the documented region-global "last-deploy-wins" behavior), but the clean-account `CreateStage` failure itself couldn't be reproduced here.


packages/bb-logger/src/index.cdk.ts
```diff
@@ -32,13 +33,14 @@ export class Logger extends Scope {
...
-			new logs.LogGroup(this, 'LogGroup', {
-				logGroupName: `/aws/lambda/${this.handler.functionName}`,
-				retention: options.retention,
-				removalPolicy: RemovalPolicy.DESTROY,
-			});
+		// Override retention on the shared handler log group when this Logger
+		// asks for a specific value; otherwise leave the stack-wide default in
+		// place. Applied via the L1 escape hatch because a per-block option must
+		// reconfigure the framework-owned group, not spawn a competing one.
+		const retention = options?.retention ?? this.defaults.logRetention;
```
sarayev: **[major]** Every `Logger` reconfigures the *same* shared group, so this unconditional write makes the last-constructed Logger win. A bare `new Logger(scope, 'x')` computes `defaults.logRetention` and overwrites a `retention` that another Logger set earlier — silently and order-dependent. Could we write only when the caller explicitly asked, so a no-op Logger doesn't stomp an explicit override?

```ts
if (options?.retention) {
    const cfnLogGroup = this.handlerLogGroup.node.defaultChild as CfnLogGroup | undefined;
    if (cfnLogGroup) cfnLogGroup.retentionInDays = options.retention;
}
```

The shared group already carries `defaults.logRetention` from `setupBlocksInfra`, so writing the default back is redundant anyway.

packages/bb-logger/src/index.cdk.ts
```diff
@@ -32,13 +33,14 @@ export class Logger extends Scope {
...
+		const cfnLogGroup = this.handlerLogGroup.node.defaultChild as CfnLogGroup | undefined;
```
sarayev: **[minor]** If `defaultChild` is ever undefined (e.g. the group resolves to an imported `ILogGroup`), the retention request is silently dropped and you'd end up with unexpectedly-infinite retention and no signal. It's always a concrete `LogGroup` today, so this is defensive — but would an `else` that throws at synth be safer than a silent skip?

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -40,10 +87,16 @@ export const BlocksPresets = {
...
+		logRetention: RetentionDays.ONE_WEEK,
+		throttling: DEFAULT_THROTTLING,
+		accessLogging: false,
 	},
 	/** Durable, protected posture for permanent deployments. */
 	production: {
 		removalPolicy: RemovalPolicy.RETAIN,
 		deletionProtection: true,
+		logRetention: RetentionDays.ONE_YEAR,
+		throttling: DEFAULT_THROTTLING,
```
sarayev: **[major]** This applies a 200 req/s (400 burst) cap to **production** by default. Before this PR the REST stage had no stage-level throttle, so existing deployments ran at the account default (~10k rps). On upgrade, anything serving >200 rps would start getting 429s with no change on their side. Is 200 intentional for production, or should prod get a higher ceiling? Either way, since it can throttle live traffic, could we call this out as a behavior change in the changeset (it currently reads as a routine `minor`)?

packages/core/src/cdk/apigateway-account.ts
```diff
@@ -0,0 +1,55 @@
...
+	if (existing) return existing;
+
+	const role = new Role(stack, 'BlocksApiGatewayCloudWatchRole', {
+		assumedBy: new ServicePrincipal('apigateway.amazonaws.com'),
+		managedPolicies: [
+			ManagedPolicy.fromAwsManagedPolicyName('service-role/AmazonAPIGatewayPushToCloudWatchLogs'),
+		],
+	});
+
+	const account = new CfnAccount(stack, 'BlocksApiGatewayAccount', {
```
sarayev, line 60: **[major]** `AWS::ApiGateway::Account` is a region singleton, but this emits one per stack. The comment notes "last deploy wins", but the teardown case is sharper: if two Blocks stacks in the same account+region both enable access logging and the one that deployed last is later destroyed, its `BlocksApiGatewayCloudWatchRole` is deleted and the account setting can end up pointing at a role that no longer exists — silently breaking logging (or blocking `CreateStage`) for the surviving stack. Should the account be importable/shared (`fromExisting`), or at least documented as a single-Blocks-stack-per-region assumption?

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -88,11 +90,22 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
     ],
   });
 
+  // The single CloudWatch log group for the shared handler. Owning it here (a
+  // real LogGroup construct passed as the function's `logGroup`) means its
+  // retention follows the stack-wide default instead of AWS's infinite default,
+  // and gives bb-logger one group to reconfigure rather than a second, colliding
+  // one. Torn down with the stack (logs are not durable state).
+  const handlerLogGroup = new logs.LogGroup(scope, 'HandlerLogGroup', {
```
sarayev: **[minor]** On an *existing* deployment the handler previously logged to Lambda's auto-created `/aws/lambda/<fn>` group (infinite retention). Pointing the function at this new CDK group leaves the old one orphaned in CloudWatch — unmanaged and still infinite, i.e. the exact cost leak this PR targets. Worth a migration note in the changeset? (The old `bb-logger` group also gets DESTROY'd on upgrade, dropping its history.)

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -126,11 +139,49 @@ export function setupBlocksInfra(scope: Construct, props: BlocksBackendProps, id
...
+  // Structured JSON access logging on the shared stage, when the stack-wide
+  // default enables it. Requires the account-level CloudWatch Logs role (see
+  // ensureApiGatewayAccount) — provisioned once here rather than per adopter.
+  let accessLogGroup: logs.LogGroup | undefined;
+  let apiGatewayAccount: apigateway.CfnAccount | undefined;
+  if (props.defaults.accessLogging) {
+    apiGatewayAccount = ensureApiGatewayAccount(cdk.Stack.of(scope));
+    accessLogGroup = new logs.LogGroup(scope, 'ApiAccessLogs', {
+      retention: props.defaults.logRetention,
+      removalPolicy: cdk.RemovalPolicy.DESTROY,
```
sarayev: **[minor]** In production these access logs are the audit trail, but the group is `RemovalPolicy.DESTROY`, so they vanish on stack teardown. Consistent with treating logs as disposable — but is that the intent for a production audit log? Might warrant `RETAIN` for the access-log group specifically.

packages/bb-data/src/infra.ts
```diff
@@ -43,6 +44,12 @@ export interface AuroraInfraConfig {
   deletionProtection?: boolean;
   /** Aurora PostgreSQL engine version, e.g. `'16.13'`. @default '16.13' */
   postgresVersion?: string;
+  /**
+   * CloudWatch retention for the migration Lambda's log group. Populated from
+   * the stack-wide `defaults.logRetention`; when omitted the log group uses the
+   * CDK `LogGroup` default retention.
+   */
+  logRetention?: cdk.aws_logs.RetentionDays;
```
sarayev, line 52: **[nit]** `materialize` is exported, and when `logRetention` is omitted the group falls back to CDK's default (`TWO_YEARS`), not the Blocks default. `Database` always passes it so this is fine today — just flagging the escape-hatch path for a direct caller.

packages/bb-distributed-data/src/index.cdk.ts
```diff
@@ -79,6 +80,12 @@ export class DistributedDatabase extends Scope {
       handler: 'handler',
       runtime: DEFAULT_NODE_RUNTIME,
       timeout: cdk.Duration.minutes(MIGRATION_LAMBDA_TIMEOUT_MINUTES),
+      // Own the migration Lambda's log group so its retention follows the
+      // stack-wide default instead of AWS's infinite retention.
+      logGroup: new LogGroup(stack, `${this.fullId}DsqlMigrationLogs`, {
```
sarayev, line 87: **[minor]** Every other adopter got a CDK retention test (bb-data, bb-distributed-table, bb-app-setting, bb-logger, realtime, core) — this DSQL migration group is the one change without one. A one-liner mirroring the bb-data test would close the parity gap:

```ts
template.hasResourceProperties('AWS::Logs::LogGroup', { RetentionInDays: 7 });
```

sarayev (commented):
Reviewed as a senior engineer — focused on correctness, edge cases, security, scaling, and tests (skipping pure style). Overall a clean, well-documented change with good synth coverage. Inline notes below: 3 majors worth resolving before merge (Logger retention clobber, the production throttle default as a silent behavior change, and the per-stack `ApiGateway::Account` singleton risk), plus a few minors/nits. None are hard blockers on their own — flagging for discussion.

packages/bb-logger/src/index.cdk.ts
osama-rizk: Good catch — fixed in 784657c8. Logger now writes retention **only when `options.retention` is explicitly set**; a bare `new Logger(...)` no longer writes `defaults.logRetention` back, so it can't clobber an explicit value from another Logger (and the redundant default write is gone). Added a test that pins this: an explicit `retention: 14` survives a later bare Logger.

packages/bb-logger/src/index.cdk.ts
osama-rizk: Fixed in 784657c8. When `retention` is explicitly requested but the shared group's `defaultChild` isn't a concrete `LogGroup` (e.g. an imported `ILogGroup`), it now throws an actionable synth error rather than silently dropping the request. (In the no-retention path there's nothing to write, so no throw.)

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Addressed in 784657c8. Production now defaults to **1000/2000** (sandbox stays 200/400) so the new stage cap doesn't 429 real prod traffic, and it's overridable per-stack via `throttling`. I also called this out explicitly as a **behavior change** in the changeset (before: no stage throttle → account default ~10k rps) with upgrade guidance.

packages/core/src/cdk/apigateway-account.ts
osama-rizk: Documented in 784657c8. `ensureApiGatewayAccount()` now spells out the region-singleton teardown hazard you describe (later stack's role deleted → account setting points at a dead role → breaks logging / blocks `CreateStage` for the survivor) and states the **single-Blocks-stack-per-region** assumption for `accessLogging`. A shared/importable account for true multi-stack support is deferred; noted that multi-stack users should manage the account role out-of-band. Also flagged in the changeset.

packages/core/src/cdk/blocks-backend.ts
osama-rizk: Good point — added an upgrade/migration note to the changeset in 784657c8: on upgrade the handler's old auto-created `/aws/lambda/<fn>` group (and any prior `bb-logger` retention group) is left orphaned/unmanaged in CloudWatch and should be deleted manually to reclaim the cost/close the infinite-retention leak.

packages/core/src/cdk/blocks-backend.ts
osama-rizk: Changed in 784657c8. The access-log groups (core REST + realtime WS) now use `removalPolicy: defaults.removalPolicy`, so **production RETAINs** the audit trail on teardown while sandbox still DESTROYs. The handler's operational stdout group stays DESTROY (not an audit trail). Reflected in the changeset.

packages/bb-data/src/infra.ts
osama-rizk: Acknowledged — `Database` always passes `logRetention` from `defaults`, so the escape-hatch (`TWO_YEARS`) path is only reachable by a direct `materialize` caller. The option's JSDoc already notes it falls back to the CDK `LogGroup` default when omitted, so a direct caller is on notice. Left as-is.

packages/bb-distributed-data/src/index.cdk.ts
osama-rizk: Added in 784657c8 — a parity test in `index.cdk.test.ts` asserting the DSQL migration Lambda's log group adopts `defaults.logRetention` (`RetentionInDays: 7` under sandbox), matching the other adopters.

osama-rizk:
Thanks @sarayev — addressed the review in `784657c8`:

**Majors**
- **Logger clobber** → writes retention only when an explicit per-Logger `retention` is set (bare Logger no longer stomps another's value); added a regression test.
- **Prod throttle** → production raised to **1000/2000** (sandbox stays 200/400) so upgrades don't 429 real traffic; overridable, and called out as a behavior change in the changeset.
- **`ApiGateway::Account` singleton** → documented the teardown hazard + single-Blocks-stack-per-region assumption (shared/imported account deferred).

**Minors/nits**
- Logger now throws (not silently skips) if the shared group isn't a concrete `LogGroup`.
- Access-log groups follow `defaults.removalPolicy` → production **RETAIN**s the audit trail.
- Orphaned legacy `/aws/lambda/<fn>` handler group documented as an upgrade note.
- Added the missing DSQL migration-Lambda retention test (parity).
- `materialize` escape-hatch default acknowledged (Database always passes `logRetention`).

`build` + `lint` + `lint:deps` + affected unit tests green.

soberm (changes requested):
This adds `logRetention`, `throttling`, and `accessLogging` to the stack-wide defaults and wires them through the core handler, the REST/WebSocket stages, and the migration/GSI Lambdas, plus a new `ensureApiGatewayAccount()` singleton for the account-level CloudWatch role. The mechanics check out — went and verified a couple of the less obvious CDK details (the v1 `AccessLogFormat` reuse on the v2 WebSocket stage, `LogGroup`'s default retention when omitted) against the docs, and they're both fine. The sandbox deploy verification in the description is a nice touch too.

Requesting changes on one point: defaulting `accessLogging` to `true` in production means this starts mutating an account/region-level singleton (the API Gateway CloudWatch role) by default on upgrade. `ensureApiGatewayAccount`'s own doc comment lays out the failure mode if a second stack in the same account+region also turns this on — repointed role on deploy, orphaned/broken role on teardown. That's a big enough blast radius that I'd want an explicit yes on it from whoever's doing the API-BR pass mentioned in the description, rather than have it ride through on the docs alone. Everything else below is non-blocking.

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -64,12 +114,23 @@ export const BlocksPresets = {
...
+	 * higher-volume APIs.
+	 */
 	production: {
 		removalPolicy: RemovalPolicy.RETAIN,
 		deletionProtection: true,
 		allowedOrigins: [],
 		pointInTimeRecovery: true,
+		logRetention: RetentionDays.ONE_YEAR,
+		throttling: { rateLimit: 1000, burstLimit: 2000 },
+		accessLogging: true,
```
soberm: [blocker] Production defaulting this to `true` means every app that upgrades and redeploys with the production preset starts provisioning or re-pointing the account-wide `AWS::ApiGateway::Account` role. That's a real region/account singleton, not a stack-scoped one, so a second Blocks stack (or anything else in the account using API Gateway CloudWatch logging) can get silently repointed on deploy or orphaned on this stack's teardown. Want to see an explicit sign-off on this before it merges, or `false` as the initial default with an easy opt-in.

packages/core/src/cdk/apigateway-account.ts
```diff
@@ -0,0 +1,65 @@
...
+ *   - on teardown of that later stack, its role is deleted and the account
+ *     setting is left pointing at a now-deleted role — silently breaking access
+ *     logging (or blocking `CreateStage`) for the surviving stack until it
+ *     redeploys.
+ * Enabling `accessLogging` is therefore safe for one Blocks stack per region.
+ * Multi-stack support (a shared/imported account) is deferred; if you run
+ * multiple Blocks stacks per region, manage the account role out-of-band and
+ * leave `accessLogging` to a single owner.
+ */
+export function ensureApiGatewayAccount(stack: cdk.Stack): CfnAccount {
```
soberm, line 47: [question] Is there a test anywhere with one stack that has both a LambdaCompute and a Realtime instance, both with access logging on, checking we still only get one `AWS::ApiGateway::Account`? That's the actual case the `Symbol.for` sharing exists for, and I only see it tested per-package in isolation.

packages/bb-logger/src/index.cdk.ts
```diff
@@ -32,13 +35,22 @@ export class Logger extends Scope {
 			registerConfig(this, 'LOG_LEVEL', options.level);
 		}
 
-		// Optionally create a LogGroup with retention
+		// Override retention on the shared handler log group ONLY when this
+		// Logger explicitly asks for one. Applied via the L1 escape hatch because
+		// a per-block option must reconfigure the framework-owned group, not spawn
+		// a competing one.
 		if (options?.retention) {
```
soberm, line 52: [minor] If two Loggers each pass a different explicit `retention`, whichever gets constructed last wins with no warning, since they share one group. Might be worth a line in the "last wins" comment/test spelling that out, or a synth-time warning.

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -47,6 +61,42 @@ export interface BlocksDefaults {
...
+
+	/**
+	 * How long CloudWatch Logs keeps log events written by Blocks-managed log
+	 * groups (the core handler Lambda, migration/GSI Lambdas, hosting compute,
+	 * `bb-logger`, and API Gateway access logs) before expiring them. Without a
+	 * retention set, AWS keeps log events forever, which grows cost unbounded —
+	 * every Blocks-managed log group reads this default so retention is applied
+	 * consistently.
+	 */
+	logRetention: RetentionDays;
```
soberm, line 73: [minor] These three fields are required, not optional. Fine for anyone spreading a preset, but anyone hand-rolling a `BlocksDefaults` object will hit a compile error after this, in what's tagged a minor bump. Worth a line in the changeset, or make them optional with an internal fallback.

packages/core/src/cdk/apigateway-account.ts
```diff
@@ -0,0 +1,65 @@
...
+		assumedBy: new ServicePrincipal('apigateway.amazonaws.com'),
+		managedPolicies: [
+			ManagedPolicy.fromAwsManagedPolicyName('service-role/AmazonAPIGatewayPushToCloudWatchLogs'),
+		],
+	});
+
+	const account = new CfnAccount(stack, 'BlocksApiGatewayAccount', {
+		cloudWatchRoleArn: role.roleArn,
+	});
+	account.node.addDependency(role);
```
soberm: [nit] This looks redundant — the `cloudWatchRoleArn: role.roleArn` reference already gives CDK an implicit dependency. Not wrong, just extra.

packages/bb-data/src/index.cdk.ts
```diff
@@ -81,6 +81,8 @@ export class Database extends Scope {
       // an override like `{ ...production, deletionProtection: false }` is honored.
       deletionProtection: this.defaults.deletionProtection,
       postgresVersion: options?.postgresVersion,
+      // Migration Lambda log retention follows the stack-wide default.
```
soberm, line 84: [nit] This comment just restates the line below it, could drop it.

soberm:
**Inline notes** (lines not reachable in the diff)

**packages/bb-distributed-table/src/index.cdk.test.ts line 409** [nit] This only checks that *some* log group has `RetentionInDays: 7`, not that both the manager and isComplete Lambda groups do. Would rather see both pinned so a partial regression can't hide behind the other resource.

**packages/core/src/cdk/blocks-backend.test.ts** [nit] `StubLambdaCompute` got the identical `logGroup` addition in both blocks-backend.test.ts and blocks-stack.test.ts. Might be worth one shared stub at some point so they can't drift apart.

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Done in 912aeef7 — production `accessLogging` now defaults to **`false`** (off in both presets), so an upgrade never provisions/re-points the account/region-wide `AWS::ApiGateway::Account` role without an explicit opt-in (`defaults: { ...BlocksPresets.production, accessLogging: true }`). Presets, tests, DESIGN docs, changeset, and the PR description are updated.

packages/core/src/cdk/apigateway-account.ts
osama-rizk: Added a test in `bb-lambda-compute/src/index.cdk.test.ts` (912aeef7): two access-logging computes in one stack synth to exactly **one** `AWS::ApiGateway::Account` (two `ApiGateway::Stage`s). That exercises the same `ensureApiGatewayAccount` `Symbol.for` path the compute + realtime share (realtime can't import bb-lambda-compute without a dep cycle, but both call the identical helper).

packages/bb-logger/src/index.cdk.ts
osama-rizk: Addressed in 912aeef7 — when two Loggers set *different* explicit `retention` on the shared handler group, the last still wins but now emits a synth warning (`Annotations.addWarningV2`) naming both values, and the class JSDoc says so. Same-value and bare-Logger cases stay silent; added tests for both.

packages/core/src/cdk/blocks-defaults.ts
osama-rizk: Noted in the changeset (912aeef7): the three new `BlocksDefaults` fields are required. Apps spreading a `BlocksPresets` preset (the documented path) are unaffected; hand-rolled literal `BlocksDefaults` objects need the new fields. Kept them required (consistent with the existing `removalPolicy`/`deletionProtection`/`pointInTimeRecovery` fields) rather than optional-with-fallback, so posture stays explicit.

packages/core/src/cdk/apigateway-account.ts
osama-rizk: Removed in 912aeef7 — the `cloudWatchRoleArn: role.roleArn` reference already gives CDK the implicit dependency, so the explicit `node.addDependency` is gone.

packages/bb-data/src/index.cdk.ts
osama-rizk: Dropped the comment in 912aeef7.

osama-rizk:
Thanks @soberm — addressed the review in `912aeef7` (rebased on the latest main merge):

- **[blocker] accessLogging default** → now **`false` in both presets** (opt-in). No account-singleton mutation on upgrade unless a stack explicitly opts in; presets/tests/docs/changeset/PR description updated.
- **[question] shared account** → added a test: two access-logging stages in one stack → exactly one `AWS::ApiGateway::Account` (the `Symbol.for` sharing path).
- **[minor] Logger retention conflict** → synth warning when two Loggers set different explicit retentions on the shared group (last still wins); + tests + JSDoc.
- **[minor] required BlocksDefaults fields** → called out in the changeset (spread a preset; hand-rolled objects need the new fields).
- **[nit] redundant `node.addDependency`** → removed. **[nit] restating comment** → dropped.

build + lint + lint:deps + affected unit tests green.

soberm (approved):

sarayev (approved):

packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -76,11 +95,50 @@ export class LambdaCompute extends Compute {
...
+		}
+
 		this.apiGateway = new apigateway.RestApi(this, 'API', {
 			restApiName: 'Blocks API',
-			deployOptions: { cachingEnabled: false },
+			// Don't let RestApi auto-create its own account-level CloudWatch role: it
+			// would collide with the one shared account we provision — a stack may
+			// have only one effective account setting. When access logging is on we
+			// point the stage at the shared account; when off, none is needed.
+			cloudWatchRole: false,
```
sarayev, line 120: [question] This unconditionally turns off `RestApi`'s auto-created account role. On a newer app that's a no-op (the `@aws-cdk/aws-apigateway:disableCloudWatchRole` feature flag is on by default), but if an existing deployer's app doesn't have that flag set, the REST API was previously provisioning an `AWS::ApiGateway::Account`, and flipping this removes it on upgrade, resetting the region's API Gateway CloudWatch role. Worth a line in the changeset's behavior-changes section next to the throttling / handler-log-group notes?

osama-rizk: Sure will address this in a follow up PR.

PR #387, opened 2026-08-18
fix(create-blocks-app): generate the client entry point on fresh checkouts

### LINKED ISSUE #386: Scaffolded templates gitignore aws-blocks/client.js, which the package exports declare as the entry point, so a fresh checkout cannot build or synthesise

@aws-blocks/create-blocks-app (templates), @aws-blocks/core (Hosting build behaviour makes the impact infrastructure-wide)

## Describe the bug

The scaffolded templates ship a local workspace package `aws-blocks` whose `package.json` maps the `browser` and `import` conditions to `./client.js`:

```json
"exports": {
  ".": {
    "types": "./index.ts",
    "browser": "./client.js",
    "react-server": "./client.js",
    "import": "./client.js",
    "default": "./index.ts"
  }
}
```

The template's `gitignore` excludes that same file (`templates/default/gitignore`, line 6: `aws-blocks/client.js`), and no script in the template regenerates it before `build`. The frontend imports the package (`import { api, authApi } from 'aws-blocks'` in `src/index.ts`).

The result is a project that builds only on the machine that scaffolded it. The dev server writes `client.js` as a side effect, so local development always works. Any fresh checkout (a teammate's clone, any CI runner) has no `client.js`, and `vite build` fails with:

```
[commonjs--resolver] Failed to resolve entry for package "aws-blocks". The package may have incorrect main/module/exports specified in its package.json.
```

The impact is larger than a broken frontend: the `Hosting` Block runs `npm run build` during CDK synthesis (`@aws-blocks/core/dist/hosting.js`, "Frontend build failed" throw), so on a clean checkout `cdk synth` itself fails. Infrastructure cannot be synthesised from a fresh clone of a committed scaffold.

Found by putting a freshly scaffolded, committed project through a CI pipeline on its first run.

## Expected behavior

A committed scaffold should build and synthesise from a fresh clone with `npm ci` alone. Either the generated client should be regenerated automatically by every path that needs it, or it should not be gitignored.

## Reproduction steps

1. `npm create @aws-blocks/blocks-app@latest repro-app` (template `default`, version 0.1.19)
2. `cd repro-app && git init && git add -A && git commit -m scaffold` (the template's gitignore excludes `aws-blocks/client.js`)
3. Clone that repository to a second directory, or `rm aws-blocks/client.js` to simulate the fresh checkout
4. `npm ci && npm run build`
5. Observe the resolver error above. `npx cdk synth` fails identically, from inside the Hosting Block's frontend build.

Verified present in the `create-blocks-app@0.1.19` tarball: `templates/default/gitignore` line 6, exports in `templates/*/aws-blocks/package.json`, and no `prebuild` in any template's scripts.

## Code snippet

Suggested fix, which is what we applied locally and verified: a `prebuild` hook in the template's `package.json`, plus a small script beside the template's existing `aws-blocks/scripts/*.ts` calling the toolkit's own public generator:

```json
"scripts": {
  "prebuild": "tsx aws-blocks/scripts/generate-client.ts"
}
```

```ts
// aws-blocks/scripts/generate-client.ts
import { writeClientCode } from '@aws-blocks/blocks/scripts';
await writeClientCode('aws-blocks/index.ts', 'aws-blocks/client.js');
```

npm runs `prebuild` before `build` on every path that builds the frontend, including the Hosting Block during `cdk synth`, so local dev, CI, `cdk synth`, `cdk deploy` and `npm run sandbox` all regenerate the entry point with nothing extra to remember. After this change, deleting `client.js`, `cdk.out` and `dist` and running `cdk synth` alone succeeds and logs the regeneration.

Note `npm run spec` does not cover this: `blocks-generate-spec` writes `blocks.spec.json` only.

A related smaller gap in the same template gitignore: `.hosting/` (the Vite output the synthesis produces) is not excluded, so a user who commits after a synth commits build artefacts.

## Log output

```
🏗️  Building frontend (spa): npm run build
> blocks-app@0.1.0 build
> tsc && vite build
vite v6.4.3 building for production...
✓ 2 modules transformed.
✗ Build failed in 98ms
error during build:
[commonjs--resolver] Failed to resolve entry for package "aws-blocks". The package may have incorrect main/module/exports specified in its package.json.
Error: Frontend build failed: npm run build
Ensure the build command is correct and all dependencies are installed.
    at new Hosting (.../node_modules/@aws-blocks/core/dist/hosting.js:101:23)
    at <anonymous> (.../aws-blocks/index.cdk.ts:37:3)
npx tsx -C cdk aws-blocks/index.cdk.ts: Subprocess exited with error 1
```

## Environment

- @aws-blocks/create-blocks-app: 0.1.19, template `default`
- @aws-blocks/blocks: 0.2.7
- Node.js: v22.23.2 (CI, CodeBuild aws/codebuild/standard:7.0) and v24.18.0 (local macOS)
- npm: 11.x
- OS: Amazon Linux (CodeBuild) and macOS 15

el-pedrito:
Fixes #386.

## The problem

The scaffolded templates declare `./client.js` as the local `aws-blocks` package's `browser` and `import` entry point, and gitignore that same file because it is generated. The dev server writes it as a side effect, so development machines always have one. A fresh checkout of a committed scaffold does not, and `vite build` fails with:

```
[commonjs--resolver] Failed to resolve entry for package "aws-blocks"
```

Because the `Hosting` block runs the frontend build during CDK synthesis, the impact is wider than the frontend: **`cdk synth` fails on any clean clone** (a teammate's machine, any CI runner). Full reproduction in #386, found by putting a freshly scaffolded project through CI on its first run.

## The fix

A `prebuild` hook in each affected template calling `writeClientCode`, the toolkit's existing public generator, from a script placed beside the templates' other `aws-blocks/scripts/*.ts`. npm runs `prebuild` before `build` on every path that builds the frontend, including the `Hosting` block during synth and deploy, so no consumer has anything to remember and no CI needs a special step.

Applied to the **six templates that have a frontend build importing the local package** (auth-cognito, bare, default, demo, nextjs, react). The `backend` and `amplify` templates have no frontend build and are deliberately untouched.

Also adds `.hosting/` to the same templates' gitignore: it is Vite build output produced during synthesis, and without that entry a user committing after a synth commits build artefacts.

## Verification

Copied the modified `default` template to a clean directory, `npm install`, deleted `aws-blocks/client.js` to reproduce the fresh-checkout state, and ran the build: the hook regenerates the file before Vite runs (`generate-client: wrote .../aws-blocks/client.js`). Also verified the hook alone via `npm run prebuild` after deleting the file again.

One note from that verification, unrelated to this change: the templates on `main` import `BlocksPresets`, which is not yet in the published 0.2.7 package, so a template copied from `main` and installed against npm fails `tsc`. Having looked at the release setup, this appears to be normal changesets mechanics rather than a defect: the pending `blocks-infrastructure-defaults` changeset bumps `@aws-blocks/blocks` and `@aws-blocks/create-blocks-app` together, so the next Version Packages release ships both sides at once and scaffolded users never see the gap. Mentioned only so the failing `tsc` in my verification is not attributed to this PR.


packages/create-blocks-app/templates/default/aws-blocks/scripts/generate-client.ts
```diff
@@ -0,0 +1,31 @@
...
+const root = join(dirname(fileURLToPath(import.meta.url)), '..', '..');
+const foundation = join(root, 'aws-blocks', 'index.ts');
+const output = join(root, 'aws-blocks', 'client.js');
+
+if (!existsSync(foundation)) {
+  console.error(`generate-client: cannot find ${foundation}`);
+  process.exit(2);
+}
+
+await writeClientCode(foundation, output);
```
Simone319: 🔴 Blocking — this must run the generator with `--conditions=aws-runtime`. The emitted `client.js` is not condition-independent: the middleware import line comes from whatever registers when the backend is imported. Under plain `tsx` (no `aws-runtime`), Realtime registers `mock-middleware` instead of `aws-middleware`:
```diff
-import '@aws-blocks/bb-realtime/aws-middleware';
+import '@aws-blocks/bb-realtime/mock-middleware';
```
This fails silently (build stays green). Since `prebuild` fires on every build, `npm run deploy` runs this *after* `deploy.ts` wrote the correct client and overwrites it with the mock version right before Vite bundles → a deployed frontend wired to mock middleware, so Realtime subscriptions never reach the API Gateway WebSocket. `dev-server.ts` already guards this case and `generate-client.test.ts` documents the hazard. Affects `bare`, `default`, `react` (incl. the repro template). Fix: invoke it the way `sandbox.ts`/`deploy.ts` do — `node --conditions=aws-runtime --import tsx <generate-client-worker.js> <foundation> <output>`. That also makes the deploy-path overwrite byte-identical, so the ordering concern dissolves. Doesn't affect `npm run dev`/`sandbox` — they use their own generators and never invoke `build`.

packages/create-blocks-app/templates/bare/aws-blocks/scripts/generate-client.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
Simone319: Minor — these 6 byte-identical `generate-client.ts` copies are a mis-copy hazard (and the reason the flag fix above would otherwise need repeating 6×). Consider a `blocks-generate-client` bin in `@aws-blocks/blocks`, mirroring the existing `blocks-generate-spec`/`blocks-vendorize`, so each template reduces to `"prebuild": "blocks-generate-client"` and the correct invocation (incl. the condition flag) lives once in the toolkit.

Simone319:
Minor — `templates/backend/gitignore` still ignores `aws-blocks/client.js` while `backend/aws-blocks/package.json` declares it as the `import` entry. Harmless today (no `build` script) but the same latent trap this PR fixes elsewhere. It's the only template left inconsistent after this change — `amplify` is overlay-only with its own generator; the other 6 are handled here.

packages/create-blocks-app/templates/default/aws-blocks/scripts/generate-client.ts
el-pedrito: Fixed in 40f3aa5, and thanks for catching this — verified empirically: the old path emits `@aws-blocks/bb-realtime/mock-middleware`, the new one emits `aws-middleware`. Rather than adding the flag to six prebuild lines, I took your bin suggestion below: the new `blocks-generate-client` bin spawns the core generator worker with `--conditions=aws-runtime` internally, so the condition is applied in one place and can't be forgotten by a template.

packages/create-blocks-app/templates/bare/aws-blocks/scripts/generate-client.ts
el-pedrito: Done in 40f3aa5 — `blocks-generate-client` on `@aws-blocks/blocks`, mirroring `blocks-generate-spec` and `blocks-vendorize`. The six template copies are removed and `prebuild` is now just the bin. `@aws-blocks/core` exports its existing `generate-client-worker` subpath so the bin resolves it through the blocks dependency; tsx is resolved from the app's devDependencies like `blocks-generate-spec` does.

el-pedrito:
Backend template covered in 40f3aa5 as well: it gets the same `"prebuild": "blocks-generate-client"` hook. It's dormant today (no `build` script), but the day one is added the client regenerates like everywhere else, so the exports/gitignore pair is no longer a latent trap. `amplify` stays untouched as discussed — overlay-only with its own generator.

Simone319 (approved):

sharonyajain (approved):

PR #390, opened 2026-08-19
fix(core): cap JSON-RPC request body size at the shared parser (mock↔AWS parity)

osama-rizk:
## What

Cap the JSON-RPC request-body size at the **shared parser** (`parseRpcRequest` in `packages/core/src/rpc.ts`). Bodies over **10 MiB** (`MAX_RPC_BODY_BYTES`) are now rejected *before* parsing or dispatch with an `InvalidRequest` error named `PayloadTooLarge`.

```ts
if (Buffer.byteLength(bodyText, 'utf8') > MAX_RPC_BODY_BYTES) {
  return { ok: false, response: errorResponse(RpcErrorCode.InvalidRequest,
    `Request body exceeds the ${MAX_RPC_BODY_BYTES} byte limit`, null, { name: 'PayloadTooLarge' }) };
}
```

## Why

Board bug-bash card (P2 / Cost-availability, 5 reports): *"core: JSON-RPC accepts unbounded request bodies (20MB wedges dev pglite)."* The dev server buffered request bodies with no cap, so an oversized body could wedge the local database (PGlite) — and there was no size guard on the shared path at all.

## Why here (mock↔AWS parity)

Both the Lambda handler (`lambda-handler.ts:502`) and the dev server (`dev-server.ts:1116`) route through the **same** `parseRpcRequest`. Enforcing the cap there — rather than only in the dev server — means the local mock rejects **exactly** what the deployed runtime would, which is the framework's established pattern for size limits (see `bb-app-setting`, which enforces the SSM 4 KB limit identically in its mock and AWS layers). The limit matches API Gateway's ~10 MB REST payload cap, so dev and prod agree. The error crosses the wire by `name` (`PayloadTooLarge`), per the framework's error convention, so callers can `isBlocksError(e, ...)` on it.

This also kept the change **out of the heavily-contended `lambda-handler.ts`** (6 open PRs) — the guard lives in `rpc.ts`, which `#231` touches only in unrelated functions (imports + `successResponse`), not `parseRpcRequest`.

## Scope note (honest)

The guard is a **post-assembly** size check (like `bb-app-setting`'s `Buffer.byteLength`), so it rejects the oversized body before it reaches a handler / the DB — fixing the card's symptom. It does not add a *streaming* byte cap in each transport (which would be transport-specific and diverge from the shared-parser pattern); preventing the transient buffering of a multi-GB body could be a follow-up, but the shared cap is the parity-correct core of the fix.

## Testing

- **Parser unit tests added** to `rpc.test.ts`: normal body passes; a body just over the limit is rejected as `PayloadTooLarge`; an oversized *and* invalid-JSON body is rejected by the size guard **before** the parse (proving it short-circuits before any dispatch). **Verified red without the guard.**
- `npm run build` (full monorepo) green; `packages/core` suite **676/676**; `biome lint --changed` clean; `check:api` clean (rpc.ts is internal, not in the public API report). Changeset included.


osama-rizk:
Thanks for the review — both points addressed in `3edafb04`.

**Should-fix (413 vs -32600):** You're right. `-32600` collapses to HTTP 500 via `decodeRpcResponse` (`code > 0 ? code : 500`), so the "413" narrative didn't hold. The guard now emits `errorResponse(413, …, { name: 'PayloadTooLarge' })`, mirroring the 504 handler-timeout path — so a caller's `e.status === 413` works and the status agrees with the name. The test now asserts `error.code === 413` and adds a `decodeRpcResponse` round-trip asserting `ApiError.status === 413` + `isBlocksError(e, 'PayloadTooLarge')`.

**Minor (parity framing):** Also fixed — the JSDoc and changeset no longer imply the deployed Lambda runs this guard for the >10 MiB case. In prod API Gateway rejects an oversized body at the edge before the Lambda is invoked, so this rejection path effectively runs on the dev/mock server; the wording now says the dev server enforces the *same* limit API Gateway enforces at the edge (so an oversized body prod would 413 upstream is likewise rejected locally instead of wedging PGlite).

packages/core/src/rpc.ts
```diff
@@ -32,6 +32,18 @@ export type RpcParseResult =
...
+ * Maximum accepted request-body size, in bytes (10 MiB).
+ *
+ * Mirrors API Gateway's hard payload limit (~10 MB for a REST API), so the
+ * local dev server rejects exactly what the deployed runtime would — an
+ * oversized body can't quietly wedge the local database (e.g. PGlite) in dev
+ * while failing in production. Enforced in `parseRpcRequest`, the single choke
+ * point both the Lambda handler and the dev server route through, so the mock
+ * and AWS paths behave identically.
+ */
+export const MAX_RPC_BODY_BYTES = 10 * 1024 * 1024;
```
Simone319, line 55: With introduction of container, the size no longer applies, as it doesn't use APIGW. We can refactor this to lambda compute only limit when introducing container.

Simone319 (dismissed):

packages/core/src/rpc.ts
osama-rizk: Good call — agreed, this is really the **Lambda + API Gateway** limit sitting in the compute-agnostic parser. For now I've annotated `MAX_RPC_BODY_BYTES` (in `756ec520`) to say exactly that: the 10 MiB is the Lambda+APIGW limit and should become compute-aware — sourced from the compute layer via the `Compute.setEnv` hook, since `core` can't import a `bb-*` compute package — when a non-APIGW compute (container/ALB) lands.

I deliberately **didn't** wire it into `bb-lambda-compute` in this PR: that would edit `index.cdk.ts` (which #341 is reshaping) and build on the compute registry #391 is still changing, so it's better folded into the multi-compute work than done in parallel here. Happy to own that refactor as a fast-follow once #341/#391 land, or leave it to the container-compute PR — whichever you prefer.

packages/core/src/rpc.ts
```diff
@@ -89,6 +111,28 @@ export function decodeRpcResponse(body: unknown): unknown {
  * anything about the JSON-RPC spec.
  */
 export function parseRpcRequest(bodyText: string): RpcParseResult {
+  // Reject oversized bodies before parsing or dispatch — an unbounded body can
+  // wedge the local database in dev and inflate compute cost in production. The
+  // limit matches API Gateway's payload cap (see MAX_RPC_BODY_BYTES), so the
+  // dev server and the deployed runtime reject identically. `name` is set so
+  // callers can match it with `isBlocksError(e, ...)` (errors cross by name).
```
soberm: [nit] This mostly repeats what's already in the `MAX_RPC_BODY_BYTES` doc comment right above (API Gateway parity, wedging PGlite, single choke point). The one new bit here is that `name` is set for `isBlocksError` matching, might be worth trimming this down to just that and letting the constant's own doc comment carry the rest.

packages/core/src/rpc.ts
```diff
@@ -89,6 +111,28 @@ export function decodeRpcResponse(body: unknown): unknown {
...
+    // Emit the real HTTP status (413) as the error code, not a reserved -32xxx.
+    // `decodeRpcResponse` maps a positive code straight to the client
+    // `ApiError.status` (reserved codes collapse to 500), mirroring the 504
+    // handler-timeout path — so a caller's `e.status === 413` check works and
+    // the status agrees with the `PayloadTooLarge` name.
+    return {
+      ok: false,
+      response: errorResponse(
+        413,
+        `Request body exceeds the ${MAX_RPC_BODY_BYTES} byte limit`,
```
soberm: [nit] This'll read as "Request body exceeds the 10485760 byte limit" since `MAX_RPC_BODY_BYTES` interpolates as a raw number. Might be worth adding the MiB figure too, something like "...10485760 byte (10 MiB) limit", since this crosses the wire as `error.message` and everything else here (docs, changeset, description) talks in MiB.

soberm (commented):
Went through this end to end. The core approach is right, and it's already been through a couple of rounds of self-correction (switching the error code from the reserved -32600 to a real 413, and just now adding the missing `@aws-blocks/blocks` changeset entry), so most of what I would've blocked on is already handled.

One thing I was about to flag as a hard blocker: the changeset originally only bumped `@aws-blocks/core`, but `@aws-blocks/blocks` re-exports everything from core (`export * from '@aws-blocks/core'`) and pins it with a caret range, so the repo's `verify-umbrella` guard wants a sibling entry too, otherwise `changeset version` leaves the umbrella's version untouched while its packed content still moves. That's already fixed now (`"@aws-blocks/blocks": patch` is in the frontmatter), so leaving this as a comment instead of blocking on it.

Smaller stuff below, left inline where I could anchor it to a line. None of it blocks.

- The dev server (`packages/core/src/scripts/dev-server.ts:1095-1097`) still buffers the whole request body into one string with no cap, and its verbose `[rpc-in]` logger JSON.parses that entire buffered body too, both before `parseRpcRequest` ever runs. So this guard reliably stops an oversized body from reaching PGlite, which is the actual bug, but a big enough body still costs real memory and CPU during buffering/logging before the guard gets a say. The description already calls this out as an intentional scope cut, so just flagging it explicitly in case it's worth a quick tracking ticket rather than letting it get lost.
- Worth a line on why this is `patch` and not `minor`, given the project's own pre-1.0 convention in `changeset-guard.ts` (patch = non-breaking, minor = the breaking channel). On the dev-server side this does change behavior for payloads that used to get processed as long as they didn't touch PGlite, and now get a hard 413 instead.
- The PR description's "What" section still shows the pre-413 snippet (`RpcErrorCode.InvalidRequest`) and doesn't mention the `decodeRpcResponse` round-trip test that got added. Worth a quick pass so it matches the current diff.

Good parts: routing the cap through the shared parser instead of duplicating it per transport, using `Buffer.byteLength(..., 'utf8')` instead of `.length` (matches the same pattern `bb-app-setting` already uses for its 4 KB check), the test proving the size check runs before `JSON.parse` by feeding it garbage that's also oversized, and the added round-trip test asserting `ApiError.status === 413`. The new note on `MAX_RPC_BODY_BYTES` flagging that this is currently Lambda/API-Gateway-specific and will need to become compute-aware later is a good call too.

packages/core/src/rpc.ts
osama-rizk: Done — trimmed the pre-guard comment to just point at `MAX_RPC_BODY_BYTES` for the rationale, and kept the non-obvious 413/`name` explanation on the return itself. `8f30aa28`.

packages/core/src/rpc.ts
osama-rizk: Good catch — added the human-readable size: the message now reads `Request body exceeds the 10485760 byte (10 MiB) limit`. Computed from the constant (`/ (1024*1024)`) so it can't drift if the limit changes, and added a test asserting the `10 MiB` figure. `8f30aa28`.

soberm (approved):

PR #391, opened 2026-08-19
[core] Retarget config registry to shared role + per-stack compute registry

Simone319:
## Problem

The config registry granted `s3:GetObject` on `blocks-config.json` to a single hardcoded handler function and stamped `BLOCKS_CONFIG_BUCKET` / `BLOCKS_CONFIG_KEY` on that one function. That couples config delivery to one Lambda and doesn't generalize to the multi-compute model, where every compute must be able to read the shared config object.

This is Issue **B1** of Multi-Compute Phase 1. It stacks on **A3** (it consumes the `computes` list and default compute introduced there; its self-registration lives in the A2 `Compute` base).

## Changes

- **Retarget `finalizeConfigRegistry`** — signature is now `finalizeConfigRegistry(root, executionRole, computes)`:
  - grants `s3:GetObject` **once to the shared execution role** (so every compute that assumes the role can read the object), instead of to one function's role;
  - stamps `BLOCKS_CONFIG_BUCKET` / `BLOCKS_CONFIG_KEY` on **every compute** via `compute.setEnv(...)`.
  - The previous `ConfigFinalizeRoot` wrapper interface is removed in favor of three explicit parameters.
- **Per-stack compute registry** (`packages/core/src/cdk/compute/compute-registry.ts`, new): `registerCompute` / `getComputes`, with state stored on the owning stack keyed by a `Symbol` and resolved via `cdk.Stack.of()` — mirroring the config registry. This keeps each stack's computes isolated in a multi-stack synth (Amplify Gen2 nested stacks, Pipelines stages); a process-global list would let one stack's finalize see another's computes.
- **`Compute` base self-registers** — `registerCompute(this)` in its constructor, so every compute joins its stack's registry the moment it's constructed (no separate discovery pass).
- **`create()` reads the stack's computes directly via `getComputes(stack)`** and passes `(stack, stack.executionRole, getComputes(stack))` to finalize. (There is no `.computes` getter on `BlocksStack` / `BlocksBackend`; the registry is read directly.)
- Updated the two CDK test callers (`bb-auth-cognito`, `bb-knowledge-base`) to the new signature.

Behavior-preserving for the default single-compute app: the same config object is written to S3, the same two coordinates reach the runtime, and the same `s3:GetObject` permission is available — now via the shared role.

## Validation

- `npm run build` ✓
- Unit tests: `@aws-blocks/core` 656/656 (incl. a new multi-stack registry-isolation test), `bb-auth-cognito` 241/241, `bb-knowledge-base` 129/129 ✓
- `npm run lint` (zero blocking errors), `npm run lint:deps`, `npm run check:api` ✓
- `npm run test:e2e:local` 8/8 ✓

## Checklist

- [x] PR description included

### DIFF
--- a/packages/core/src/cdk/compute/compute-registry.ts
+++ b/packages/core/src/cdk/compute/compute-registry.ts
@@ -1,0 +1,45 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as cdk from 'aws-cdk-lib';
+import type { Construct } from 'constructs';
+import type { Compute } from './compute.js';
+
+const REGISTRY_KEY = Symbol.for('BLOCKS_COMPUTE_REGISTRY');
+
+/**
+ * Get or create the compute list for a given stack. The list is stored on the
+ * stack object (keyed by a Symbol), so each stack in a multi-stack synth gets
+ * its own — a compute never leaks into another stack's list. Mirrors the config
+ * registry (`config-registry.ts`), which scopes its state the same way.
+ */
+function getRegistry(stack: cdk.Stack): Compute[] {
+	let list = (stack as any)[REGISTRY_KEY] as Compute[] | undefined;
+	if (!list) {
+		list = [];
+		(stack as any)[REGISTRY_KEY] = list;
+	}
+	return list;
+}
+
+/**
+ * Register a compute on its owning stack. Called from the {@link Compute} base
+ * constructor, so every compute self-registers the moment it is constructed —
+ * the finalize steps then enumerate them without a separate discovery pass
+ * (mirrors how `registerConfig` accumulates config during the backend import).
+ *
+ * @param compute - The compute to register (used to locate its stack).
+ */
+export function registerCompute(compute: Compute): void {
+	getRegistry(cdk.Stack.of(compute)).push(compute);
+}
+
+/**
+ * The computes registered on the stack that owns `scope`, in construction
+ * order. Returns an empty array before any compute is constructed.
+ *
+ * @param scope - Any construct in the stack (used to locate the stack).
+ */
+export function getComputes(scope: Construct): readonly Compute[] {
+	return getRegistry(cdk.Stack.of(scope));
+}
--- a/packages/core/src/cdk/config-registry.ts
+++ b/packages/core/src/cdk/config-registry.ts
@@ -4,8 +4,8 @@
 import * as cdk from 'aws-cdk-lib';
 import * as s3 from 'aws-cdk-lib/aws-s3';
 import * as s3deploy from 'aws-cdk-lib/aws-s3-deployment';
-import * as iam from 'aws-cdk-lib/aws-iam';
 import type { Construct } from 'constructs';
+import type { Compute } from './compute/compute.js';
 
 const REGISTRY_KEY = Symbol.for('BLOCKS_CONFIG_REGISTRY');
 
@@ -47,19 +47,29 @@ export function registerConfig(scope: Construct, key: string, value: unknown): v
 
 /**
  * Finalize the config registry: create an S3 bucket, upload the config JSON,
- * set env vars on the handler, and grant read access.
+ * grant read to the shared execution role, and stamp the config coordinates
+ * (`BLOCKS_CONFIG_BUCKET` / `BLOCKS_CONFIG_KEY`) onto every compute.
+ *
+ * Read access is granted once to the shared role (`root.executionRole`) rather
+ * than to a single function, so every compute that assumes the role can read
+ * the object. The bucket/key coordinates can't live on a role (env vars are
+ * per-compute), so they are set on each compute via `setEnv`.
  *
  * Must be called after all BBs are constructed (i.e., after the backendCDKPath
  * import completes in BlocksStack.create() / BlocksBackend.create()).
  *
- * @param scope - The CDK construct to create resources under
- * @param handler - The Lambda function that needs to read the config
+ * @param root - The construct to create the config resources under (also used
+ *   to locate the owning stack).
+ * @param executionRole - The shared role every compute assumes; config read is
+ *   granted to it once.
+ * @param computes - The computes to stamp `BLOCKS_CONFIG_BUCKET` / `BLOCKS_CONFIG_KEY` on.
  */
 export function finalizeConfigRegistry(
-	scope: Construct,
-	handler: cdk.aws_lambda.IFunction,
+	root: Construct,
+	executionRole: cdk.aws_iam.IRole,
+	computes: readonly Compute[],
 ): void {
-	const stack = cdk.Stack.of(scope);
+	const stack = cdk.Stack.of(root);
 	const registry = getRegistry(stack);
 
 	if (registry.finalized) return;
@@ -67,7 +77,7 @@ export function finalizeConfigRegistry(
 
 	if (registry.entries.size === 0) return;
 
-	const configBucket = new s3.Bucket(scope, 'BlocksConfigBucket', {
+	const configBucket = new s3.Bucket(root, 'BlocksConfigBucket', {
 		removalPolicy: cdk.RemovalPolicy.DESTROY,
 		autoDeleteObjects: true,
 		encryption: s3.BucketEncryption.S3_MANAGED,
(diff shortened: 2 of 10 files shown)


packages/core/src/cdk/compute/compute-registry.ts
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Register a compute on its owning stack. Called from the {@link Compute} base
+ * constructor, so every compute self-registers the moment it is constructed —
+ * the finalize steps then enumerate them without a separate discovery pass
+ * (mirrors how `registerConfig` accumulates config during the backend import).
+ *
+ * @param compute - The compute to register (used to locate its stack).
+ */
+export function registerCompute(compute: Compute): void {
+	getRegistry(cdk.Stack.of(compute)).push(compute);
```
osama-rizk, line 34: Scope check worth confirming. This keys the registry on `cdk.Stack.of(compute)` — the CDK stack — matching the config registry. But #341 anchors `defaults` per-`BlocksBackend` (a Construct embedded in a stack), and that's an explicitly supported, tested topology (`blocks-backend.test.ts:325` — "two backends in one stack must NOT clobber each other").

For a two-`BlocksBackend`s-in-one-CDK-stack synth, config + compute registries are therefore stack-scoped while `defaults` are backend-scoped. Because `finalizeConfigRegistry` has a per-stack `finalized` flag, backend A finalizes first (bucket created, `s3:GetObject` granted to **A's** role, computes-present-then stamped), and backend B's finalize returns early — so **B's execution role never gets the grant and B's compute never gets `BLOCKS_CONFIG_BUCKET`/`KEY`**, silently breaking B's runtime config read (B's entries still merge into A's object via `Lazy`).

This is pre-existing for config (the finalize-once + single grant predate this PR); B1 just extends the stack-scoped assumption to computes. Not a blocker — but worth either confirming multi-backend-per-stack is unsupported *for config* (only `defaults` isolate), or scoping these registries per-backend to match `defaults`. Note the new isolation test covers multi-**stack**, not multi-**backend-in-one-stack**, so this path is untested.

packages/core/src/cdk/index.ts
```diff
@@ -124,7 +125,7 @@ export class BlocksStack extends cdk.Stack implements BaseBlocksStack {
       }
     }
     // Finalize BB config → S3 (after all BBs have registered their config)
-    finalizeConfigRegistry(stack, stack.handler);
+    finalizeConfigRegistry(stack, stack.executionRole, getComputes(stack));
```
osama-rizk, line 128: Minor doc nit: the PR description and changeset say "`BlocksStack.computes` / `BlocksBackend.computes` now read from the registry," but `create()` calls `getComputes(stack)` directly here rather than going through a `.computes` getter. Worth confirming that public getter actually exists and is used, or dropping the claim so the description matches the code.

osama-rizk (commented):
Refactor (B1) — verified behavior-preserving for the default single-compute app: `grantRead(handler)` previously targeted the handler's role, which is the shared `executionRole`, so the new `grantRead(executionRole)` is the identical grant, and the config coordinates are stamped on the same single compute. Self-registration timing is correct (the construct is attached via `super()` before `cdk.Stack.of` runs, and finalize runs after the backend import, so `getComputes` enumerates every compute including the default). Multi-stack isolation holds and is tested.

Two low-severity items inline: a registry-scope-vs-`defaults`-scope inconsistency that affects the multi-backend-in-one-stack topology (pre-existing for config, now extended to computes), and a description/code mismatch on the `.computes` getter.

packages/core/src/cdk/config-registry.ts
```diff
@@ -83,24 +93,20 @@ export function finalizeConfigRegistry(
...
-		'BLOCKS_CONFIG_BUCKET',
-		configBucket.bucketName,
-	);
-	(handler as cdk.aws_lambda.Function).addEnvironment(
-		'BLOCKS_CONFIG_KEY',
-		configKey,
-	);
+	// Grant read once to the shared role (scoped to the config key), so every
+	// compute assuming the role can read it.
+	configBucket.grantRead(executionRole, configKey);
```
osama-rizk, line 109: The refactor swaps the old single-action `s3:GetObject` `PolicyStatement` for `configBucket.grantRead(executionRole, configKey)`. `grantRead` grants a broader action set (`s3:GetObject*`, `s3:GetBucket*`, `s3:List*`) than the previous single `s3:GetObject`. On this dedicated, block-public-access config bucket that holds only `blocks-config.json`, the practical exposure is negligible (the shared role can list/head a bucket that contains only the config it already reads), so this is not blocking — but it's a small, real widening of least-privilege worth a conscious call. The runtime only ever needs `GetObject` on the one key.

**Fix:** to keep the grant as tight as the prior code, replace it with an explicit `executionRole.addToPrincipalPolicy(new iam.PolicyStatement({ actions: ['s3:GetObject'], resources: [`${configBucket.bucketArn}/${configKey}`] }))` (re-adding the `aws-iam` import). Otherwise, a one-line comment noting the intentional acceptance of `grantRead`'s broader set on a config-only bucket would settle it.

.changeset/bb-auth-cognito-config-registry-test.md
```diff
@@ -0,0 +1,9 @@
+---
+"@aws-blocks/bb-auth-cognito": patch
```
osama-rizk: This PR ships a `@aws-blocks/bb-auth-cognito` changeset for a test-only edit (`index.cdk.test.ts`) but ships no changeset for the equivalent test-only edit to `@aws-blocks/bb-knowledge-base` (`index.cdk.test.ts`). Both are release-irrelevant test adaptations to the new `finalizeConfigRegistry` signature, so neither strictly needs a changeset — but publishing one for one package and not the other is inconsistent and will produce a spurious patch bump + changelog line for `bb-auth-cognito` only. Not blocking either way.

**Fix:** pick one convention — either drop the `bb-auth-cognito` changeset (test-only, no published behavior changed) or add a matching `bb-knowledge-base` changeset. Dropping it is the cleaner choice since no runtime/public API of either package changed.

packages/core/src/cdk/compute/compute-registry.ts
Simone319: Good catch, and the analysis is right — I confirmed the infrastructure defaults are anchored per backend, while the config and compute registries are scoped per CDK stack with a finalize-once flag. So a second backend constructed inside the same CDK stack would hit the early-return in finalize and never receive the config read grant or the config coordinates.

This is pre-existing behavior for the config registry: the finalize-once flag and the single grant predate this change, which only extends that same stack scoping to the new compute list. It stays behavior-preserving for the default single-backend app, which is this PR's scope. Multiple backends sharing one CDK stack isn't a defined topology for config delivery today — only the infrastructure defaults isolate per backend.

Rather than widen this PR, I'll track scoping the config and compute registries per backend (to match the defaults) as a separate follow-up, so the fix lands as one coherent change with a dedicated multiple-backends-in-one-stack test. The isolation test added here intentionally covers the multiple-stacks case; I'll extend it to the multiple-backends case in that follow-up.

packages/core/src/cdk/index.ts
Simone319: You're right — there's no `.computes` getter; `create()` calls `getComputes(stack)` directly (a later commit dropped the getters in favor of reading the registry directly). I've corrected the PR description to match the code. Thanks for catching the stale wording.

packages/core/src/cdk/config-registry.ts
Simone319: Intentional, and I've added a comment in the code to make that explicit. I kept `grantRead(role, key)` over a hand-rolled `GetObject`-only statement because it's the idiomatic form, it's already resource-scoped to the single config object, and it stays correct automatically if the config bucket ever gains KMS encryption (it would add `kms:Decrypt`). The extra list/get-bucket actions are scoped to a dedicated, block-all-public, config-only bucket, so the practical exposure is negligible.

.changeset/bb-auth-cognito-config-registry-test.md
Simone319: Agreed — dropped the changeset. It was a test-only adaptation with no published behavior change, and the equivalent test edit in the other package ships no changeset, so removing it keeps the convention consistent and avoids a spurious version bump.

Simone319: Correction on this — dropping the changeset actually turned the changeset-coverage check red. That guard requires a changeset entry for **every** changed package, including test-only edits, so with the changeset removed neither this package nor the other test-adapted one was covered, and CI flagged both.

So instead of dropping it, I added a single changeset covering **both** test-adapted packages (patch each). That gives the consistent outcome you were after — both are covered the same way, rather than one having a changeset and the other not. I also added the umbrella package to the primary changeset, since the release guard requires it to bump alongside the packages it re-exports.

osama-rizk (approved):

soberm (approved):

PR #393, opened 2026-08-19
fix(hosting): drop cloudfront-keyvaluestore:ListKeys from KvKeys custom resource

osama-rizk:
## Problem

AWS Blocks **Hosting** fails to deploy into accounts governed by a restrictive Service Control Policy (SCP) whose managed policy **denies `cloudfront-keyvaluestore:ListKeys`**. The `KvKeys` custom resource requests that action, so the stack can't be granted it and the deploy fails.

This blocks affected users from deploying stock Blocks Hosting ahead of an upcoming launch. A user worked around it by editing the hosting code locally; this PR fixes it properly in the package.

## Root cause

`KvKeys` (`packages/hosting/src/constructs/kv_keys.ts`) granted **six** KVS data-plane actions, but the handler (`kv_keys_handler.ts`) only ever calls **two**:

| Action | Used by handler? |
|---|---|
| `DescribeKeyValueStore` | ✅ `currentEtag()` |
| `UpdateKeys` | ✅ `applyUpdate()` |
| `ListKeys` | ⚠️ only `listAll()`, on the **Create** path |
| `GetKey` | ❌ never |
| `PutKey` | ❌ never |
| `DeleteKey` | ❌ never |

`ListKeys` was used solely to diff the *desired* entries against the *live* store on Create. But the `KeyValueStore` (`cdn_construct.ts`) is created **fresh with no `ImportSource`**, so at Create time the store is empty and the live read returns nothing. The normal redeploy (Update) path already diffs against `OldResourceProperties.Entries`, and Delete uses `deleteDrainSet()` — neither touches `ListKeys`.

## Fix

- Trim the IAM policy to **`DescribeKeyValueStore` + `UpdateKeys`** (least privilege; also drops the 3 dead actions).
- On **Create**, treat previous state as `{}` instead of a live `ListKeys` read.
- Remove the now-unused `listAll()` and the `ListKeysCommand` import.

## Trade-off

A re-create over a store that still holds **orphan** keys (e.g. an out-of-band manual write) will no longer auto-drain them. The route table is regenerated wholesale each deploy, so new/changed keys are still written — only keys absent from `desired` could linger, which does not happen for a CDK-managed store created fresh per stack. If we want to keep that safety net, we can gate a `ListKeys` drain behind an opt-in flag that is **off by default**.

## Verification

- `npm run build -w packages/hosting` — typecheck + handler bundle ✅
- `npm test -w packages/hosting` — **842 passed, 0 failed** ✅
- No test pinned the six IAM actions (no `kv_keys.test.ts`), and the handler unit tests only cover `computeDiff`/`batches`, so nothing else depends on the removed code.

## Open item before merge

Confirm with the reporter that the SCP denies **only** `ListKeys`, not also `DescribeKeyValueStore`/`UpdateKeys`. If those are denied too, KVS can't be written under that SCP at all and a different approach (or an SCP exception) is needed. Opening as **draft** pending that confirmation.

---

## End-to-end validation

Deployed a real Nuxt SSR app whose hosting stack links this branch's `@aws-blocks/hosting`, to confirm the reduced permission set still produces a working deploy.

- **Linked & confirmed at byte level** — the app loads the patched `kv_keys.js`; deploy-time IAM policy grants only `cloudfront-keyvaluestore:DescribeKeyValueStore` + `UpdateKeys` (no `ListKeys`/`GetKey`/`PutKey`/`DeleteKey`), and the handler carries no `ListKeysCommand`/`listAll`.
- **Deployed** to stack `stress-nuxt-prod-default` (us-west-2) → `UPDATE_COMPLETE`. The synthesized CloudFormation template grants the KvKeys resource only the two actions, and the **`RouteStoreKeys` custom resource completed successfully** — i.e. the KVS route table was written **without `ListKeys`**.
- **App serves correctly** — `/` → 308 → `/myapp/` → 200 (Nuxt SSR), confirming the KVS-driven edge routing works with the trimmed policy. Live: https://drz7zx8m0esrg.cloudfront.net/myapp/ (disposable sandbox stack).
- **E2e (`ssr-hosting-e2e`, Nuxt suite): 77 passed / 6 skipped / 2 failed.** All routing, CloudFront-behavior, asset, cache, redirect, and URL-safety tests passed (these exercise the KVS route table end-to-end). The 6 skipped are non-Nuxt framework-gated tests. The 2 failures are a pre-existing Vue SSR **hydration mismatch** in the sample app — unrelated to this change, which touches only an IAM policy and the handler's Create-path diff (no path affects Vue hydration, and no routing/KVS test failed).

**Conclusion:** the fix deploys cleanly and the route-table custom resource succeeds with the reduced permission set — so it will deploy under a restrictive SCP that denies `cloudfront-keyvaluestore:ListKeys`.

### DIFF
--- a/.changeset/hosting-drop-kvs-listkeys.md
+++ b/.changeset/hosting-drop-kvs-listkeys.md
@@ -1,0 +1,9 @@
+---
+"@aws-blocks/hosting": patch
+---
+
+Trim the `KvKeys` custom resource IAM policy to true least privilege: it now grants only `cloudfront-keyvaluestore:DescribeKeyValueStore` and `UpdateKeys` — the two actions the deploy-time handler actually calls. The previously-granted `ListKeys`, `GetKey`, `PutKey`, and `DeleteKey` are dropped.
+
+This also makes the hosting stack deployable under restrictive Service Control Policies (SCPs) / permission boundaries that deny `cloudfront-keyvaluestore:ListKeys`, which previously blocked the deploy.
+
+Behavior is preserved: `ListKeys` was only used to diff against the live store on Create, but the route-table `KeyValueStore` is created fresh with no `ImportSource`, so it is empty at Create time — the handler now diffs Create against `{}`. The Update path still diffs against the prior template's entries and Delete still drains via `deleteDrainSet()`, neither of which used `ListKeys`.
--- a/packages/hosting/src/constructs/kv_keys.ts
+++ b/packages/hosting/src/constructs/kv_keys.ts
@@ -29,9 +29,9 @@ export type KvKeysProps = {
   /** The CloudFront KeyValueStore to write into. */
   store: IKeyValueStore;
   /**
-   * Desired key→value map. The custom resource diffs this against what's in
-   * the store and applies the minimal set of put/delete operations (chunked to
-   * the 50-key / 3 MB UpdateKeys ceiling).
+   * Desired key→value map. The custom resource diffs this against the
+   * previously deployed entries (empty on Create) and applies the minimal set
+   * of put/delete operations, chunked to the 50-key / 3 MB UpdateKeys ceiling.
    */
   entries: Record<string, string>;
 };
@@ -65,15 +65,15 @@ export class KvKeys extends Construct {
       timeout: Duration.minutes(5),
     });
 
-    // Data-plane KVS access: describe/list to diff, update to apply.
+    // Data-plane KVS access. The handler reads the store's current ETag
+    // (DescribeKeyValueStore) and writes the route-table diff in batches
+    // (UpdateKeys) — the only two actions it calls. Previous state is read from
+    // this custom resource's CloudFormation properties, not from the store, so
+    // no key-read (ListKeys/GetKey) or single-key write (PutKey/DeleteKey) is used.
     handler.addToRolePolicy(
       new iam.PolicyStatement({
         actions: [
           'cloudfront-keyvaluestore:DescribeKeyValueStore',
-          'cloudfront-keyvaluestore:ListKeys',
-          'cloudfront-keyvaluestore:GetKey',
-          'cloudfront-keyvaluestore:PutKey',
-          'cloudfront-keyvaluestore:DeleteKey',
           'cloudfront-keyvaluestore:UpdateKeys',
         ],
         resources: [props.store.keyValueStoreArn],
--- a/packages/hosting/src/constructs/kv_keys_handler.ts
+++ b/packages/hosting/src/constructs/kv_keys_handler.ts
@@ -25,7 +25,6 @@
 import {
   CloudFrontKeyValueStoreClient,
   DescribeKeyValueStoreCommand,
-  ListKeysCommand,
   UpdateKeysCommand,
 } from '@aws-sdk/client-cloudfront-keyvaluestore';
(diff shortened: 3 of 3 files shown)


pranavosu (dismissed):

pranavosu (approved):

Simone319 (approved):

soberm (commented):
Overall this is a well-scoped, well-reasoned change. The IAM trim is a genuine least-privilege win, and I verified the core premise independently: `RouteStore` is constructed in `cdn_construct.ts:713-716` with no `ImportSource`, so the KeyValueStore really is empty at Create time. Dropping the live `ListKeys` read also removes an unnecessary paginated full-store scan from the Create hot path. No critical issues found. A few suggestions below.

### Suggestions

**1. `kv_keys_handler.ts:184-193` — the Create-path behavior change ships untested.** The existing unit tests only exercise `computeDiff`, `batches`, and `deleteDrainSet` — none call `handler()` or assert on the new `previous = {}` branch. Consider extracting the previous-state selection into a small exported pure function, mirroring the `deleteDrainSet` pattern already in this file:
```ts
export function previousEntries(event: Event): Entries {
  return event.RequestType === 'Update' && event.OldResourceProperties?.Entries
    ? (JSON.parse(event.OldResourceProperties.Entries) as Entries)
    : {};
}
```
Then add cases for Create (→ {}) and Update with/without `OldResourceProperties.Entries`. This pins the exact regression this PR is designed to prevent, without mocking the AWS SDK.

**2. `kv_keys.ts:73-81` — nothing pins the trimmed IAM action list.** A future edit could silently re-widen it (e.g. copy-pasting `ListKeys` back in) and reopen the exact SCP failure this fixes. A small cdk-assertions synth test would turn least-privilege into an enforced invariant:
```ts
template.hasResourceProperties('AWS::IAM::Policy', {
  PolicyDocument: Match.objectLike({
    Statement: Match.arrayWith([Match.objectLike({
      Action: ['cloudfront-keyvaluestore:DescribeKeyValueStore', 'cloudfront-keyvaluestore:UpdateKeys'],
    })]),
  }),
});
```

**3. Orphan-key cleanup edge case.** Beyond the manual-write scenario noted in the description: if the KvKeys construct's logical id ever changes (rename / re-scope) while the underlying KeyValueStore is untouched, the new resource's Create sees `previous = {}` and won't delete pre-existing keys not in `desired`. Cleanup then depends on delete/create ordering vs. the old resource's own drain, which isn't obviously deterministic. The proposed opt-in `ListKeys`-drain flag (off by default) would be cheap to land now; if not, worth a follow-up ticket before merge rather than leaving it as a description-only note.

**4. Changeset.** This is a behavior change on Create, not just a permissions trim. The changeset only frames it as an IAM cleanup — worth a line noting the orphan-key trade-off so it surfaces in the published changelog, not just the PR description.

### Nitpicks
- The Create-path rationale is duplicated between the `KvKeysProps.entries` JSDoc and the handler's inline comment. Consider having one reference the other so they don't drift.
- `previous` at `kv_keys_handler.ts:184` reads cleaner as a helper call at the call site — folds into suggestion 1.

Nice touches worth calling out: the dead-code removal (`GetKey`/`PutKey`/`DeleteKey` were never called — confirmed no other references in the repo), the thorough root-cause table and live-sandbox deploy verification in the description, and the safe upgrade path (existing stacks hit this as an Update, so no behavior change on their next deploy).

soberm (approved):

PR #418, opened 2026-08-20
feat(core): CloudFormation Express Mode (--express) for sandbox deploys

osama-rizk:
## What

Enables **CloudFormation Express Mode** for **sandbox** deploys. `npm run sandbox` now runs:

```
cdk deploy --all --method direct --express --require-approval never …
```

Express Mode (`--express`) reports each stack operation complete as soon as the resource's configuration is applied, **without waiting for full stabilization** — a substantial speedup for the sandbox iteration loop. `--method direct` (skip change-set creation) is the baseline method it builds on.

## Scope: sandbox only

- ✅ `packages/core/src/scripts/sandbox.ts` (`npm run sandbox`) — gets `--method direct --express`.
- ❌ `packages/core/src/scripts/deploy.ts` (`npm run deploy`, production) — **unchanged**. Production keeps a reviewable change set and full stabilization with automatic rollback. A test asserts production uses neither `--express` nor `--method`.
- `cdk watch` (sandbox iterate loop) is unchanged.

## Rollback

Express Mode **disables automatic rollback by default**. For the throwaway sandbox loop we keep that default (no `--rollback`), trading rollback safety for speed — a failed sandbox deploy may leave the stack in a failed state until the next deploy. Production is untouched and still rolls back.

## CDK CLI version

`--express` is not present in older CLIs (our previous lockfile pinned `aws-cdk` 2.1122.0, which only has `--rollback`). It lands in `aws-cdk` **2.1138.0**, so the dev-dependency floor is bumped `^2.1117.0` → `^2.1138.0`. Lockfile change is scoped to the `aws-cdk` version fields only (no dependency-tree churn).

## Refactor

The sandbox deploy argv is built by a pure, unit-tested `buildSandboxDeployArgs` helper, mirroring the existing `buildCdkDeployArgs` used for production.

## Tests

`packages/core/src/scripts/sandbox-args.test.ts` (7 cases):
- enables `--express`
- pairs with `--method direct`
- leaves rollback at its default (no `--rollback`)
- `--all` present (Lambda@Edge second stack)
- non-interactive contract + `sandboxMode=true` + `--app` preserved
- `outDir` / `projectRoot` interpolation
- **guard:** production `buildCdkDeployArgs` includes neither `--express` nor `--method`

## History

An earlier revision of this PR used `--method direct` alone and described it as "express mode". That was inaccurate — `--method direct` only skips change-set creation and still waits for stabilization. This revision switches to the real `--express` flag.

## Checklist
- [x] Changeset added (`@aws-blocks/core`, patch)
- [x] `biome lint` clean on changed files
- [x] Build + unit tests pass
- [ ] Draft — pending review

### DIFF
--- a/packages/core/src/scripts/sandbox-args.test.ts
+++ b/packages/core/src/scripts/sandbox-args.test.ts
@@ -26,8 +26,19 @@ describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
     backendPath: 'aws-blocks/index.ts',
   });
 
-  it('deploys via direct UpdateStack (express mode) instead of a change set', () => {
+  it('enables CloudFormation Express Mode via --express', () => {
+    assert.ok(args.includes('--express'), `expected --express in: ${args.join(' ')}`);
+  });
+
+  it('pairs express mode with the direct deployment method', () => {
     assert.strictEqual(valueAfter(args, '--method'), 'direct');
+  });
+
+  it('leaves express-mode rollback at its default (off) — no --rollback', () => {
+    // Express Mode disables automatic rollback by default; we keep that default
+    // for the throwaway sandbox loop. Passing --rollback here would silently
+    // give up the speedup.
+    assert.ok(!args.includes('--rollback'), `sandbox must not force --rollback: ${args.join(' ')}`);
   });
 
   it('deploys every stack so Lambda@Edge apps synth cleanly', () => {
@@ -49,9 +60,11 @@ describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
   });
 
   it('does NOT leak express mode into the production deploy path', () => {
-    // A production deploy must keep a reviewable change set: `--method direct`
-    // is a sandbox-only speedup and must never appear in buildCdkDeployArgs.
+    // A production deploy must keep a reviewable change set and full
+    // stabilization with rollback: neither `--express` nor `--method direct`
+    // (both sandbox-only speedups) may appear in buildCdkDeployArgs.
     const prod = buildCdkDeployArgs({ projectRoot: '/app', outputsFile: '.blocks-sandbox/outputs.json' });
+    assert.ok(!prod.includes('--express'), `production deploy must not use --express: ${prod.join(' ')}`);
     assert.ok(!prod.includes('--method'), `production deploy must not use --method: ${prod.join(' ')}`);
   });
 });
--- a/packages/core/src/scripts/sandbox-args.test.ts
+++ b/packages/core/src/scripts/sandbox-args.test.ts
@@ -7,6 +7,18 @@ import assert from 'node:assert';
 import { buildSandboxDeployArgs } from './sandbox.js';
 import { buildCdkDeployArgs } from './deploy-stream.js';
 
+/**
+ * Return the value that follows `flag` in `args`, failing with a clear message
+ * if the flag is absent. Reading `args[indexOf(flag) + 1]` directly silently
+ * reads `args[0]` when the flag is missing (indexOf returns -1), which turns a
+ * dropped flag into a confusing value mismatch instead of a "flag missing".
+ */
+function valueAfter(args: string[], flag: string): string {
+  const i = args.indexOf(flag);
+  assert.ok(i !== -1, `expected ${flag} in: ${args.join(' ')}`);
+  return args[i + 1];
+}
+
 describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
   const args = buildSandboxDeployArgs({
     outDir: '.blocks-sandbox',
@@ -15,7 +27,7 @@ describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
   });
 
   it('deploys via direct UpdateStack (express mode) instead of a change set', () => {
-    assert.strictEqual(args[args.indexOf('--method') + 1], 'direct', `expected --method direct in: ${args.join(' ')}`);
+    assert.strictEqual(valueAfter(args, '--method'), 'direct');
   });
 
   it('deploys every stack so Lambda@Edge apps synth cleanly', () => {
@@ -24,16 +36,16 @@ describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
 
   it('keeps the non-interactive deploy contract and sandbox context', () => {
     assert.deepStrictEqual(args.slice(0, 4), ['exec', 'cdk', '--', 'deploy']);
-    assert.strictEqual(args[args.indexOf('--require-approval') + 1], 'never');
-    assert.strictEqual(args[args.indexOf('--outputs-file') + 1], '.blocks-sandbox/outputs.json');
+    assert.strictEqual(valueAfter(args, '--require-approval'), 'never');
+    assert.strictEqual(valueAfter(args, '--outputs-file'), '.blocks-sandbox/outputs.json');
     assert.ok(args.includes('sandboxMode=true'), `expected sandboxMode=true in: ${args.join(' ')}`);
-    assert.strictEqual(args[args.indexOf('--app') + 1], 'npm exec tsx -- -C cdk aws-blocks/index.ts');
+    assert.strictEqual(valueAfter(args, '--app'), 'npm exec tsx -- -C cdk aws-blocks/index.ts');
   });
 
   it('interpolates the provided outDir and projectRoot', () => {
     const custom = buildSandboxDeployArgs({ outDir: 'out', projectRoot: '/other', backendPath: 'b.ts' });
-    assert.strictEqual(custom[custom.indexOf('--outputs-file') + 1], 'out/outputs.json');
-    assert.strictEqual(custom[custom.indexOf('--context') + 1], 'projectRoot=/other');
+    assert.strictEqual(valueAfter(custom, '--outputs-file'), 'out/outputs.json');
+    assert.strictEqual(valueAfter(custom, '--context'), 'projectRoot=/other');
   });
 
   it('does NOT leak express mode into the production deploy path', () => {
--- a/packages/core/src/scripts/sandbox-args.test.ts
+++ b/packages/core/src/scripts/sandbox-args.test.ts
@@ -12,6 +12,11 @@ import { buildCdkDeployArgs } from './deploy-stream.js';
  * if the flag is absent. Reading `args[indexOf(flag) + 1]` directly silently
  * reads `args[0]` when the flag is missing (indexOf returns -1), which turns a
  * dropped flag into a confusing value mismatch instead of a "flag missing".
+ *
+ * NOTE: resolves only the FIRST occurrence of `flag`. `--context` legitimately
+ * appears twice in this argv (projectRoot, then sandboxMode=true), so do not
+ * use `valueAfter(args, '--context')` expecting the second value — it silently
+ * returns the first. Assert repeated flags with `args.includes(...)` instead.
  */
 function valueAfter(args: string[], flag: string): string {
   const i = args.indexOf(flag);
--- a/packages/core/src/scripts/sandbox-args.test.ts
+++ b/packages/core/src/scripts/sandbox-args.test.ts
@@ -4,7 +4,7 @@
 import { describe, it } from 'node:test';
 import assert from 'node:assert';
 
-import { buildSandboxDeployArgs } from './sandbox.js';
+import { buildSandboxDeployArgs, sandboxFailureRecoveryHint } from './sandbox.js';
 import { buildCdkDeployArgs } from './deploy-stream.js';
 
 /**
@@ -71,5 +71,22 @@ describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
     const prod = buildCdkDeployArgs({ projectRoot: '/app', outputsFile: '.blocks-sandbox/outputs.json' });
     assert.ok(!prod.includes('--express'), `production deploy must not use --express: ${prod.join(' ')}`);
     assert.ok(!prod.includes('--method'), `production deploy must not use --method: ${prod.join(' ')}`);
+  });
+});
+
+describe('sandboxFailureRecoveryHint — no silent dead-end after a failed express deploy', () => {
+  const hint = sandboxFailureRecoveryHint();
+
+  it('points to the repo-native destroy + redeploy recovery', () => {
+    assert.match(hint, /npm run sandbox:destroy/);
+    assert.match(hint, /npm run sandbox\b/);
+  });
+
+  it('covers the manual UPDATE_ROLLBACK_FAILED escape hatch', () => {
+    assert.match(hint, /continue-update-rollback/);
+  });
+
+  it('explains why (rollback is off under express mode)', () => {
+    assert.match(hint, /rollback is off|automatic rollback/i);
   });
 });
--- a/packages/core/src/scripts/sandbox-args.test.ts
+++ b/packages/core/src/scripts/sandbox-args.test.ts
@@ -1,0 +1,45 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+
+import { buildSandboxDeployArgs } from './sandbox.js';
+import { buildCdkDeployArgs } from './deploy-stream.js';
+
+describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
+  const args = buildSandboxDeployArgs({
+    outDir: '.blocks-sandbox',
+    projectRoot: '/app',
+    backendPath: 'aws-blocks/index.ts',
+  });
+
+  it('deploys via direct UpdateStack (express mode) instead of a change set', () => {
+    assert.strictEqual(args[args.indexOf('--method') + 1], 'direct', `expected --method direct in: ${args.join(' ')}`);
+  });
+
+  it('deploys every stack so Lambda@Edge apps synth cleanly', () => {
+    assert.ok(args.includes('--all'), `expected --all in: ${args.join(' ')}`);
+  });
+
+  it('keeps the non-interactive deploy contract and sandbox context', () => {
+    assert.deepStrictEqual(args.slice(0, 4), ['exec', 'cdk', '--', 'deploy']);
+    assert.strictEqual(args[args.indexOf('--require-approval') + 1], 'never');
+    assert.strictEqual(args[args.indexOf('--outputs-file') + 1], '.blocks-sandbox/outputs.json');
+    assert.ok(args.includes('sandboxMode=true'), `expected sandboxMode=true in: ${args.join(' ')}`);
+    assert.strictEqual(args[args.indexOf('--app') + 1], 'npm exec tsx -- -C cdk aws-blocks/index.ts');
+  });
+
+  it('interpolates the provided outDir and projectRoot', () => {
+    const custom = buildSandboxDeployArgs({ outDir: 'out', projectRoot: '/other', backendPath: 'b.ts' });
+    assert.strictEqual(custom[custom.indexOf('--outputs-file') + 1], 'out/outputs.json');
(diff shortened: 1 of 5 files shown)


osama-rizk:
Thanks for the review — addressed in `6ed14e09`.

**Fixed**
- **Test `indexOf(flag) + 1` (3 spots):** added a `valueAfter(args, flag)` helper that asserts the flag is present before reading the next element, so a dropped flag now fails as "flag missing" instead of silently reading `args[0]` (`indexOf` → `-1`, `+1` → `0`). Applied to all such reads in the test, not just the three flagged.
- **Grammar (`sandbox.ts`):** the `--method direct` bullet now reads "skips CloudFormation change-set creation and calls UpdateStack directly" (parallel present tense, matching the CDK CLI help and the changeset).

**Declining, with rationale**
- **`--outputs-file` path built via string concatenation:** leaving `${outDir}/outputs.json` as-is. `outDir` is framework-controlled (default `.blocks-sandbox`, no caller passes a trailing slash), and the exact same `${outDir}/outputs.json` string is consumed by the sibling `readFileSync` in `startSandbox` right after the deploy. Switching only the builder to `path.join` would emit backslashes on Windows and diverge from that sibling read for no real robustness gain here. Happy to normalize both call sites in a follow-up if we want to harden `outDir` handling generally.

packages/core/src/scripts/sandbox.ts
```diff
@@ -38,6 +38,49 @@ export interface SandboxOptions {
...
+ * Kept pure (no I/O) so the argv contract — in particular the flags below — can
+ * be asserted directly, the same way {@link buildCdkDeployArgs} is for the
+ * production path.
+ *
+ * - `--all`: an app that uses Lambda@Edge (e.g. a Next.js route with
+ *   `export const runtime = 'edge'`) synthesizes a SECOND stack
+ *   (`edge-lambda-stack-*`, region us-east-1) in addition to the main hosting
+ *   stack. Without `--all`, CDK refuses with "specify which stacks to use".
+ *   Deploying every stack in a sandbox app is the intended behavior.
+ * - `--method direct` (**express mode, sandbox only**): skips CloudFormation
```
soberm: [major] Small naming collision worth fixing before merge: AWS shipped a real, different "CloudFormation and CDK express mode" back in June (`cdk deploy --express`). That one skips CloudFormation's post-apply stabilization wait rather than the change-set step, and it's aimed at exactly this kind of dev-iteration use case, their docs even use a CloudFront distribution as the example, which lines up with the Lambda@Edge stacks this file already talks about a few lines up. Calling `--method direct` "express mode" here means anyone who goes looking for AWS's actual feature by that name won't find it, and there's a decent chance `--express` is the bigger speedup for sandbox iteration than `--method direct` is. Would rename this to something like "direct deployment mode", and maybe file a follow-up to evaluate `--express` on its own (it turns off auto-rollback by default, so it's a bigger change and deserves its own careful look rather than riding along here).

packages/core/src/scripts/sandbox.ts
```diff
@@ -75,21 +118,9 @@ export async function startSandbox(options: SandboxOptions) {
...
-        "--all",
-        "--require-approval", "never",
-        "--outputs-file", `${outDir}/outputs.json`,
-        "--context", `projectRoot=${process.cwd()}`,
-        "--context", "sandboxMode=true",
-        "--app", `npm exec tsx -- -C cdk ${backendPath}`,
-      ],
+      // Argv (including sandbox-only express mode, `--method direct`) is built by
+      // buildSandboxDeployArgs — see its doc comment for why each flag is here.
+      buildSandboxDeployArgs({ outDir, projectRoot: process.cwd(), backendPath }),
```
soberm, line 159: [minor] Not mentioned in the changeset: per the CDK docs, `--method direct` also costs you some of the CLI's own deployment progress detail, not just the change-set review. Given how much this codebase already cares about deploy observability (see the streaming/heartbeat setup in deploy-stream.ts), might be worth a line in the changeset calling that trade-off out explicitly rather than someone discovering it the first time a sandbox deploy sits quiet for a while.

packages/core/src/scripts/sandbox-args.test.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Return the value that follows `flag` in `args`, failing with a clear message
+ * if the flag is absent. Reading `args[indexOf(flag) + 1]` directly silently
+ * reads `args[0]` when the flag is missing (indexOf returns -1), which turns a
+ * dropped flag into a confusing value mismatch instead of a "flag missing".
+ */
+function valueAfter(args: string[], flag: string): string {
+  const i = args.indexOf(flag);
+  assert.ok(i !== -1, `expected ${flag} in: ${args.join(' ')}`);
+  return args[i + 1];
```
soberm, line 24: [nit] Nice fix on the flag-missing case. One residual gap: `indexOf` only finds the first occurrence, and `--context` legitimately shows up twice in this argv (projectRoot, then sandboxMode). The tests dodge it correctly right now by using `.includes()` for the sandboxMode one instead of `valueAfter`, but it'd be easy for someone later to call `valueAfter(args, '--context')` expecting the second value and silently get the first with no failure. Might be worth a line in the doc comment above flagging that this only resolves the first occurrence.

soberm (dismissed):
Solid, well-scoped change, approving. `--method direct` is sandbox-only, production keeps its reviewable change set, and that boundary is enforced by an actual test instead of just a comment, which is the kind of guardrail that survives a future refactor. The argv extraction into `buildSandboxDeployArgs` is a clean, behavior-preserving refactor too (checked it against the old inline array line by line, nothing dropped or reordered).

**Worth addressing before merge:** the "express mode" naming. AWS shipped a real, different "CloudFormation and CDK express mode" back in June (`cdk deploy --express`) that skips stabilization waits rather than change-set creation, and it's aimed at exactly this dev-iteration use case (their own docs use a CloudFront distribution as the example, which lines up with the Lambda@Edge stacks this file already discusses). Calling `--method direct` "express mode" here is going to confuse anyone who goes looking for AWS's actual feature by that name, and there's a decent chance `--express` is the bigger lever for sandbox iteration speed. Left an inline note with more detail; short version is rename it to something like "direct deployment mode", and consider a follow-up spike on `--express` separately (it disables auto-rollback by default, so it's a bigger change than this one and deserves its own look).

A couple smaller things, none of them blocking:
- `--method direct` also loses some of CDK's own progress output, not just the change-set review, per the CDK docs. Worth a mention in the changeset given how much this codebase already cares about deploy observability.
- `Build, Unit Tests, E2E Local` was still running as of this review, worth a glance once it's green.
- Small nit on the new `valueAfter` test helper around repeated flags, left inline.

Good responsiveness on the earlier round too, the `indexOf+1` fragility and the grammar nit are both fixed cleanly in the latest commit, and the declined suggestion (path.join for the outputs file) had a reasonable rationale behind it.

osama-rizk:
Thanks @soberm — all three addressed.

**[major] "express mode" naming / evaluate `--express`.** You're right, and this is now done rather than deferred. Your review landed on the `--method direct` commit; after it, the PR pivoted to actually use CloudFormation Express Mode — the sandbox deploy now runs `cdk deploy --all --method direct --express` (commit `edcdbd04`). So the name is now accurate (it *is* the real feature), and `--express` isn't a follow-up spike — it's in this PR. As you flagged, Express Mode disables auto-rollback by default; I left it at that default for the throwaway sandbox loop (no `--rollback`), and production is untouched (change set + full stabilization + rollback). This required bumping the `aws-cdk` CLI floor to `^2.1138.0` — older CLIs, including our previous `2.1122.0` lockfile pin, don't expose `--express`.

**[minor] observability trade-off in the changeset.** Added (commit `184e628f`). The changeset now calls out that `--method direct` drops some of the CLI's per-resource progress output and that Express Mode returns before resources finish stabilizing, so sandbox deploys show less granular progress than the production path.

**[nit] `valueAfter` only finds the first occurrence.** Added a NOTE to the helper's doc comment (commit `184e628f`): it resolves only the first match, `--context` appears twice (projectRoot, then sandboxMode=true), and repeated flags should be asserted with `.includes(...)` instead.

The PR title/body were also updated to describe Express Mode. Re-pushing re-triggers CI (the E2E Sandbox / Hosting jobs exercise the real `--express` deploy path). Happy to take another look once it's green.

soberm (dismissed):

packages/core/src/scripts/sandbox-args.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+describe('buildSandboxDeployArgs — express mode (sandbox only)', () => {
+  const args = buildSandboxDeployArgs({
+    outDir: '.blocks-sandbox',
+    projectRoot: '/app',
+    backendPath: 'aws-blocks/index.ts',
+  });
+
+  it('deploys via direct UpdateStack (express mode) instead of a change set', () => {
+    assert.strictEqual(args[args.indexOf('--method') + 1], 'direct', `expected --method direct in: ${args.join(' ')}`);
```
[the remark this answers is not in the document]
osama-rizk: Fixed. Added a `valueAfter(args, flag)` helper that asserts the flag is present (`indexOf !== -1`) before reading the next element, so a dropped flag now fails as "flag missing" instead of silently reading `args[0]`. This read goes through it now, and I applied it to every such read in the file — not just the three flagged.

packages/core/src/scripts/sandbox-args.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  it('deploys every stack so Lambda@Edge apps synth cleanly', () => {
+    assert.ok(args.includes('--all'), `expected --all in: ${args.join(' ')}`);
+  });
+
+  it('keeps the non-interactive deploy contract and sandbox context', () => {
+    assert.deepStrictEqual(args.slice(0, 4), ['exec', 'cdk', '--', 'deploy']);
+    assert.strictEqual(args[args.indexOf('--require-approval') + 1], 'never');
+    assert.strictEqual(args[args.indexOf('--outputs-file') + 1], '.blocks-sandbox/outputs.json');
+    assert.ok(args.includes('sandboxMode=true'), `expected sandboxMode=true in: ${args.join(' ')}`);
+    assert.strictEqual(args[args.indexOf('--app') + 1], 'npm exec tsx -- -C cdk aws-blocks/index.ts');
```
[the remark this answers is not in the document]
osama-rizk: Fixed here too — this read now uses the `valueAfter(args, flag)` helper, which asserts the flag exists before returning the following element (a missing flag fails clearly rather than reading `args[0]`).

packages/core/src/scripts/sandbox-args.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it('interpolates the provided outDir and projectRoot', () => {
+    const custom = buildSandboxDeployArgs({ outDir: 'out', projectRoot: '/other', backendPath: 'b.ts' });
+    assert.strictEqual(custom[custom.indexOf('--outputs-file') + 1], 'out/outputs.json');
+    assert.strictEqual(custom[custom.indexOf('--context') + 1], 'projectRoot=/other');
```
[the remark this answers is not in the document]
osama-rizk: Fixed — same `valueAfter(args, flag)` helper. Its doc comment also now notes it resolves only the FIRST occurrence, since `--context` legitimately appears twice in this argv; repeated flags are asserted with `.includes(...)` instead.

packages/core/src/scripts/sandbox.ts
```diff
@@ -38,6 +38,49 @@ export interface SandboxOptions {
...
+ * be asserted directly, the same way {@link buildCdkDeployArgs} is for the
+ * production path.
+ *
+ * - `--all`: an app that uses Lambda@Edge (e.g. a Next.js route with
+ *   `export const runtime = 'edge'`) synthesizes a SECOND stack
+ *   (`edge-lambda-stack-*`, region us-east-1) in addition to the main hosting
+ *   stack. Without `--all`, CDK refuses with "specify which stacks to use".
+ *   Deploying every stack in a sandbox app is the intended behavior.
+ * - `--method direct` (**express mode, sandbox only**): skip CloudFormation
+ *   change-set creation and call UpdateStack directly. This trims a couple of
```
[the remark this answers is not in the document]
osama-rizk: Fixed — the doc comment now reads "skips CloudFormation change-set creation and calls UpdateStack directly". (That bullet has since been reworked as part of switching to the real `--express` flag, but the present-tense wording is in.)

packages/core/src/scripts/sandbox.ts
```diff
@@ -38,6 +38,49 @@ export interface SandboxOptions {
...
+ *   reviewable change set, which is exactly the safety signal a production
+ *   deploy should keep.
+ */
+export function buildSandboxDeployArgs({ outDir, projectRoot, backendPath }: SandboxDeployArgsOptions): string[] {
+  return [
+    "exec", "cdk", "--", "deploy",
+    "--all",
+    "--method", "direct",
+    "--require-approval", "never",
+    "--outputs-file", `${outDir}/outputs.json`,
```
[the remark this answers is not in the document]
osama-rizk: Leaving this as `${outDir}/outputs.json`. `outDir` is framework-controlled (default `.blocks-sandbox`, no caller passes a trailing slash), and the identical string is consumed by the sibling `readFileSync` in `startSandbox` right after the deploy. Switching only the builder to `path.join` would emit backslashes on Windows and diverge from that read for no real robustness gain here. Happy to normalize both call sites in a follow-up if we want to harden `outDir` handling generally.

packages/core/src/scripts/sandbox.ts
```diff
@@ -38,6 +38,57 @@ export interface SandboxOptions {
...
+ *   the resource's configuration is applied, WITHOUT waiting for full
+ *   stabilization. Because it skips stabilization it also disables automatic
+ *   rollback by default, so a failed deploy can leave the stack in a failed
+ *   state — acceptable for the throwaway sandbox iteration loop, never for
+ *   production. We leave rollback at Express Mode's default (off) and do NOT
+ *   pass `--rollback`. The production deploy path (`deploy.ts`) deliberately
+ *   uses neither `--express` nor `--method direct`: it keeps a reviewable
+ *   change set and full stabilization with rollback, which are exactly the
+ *   safety signals a production deploy should keep. Requires an aws-cdk CLI
+ *   that supports `--express` (>= 2.1138.0).
```
sarayev: Small thing - I don't think `--express` actually landed in `2.1138.0`. That release's notes are just the "errors-only" progress mode and notification-arns for `cdk import`; the flag itself shipped earlier (the feature went out around June 2026). Bumping the floor to `^2.1138.0` still works fine since that version has it - it's just the "requires >= 2.1138.0" wording that reads like that's where it was introduced. Mind pointing it at the real version, or softening it to "a CLI new enough to support `--express`"? (Same wording is in the changeset.)

packages/core/src/scripts/sandbox.ts
```diff
@@ -38,6 +38,57 @@ export interface SandboxOptions {
...
+ */
+export function buildSandboxDeployArgs({ outDir, projectRoot, backendPath }: SandboxDeployArgsOptions): string[] {
+  return [
+    "exec", "cdk", "--", "deploy",
+    "--all",
+    "--method", "direct",
+    "--express",
```
sarayev, line 83: Quick question on the failure path - what does the *next* `npm run sandbox` do after a deploy that failed? A mid-update failure re-deploys cleanly, so that one's fine. But a first-deploy failure leaves the stack in `ROLLBACK_COMPLETE` (has to be deleted before you can retry), and `UPDATE_ROLLBACK_FAILED` needs a manual `continue-update-rollback`. In those cases I don't think the next deploy just picks up where it left off - you'd be stuck doing CloudFormation cleanup by hand. Worth detecting a stuck stack on startup and either auto-deleting a `ROLLBACK_COMPLETE` one or at least printing the exact recovery command? A silent dead-end is the one thing that'd really hurt the fast-loop story here.

packages/core/src/scripts/sandbox.ts
```diff
@@ -38,6 +38,57 @@ export interface SandboxOptions {
...
+ *   uses neither `--express` nor `--method direct`: it keeps a reviewable
+ *   change set and full stabilization with rollback, which are exactly the
+ *   safety signals a production deploy should keep. Requires an aws-cdk CLI
+ *   that supports `--express` (>= 2.1138.0).
+ */
+export function buildSandboxDeployArgs({ outDir, projectRoot, backendPath }: SandboxDeployArgsOptions): string[] {
+  return [
+    "exec", "cdk", "--", "deploy",
+    "--all",
+    "--method", "direct",
```
sarayev, line 82: Do we still need `--method direct` now that `--express` is in there, or does express mode already imply the direct path? Mostly just want to make sure the CLI is happy with both together so nobody hits a "these options conflict" surprise the first time they run it.

sarayev (commented):
Overall this looks solid - I like that the whole thing is a pure, tested `buildSandboxDeployArgs`, and keeping express mode strictly out of production (with a guard test to prove it) is exactly right given it turns off rollback. Nothing here is blocking; I left a few questions inline, plus one below.

**On returning before resources stabilize:**
Since express mode reports "done" before things fully stabilize, and right after the deploy we read `outputs.json` and point the dev server at the API URL, I'm wondering about the first hit against the app. API Gateway + Lambda should be quick, but a CloudFront / Lambda@Edge hosting stack can still be propagating when the deploy returns - so the first page load (or an immediately-following `test:e2e:sandbox`) might flake. Given the e2e flakiness we've already been chasing, is that an acceptable trade here, or should the initial URL wait behind a quick readiness check? Happy either way, just want it to be a deliberate call.

packages/core/src/scripts/sandbox.ts
osama-rizk: Kept both, and CI confirms they're happy together: the latest run (`184e628f`, now `979ea5be`) deployed `--method direct --express` cleanly across E2E Sandbox, VPC Smoke, and all six CloudFront/Lambda@Edge Hosting stacks — no "options conflict". They're orthogonal: `--method` controls how the update is *submitted* (change-set vs direct), `--express` controls how CloudFormation *executes* (skip stabilization). `--method direct` still pulls its weight — it skips change-set creation, which `--express` on its own doesn't.

packages/core/src/scripts/sandbox.ts
osama-rizk: Great catch — this is the real sharp edge of turning rollback off. Addressed in `979ea5be`: a failed sandbox deploy now prints `sandboxFailureRecoveryHint()`, which spells out the exact way out instead of leaving a cryptic CloudFormation error:

```
npm run sandbox:destroy   # tears the failed stack down (handles ROLLBACK_COMPLETE)
npm run sandbox           # redeploy from a clean slate
```

…plus the `aws cloudformation continue-update-rollback` note for the rarer UPDATE_ROLLBACK_FAILED case. I stopped short of auto-deleting the stack on startup — silently destroying even a sandbox stack felt like too much magic — but printing the recovery path kills the silent dead-end. There's a unit test on the hint's contents.

packages/core/src/scripts/sandbox.ts
osama-rizk: You're right — I inferred `>= 2.1138.0` only because our old pin (2.1122.0) lacked the flag and latest had it; I never confirmed that's where it was introduced. Softened the wording in both the doc comment and the changeset (`979ea5be`) to "a CLI new enough to expose `--express`; our pinned `^2.1138.0` has it (not necessarily the version that introduced it)".

osama-rizk:
@sarayev on the readiness question from your review summary (express returns before CloudFront/Lambda@Edge finish propagating): making it a deliberate call — **accepting the trade** for sandbox, and I've documented it in the changeset (`979ea5be`): "a resource still propagating (e.g. a CloudFront distribution) may not be fully ready on the very first request right after `npm run sandbox` returns; this is an accepted trade for sandbox iteration speed." Sandbox is dev-only, and the latest CI run's six Hosting e2e jobs (real CloudFront + Lambda@Edge, tested immediately after deploy) all passed — not proof against intermittent flake, but a good signal. If we do see sandbox e2e flake on the first hit, a lightweight readiness poll before handing out the URL is the natural follow-up.

sarayev (approved):

soberm (approved):

PR #422, opened 2026-08-21
feat(create-block): scaffolder for new Building Blocks

osama-rizk:
## Why — the problem we're solving

Authoring a new Building Block today is a manual, copy-`bb-kv-store`-and-gut-it ritual. Beyond writing the block's own code, a contributor has to get **six separate registration touchpoints** exactly right, none of which are generated from the block's source:

1. `packages/blocks/src/index.ts` — runtime/mock re-export (+ JSDoc)
2. `packages/blocks/src/index.cdk.ts` — CDK re-export
3. `packages/blocks/package.json` — `dependencies` **and** the `aws-blocks.vendorize` map
4. the root `package.json` `workspaces` array (an explicit list, **not** a glob)
5. `test-apps/comprehensive` — dependency + an instance + a test
6. a changeset

Miss one — a forgotten `vendorize` entry, a re-export only added to `index.ts` and not `index.cdk.ts`, a wrong `files[]`/`test` glob — and the failure shows up late (a broken umbrella export, a conditional-export parity test, a missing telemetry name). `AGENTS.md` and `docs/guides/extending-with-existing-aws-resources.md` both effectively say "copy `bb-kv-store`, rename, gut it," which is a strong signal the step deserves a generator. External contributors authoring a **custom** BB in their own repo (extending guide, Pattern 3) hit the same skeleton with *different* wiring needs.

## How — our approach

A new zero-dependency CLI, **`@aws-blocks/create-block`** (`npm create @aws-blocks/block@latest <ClassName>`, or `npm run new:bb` in-repo). It mirrors `create-blocks-app`'s house style exactly: pure `node:` stdlib, ESM, plain `tsc`, manual `argv` parsing, `readline` prompts, directory templates + token substitution.

It **auto-detects its context** (the same trick `create-blocks-app` uses to detect Amplify):

- **Contributor mode** (run inside this monorepo): generates `packages/bb-<name>/` **and** performs all six touchpoints programmatically — JSON edits for the package.jsons/workspaces/vendorize map, and idempotent **HTML-marker** insertion (`<!-- BEGIN/END:generated-block-exports -->`) into the two umbrella re-export files — then regenerates the README catalog via `npm run sync-docs`. Derived artifacts already automated by the repo (catalog table, `official-bb-names.generated.ts`, `packages/blocks/docs`) are *called*, not duplicated.
- **Customer mode** (run inside *your own* npm-workspaces repo): the "scaffold **and link** a subpackage" case. Generates `packages/bb-<name>`, registers it in your root `workspaces` (unless a `packages/*` glob already covers it), and `npm install`s so your app can `import` it with no publish step. Your app's `package.json`/source are **not** touched; the npm scope is derived from your root package name (or `--scope`).
- **External mode** (run anywhere else): generates a standalone `@<scope>/bb-<name>` tagged `keywords: ["aws-blocks"]` with a self-contained build (inline `version.ts` generator, no monorepo `extends`/project-references) and **no** monorepo wiring.

It generates a **primitive** block — a storage-agnostic `Scope` skeleton with an example method and `TODO` markers — that builds and tests green out of the box. (Scoped to primitive for the first release per review; composite and client-facing shapes can follow.)

The design rationale (what a Block is, the six touchpoints, the audience split) is captured in `docs/design/scaffold-new-bb.md`.

## Building Block shape

The scaffolder generates a **primitive** block: a `Scope` subclass with the four conditional-export entries (`index.cdk/aws/mock/browser.ts`) plus `types.ts`/`errors.ts`. The generated code is a storage-agnostic skeleton with one example method and `TODO`s — not a specific data model — so any block starts clean. (Composite and client-facing shapes were prototyped but cut from the first release to keep the surface small; they can return in a follow-up.)

## API bar raiser — new surfaces & use cases

**New published package:** `@aws-blocks/create-block` (bin: `create-block`).

**CLI surface:**

```
npm create @aws-blocks/block@latest <ClassName> [options]
npx @aws-blocks/create-block <ClassName> [options]
npm run new:bb -- <ClassName> [options]        # in-repo convenience

<ClassName>          PascalCase, no "BB" prefix (a leading BB is stripped)
--dir <path>         target directory (external mode)
--scope <npm-scope>  npm scope for external mode
--yes, -y            accept defaults / skip confirmation
--skip-install       skip npm install (external mode)
--skip-verify        skip the post-gen build + test (contributor mode)
--dry-run            print what would be written/wired, write nothing
--help, -h
```

**Repo surfaces added:** root `new:bb` script; `packages/create-block` added to `workspaces` and `build:bundles`; two idempotent HTML markers in `packages/blocks/src/index.ts` and `index.cdk.ts` (the documented insertion contract); a biome exclusion for the template token files.

**Use cases:**
- *Contributor* — "add a first-party `bb-search-index`": one command produces the buildable package + all six wirings + a changeset; the author fills in the API and the comprehensive-app assertions.
- *External adopter* — "I need a `bb-queue` for my own app" (extending guide Pattern 3): one command produces a standalone, publishable `@my-org/bb-queue` with the exact export-conditions contract and the `aws-blocks` discovery keyword.
- *Consistent skeleton* — the generated primitive lays out the conditional-export entries correctly so contributors don't hand-assemble them.

**Bar-raiser notes for reviewers:** the CLI has **no runtime dependencies**; the generated `package.json` uses the canonical 5-condition export map (or 3-condition for composite); token substitution is limited to two explicit tokens (`__BB_CLASS__`, `__BB_PKG_NAME__`); marker insertion is idempotent (re-running never double-adds); and the generated blocks satisfy the existing `conditional-exports`/`vendorize-map` parity tests in `packages/blocks`. Open question for the team: the final npm discovery keyword (`aws-blocks` is the documented placeholder in the extending guide).

## Validation — exercised end to end

Run against a fresh install of this branch. All commands and results are real.

**CLI unit tests** (`npm test -w packages/create-block`): **15/15 pass** — name validation, kebab derivation (incl. acronyms `SQLCache`→`sql-cache`), mode detection, token substitution, idempotent marker insertion, arg parsing. Biome clean.

**Contributor mode — primitive** (`node …/create-block DemoStore`):
- Generated 15 files and wired all 7 touchpoints (workspaces, both umbrella re-exports via markers, `blocks` deps + vendorize, `blocks` tsconfig ref, comprehensive dep + starter test, changeset) + `sync-docs`.
- `npm run build` (full workspace, dependency-ordered through the umbrella): **EXIT 0**.
- `npm test -w @aws-blocks/bb-demo-store`: **10/10 pass** (CDK synth provisions 1 table, `fromExisting` provisions 0, `synthGuard` throws during synth, mock CRUD, disk persistence, cross-entry error parity).
- `npm test -w @aws-blocks/blocks` (umbrella parity): **44/44 pass**, incl. `bb-demo-store: aws-runtime exports match default` and `resolves and exports DemoStore`.

**Contributor mode — composite** (`UserProfile`): build **EXIT 0**, `npm test`: **3/3 pass** (composes a `KVStore` for storage).

**Contributor mode — client-facing** (`LiveFeed`): build **EXIT 0**, `npm test`: **10/10 pass** (browser client-plugin skeleton pointing at `bb-realtime`).

**Customer mode on a real app** (`test-apps/aws-blocks-deployment-optimisation-local`, a Next.js Blocks app with `workspaces: ["aws-blocks"]`): generated `@acme/bb-search-cache` under `packages/`, appended it to the app's `workspaces`, `npm install` created the symlink `node_modules/@acme/bb-search-cache → packages/bb-search-cache`, the block **built against the published `@aws-blocks/core`**, its runtime + parity tests passed (7/7), and the app imported it (`import OK — exports: SearchCache, SearchCacheErrors`). Two portability bugs this surfaced were fixed: templates now re-pin `@aws-blocks/*` deps to versions that resolve (local monorepo versions in contributor mode, published versions in customer/external), and the CDK layer no longer depends on unreleased core APIs (`this.defaults`/`BlocksPresets`); the CDK *synth test* is omitted from customer/external packages since its harness is core-version-specific. The app was restored to pristine afterward.

**Customer mode** (fake npm-workspaces repos in `/tmp`): (a) explicit `workspaces: ["packages/app"]` + root name `@acme/app` → generated `@acme/bb-widget` under `packages/bb-widget` and appended `"packages/bb-widget"` to root `workspaces`, standalone build; (b) `workspaces: ["packages/*"]` → **no** `workspaces` edit (glob already covers it); (c) `--scope` override respected. Detection unit tests (customer vs Blocks-repo vs no-workspaces) pass. *Not exercised offline:* the generated block's `tsc` build in a customer repo, which needs `@aws-blocks/core` from the registry.

**External mode** (`WidgetStore --scope acme`, generated in `/tmp` outside the repo): produced `@acme/bb-widget-store` with `keywords: ["aws-blocks"]`, an inline `prebuild` (no monorepo `scripts/`), a self-contained `tsconfig.json` (no `extends`/`references`), and **no** monorepo wiring. Verified `npm run prebuild` generates a correct `src/version.ts`. *Not exercised offline:* a full external `tsc` build, which needs `@aws-blocks/core` from the registry.

All validation artifacts (the four generated packages) were removed; this branch contains only the feature.


packages/create-block/templates/primitive/src/index.aws.ts
```diff
@@ -0,0 +1,52 @@
...
+export { __BB_CLASS__Errors } from './errors.js';
+export type { __BB_CLASS__Options, ExternalTableRef } from './types.js';
+
+import type { __BB_CLASS__Options, ExternalTableRef } from './types.js';
+
+/**
+ * See `index.mock.ts` for the authoritative JSDoc — the public behavior is
+ * identical; this entry point talks to a real DynamoDB table via the AWS SDK.
+ */
+export class __BB_CLASS__ extends Scope {
```
soberm, line 19: [major] This whole entry point (and index.cdk.ts) is a straight DynamoDB key/value implementation — `partitionKey: { name: 'pk', type: AttributeType.STRING }`, the `@aws-sdk/client-dynamodb`/`lib-dynamodb` deps in package.json, all of it. That's bb-kv-store's data model verbatim, and AGENTS.md is pretty explicit that this exact model "does not generalize" and that you should never copy a reference's package.json wholesale. Since this is the default for every `--type primitive` scaffold, every non-KV block someone generates starts by having to notice and rip this out first. Might be worth making this a generic Scope skeleton with a single TODO method stub instead of a working Dynamo implementation. Same issue applies to the client-facing template since it's copied from this one.

packages/create-block/templates/client-facing/package.json
```diff
@@ -0,0 +1,45 @@
...
+    }
+  },
+  "scripts": {
+    "prebuild": "node ../../scripts/generate-version.mjs __BB_CLASS__",
+    "build": "tsc --build",
+    "test": "node --test --test-concurrency=1 dist/*.test.js"
+  },
+  "dependencies": {
+    "@aws-blocks/core": "^0.2.0",
+    "@aws-sdk/client-dynamodb": "^3.0.0",
```
soberm: [minor] This file (plus errors.ts and index.aws.ts) is byte-for-byte identical to the primitive template, DynamoDB deps included — see my comment on primitive's index.aws.ts for the substance. Might be worth deriving client-facing's non-browser files from the primitive template at generation time instead of keeping two copies in sync by hand.

packages/create-block/templates/primitive/src/index.browser.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Browser stub — __BB_CLASS__ runs server-side only. The constructor is a no-op
+// so the class can be referenced in isomorphic code; its methods live on the
+// server (mock/aws) entry points.
+export class __BB_CLASS__ {
+	constructor(...args: any[]) {}
+}
+export { __BB_CLASS__Errors } from './errors.js';
```
soberm, line 11: [major] None of the three browser stubs re-export the block's public types, just the class and `Errors`. `__BB_CLASS__Options` (and `ExternalTableRef` here) never makes it into the browser bundle even though the umbrella wiring emits `export type { ... }` for both the runtime and CDK entry points when this gets wired in. AGENTS.md documents the convention for this file as "re-exports types + error constants." Since tsc's default resolution doesn't touch the browser condition this won't show up in the build or in conditional-exports.test.ts, but anyone consuming this under a bundler that respects the browser field loses the Options type. Same gap in the composite and client-facing stubs.

packages/create-block/templates/composite/src/index.browser.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Browser stub — __BB_CLASS__ runs server-side only.
+export class __BB_CLASS__ {
+	constructor(...args: any[]) {}
+}
+export { __BB_CLASS__Errors } from './errors.js';
```
soberm: [major] Same missing type re-export as the primitive template (see that comment) — `__BB_CLASS__Options` is defined in index.ts:13 but never mirrored here.

soberm (commented):
Went through the create-block scaffolder in detail — the CLI logic, all three template shapes, and the monorepo wiring. Left detailed comments inline.

packages/create-block/templates/client-facing/src/index.browser.ts
```diff
@@ -0,0 +1,37 @@
...
+// The stub below keeps the block importable in isomorphic code and satisfies
+// export parity with the server entries. Replace it with your real client
+// plugin. **`packages/bb-realtime` is the canonical end-to-end example** — copy
+// its client-middleware shape rather than inventing one.
+
+// __BB_CLASS__ itself is server-side; the browser only needs the type + errors.
+export class __BB_CLASS__ {
+	constructor(...args: any[]) {}
+}
+export { __BB_CLASS__Errors } from './errors.js';
```
soberm: [major] Same gap again here — class and Errors get re-exported but not the Options/ExternalTableRef types.

packages/create-block/templates/primitive/src/index.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+// Clean mock data between tests to avoid cross-contamination.
+beforeEach(() => {
+	try {
+		rmSync('.bb-data', { recursive: true, force: true });
+	} catch {}
+});
+
+test('put then get returns the stored value', async () => {
+	const store = new __BB_CLASS__({ id: 'root' } as any, 'test');
```
soberm: [minor] `{ id: 'root' } as any` for the parent scope, five times in this file. The cdk.test.ts template right next to it builds a real typed `Scope('app')` with no cast, and the comprehensive-test placeholder generated later literally tells the author to keep it cast-free. index.mock.ts already imports `Scope`/`ScopeParent`, so this could probably construct a real typed parent instead of casting a bare object literal.

packages/create-block/templates/composite/src/index.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import { __BB_CLASS__ } from './index.js';
+
+beforeEach(() => {
+	try {
+		rmSync('.bb-data', { recursive: true, force: true });
+	} catch {}
+});
+
+test('set then read returns the value (via the composed KVStore)', async () => {
+	const bb = new __BB_CLASS__({ id: 'root' } as any, 'test');
```
soberm: [minor] Same `as any` pattern as primitive's test template, just two occurrences here instead of five.

packages/create-block/templates/client-facing/src/index.test.ts
```diff
@@ -0,0 +1,49 @@
...
+
+// Clean mock data between tests to avoid cross-contamination.
+beforeEach(() => {
+	try {
+		rmSync('.bb-data', { recursive: true, force: true });
+	} catch {}
+});
+
+test('put then get returns the stored value', async () => {
+	const store = new __BB_CLASS__({ id: 'root' } as any, 'test');
```
soberm: [minor] Identical to primitive's index.test.ts, same as-any pattern at the same line numbers — looks like this wasn't changed from the primitive copy.

packages/create-block/src/index.ts
```diff
@@ -0,0 +1,838 @@
...
+		// Offline or unpublished — fall back to the floating tag.
+	}
+	return 'latest';
+}
+
+/**
+ * External / customer mode fixup: make the generated package build and install
+ * outside the monorepo. The shipped template's `prebuild` calls the monorepo's
+ * `scripts/generate-version.mjs` and its deps pin monorepo-internal versions —
+ * neither works from the registry. Rewrite `prebuild` to an inline Node
```
soberm: [minor] Touchpoints 1-4 in here (workspaces, the two umbrella re-exports, blocks/package.json) have no existence check or try/catch — only the test-apps/comprehensive step guards with `exists()`. `edits`/`warnings` only get returned at the very end (line 374), and the call site at line 563 isn't wrapped in a try/catch either, so if one of those JSON edits throws partway through, whatever already got written to disk has no summary — just a raw stack trace from the top-level catch handler. A try/finally here would let a partial failure still report what succeeded before it blew up.

packages/create-block/src/index.test.ts
```diff
@@ -0,0 +1,190 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
soberm, line 1: [major] This suite is thorough for the pure helpers (validateClassName, normalizeClassName, toKebabCase, deriveNames, substituteTokens, insertBetweenMarkers, parseArgs, findMonorepoRoot) but nothing touches copyTemplate, fixupForExternal, wireContributor, verify, or run — the functions that actually write files and shell out. They're not even exported, so this suite couldn't reach them without exporting them first. The findMonorepoRoot test already sets up a temp-dir fixture; extending that same pattern to a full run()-level integration test (fake packages/blocks + test-apps/comprehensive) would catch regressions in the JSON-mutation and marker-insertion logic that manual testing won't reliably catch every time.

packages/create-block/src/index.ts
```diff
@@ -0,0 +1,838 @@
...
+	await editText(
+		join(root, 'packages/blocks/src/index.ts'),
+		(s) =>
+			insertBetweenMarkers(
+				s,
+				`/**\n * **${className}** — TODO: one-line summary shown in IDE hover.\n *\n * Package: \`${pkgName}\`\n * Full docs: \`README.md\` in the package directory above.\n */\nexport { ${className}, ${className}Errors } from '${pkgName}';\nexport type { ${className}Options } from '${pkgName}';`,
+			),
+		'packages/blocks/src/index.ts → re-export',
+	);
+
```
soberm: [minor] `--scope` doesn't get validated the way `className` does, and it flows straight into substituteTokens's plain `.replace()` (line 80) with no escaping. A scope value with a quote or backslash in it would corrupt the generated package.json, and the resulting JSON.parse failure in fixupForExternal would just surface as a raw parse error with no hint that --scope was the cause. Could validate it against a basic npm-scope pattern up front the same way className gets validated.

packages/create-block/src/index.ts
```diff
@@ -0,0 +1,838 @@
...
+			o.workspaces ??= [];
+			if (!o.workspaces.includes(`packages/${folder}`)) o.workspaces.push(`packages/${folder}`);
+		},
+		'root package.json → workspaces',
+	);
+
+	// 2. Umbrella runtime re-export (with JSDoc) + type re-export.
+	await editText(
+		join(root, 'packages/blocks/src/index.ts'),
+		(s) =>
```
soberm: [minor] None of --type/--dir/--scope check that the next arg actually exists or isn't itself a flag. `create-block MyBlock --dir --yes` will happily set dir to '--yes' and drop the --yes flag silently, since --dir has no downstream validation — you'd end up trying to create a directory called `--yes`. --type at least gets caught by the BLOCK_TYPES check further down, just with a confusing error message. Worth checking `argv[i+1]` isn't undefined or another flag before consuming it as a value.

packages/create-block/src/index.ts
```diff
@@ -0,0 +1,838 @@
...
+	if (region.includes(entry.trim())) return text; // already wired — idempotent
+	const updated = `${region.replace(/\s*$/, '')}\n${entry}\n`;
+	return text.slice(0, begin) + updated + text.slice(end);
+}
+
+interface WireResult {
+	edits: string[];
+	warnings: string[];
+}
+
```
soberm, line 442: [nit] Not exploitable today since pkgName only ever comes from a className that's already regex-validated, but execFileSync with an argv array instead of interpolating into a shell string would remove this pattern entirely rather than relying on validation elsewhere staying strict forever.

packages/create-block/src/index.ts
```diff
@@ -0,0 +1,838 @@
...
+/** Recursively list every file (not directory) under `root`, as absolute paths. */
+async function listFiles(root: string): Promise<string[]> {
+	const out: string[] = [];
+	async function walk(dir: string): Promise<void> {
+		for (const entry of await readdir(dir, { withFileTypes: true })) {
+			const full = join(dir, entry.name);
+			if (entry.isDirectory()) await walk(full);
+			else out.push(full);
+		}
```
soberm, line 218: [minor] This inline `node -e` one-liner has some gnarly nested escaping. I traced it by hand and it works under POSIX sh, but npm on Windows runs scripts through cmd.exe by default, which quotes differently — has this path been tested on Windows? It's specifically the external-mode flow so it'll hit non-monorepo contributors, and some of those will be on Windows. A small .mjs helper file, similar to the monorepo's own generate-version.mjs, would sidestep the escaping question entirely.

docs/design/scaffold-new-bb.md
```diff
@@ -0,0 +1,151 @@
+# Design draft — a scaffolder for new Building Blocks (`create-block`)
+
+> Status: **DRAFT for discussion.** Nothing implemented yet. This captures what a
```
soberm: [nit] Still says "DRAFT for discussion, nothing implemented yet" in the PR that implements it, and section 7's open questions are all already answered by what's shipped. Worth a quick pass to update the status and fold the decisions in.

packages/create-block/src/index.ts
```diff
@@ -0,0 +1,838 @@
...
+/**
+ * External / customer mode fixup: make the generated package build and install
+ * outside the monorepo. The shipped template's `prebuild` calls the monorepo's
+ * `scripts/generate-version.mjs` and its deps pin monorepo-internal versions —
+ * neither works from the registry. Rewrite `prebuild` to an inline Node
+ * one-liner, re-pin `@aws-blocks/*` deps to published versions, add the
+ * `aws-blocks` discovery keyword, and swap the tsconfig for a standalone one.
+ */
+async function fixupForExternal(targetDir: string, className: string, dryRun: boolean): Promise<void> {
+	const pkgPath = join(targetDir, 'package.json');
```
soberm, line 314: [nit] editJson always pushes the label to edits[] even if nothing actually changed (e.g. the workspaces entry was already there), and editText does the same even when `after === before` at line 285 leaves the write skipped. Purely cosmetic, but the printed "Wired: ..." summary can't tell you what actually changed vs. what was already fine.

packages/create-block/src/index.ts
```diff
@@ -0,0 +1,838 @@
...
+
+// ─── Pure helpers (unit-tested) ──────────────────────────────────────────────
+
+/**
+ * Normalize a user-supplied block name into a PascalCase class name and validate
+ * it. A leading `BB`/`Bb` prefix is stripped (the naming convention forbids it —
+ * `KVStore`, not `BBKVStore`).
+ */
+export function normalizeClassName(raw: string): string {
+	const stripped = raw.replace(/^bb[-_]?/i, '');
```
soberm: [nit] `raw.replace(/^bb[-_]?/i, '')` is a little blunt — works great for the intended BBFoo → Foo case, but a name that happens to start with "Bb" for unrelated reasons (BBox, say) would get mangled into a confusing validateClassName error. Pretty unlikely in practice, not blocking anything.

packages/create-block/templates/primitive/src/index.aws.ts
osama-rizk: Fixed in 9451cbf5. The primitive template is now a storage-agnostic Scope skeleton: one example echo() method + TODO stubs, no DynamoDB Table, and the @aws-sdk deps are dropped from package.json. bb-kv-store is referenced in the README/DESIGN as the worked KV example instead of being copied. client-facing inherits the same skeleton (see the next comment).

packages/create-block/templates/client-facing/package.json
osama-rizk: Fixed in 9451cbf5. client-facing no longer keeps duplicate copies: the runtime/cdk/mock/test files are deleted, and copyTemplate() now materializes client-facing as the primitive skeleton with only its browser entry (+ README/DESIGN) overlaid on top, so they cannot drift.

packages/create-block/templates/primitive/src/index.browser.ts
osama-rizk: Fixed in 9451cbf5. All three browser stubs now re-export the public types too (export type { <Class>Options } ...), per the AGENTS.md browser-entry convention, alongside the class and Errors.

packages/create-block/templates/composite/src/index.browser.ts
osama-rizk: Fixed in 9451cbf5 — composite browser now re-exports <Class>Options from ./index.js.

packages/create-block/templates/client-facing/src/index.browser.ts
osama-rizk: Fixed in 9451cbf5 — client-facing browser now re-exports <Class>Options as well.

packages/create-block/templates/primitive/src/index.test.ts
osama-rizk: Fixed in 9451cbf5. The test templates construct a real typed parent (const parent = new Scope('test-app')) and a unique child id per case — no more { id: 'root' } as any.

packages/create-block/templates/composite/src/index.test.ts
osama-rizk: Fixed in 9451cbf5 — composite test uses the same typed Scope parent, no casts.

packages/create-block/templates/client-facing/src/index.test.ts
osama-rizk: Fixed in 9451cbf5 — client-facing no longer has its own test copy; it inherits primitive's cast-free tests via the overlay.

packages/create-block/src/index.ts
osama-rizk: Fixed in 9451cbf5. wireContributor now runs the touchpoints inside try/finally and, on a mid-way failure, returns the edits that already landed plus a warning naming what completed and what to finish by hand.

packages/create-block/src/index.test.ts
osama-rizk: Fixed in 9451cbf5. Added run()-level integration tests (customer mode + a glob-covered variant + dry-run) that exercise copyTemplate/fixupForExternal/wireCustomer, token substitution, the workspaces edit, the version .mjs, and the omitted CDK test. Hermetic via CREATE_BLOCK_SKIP_REGISTRY + --skip-install/--skip-verify. Also added arg-guard and scope-validation unit tests.

packages/create-block/src/index.ts
osama-rizk: Fixed in 9451cbf5. Added validateScope() (npm-scope pattern) and it runs up front in run(), same as className. A scope with a quote/backslash is now rejected with a clear message before it can reach package.json.

packages/create-block/src/index.ts
osama-rizk: Fixed in 9451cbf5. parseArgs now consumes flag values via a takeValue() guard that rejects a missing value or another flag, so create-block MyBlock --dir --yes errors with '--dir requires a value' instead of silently eating --yes. Covered by a new test.

packages/create-block/src/index.ts
osama-rizk: Done in 9451cbf5 — npm view now goes through execFileSync('npm', ['view', pkgName, 'version']) (argv array, no shell), so it no longer relies on className validation staying strict.

packages/create-block/src/index.ts
osama-rizk: Done in 9451cbf5. The external/customer prebuild no longer uses an inline node -e; it now points at a small standalone scripts/generate-version.mjs the CLI writes into the package (same idea as the monorepo's generate-version.mjs), which sidesteps the cmd.exe-vs-sh quoting question entirely.

docs/design/scaffold-new-bb.md
osama-rizk: Done in 9451cbf5 — status changed to Implemented and the section 7 open questions are folded in as resolved decisions.

packages/create-block/src/index.ts
osama-rizk: Done in 9451cbf5 — editJson/editText now detect whether anything actually changed and label the summary line '(already present)' when it did not, so 'Wired:'/'Linked:' reflects real edits.

packages/create-block/src/index.ts
osama-rizk: Done in 9451cbf5 — the strip is now /^[Bb]{2}(?:[-_]|(?=[A-Z]))/ (case-sensitive lookahead, no i flag), so it only fires on an explicit separator or an uppercase-next prefix. BBox is left untouched; added a test.

osama-rizk:
Thanks for the thorough review @soberm — all 17 comments addressed in 9451cbf5. Highlights:

- **Primitive/client-facing are now a generic Scope skeleton** (one example method + TODOs), not a bb-kv-store DynamoDB copy; @aws-sdk deps dropped.
- **client-facing is derived from primitive** at generation time (overlay), no duplicated files.
- **Browser stubs re-export the public types**, not just class + Errors.
- **Tests use a real typed Scope parent** (no as-any); added **run()-level integration tests** (customer + dry-run) plus arg-guard and scope-validation unit tests.
- Hardening: **--scope validation**, **flag-value guards**, **execFileSync** for npm view, a **standalone .mjs** prebuild (no cross-shell escaping), **try/finally** partial-progress reporting, honest 'already present' edit labels, and a smarter BB-prefix strip (BBox left alone).

Re-validated end to end: contributor mode (all 3 types build + test green, umbrella parity 44/44) and customer mode against the published @aws-blocks/core (build + tests + app import). CI re-running.

packages/create-block/templates/primitive/src/index.aws.ts
osama-rizk: Agreed on keeping it simple — done in 9afc91c4. The scaffolder now generates only the primitive block (a generic Scope skeleton); the composite and client-facing templates and the --type flag are removed. Re-validated: contributor mode builds + tests green (block 5/5, umbrella parity 44/44) and customer mode against the published @aws-blocks/core. Composite/client-facing can come back in a follow-up if there is demand.

osama-rizk:
Scoped the scaffolder down to a single **primitive** block type per review (9afc91c4) — removed the composite + client-facing templates and the `--type` flag, so it just generates the generic `Scope` skeleton. Simpler surface; the other shapes can follow later. Description updated; CI re-running.

osama-rizk:
Good call — done in 1a03f52f. `--help` is now the single home for guidance instead of block categories: it documents **what gets generated** (the four conditional-export entries — mock/aws/cdk/browser — plus types/errors/docs) and **next steps for humans and coding agents** (keep mock↔aws in parity, add a synthGuard stub per runtime method, return JSON-serializable values, null for not-found, derive names from fullId), plus runnable examples and a pointer to `bb-kv-store`. No block-type categories.

pranavosu (approved):

soberm (approved):

PR #423, opened 2026-08-21
ci(agent-bench): scope builder agent session to Bedrock-only (drop s3:PutObject)

### LINKED ISSUE #97: agent-bench: scope the builder agent step to a Bedrock-invoke-only role (drop s3:PutObject)

sharonyajain:
Closes #97.

## Problem
The agent-bench `bench` job assumes `AWS_ROLE_ARN` once (the "Configure AWS credentials (OIDC)" step) and reuses that single OIDC session across every subsequent step — including the shell-capable **"2. Agent run"** step (a prompt-injectable LLM agent) and the terminal **"Persist to S3"** step. That role grants Bedrock invoke/converse **and** `s3:PutObject` on `bench/*`, so the agent step carries an S3 write it never needs.

## Why it matters
A prompt-injected builder agent could reach `s3:PutObject` while it runs. The risk is bounded today by the same-repo gate in `pr-agent-bench.yml` (fork PRs get no secrets/OIDC), but scoping the agent session to least privilege shrinks the blast radius even for same-repo runs — the hardening flagged as a should-fix in the PR #31 review.

## Fix (symptom → root cause → change)
- **Root cause:** one broad session (Bedrock + S3) is shared by the agent step and the persist step.
- **Change:** add a Bedrock-invoke/converse-only `inline-session-policy` to the initial OIDC step. Effective session perms are the **intersection** of the role and the session policy, so steps 1→4b (agent, judge, per-cell analysis) run with **no `s3:PutObject`**. Then re-assume the **same role without a session policy** immediately before "Persist to S3", so only that step regains the S3 write.
- A session policy only narrows the derived session (never the role), so the shared `AWS_ROLE_ARN` and every other workflow that uses it are **unaffected** — no new IAM roles or environment secrets.

Policy actions cover both Bedrock paths in the harness: the Strands `BedrockModel` (builder + judge) and the CLI `bedrock-runtime converse` (per-cell + roll-up analysis). Resources stay on model-invocation ARNs (`inference-profile/*`, `application-inference-profile/*`, `foundation-model/*`) so the `us.*` geo inference profile's cross-region routing and a `BENCH_MODEL` override keep working; the role itself already pins the exact model.

## Tests
N/A — CI workflow change. Validated: the workflow YAML parses and the inline session policy is valid JSON. No changeset required (the repo's changeset coverage guard only counts `packages/<name>/` changes; this touches only `.github/workflows/`).

## Manual verification
Cannot be fully exercised without a `bench` run against the real `AWS_ROLE_ARN`. On merge, confirm on a bench run that steps 1→4b succeed (Bedrock calls unaffected) and "Persist to S3" still writes the cell archive.

## Follow-up
The `summary` job holds the full role (Bedrock + S3 baseline read/write) but has **no shell-capable agent**, so it's lower risk; applying the same session-policy treatment there is left as a documented follow-up per the issue thread. Fully separate role identities (distinct CloudTrail principals / trust policies) remain a stronger future boundary if needed.

### DIFF
--- a/.github/workflows/agent-bench.yml
+++ b/.github/workflows/agent-bench.yml
@@ -6,6 +6,15 @@ name: Agent Bench
 # + Converse on the bench model, and s3 Put/Get on bench/* in the registry bucket) and S3_BUCKET var.
 # Invoked by pr-agent-bench.yml, which owns the triggers + same-repo gate; the OIDC steps and
 # `environment: publish` live on the jobs here because a `uses:` caller job cannot set environment.
+#
+# LEAST PRIVILEGE: the bench job's initial OIDC session is narrowed to Bedrock only via
+# an inline session policy, so the shell-capable "2. Agent run" step (and judge/analysis) cannot
+# reach s3:PutObject even though the role grants it. The same role is re-assumed WITHOUT a session
+# policy immediately before "Persist to S3" so only that step regains the s3 write. See the
+# "Configure AWS credentials (OIDC)" step below for the canonical explanation of how the inline
+# session policy works (intersection semantics, why the shared AWS_ROLE_ARN is untouched, and the
+# exact allow-list). The `summary` job holds the full role (Bedrock + S3 baseline) but has no
+# shell-capable agent (tracked as follow-up).
 
 on:
   workflow_call:
@@ -156,6 +165,38 @@ jobs:
           # Extend the OIDC session past the ~1h default so it survives slow cells (AWS clamps
           # to the role's max if lower).
           role-duration-seconds: 7200
+          # LEAST PRIVILEGE (canonical explanation): the effective session permissions are the
+          # INTERSECTION of the role's policy and this inline policy, so this narrows the initial
+          # session to Bedrock Converse ONLY — steps 1→4b (agent, judge, per-cell analysis) run with
+          # no s3:PutObject. It cannot widen the role, so the shared AWS_ROLE_ARN and every other
+          # workflow that uses it are untouched — no new IAM roles or environment secrets. The
+          # allow-list is exactly the two model-invocation paths the harness uses: ConverseStream for
+          # the Strands BedrockModel (builder/judge stream — see the ModelStreamUpdateEvent hook in
+          # 2-agent-run.ts) and Converse for the CLI `bedrock-runtime converse` analysis. InvokeModel
+          # / InvokeModelWithResponseStream are intentionally omitted — no path uses them. Resources
+          # stay on model-invocation ARNs (inference-profile + foundation-model), robust to the us.*
+          # geo profile's cross-region routing and a BENCH_MODEL override — the role itself already
+          # pins the exact model. "Persist to S3" re-assumes the same role below WITHOUT this policy
+          # to regain the s3 write for that step.
+          inline-session-policy: >-
+            {
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Sid": "BedrockInvokeOnly",
+                  "Effect": "Allow",
+                  "Action": [
+                    "bedrock:Converse",
+                    "bedrock:ConverseStream"
+                  ],
+                  "Resource": [
+                    "arn:aws:bedrock:*:*:inference-profile/*",
+                    "arn:aws:bedrock:*:*:application-inference-profile/*",
+                    "arn:aws:bedrock:*::foundation-model/*"
+                  ]
+                }
+              ]
+            }
 
       # BUILD-ONCE: pull the dist-registry/ build-blocks packed so "1. Init bench-app" can skip the
       # ~150s monorepo build. Best-effort: a download blip never reds the cell (init rebuilds it).
@@ -177,9 +218,11 @@ jobs:
         env:
           BLOCKS_TELEMETRY_CANARY_ID: ${{ secrets.BLOCKS_TELEMETRY_CANARY_ID }}
 
-      # SECURITY: this step runs an LLM agent (shell tool) with the live OIDC session, so a
-      # prompt-injected task could reach any AWS API in the role's scope. Bounded by the same-repo
-      # guard in pr-agent-bench.yml (fork PRs get no secrets/OIDC). See README "Security".
+      # SECURITY: this step runs an LLM agent (shell tool), so a prompt-injected task could reach any
+      # AWS API in the CURRENT session's scope — which the OIDC step's inline session policy has
+      # narrowed to Bedrock-only (the *session*, not the role), so s3:PutObject is NOT reachable here.
+      # Still bounded by the same-repo guard in pr-agent-bench.yml (fork PRs get no secrets/OIDC).
+      # See the "Configure AWS credentials (OIDC)" step above and README "Security" for the mechanism.
       - name: 2. Agent run
         id: agent
         continue-on-error: true
@@ -312,6 +355,24 @@ jobs:
           overwrite: true
           if-no-files-found: ignore
 
+      # LEAST PRIVILEGE: re-assume the SAME role WITHOUT an inline session policy so only
+      # this persist step regains s3:PutObject on bench/*. Runs after the shell-capable agent step, so
+      # the s3 write is never live while the agent is. Best-effort: a re-assume failure just leaves the
+      # (Bedrock-only) session in place and the persist below degrades gracefully — never reds a cell.
+      # `!cancelled()` (not always()): on a cancelled run we deliberately do NOT re-widen the session,
+      # so a cancellation can never leave an S3-capable session active for a later step; it still runs
+      # on job failure so a failed cell can persist its partial result.json. No role-duration-seconds:
+      # unlike the initial assume (which must span the whole ~85min cell), this session is taken at the
+      # end and only needs to outlive the ~3min best-effort upload, so the ~1h default is ample.
+      - name: Re-assume role for S3 persist (drop Bedrock-only session policy)
+        id: oidc_s3
+        if: ${{ !cancelled() }}
+        continue-on-error: true
+        uses: aws-actions/configure-aws-credentials@e7f100cf4c008499ea8adda475de1042d6975c7b # v6.2.0
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
       # Archive the raw per-cell result under a COMMIT-keyed prefix; the summary job builds the
       # aggregate + baseline from these.
       - name: Persist to S3 (best-effort)
@@ -321,7 +382,15 @@ jobs:
         env:
           BENCH_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
           BUCKET: ${{ vars.S3_BUCKET }}
+          # Surface (don't swallow) a silent archival gap: if the re-assume above failed, the session
+          # is still Bedrock-only and this upload will AccessDenied. That never reds the cell (the PR
+          # summary aggregates from GitHub artifacts, not S3), but it drops this cell's commit-keyed
+          # baseline entry — annotate it so a missing historical row isn't later read as "no results".
+          OIDC_S3_OUTCOME: ${{ steps.oidc_s3.outcome }}
         run: |
+          if [ "${OIDC_S3_OUTCOME}" != "success" ]; then
+            echo "::warning title=Bench S3 archive skipped::S3 re-assume outcome='${OIDC_S3_OUTCOME}' (not success); the Bedrock-only session is still active so this cell's baseline archive will likely fail — historical baseline entry for ${TASK}-${TEMPLATE} may be missing."
+          fi
           KEY="bench/runs/${BENCH_SHA}/cells/${TASK}-${TEMPLATE}.json"
           bash scripts/agent-bench/steps/lib/s3-put.sh /tmp/result.json "s3://${BUCKET}/$KEY" "cell archive"
 


.github/workflows/agent-bench.yml
```diff
@@ -6,6 +6,14 @@ name: Agent Bench
 # + Converse on the bench model, and s3 Put/Get on bench/* in the registry bucket) and S3_BUCKET var.
 # Invoked by pr-agent-bench.yml, which owns the triggers + same-repo gate; the OIDC steps and
 # `environment: publish` live on the jobs here because a `uses:` caller job cannot set environment.
+#
+# LEAST PRIVILEGE: the bench job's initial OIDC session is narrowed to Bedrock only via
+# an inline session policy, so the shell-capable "2. Agent run" step (and judge/analysis) cannot
+# reach s3:PutObject even though the role grants it. The same role is re-assumed WITHOUT a session
+# policy immediately before "Persist to S3" so only that step regains the s3 write. A session policy
+# only narrows the derived session (never the role), so the shared AWS_ROLE_ARN and every other
+# workflow that uses it are unaffected — no new IAM roles or environment secrets. The `summary` job
```
soberm: This basically previews the same intersection-of-permissions explanation that shows up again almost word-for-word at the OIDC step and the agent-run step below. Not wrong, just a lot of repetition for one idea — could shrink this to a one-liner pointing at the detailed version further down.

.github/workflows/agent-bench.yml
```diff
@@ -156,6 +164,36 @@ jobs:
           # Extend the OIDC session past the ~1h default so it survives slow cells (AWS clamps
           # to the role's max if lower).
           role-duration-seconds: 7200
+          # LEAST PRIVILEGE: the effective session permissions are the INTERSECTION of
+          # the role's policy and this inline policy, so this narrows the initial session to Bedrock
```
soberm: Same rationale as the header comment, restated again here. Three copies of the same explanation in one file is more than it needs.

.github/workflows/agent-bench.yml
```diff
@@ -156,6 +164,36 @@ jobs:
...
+          inline-session-policy: >-
+            {
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Sid": "BedrockInvokeOnly",
+                  "Effect": "Allow",
+                  "Action": [
+                    "bedrock:InvokeModel",
+                    "bedrock:InvokeModelWithResponseStream",
```
soberm: Are InvokeModelWithResponseStream / ConverseStream actually used anywhere in the harness (builder/judge/analysis), or are they in defensively? If nothing streams, you could drop them and keep the allow-list a bit tighter. Not a blocker either way.

.github/workflows/agent-bench.yml
```diff
@@ -177,9 +215,11 @@ jobs:
         env:
           BLOCKS_TELEMETRY_CANARY_ID: ${{ secrets.BLOCKS_TELEMETRY_CANARY_ID }}
 
-      # SECURITY: this step runs an LLM agent (shell tool) with the live OIDC session, so a
-      # prompt-injected task could reach any AWS API in the role's scope. Bounded by the same-repo
-      # guard in pr-agent-bench.yml (fork PRs get no secrets/OIDC). See README "Security".
+      # SECURITY: this step runs an LLM agent (shell tool), so a prompt-injected task could reach any
+      # AWS API in the CURRENT session's scope. That session is narrowed to Bedrock-only by the OIDC
+      # step's inline session policy, so s3:PutObject is NOT reachable here even though
```
soberm: Good catch updating this to say the *session* is Bedrock-only rather than implying the role itself changed — that distinction matters and the old wording could've been misread.

.github/workflows/agent-bench.yml
```diff
@@ -312,6 +352,19 @@ jobs:
           overwrite: true
           if-no-files-found: ignore
 
+      # LEAST PRIVILEGE: re-assume the SAME role WITHOUT an inline session policy so only
+      # this persist step regains s3:PutObject on bench/*. Runs after the shell-capable agent step, so
+      # the s3 write is never live while the agent is. Best-effort: a re-assume failure just leaves the
+      # (Bedrock-only) session in place and the persist below degrades gracefully — never reds a cell.
+      - name: Re-assume role for S3 persist (drop Bedrock-only session policy)
```
soberm, line 367: Just to confirm from the diff alone — is there anything between this re-assume and the actual "Persist to S3" step that could run with the widened, S3-capable session? Looks adjacent here but wanted to double check nothing slots in between later.

.github/workflows/agent-bench.yml
```diff
@@ -312,6 +352,19 @@ jobs:
...
+        id: oidc_s3
+        if: always()
```
soberm: if: always() means this re-assume (and the S3 write it grants) fires even if the job already failed/was cancelled upstream. Probably intentional for the best-effort persist, but worth a sanity check that there's no path where a cancelled/failed job leaves the S3-capable session active for some later step that isn't part of this change. Might be worth `always() && !cancelled()` if cancellation is a real scenario here.

.github/workflows/agent-bench.yml
```diff
@@ -312,6 +352,19 @@ jobs:
...
+        continue-on-error: true
```
soberm, line 370: continue-on-error here means a silent re-assume failure just leaves the Bedrock-only session in place. That's fine as long as "Persist to S3" below tolerates a stale/insufficient session gracefully rather than hard-failing — looks like it does given the "best-effort" naming, just flagging the dependency.

.github/workflows/agent-bench.yml
```diff
@@ -312,6 +352,19 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@e7f100cf4c008499ea8adda475de1042d6975c7b # v6.2.0
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
```
soberm, line 374: The original OIDC step bumps role-duration-seconds to 7200 specifically to survive slow cells. This re-assume doesn't set a duration, so it falls back to default (often 1h). Probably not an issue since persist looks like a quick single upload, but wanted to flag it since the original step clearly anticipated duration problems.

.github/workflows/agent-bench.yml
```diff
@@ -312,6 +352,19 @@ jobs:
...
 
+      # LEAST PRIVILEGE: re-assume the SAME role WITHOUT an inline session policy so only
+      # this persist step regains s3:PutObject on bench/*. Runs after the shell-capable agent step, so
+      # the s3 write is never live while the agent is. Best-effort: a re-assume failure just leaves the
+      # (Bedrock-only) session in place and the persist below degrades gracefully — never reds a cell.
+      - name: Re-assume role for S3 persist (drop Bedrock-only session policy)
+        id: oidc_s3
+        if: always()
+        continue-on-error: true
+        uses: aws-actions/configure-aws-credentials@e7f100cf4c008499ea8adda475de1042d6975c7b # v6.2.0
```
soberm, line 371: Nice that this is still pinned to the same full commit SHA as the original assume step — keeps the supply-chain story consistent.

soberm (commented):
Nice, focused hardening change. Narrowing the bench job's OIDC session to Bedrock-only via an inline session policy (and re-assuming without it right before the S3 persist step) is the right mechanism for this — it only narrows the derived session, never widens the underlying role, so nothing else that shares `AWS_ROLE_ARN` is affected. Left a few inline questions/nits, nothing blocking.

One cross-cutting nit that doesn't anchor well to a single line: the "session policy = intersection, never widens the role" rationale gets repeated almost verbatim in three places (workflow header, the OIDC step, and the agent-run step comment). Might be worth trimming two of the three down to a pointer at the canonical explanation, just to cut the repetition.

.github/workflows/agent-bench.yml
osama-rizk, line 370: The best-effort design here (`continue-on-error: true` + `if: always()`) is a reasonable "never red a cell" choice, but it makes this re-assume a new *silent*-failure point in the S3 archival path: if it ever fails (a transient STS hiccup, or a future misconfig of `AWS_ROLE_ARN`/region), the Bedrock-only session stays in place, "Persist to S3" below then fails with AccessDenied — silently — and that cell's commit-keyed archive is dropped with no CI signal. The immediate PR summary is unaffected (it aggregates from GitHub artifacts), so the damage is a quietly-missing historical baseline entry rather than a broken run. Not a blocker — worth either emitting a warning when `steps.oidc_s3.outcome != 'success'`, or noting the accepted risk so a silent archival gap isn't later mistaken for "no results."

osama-rizk (commented):
Reviewed the full workflow, not just the diff — this is a correct, genuinely behavior-preserving least-privilege hardening. **Recommend approve.**

The key risk in any "narrow the session" change is that a session policy cuts *every* non-Bedrock permission (including `s3:GetObject`), not just the `PutObject` you meant to drop. I verified it's safe: the only S3 operation in the entire `bench` job is "Persist to S3" (PutObject), and it runs *after* the re-assume — the build-cache pull (`1a`) and all result uploads (`5/6/6b`) are `actions/{download,upload}-artifact` (GitHub, no AWS), and steps 2/4/4b are Bedrock-only. Reinforcing that: since the role only grants Bedrock + S3 and the bench passes today, the build/test steps can't be hitting other AWS services against real infra, so cutting everything-but-Bedrock can't strand them.

Bedrock coverage holds (session-policy actions are a superset of the role's Bedrock grant; wildcard resources cover the `us.*` geo profile's cross-region routing + a `BENCH_MODEL` override), and the re-assume is consistent (same pinned action SHA, role, `aws-region: us-east-1`). The `summary` job is correctly left as a documented follow-up. Comments and PR write-up are exemplary.

One operability note inline.

.github/workflows/agent-bench.yml
sharonyajain: Fixed in 85a38e2: trimmed this header block to a pointer at the "Configure AWS credentials (OIDC)" step, which is now the single canonical explanation of the intersection semantics + allow-list.

.github/workflows/agent-bench.yml
sharonyajain: Fixed in 85a38e2, kept this OIDC-step comment as the one canonical explanation and reduced the two other copies (workflow header + "2. Agent run" comment) to pointers here.

.github/workflows/agent-bench.yml
sharonyajain: Partly dropped in 85a38e2. `ConverseStream` is actually used: the Strands `BedrockModel` streams for builder/judge. The `ModelStreamUpdateEvent` hook in `2-agent-run.ts` consumes `modelMetadataEvent` stream events — so it stays. But `InvokeModel`/`InvokeModelWithResponseStream` are used by no path (Strands is Converse-family only; the CLI analysis calls `bedrock-runtime converse`), so I removed both. Allow-list is now just `bedrock:Converse` + `bedrock:ConverseStream`.

.github/workflows/agent-bench.yml
sharonyajain: Thanks — kept that "session, not role" framing intact when trimming the comment down to a pointer.

.github/workflows/agent-bench.yml
sharonyajain: Confirmed. The re-assume step is immediately followed by "Persist to S3" with nothing in between, so the only step that ever runs under the widened session is the persist itself. (The warning I added in 85a38e2 lives inside the Persist step's env/run, not as a step between them, to keep that adjacency.)

.github/workflows/agent-bench.yml
sharonyajain: Fixed in 85a38e2 — changed to `if: ${{ !cancelled() }}`. A cancelled run now never re-widens the session (so a cancellation can't leave an S3-capable session active for any later step); it still runs on job failure so a failed cell can persist its partial result.json.

.github/workflows/agent-bench.yml
sharonyajain: Confirmed — "Persist to S3" is `continue-on-error` and `s3-put.sh` degrades gracefully, so a stale Bedrock-only session just yields a failed upload, never a hard cell failure. Also addressed the silent-gap angle in your other thread (warning added in 85a38e2).

.github/workflows/agent-bench.yml
sharonyajain: Rebutted + documented in 85a38e2. Unlike the initial assume (which must span the whole ~85min cell, hence 7200s), this re-assume happens at the very end and only needs to outlive the ~3min best-effort upload, so the ~1h default is ample. Added an inline note spelling that out so the omission reads as intentional.

.github/workflows/agent-bench.yml
sharonyajain: Thanks — yep, deliberately pinned to the same full commit SHA as the original assume step to keep the supply-chain story consistent.

.github/workflows/agent-bench.yml
sharonyajain: Good call — addressed in 85a38e2 rather than accepting the risk. The Persist step now emits a `::warning::` annotation when `steps.oidc_s3.outcome != 'success'`, naming the task-template cell, so a silently-missing baseline archive surfaces in the run summary instead of being mistaken for "no results". Folded it into the Persist step's run block (not a new step) to preserve the re-assume→persist adjacency from your other thread.

soberm (approved):

osama-rizk (approved):

PR #424, opened 2026-08-21
fix(core): fail fast on missing AWS credentials before sandbox/deploy synth

osama-rizk:
## What

`npm run sandbox` (`startSandbox`) and `npm run deploy` (`deploy`) spend ~10 seconds synthesizing the CDK app before making their first AWS call. When credentials are missing or expired, that surfaces only *after* the wasted synth — as an opaque CDK/CloudFormation error that never names the real cause.

This adds a fail-fast **STS `GetCallerIdentity`** pre-check at the very start of both commands. If credentials don't resolve, the command exits immediately with actionable guidance instead of burning the synth.

## Why

Board card: **`[Bug Bash] core: npm run sandbox has no credential pre-check (10s wasted on synth)`** (P3, DX). Verified against `main`: for the common DynamoDB template, `ensureSecrets` returns early with no AWS call, so the app does zero AWS work before `cdk deploy` — exactly the ~10s-wasted-then-opaque-error path the card describes.

## How

- New `packages/core/src/scripts/preflight-credentials.ts` — `assertAwsCredentials(command, probe?)`. The default probe calls STS `GetCallerIdentity` via the SDK's standard credential chain, dynamically imported (matching the existing `@aws-sdk/client-ssm` pattern in `ensure-secrets.ts`) so it's only loaded when a deploy actually runs. A region is always supplied (falling back to `us-east-1`) so the probe never fails for a *missing region* — `GetCallerIdentity` is identity-only, and region correctness is validated later by the deploy itself.
- The `probe` is injectable so the guard is unit-tested without network or real credentials.
- Wired into `startSandbox` and `deploy`, right after `BLOCKS_STAGE` is set and before any synth/secret work.
- Declared `@aws-sdk/client-sts` as a `@aws-blocks/core` dependency (it was already resolved transitively; sibling of the `client-s3`/`client-ssm` already used here).

### Error message

```
No valid AWS credentials found for `npm run sandbox`.
Configure credentials before deploying — for example:
  • run `aws configure` (or `aws sso login`), or
  • set AWS_PROFILE to a configured profile, or
  • export AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY (and AWS_SESSION_TOKEN for temporary credentials).
Then re-run `npm run sandbox`.
(underlying error: <cause>)
```

## Testing

- New `preflight-credentials.test.ts` (4 cases): silent pass on success; actionable throw naming the command; command-name interpolation (`deploy` vs `sandbox`); non-`Error` rejection handled.
- `npm run build`, full `@aws-blocks/core` suite (680 tests) ✅, `lint:deps` ✅, biome ✅, umbrella-changeset guard ✅.

## Changeset

`@aws-blocks/core` patch + `@aws-blocks/blocks` patch (umbrella re-exports `scripts`).

### DIFF
--- a/packages/core/src/scripts/deploy.ts
+++ b/packages/core/src/scripts/deploy.ts
@@ -6,6 +6,7 @@ import { readFileSync, writeFileSync, mkdirSync } from 'node:fs';
 import { join, resolve, dirname } from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import { ensureSecrets, loadProductionEnv } from './ensure-secrets.js';
+import { assertAwsCredentials } from './preflight-credentials.js';
 import { applyExternalMigrations } from './external-migrations-step.js';
 import { trackCommand } from '../telemetry/trackCommand.js';
 import { getCdkTelemetryEnv } from './cdk-telemetry-env.js';
@@ -24,6 +25,10 @@ export async function deploy(options: DeployOptions) {
     loadProductionEnv();
 
     process.env.BLOCKS_STAGE = 'production';
+
+    // Fail fast if AWS credentials are missing/expired, before generating the
+    // client and spending time in synth only to hit an opaque CDK credential error.
+    await assertAwsCredentials('deploy');
 
     // Provision secrets for production. projectRoot must match the root cdk
     // synth uses (passed as --context below) so the written parameter name
--- a/packages/core/src/scripts/preflight-credentials.test.ts
+++ b/packages/core/src/scripts/preflight-credentials.test.ts
@@ -2,64 +2,109 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import assert from 'node:assert';
-import { describe, it } from 'node:test';
-import { assertAwsCredentials } from './preflight-credentials.js';
+import { describe, it, mock } from 'node:test';
+import { assertAwsCredentials, resolveProbeRegion } from './preflight-credentials.js';
+
+const REGION_ENV = { AWS_REGION: 'us-east-1' } as NodeJS.ProcessEnv;
+
+/** Build an Error with a specific `name`, as the AWS SDK throws. */
+function namedError(name: string, message = 'raw sdk detail'): Error {
+	const err = new Error(message);
+	err.name = name;
+	return err;
+}
+
+describe('resolveProbeRegion', () => {
+	it('prefers AWS_REGION', () => {
+		assert.strictEqual(resolveProbeRegion({ AWS_REGION: 'eu-west-1', AWS_DEFAULT_REGION: 'us-east-2' }), 'eu-west-1');
+	});
+
+	it('falls back to AWS_DEFAULT_REGION', () => {
+		assert.strictEqual(resolveProbeRegion({ AWS_DEFAULT_REGION: 'ap-south-1' }), 'ap-south-1');
+	});
+
+	it('returns null when neither is set (caller then skips the probe)', () => {
+		assert.strictEqual(resolveProbeRegion({}), null);
+	});
+});
 
 describe('assertAwsCredentials', () => {
 	it('resolves silently when the credential probe succeeds', async () => {
-		let called = false;
+		let calledRegion: string | undefined;
 		await assert.doesNotReject(() =>
-			assertAwsCredentials('sandbox', async () => {
-				called = true;
-			}),
+			assertAwsCredentials(
+				'sandbox',
+				async (region) => {
+					calledRegion = region;
+				},
+				REGION_ENV,
+			),
 		);
-		assert.strictEqual(called, true);
+		assert.strictEqual(calledRegion, 'us-east-1');
 	});
 
-	it('throws an actionable error naming the command when the probe rejects', async () => {
-		const probeError = new Error('Could not load credentials from any providers');
+	it('throws actionable guidance on a real credential error, naming the command', async () => {
 		await assert.rejects(
-			() =>
-				assertAwsCredentials('sandbox', async () => {
-					throw probeError;
-				}),
+			() => assertAwsCredentials('sandbox', async () => { throw namedError('CredentialsProviderError'); }, REGION_ENV),
 			(err: Error) => {
-				// Names the failing command so the message is copy-pasteable.
 				assert.match(err.message, /npm run sandbox/);
-				// Points at concrete remediation, not just "failed".
 				assert.match(err.message, /aws configure|AWS_PROFILE|AWS_ACCESS_KEY_ID/);
-				// Preserves the underlying cause for debugging.
-				assert.match(err.message, /Could not load credentials from any providers/);
+				assert.match(err.message, /CredentialsProviderError/); // the name, for debugging
 				return true;
 			},
 		);
 	});
 
-	it('interpolates the command name (deploy vs sandbox)', async () => {
+	it('treats an expired token as a credential error', async () => {
 		await assert.rejects(
-			() =>
-				assertAwsCredentials('deploy', async () => {
-					throw new Error('boom');
-				}),
+			() => assertAwsCredentials('deploy', async () => { throw namedError('ExpiredToken'); }, REGION_ENV),
 			(err: Error) => {
 				assert.match(err.message, /npm run deploy/);
-				assert.doesNotMatch(err.message, /npm run sandbox/);
+				assert.match(err.message, /aws sso login|AWS_PROFILE/);
 				return true;
 			},
 		);
 	});
 
-	it('handles a non-Error rejection value without crashing', async () => {
+	it('does NOT leak the raw error message (ARN / account id)', async () => {
+		const arnMessage = 'User: arn:aws:iam::123456789012:user/alice is not authorized to perform sts:GetCallerIdentity';
 		await assert.rejects(
-			// Throw a non-Error to exercise the String(error) fallback branch.
-			() =>
-				assertAwsCredentials('deploy', async () => {
-					throw 'string failure';
-				}),
+			() => assertAwsCredentials('deploy', async () => { throw namedError('InvalidClientTokenId', arnMessage); }, REGION_ENV),
 			(err: Error) => {
-				assert.match(err.message, /string failure/);
+				assert.doesNotMatch(err.message, /arn:aws:iam/);
+				assert.doesNotMatch(err.message, /123456789012/);
 				return true;
 			},
 		);
+	});
+
+	it('does not throw on a network/service error — warns and continues', async () => {
+		const warn = mock.method(console, 'warn', () => {});
+		try {
+			await assert.doesNotReject(() =>
+				assertAwsCredentials('sandbox', async () => { throw namedError('TimeoutError'); }, REGION_ENV),
+			);
+			assert.ok(
+				warn.mock.calls.some((c) => String(c.arguments[0]).includes('Could not verify')),
+				'should warn that credentials could not be verified',
+			);
+		} finally {
+			warn.mock.restore();
+		}
+	});
+
+	it('skips the probe (with a warning) when no region is resolvable', async () => {
+		const warn = mock.method(console, 'warn', () => {});
+		let probed = false;
+		try {
+			await assertAwsCredentials('sandbox', async () => { probed = true; }, {});
+			assert.strictEqual(probed, false, 'probe must not run without a region');
+			assert.ok(
+				warn.mock.calls.some((c) => String(c.arguments[0]).includes('Skipping the AWS credential pre-check')),
+				'should warn that the pre-check was skipped',
+			);
+		} finally {
+			warn.mock.restore();
+		}
 	});
 });
--- a/packages/core/src/scripts/preflight-credentials.test.ts
+++ b/packages/core/src/scripts/preflight-credentials.test.ts
@@ -1,0 +1,65 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+import { assertAwsCredentials } from './preflight-credentials.js';
+
+describe('assertAwsCredentials', () => {
+	it('resolves silently when the credential probe succeeds', async () => {
+		let called = false;
+		await assert.doesNotReject(() =>
+			assertAwsCredentials('sandbox', async () => {
+				called = true;
+			}),
+		);
+		assert.strictEqual(called, true);
+	});
+
+	it('throws an actionable error naming the command when the probe rejects', async () => {
+		const probeError = new Error('Could not load credentials from any providers');
+		await assert.rejects(
+			() =>
+				assertAwsCredentials('sandbox', async () => {
+					throw probeError;
+				}),
+			(err: Error) => {
+				// Names the failing command so the message is copy-pasteable.
+				assert.match(err.message, /npm run sandbox/);
+				// Points at concrete remediation, not just "failed".
+				assert.match(err.message, /aws configure|AWS_PROFILE|AWS_ACCESS_KEY_ID/);
+				// Preserves the underlying cause for debugging.
+				assert.match(err.message, /Could not load credentials from any providers/);
+				return true;
+			},
+		);
+	});
+
+	it('interpolates the command name (deploy vs sandbox)', async () => {
+		await assert.rejects(
+			() =>
+				assertAwsCredentials('deploy', async () => {
+					throw new Error('boom');
+				}),
+			(err: Error) => {
+				assert.match(err.message, /npm run deploy/);
+				assert.doesNotMatch(err.message, /npm run sandbox/);
+				return true;
+			},
+		);
+	});
+
+	it('handles a non-Error rejection value without crashing', async () => {
+		await assert.rejects(
+			// Throw a non-Error to exercise the String(error) fallback branch.
+			() =>
+				assertAwsCredentials('deploy', async () => {
+					throw 'string failure';
+				}),
+			(err: Error) => {
+				assert.match(err.message, /string failure/);
+				return true;
+			},
+		);
+	});
+});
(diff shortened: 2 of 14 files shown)


packages/core/src/scripts/preflight-credentials.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Dynamically imported so the STS client is only loaded when a deploy actually
+ * runs — never during dev-server startup or a mock unit test. A region is
+ * always supplied (falling back to `us-east-1`) so the probe never fails for a
+ * *missing region* — GetCallerIdentity is identity-only, so any region answers,
+ * and region correctness is validated later by the deploy itself.
+ */
+const stsProbe: CredentialProbe = async () => {
+	const { STSClient, GetCallerIdentityCommand } = await import('@aws-sdk/client-sts');
+	const region = process.env.AWS_REGION || process.env.AWS_DEFAULT_REGION || 'us-east-1';
+	const client = new STSClient({ region });
```
sarayev: No timeout here means the precheck can hang longer than the synth it's replacing. The STS client uses SDK defaults (`maxAttempts: 3`, no request/connection timeout), so on a bad network (VPN drop, locked-down VPC, dead proxy) this can stall for tens of seconds — the exact slow-and-confusing experience this change is meant to remove. The repo already has this convention in `common/config.ts` (`requestTimeout: 5000`, `maxAttempts: 3`); could we match it?

```ts
const client = new STSClient({
    region,
    maxAttempts: 1,
    requestHandler: new NodeHttpHandler({ connectionTimeout: 2000, requestTimeout: 3000 }),
});
```

packages/core/src/scripts/preflight-credentials.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Dynamically imported so the STS client is only loaded when a deploy actually
+ * runs — never during dev-server startup or a mock unit test. A region is
+ * always supplied (falling back to `us-east-1`) so the probe never fails for a
+ * *missing region* — GetCallerIdentity is identity-only, so any region answers,
+ * and region correctness is validated later by the deploy itself.
+ */
+const stsProbe: CredentialProbe = async () => {
+	const { STSClient, GetCallerIdentityCommand } = await import('@aws-sdk/client-sts');
+	const region = process.env.AWS_REGION || process.env.AWS_DEFAULT_REGION || 'us-east-1';
```
sarayev: The `us-east-1` fallback breaks GovCloud/China. If neither `AWS_REGION` nor `AWS_DEFAULT_REGION` is set (common when region comes from a profile or cdk context rather than env), the probe hits `us-east-1` — a different partition than `us-gov-west-1` or `cn-north-1` — so it can fail against a partition the user isn't even deploying to and block a deploy that would have worked. Should we skip the probe with a warning rather than hard-fail when no region can be resolved?

packages/core/src/scripts/preflight-credentials.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param command - The npm script name, used in the error message (e.g. `sandbox`, `deploy`).
+ * @param probe - Credential probe; defaults to STS GetCallerIdentity. Override in tests.
+ * @throws {Error} With actionable guidance when the probe rejects.
+ */
+export async function assertAwsCredentials(command: string, probe: CredentialProbe = stsProbe): Promise<void> {
+	try {
+		await probe();
+	} catch (error) {
+		const detail = error instanceof Error ? error.message : String(error);
```
sarayev: The raw `error.message` can leak an ARN and account id here. An STS authorization failure puts the caller ARN in the message (`User: arn:aws:iam::<acct>:user/... is not authorized...`), and this Error propagates up through `trackCommand` telemetry — which runs into the project rule about not leaking ARNs / `$metadata`. Could we interpolate `error.name` instead of the full message (or scrub `arn:`/account ids before embedding)?

```ts
const detail = error instanceof Error ? error.name : 'unknown error';
```

packages/core/src/scripts/preflight-credentials.ts
```diff
@@ -0,0 +1,65 @@
...
+		throw new Error(
```
sarayev: Right now every failure reports as "no credentials" — a throttle, a DNS/timeout, or STS not being enabled in a region all produce the same `aws configure` / `AWS_PROFILE` guidance. When the user's credentials are actually fine, that sends them down the wrong path. Could we branch on `error.name` so only the real credential cases (`CredentialsProviderError`, expired token) get the remediation text, and network/service errors surface as themselves? Also, since GetCallerIdentity proves identity but not deploy-ability (creds can still lack assume-role on the CDK bootstrap roles, or an SSO session can expire right after), softening "No valid AWS credentials found" to "AWS credentials could not be verified" would keep the message honest.

packages/core/src/scripts/deploy.ts
```diff
@@ -25,6 +26,10 @@ export async function deploy(options: DeployOptions) {
 
     process.env.BLOCKS_STAGE = 'production';
 
+    // Fail fast if AWS credentials are missing/expired, before generating the
+    // client and spending time in synth only to hit an opaque CDK credential error.
+    await assertAwsCredentials('deploy');
```
sarayev, line 31: Is env loaded before this runs? The check is placed right after `BLOCKS_STAGE` but before `loadProductionEnv()`. If a project's `.env` is what supplies `AWS_PROFILE` or `AWS_REGION`, the probe runs before those are applied and could use the wrong/empty profile or region (which also feeds the us-east-1 partition edge case). Would it be safer to run this right after the env is loaded? Nothing else expensive runs in between, so moving it down a few lines is cheap.

packages/core/src/scripts/sandbox.ts
```diff
@@ -128,6 +129,10 @@ export async function startSandbox(options: SandboxOptions) {
 
   process.env.BLOCKS_STAGE = 'sandbox';
 
+  // Fail fast if AWS credentials are missing/expired, before spending ~10s in
+  // synth only to hit an opaque CDK credential error.
+  await assertAwsCredentials('sandbox');
```
sarayev: Heads up for the sandbox path specifically: the generated template entrypoint for `sandbox` has no `.catch` (unlike `deploy`, which catches and does `process.exit(1)`), so a thrown Error from this check surfaces as a raw unhandled-rejection stack trace rather than the clean, actionable message this change is aiming for. Worth confirming whether sandbox should get the same friendly handling. (Minor: the shared message says "Configure credentials before deploying", which reads a little off for `sandbox`.)

packages/core/src/scripts/preflight-credentials.test.ts
```diff
@@ -0,0 +1,65 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+import { assertAwsCredentials } from './preflight-credentials.js';
+
+describe('assertAwsCredentials', () => {
```
sarayev, line 31: The tests only ever inject a fake probe, so the default `stsProbe` — the highest-risk logic (region fallback, `client.destroy()`, the dynamic import) — has no coverage. Pulling region resolution into a small exported pure function and testing it (env precedence + `us-east-1` fallback) would cover the riskiest branch without needing network or real credentials. Worth adding?

packages/core/src/scripts/preflight-credentials.ts
osama-rizk: Done — bounded the STS client to fail fast: `maxAttempts: 1` + `requestHandler: { connectionTimeout: 2000, requestTimeout: 3000 }` (matching the `requestHandler` shorthand in `common/config.ts`). Worst case is a few seconds, and a network/timeout error no longer blocks — it warns and lets the deploy proceed (see the reply on the error-classification thread).

packages/core/src/scripts/preflight-credentials.ts
osama-rizk: Done — dropped the `us-east-1` fallback entirely. Region now comes from `AWS_REGION`/`AWS_DEFAULT_REGION` only (new exported `resolveProbeRegion()`); when neither is set the probe is **skipped with a warning** and the deploy proceeds, so we never probe a partition the user isn't deploying to (GovCloud/China).

packages/core/src/scripts/preflight-credentials.ts
osama-rizk: Done — the thrown Error now interpolates `error.name` only, never `error.message`, so an authorization message like `User: arn:aws:iam::<acct>:...` can't leak (into the client or through `trackCommand` telemetry). Added a regression test asserting an ARN/account-id message does not appear in the thrown error.

packages/core/src/scripts/preflight-credentials.ts
osama-rizk: Done — added a `CREDENTIAL_ERROR_NAMES` allowlist (`CredentialsProviderError`, `ExpiredToken(Exception)`, `InvalidClientTokenId`, `UnrecognizedClientException`, `SignatureDoesNotMatch`, `TokenRefreshRequired`). Only those block, with the `aws configure`/`AWS_PROFILE` guidance. Everything else — network, throttling, and notably `AccessDenied` (which actually proves the identity resolved) — warns and lets the deploy proceed. Also softened the wording to **"AWS credentials could not be verified"** per your point that GetCallerIdentity proves identity, not deploy-ability.

packages/core/src/scripts/deploy.ts
osama-rizk: Good to check — but env is already loaded before the probe in both paths, so it's using the right profile/region: `deploy.ts` calls `loadProductionEnv()` at line 25, before the check at line 31; `sandbox.ts` loads `.env.local` at the top before it. So `AWS_PROFILE`/`AWS_REGION` from those files are applied first. (Neither loads a plain `.env` — that's pre-existing behavior, out of scope here.) Left the placement as-is; let me know if you'd still prefer it moved.

packages/core/src/scripts/sandbox.ts
osama-rizk: Done, both parts: (1) added `.catch((error) => { console.error(error); process.exit(1); })` to all 7 scaffolded `sandbox` template entrypoints, mirroring `deploy`, so the guidance prints cleanly instead of as an unhandled-rejection stack; (2) made the message command-neutral — it no longer says "before deploying" and reads "Configure AWS credentials … then re-run `npm run <command>`".

packages/core/src/scripts/preflight-credentials.test.ts
osama-rizk: Done — extracted and exported `resolveProbeRegion(env)` and added unit tests for it: `AWS_REGION` precedence, `AWS_DEFAULT_REGION` fallback, and the `null` (skip) case. Also added tests for the skip-with-warning path and the network-error warn-and-continue path, so the branching logic is covered without network or real credentials.

packages/core/src/scripts/sandbox.ts
soberm: [blocker] Building on the note above about the missing `.catch` — this isn't just a UX gap, it's actively failing CI on this PR. `Telemetry E2E` is red right now:

```
✖ FAIL: no creds emits sandbox/FAIL (11606.11741ms)
  AssertionError [ERR_ASSERTION]: sandbox FAIL should emit telemetry
```

Root cause: `startSandbox` is the only one of the four commands not wrapped in `trackCommand`. Its telemetry fires manually, via the `buildAndSendEvent` calls around the CDK deploy call and on success further down, and `assertAwsCredentials` now throws before either of those run, so no FAIL event ever gets sent. `deploy()` doesn't hit this because its whole body sits inside `trackCommand('deploy', ...)`.

It's worse than the missing telemetry too: every generated app's `aws-blocks/scripts/sandbox.ts` calls `startSandbox(...)` as a bare floating promise, no `await`, no `.catch`, unlike its `deploy.ts` sibling which has one. So this throw surfaces as a raw unhandled-rejection stack trace instead of the actionable message this PR is adding.

Smallest fix I can see, matching how the CDK-deploy failure below already reports telemetry:

```ts
try {
  await assertAwsCredentials('sandbox');
} catch (error) {
  buildAndSendEvent({
    command: 'sandbox', state: 'FAIL',
    duration: Date.now() - sandboxStartTime,
    error: classifyError(error),
  });
  throw error;
}
```

Fixes the telemetry gap for every caller without touching the template files. The missing `.catch()` in the templates is a separate pre-existing issue, but this PR is what makes it fire on a really common path (no local creds configured yet), so I'd want to see it closed before merging.

packages/core/src/scripts/preflight-credentials.ts
soberm: [major] Agreed on the timeout point above. With nothing bounding this, a dead VPN or proxy can make it hang way longer than the ~10s synth it's replacing, which works against the whole point of the change. `common/config.ts` already has a convention for this on the S3 client (`requestHandler: { requestTimeout: 5000 }, maxAttempts: 3`), so this doesn't need a new pattern, just reuse it:

```ts
const client = new STSClient({
  region,
  maxAttempts: 1,
  requestHandler: { requestTimeout: 3000 },
});
```

packages/core/src/scripts/preflight-credentials.ts
soberm: [major] +1 on the partition point above. `GetCallerIdentity` only needs a same-partition region to answer, so this could resolve the fallback partition-aware, or just skip the probe with a warning when nothing's set instead of guessing `us-east-1` — otherwise a gov/china account gets probed against a completely different partition.

packages/core/src/scripts/preflight-credentials.ts
soberm: [major] On the ARN/account-id point above: I traced it through `trackCommand`, and the telemetry payload itself is fine, `classifyError` only ever sends `{code, phase}`, never the raw message, so it's not reaching the telemetry backend. It does still land in stderr/CI logs though, since the templates do `console.error(error)` on catch, and there's already a `redact.ts` in this package for exactly this kind of thing. `error.name` gets the same "what kind of error" signal for debugging without carrying the ARN:

```ts
const detail = error instanceof Error ? error.name : 'unknown error';
```

packages/core/src/scripts/preflight-credentials.test.ts
soberm, line 31: [minor] Since all four tests inject a fake probe, the default `stsProbe` itself (region fallback, the dynamic import, `client.destroy()`) has zero coverage, and that's exactly the logic the region/partition comment above is about. Pulling region resolution into a small exported helper would let that get unit-tested without needing real credentials or network.

soberm (changes requested):
## Summary
Adds a fail-fast STS `GetCallerIdentity` check before `startSandbox`/`deploy` synth, so missing or expired credentials show up immediately with an actionable message instead of after a wasted ~10s synth. The core logic is small and has decent unit tests, but the wiring is uneven between the two call sites: `deploy()` is covered by `trackCommand`'s catch-all, `startSandbox()` isn't, and that's currently breaking CI (see the inline comment on sandbox.ts).

packages/core/src/scripts/sandbox.ts
osama-rizk: Great catch, and thanks for the precise root-cause — fixed in c04274b0. Wrapped the check in a try/catch that emits the FAIL event before rethrowing, using `classifyError` and mirroring the existing CDK-deploy failure path, so every caller gets telemetry without touching the templates:

```ts
try {
  await assertAwsCredentials('sandbox');
} catch (error) {
  buildAndSendEvent({ command: 'sandbox', state: 'FAIL', duration: Date.now() - sandboxStartTime, error: classifyError(error) });
  throw error;
}
```

Verified locally against the exact `Telemetry E2E` case (had to disable ambient profile/SSO creds, not just the env vars, to reproduce CI's no-creds state): `ok 1 - FAIL: no creds emits sandbox/FAIL (6.4s)` — fast-fail + `sandbox/FAIL` emitted. (The template floating-promise `.catch` is also added in this PR, per the sandbox.ts thread.)

packages/core/src/scripts/preflight-credentials.ts
osama-rizk: Agreed — already done in 26ed0d76 (reusing the `common/config.ts` convention exactly as you suggested): `maxAttempts: 1` + `requestHandler: { connectionTimeout: 2000, requestTimeout: 3000 }`. A network/timeout error now also warns-and-continues rather than blocking, so it can't hang longer than the synth.

packages/core/src/scripts/preflight-credentials.ts
osama-rizk: Agreed — already done in 26ed0d76. Dropped the `us-east-1` guess entirely: region resolves from `AWS_REGION`/`AWS_DEFAULT_REGION` only, and when neither is set the probe is skipped with a warning, so a gov/china account is never probed against the wrong partition.

packages/core/src/scripts/preflight-credentials.ts
osama-rizk: Thanks for tracing the telemetry path — that matches what I found (`classifyError` only sends `{code, phase}`). Already switched to `error.name` in 26ed0d76 for exactly the stderr/CI-log reason you note (the templates `console.error(error)`), so the ARN never lands there either. Added a regression test asserting an `arn:aws:iam::<acct>:...` message doesn't appear in the thrown error.

packages/core/src/scripts/preflight-credentials.test.ts
osama-rizk: Agreed — already done in 26ed0d76: pulled region resolution into an exported `resolveProbeRegion(env)` with unit tests for `AWS_REGION` precedence, `AWS_DEFAULT_REGION` fallback, and the null (skip) case, plus tests for the warn-and-continue and skip-with-warning branches.

soberm (approved):

sarayev (approved):

PR #435, opened 2026-08-24
feat(core): default the shared handler Lambda to arm64 (Graviton)

osama-rizk:
## What

Makes the framework-provisioned Blocks handler Lambda run on **arm64 (AWS Graviton)** by default, via a new `lambdaArchitecture` field on `BlocksDefaults` (set to `Architecture.ARM_64` in both `BlocksPresets`).

## Why

Board card: **`[Bug Bash] P3: All Lambdas pinned to x86_64 (Graviton ~20% cheaper)`** (P3, Cost/availability). Verified against `main`: `blocks-backend.ts` created the shared `Handler` with no `architecture` prop → CDK's `x86_64` default. arm64 Lambda is ~20% cheaper per GB-second at equivalent performance, and the handler is a pure-JavaScript esbuild bundle (`--conditions=aws-runtime`, `minify`) with no architecture-specific native dependencies — so the switch carries no runtime risk.

## How

- **`BlocksDefaults`** gains `lambdaArchitecture: Architecture` (fully documented on the interface). Both presets — `sandbox` and `production` — default it to `Architecture.ARM_64`. This follows the exact precedent of the recently-added `pointInTimeRecovery` default field.
- **`setupBlocksInfra`** applies `props.defaults.lambdaArchitecture` to the shared `Handler`.
- Override per stack:
  ```ts
  import { Architecture } from 'aws-cdk-lib/aws-lambda';
  defaults: { ...BlocksPresets.production, lambdaArchitecture: Architecture.X86_64 }
  ```

### Scope (deliberately narrow)

This changes **only the shared handler** — the one Lambda every Blocks app runs, and the biggest cost lever. It intentionally does **not** touch:
- **Hosting SSR compute** (`@aws-blocks/hosting`) — those adapters can bundle architecture-specific native binaries (e.g. `sharp` for Next image optimization), so flipping them to arm64 needs its own verification.
- **BB-owned auxiliary Lambdas** (`bb-data` / `bb-distributed-data` migration fns, `bb-app-setting` init fn, `bb-lambda-compute`) — these can read the same `defaults.lambdaArchitecture` in a follow-up, mirroring how `#302` adopted `BlocksDefaults` in one block first.

## Testing

- New synth assertions in `blocks-backend.test.ts`: handler synthesizes with `Architectures: ['arm64']` by default, and `['x86_64']` when overridden.
- `blocks-defaults.test.ts`: both presets assert `Architecture.ARM_64`.
- Full `@aws-blocks/core` suite (689 tests) ✅, `npm run lint` ✅, `lint:deps` ✅, `check:api` ✅, umbrella-changeset guard ✅.

## Changeset

`@aws-blocks/core` **minor** + `@aws-blocks/blocks` **minor** (umbrella re-exports `BlocksDefaults`/`BlocksPresets`).

> **Maintainer note:** Two things worth an explicit call before release — (1) making `arm64` the *default* is a behavior change (the handler's architecture flips `x86_64`→`arm64` on next deploy; in-place update, no new function), and (2) it adds a required field to the public `BlocksDefaults` `@aws-blocks/core/cdk` surface (both presets set it, matching the `pointInTimeRecovery` precedent), which wants API-BR sign-off.

### DIFF
--- a/packages/core/src/cdk/blocks-defaults.ts
+++ b/packages/core/src/cdk/blocks-defaults.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { RemovalPolicy } from 'aws-cdk-lib';
+import { Architecture } from 'aws-cdk-lib/aws-lambda';
 
 /**
  * Default values for every Amazon-authored Building Block created within a
@@ -39,6 +40,24 @@ export interface BlocksDefaults {
 	 * 1–35 days). Backups only have a window when on, so the two are one field.
 	 */
 	pointInTimeRecovery: boolean | { retentionDays: number };
+
+	/**
+	 * The instruction-set architecture for framework-provisioned Lambda compute
+	 * (the shared Blocks handler). Defaults to **`Architecture.ARM_64`** (AWS
+	 * Graviton) in every preset: arm64 Lambda is ~20% cheaper per GB-second than
+	 * x86_64 at the same performance, and the Blocks handler is a pure-JavaScript
+	 * esbuild bundle with no architecture-specific native dependencies, so the
+	 * switch is free.
+	 *
+	 * Override to `Architecture.X86_64` if you bundle an x86-only native addon
+	 * into your backend:
+	 *
+	 * ```ts
+	 * import { Architecture } from 'aws-cdk-lib/aws-lambda';
+	 * defaults: { ...BlocksPresets.production, lambdaArchitecture: Architecture.X86_64 }
+	 * ```
+	 */
+	lambdaArchitecture: Architecture;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -55,11 +74,13 @@ export const BlocksPresets = {
 		removalPolicy: RemovalPolicy.DESTROY,
 		deletionProtection: false,
 		pointInTimeRecovery: false,
+		lambdaArchitecture: Architecture.ARM_64,
```
Simone319, line 77: At this point I'm starting to think what genuinely belongs to the blocks-defaults. As lambda will be turned into a block that customers can create and configure, does it make all blocks configuration part of the defaults? 

Imo the defaults should hold settings that span across building blocks, or setting that differs between dev/sandbox/prod. Other individual building blocks setting should remain managed by each building blocks. Wdyt?

osama-rizk: Great point — agreed. Architecture isn't really a `BlocksDefaults` concern: it doesn't vary by stage (both presets would hold the same value), and `memorySize`/`timeout`/`runtime` already live on the compute construct rather than in the defaults. Architecture is the same kind of per-compute knob.

I've moved it onto the compute block instead: #439 defaults `LambdaCompute` to arm64 and adds `architecture?` to `LambdaComputeProps` (the placeholder reserved for exactly this). It's stacked on #341 so it lands with the compute-as-default work. Closing this PR in favor of that one — thanks for the steer!

osama-rizk:
Superseded by #439, which implements the same arm64 (Graviton) default on the compute block (`LambdaCompute`) instead of `BlocksDefaults` — per @Simone319's review above. #439 is stacked on #341 so it lands with the compute-as-default work. Closing this in favor of #439.

PR #436, opened 2026-08-24
fix(bb-auth-oidc): requireAuth throws ApiError(401) so handle401 redirect works

sharonyajain:
## Summary
AuthOIDC `requireAuth()` now throws `ApiError(401, NotAuthenticatedException)` instead of a bare `Error`, so the README's documented `if (handle401(e, provider)) return;` 401-redirect pattern actually fires for a signed-out protected call.

## Root cause
- `notAuthenticated()` (`packages/bb-auth-oidc/src/auth-oidc.ts`) returned a bare `Error` with `name: 'NotAuthenticatedException'` and **no `status`**.
- Across the JSON-RPC boundary, `errorResponseFromCatch` (`packages/core/src/rpc.ts`) serializes a non-`ApiError` as code **500**, so the browser receives an `ApiError` whose `status === 500`.
- `handle401()` (`packages/bb-auth-oidc/src/index.browser.ts`) only acts on `err instanceof ApiError && err.status === 401`, so it never matched — the error surfaced and the app never redirected to sign-in.

This mirrors the behavior already present in AuthCognito, whose `requireAuth()` throws `ApiError('Authentication required', 401, { name: NotAuthenticated })`.

## Fix
`notAuthenticated()` now returns `new ApiError('Authentication required', 401, { name: 'NotAuthenticatedException' })`. The error name is preserved, so existing `isBlocksError(e, AuthOIDCErrors.NotAuthenticated)` checks are unaffected; both the in-process throw and the wire-decoded error now carry `status === 401`.

## Tests
- `index.test.ts` — regression: `requireAuth()` rejects with an `ApiError` whose `status === 401` and name `NotAuthenticatedException`.
- `index.browser.test.ts` — new `handle401` block: redirects to `/aws-blocks/auth/signin/<provider>` on a 401 `ApiError`, returns `false` on a non-401, and (regression) does not match the pre-fix bare `Error` shape.
- Full `bb-auth-oidc` suite: 116/116 pass; repo-wide `tsc --build` clean.

Found during OIDC E2E (friction log #20; `docs/auth-testability-gaps.md` G-4).

### DIFF
--- a/packages/bb-auth-oidc/src/index.test.ts
+++ b/packages/bb-auth-oidc/src/index.test.ts
@@ -7,7 +7,7 @@ import { rmSync, mkdtempSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { createServer, type Server } from 'node:http';
-import { clearRouteRegistry } from '@aws-blocks/core';
+import { clearRouteRegistry, ApiError } from '@aws-blocks/core';
 import type { BlocksContext } from '@aws-blocks/core';
 import { AuthOIDC, AuthOIDCErrors, stubIdp, google, customOidc, customOauth2, cognitoFederated } from './index.mock.js';
 import { handleDiscovery, handleJwks, handleAuthorize, handleAuthorizeSubmit, stubIssuerUrl } from './engines/stub-idp.js';
@@ -578,6 +578,28 @@ describe('requireAuth / checkAuth / getCurrentUser without session', () => {
 		await assert.rejects(
 			() => auth.requireAuth(ctx),
 			(e: Error) => e.name === AuthOIDCErrors.NotAuthenticated,
+		);
+	});
+
+	test('requireAuth throws an ApiError with status 401 (regression: handle401 / e.status===401 must match)', async () => {
+		// Regression for the "handle401 silently does nothing" bug. requireAuth()
+		// used to throw a bare Error (no status) named NotAuthenticatedException;
+		// over the JSON-RPC wire that serializes as code 500, so the documented
+		// `if (handle401(e, provider)) return;` redirect pattern never fired.
+		// The error MUST be an ApiError(401) — mirroring AuthCognito.requireAuth() —
+		// so both the in-process throw and the wire-decoded ApiError carry status 401.
+		const auth = new AuthOIDC(ROOT, unique('noauth401'), {
+			providers: [stubIdp({ name: unique('p') })],
+		});
+		const ctx = freshContext();
+		await assert.rejects(
+			() => auth.requireAuth(ctx),
+			(e: unknown) => {
+				assert.ok(e instanceof ApiError, `expected ApiError, got ${(e as { constructor?: { name?: string } })?.constructor?.name}`);
+				assert.strictEqual(e.status, 401, 'status must be 401 so handle401() and e.status===401 checks match');
+				assert.strictEqual(e.name, AuthOIDCErrors.NotAuthenticated, 'name must be preserved for isBlocksError()');
+				return true;
+			},
 		);
 	});
 
--- a/packages/bb-auth-oidc/src/index.test.ts
+++ b/packages/bb-auth-oidc/src/index.test.ts
@@ -662,12 +662,8 @@ describe('requireAuth / checkAuth / getCurrentUser without session', () => {
 	});
 
 	test('requireAuth throws an ApiError with status 401 (regression: handle401 / e.status===401 must match)', async () => {
-		// Regression for the "handle401 silently does nothing" bug. requireAuth()
-		// used to throw a bare Error (no status) named NotAuthenticatedException;
-		// over the JSON-RPC wire that serializes as code 500, so the documented
-		// `if (handle401(e, provider)) return;` redirect pattern never fired.
-		// The error MUST be an ApiError(401) — mirroring AuthCognito.requireAuth() —
-		// so both the in-process throw and the wire-decoded ApiError carry status 401.
+		// A bare Error serializes as 500 over JSON-RPC, so handle401 (status === 401)
+		// never fired. Must be ApiError(401); name preserved for isBlocksError().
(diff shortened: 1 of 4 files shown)


pranavosu (approved):

sharonyajain:
## CI E2E failures here are infra (IAM role quota), not this PR

Flagging for maintainers/infra owners — the failing checks on this PR are **not** caused by the change.

### What's failing
`Build and Test` is an aggregation gate; the real failures are the E2E jobs that deploy a CDK stack (E2E Sandbox, Telemetry, and all Hosting frameworks). Every one dies at **stack create** with:

```
CREATE_FAILED  AWS::IAM::Role  ...
"Cannot exceed quota for RolesPerAccount: 1500 (Status Code: 409,
 HandlerErrorCode: ServiceLimitExceeded)"
→ DeploymentError → Deployment failed
```

The shared CI account is pinned at the hard **1500 IAM roles-per-account** cap, so no new e2e stack can create its roles. Confirmed on both the original run and a fresh re-run, across all 7 failing E2E jobs.

This PR's own unit + local job (`Build, Unit Tests, E2E Local`) **passed**, and the change is isolated to AuthOIDC's `requireAuth` error type — it touches nothing in hosting/telemetry/sandbox deploys.

### Root cause of the quota exhaustion
`Cleanup Stale E2E Stacks` (`cleanup-stacks.yml`) has been failing every 4h, and a large backlog of stale stacks is stuck in `DELETE_FAILED` — their IAM roles never get released. From the latest completed cleanup run, **~134 distinct stacks are in `DELETE_FAILED`**:

- **~130** `blocks-hosting-*` / `bb-test-prod-*` e2e stacks (logged as `unexpected status: DELETE_FAILED`; the step skips rather than resolves them), spanning old PRs — e.g. `bb-test-prod-pr-282-*` (×9), `blocks-hosting-ssr-pr-161/162` (×5 each), plus hosting stacks going back to PR #20 / #46 / #107.
- **4** `amplify-*-sandbox-*` stacks (`Deleted: 0, Still failing: 4`), one dating to 2026-06-19.

(Snapshot from the 16:07 UTC cleanup run; a live count would come from `aws cloudformation list-stacks --stack-status-filter DELETE_FAILED`.)

That ~134-stack backlog is what's holding the account at the role cap.

### Suggested remediation (needs CI-account access)
1. Clear the `DELETE_FAILED` backlog — for each stack, `describe-stack-events` to find the blocking resource, then `delete-stack --retain-resources <blockingLogicalId>` and remove the orphaned IAM roles. Given the volume, script it over the full `DELETE_FAILED` list.
2. Re-run `Cleanup Stale E2E Stacks` so its later steps reap the remaining stale `blocks-*` / `bb-test-*` stacks.
3. Optionally raise the Service Quota for IAM **Roles per account** (`L-FE177D64`) for headroom.
4. Longer-term: make the janitor actually resolve (not just log/skip) `DELETE_FAILED` stacks and not fail the whole run on a single stuck stack; and/or reduce roles-per-stack in the e2e apps.

Once the account is back under the cap, re-running the failed E2E jobs on this PR should pass. Happy to help with any of the above.

packages/bb-auth-oidc/src/index.test.ts
```diff
@@ -661,6 +661,24 @@ describe('requireAuth / checkAuth / getCurrentUser without session', () => {
 		);
 	});
 
+	test('requireAuth throws an ApiError with status 401 (regression: handle401 / e.status===401 must match)', async () => {
```
osama-rizk, line 664: This test (and the browser `handle401` test) pin the two *ends* — server throws `ApiError(401)`, client `handle401` acts on `ApiError(401)` — but not the *middle*, which is where the bug actually lived: a bare `Error` was serialized as code 500 by `errorResponseFromCatch` and decoded back as `status 500`, so `handle401` never matched. Nothing here exercises that serialize→deserialize round-trip for this error. It's covered by core's `rpc.ts` contract (and I confirmed the round-trip preserves both status and name), so this is adequate — but since the defect was *specifically* a wire-serialization behavior, the single strongest regression guard would be one integration-style assertion that `requireAuth`'s error, passed through `errorResponseFromCatch` + `decodeRpcResponse`, arrives with `status === 401`. Optional / non-blocking.

osama-rizk (commented):
Reviewed end to end — correct, minimal, well-reasoned fix. **Recommend approve.**

I traced the full JSON-RPC round-trip from source rather than trusting the root-cause narrative, since the whole bug lived in the serialization *between* the two endpoints: `requireAuth` → `ApiError(401)` → `errorResponseFromCatch` emits code 401 → `decodeRpcResponse` rebuilds `status 401` → `handle401` matches (before: bare `Error` → code 500 → no match — exactly the reported bug). Backward compat holds over the wire too: the error name is carried via `data.name` and reconstructed client-side, so `isBlocksError(e, AuthOIDCErrors.NotAuthenticated)` keeps working — the fix strictly improves (adds the correct status, preserves everything name-based), no regression. Changeset `patch` is right for a genuine defect fix. Tests are meaningful — the browser suite's regression test pinning the pre-fix bare-`Error` shape is the strongest part. One optional test-completeness note inline.

packages/bb-auth-oidc/src/index.test.ts
sharonyajain: Thanks, Osama. The serialize→deserialize step is where the defect actually lived, so it's the right thing to scrutinize.

That round-trip is covered one layer down, in `packages/core/src/rpc.test.ts` ("round-trips status, name and retriable back into an ApiError on the client"), which drives errorResponseFromCatch → decodeRpcResponse and asserts both status and name survive the wire. The two tests here pin the `bb-auth-oidc` ends against that contract.

To assert the full requireAuth → wire → handle401 path from within bb-auth-oidc, we'd need to surface errorResponseFromCatch/decodeRpcResponse on core's public exports (they're currently internal), which trades a bit of encapsulation for the stronger guard. I'm happy to go either way if you'd prefer that end-to-end assertion, I'll expose the helpers and add it. Otherwise I'll leave the layering as-is.

osama-rizk (approved):

PR #439, opened 2026-08-25
feat(bb-lambda-compute): default the compute Lambda to arm64 (Graviton)

osama-rizk:
> **Stacked on #341** (`zimzha/multi-compute-a3-compute-resolution`). Review/merge #341 first; this PR's diff is only the arm64 delta on top. Base will retarget to `main` once #341 merges.

## What

Defaults `LambdaCompute`'s Lambda function to **arm64 (AWS Graviton)** and adds `architecture?: Architecture` to `LambdaComputeProps` as the override.

## Why

Board card: **`[Bug Bash] P3: All Lambdas pinned to x86_64 (Graviton ~20% cheaper)`** (P3, Cost/availability). arm64 Lambda is ~20% cheaper per GB-second at equivalent performance, and the Blocks backend is a pure-JavaScript esbuild bundle with no architecture-specific native dependencies — so the switch carries no runtime risk. Since `LambdaCompute` is the default compute injected by `@aws-blocks/blocks`, every app's shared handler runs on Graviton out of the box.

## Why here (not `BlocksDefaults`) — addressing @Simone319's review on #435

This supersedes #435, which put `lambdaArchitecture` in `BlocksDefaults`. Per Simone's feedback there, architecture is a **per-compute setting**, not a stack-wide default:
- `BlocksDefaults` fields either vary by stage (`removalPolicy`, `deletionProtection`, `pointInTimeRecovery`) or are cross-cutting posture. Architecture is neither — both presets would hold the *same* value.
- `memorySize`, `timeout`, and `runtime` already live on the compute construct, not in `BlocksDefaults`. Architecture is the same kind of knob, so it belongs alongside them — and `LambdaComputeProps` was already a placeholder explicitly reserved for "future options (e.g. memory, timeout)".

## How

- `LambdaComputeProps` gains `architecture?: Architecture` (documented, with an x86_64 override example).
- `LambdaCompute` applies `options?.architecture ?? Architecture.ARM_64` to its `NodejsFunction`.

## Testing

- New synth assertions in `index.cdk.test.ts`: the function defaults to `Architectures: ['arm64']`, and honors an `Architecture.X86_64` override.
- `@aws-blocks/bb-lambda-compute` suite (10 tests) ✅, `npm run lint` ✅, umbrella-changeset guard ✅.

## Changeset

`@aws-blocks/bb-lambda-compute` **minor** + `@aws-blocks/blocks` **patch** (umbrella injects `LambdaCompute` as the default compute).

> **Behavior change on next deploy:** the handler function's architecture is `arm64` (in-place update on an existing function).

### DIFF
--- a/.changeset/lambda-compute-arm64.md
+++ b/.changeset/lambda-compute-arm64.md
@@ -1,0 +1,12 @@
+---
+"@aws-blocks/bb-lambda-compute": minor
+"@aws-blocks/blocks": patch
+---
+
+`LambdaCompute`: default the function to **arm64 (AWS Graviton)**, with an `architecture` override.
+
+The compute's Lambda now defaults to `Architecture.ARM_64` instead of CDK's `x86_64` default. arm64 Lambda is ~20% cheaper per GB-second than x86_64 at equivalent performance, and the Blocks backend is a pure-JavaScript esbuild bundle with no architecture-specific native dependencies, so the switch carries no runtime risk. Because `LambdaCompute` is the default compute for every `@aws-blocks/blocks` app, the shared handler runs on Graviton out of the box.
+
+Adds `architecture?: Architecture` to `LambdaComputeProps` to override it (e.g. `Architecture.X86_64` when bundling an x86-only native addon).
+
+> **Behavior change on next deploy:** the handler function's architecture is `arm64` (an in-place update on an existing function).
--- a/.changeset/lambda-compute-arm64.md
+++ b/.changeset/lambda-compute-arm64.md
@@ -3,10 +3,10 @@
 "@aws-blocks/blocks": patch
 ---
 
-`LambdaCompute`: default the function to **arm64 (AWS Graviton)**, with an `architecture` override.
+`LambdaCompute`: default the function to **arm64 (AWS Graviton)**.
 
-The compute's Lambda now defaults to `Architecture.ARM_64` instead of CDK's `x86_64` default. arm64 Lambda is ~20% cheaper per GB-second than x86_64 at equivalent performance, and the Blocks backend is a pure-JavaScript esbuild bundle with no architecture-specific native dependencies, so the switch carries no runtime risk. Because `LambdaCompute` is the default compute for every `@aws-blocks/blocks` app, the shared handler runs on Graviton out of the box.
+The compute's Lambda now defaults to `Architecture.ARM_64` instead of CDK's `x86_64` default. arm64 Lambda is ~20% cheaper per GB-second than x86_64 at equivalent performance. Because `LambdaCompute` is the default compute for every `@aws-blocks/blocks` app, the shared handler runs on Graviton out of the box.
 
-Adds `architecture?: Architecture` to `LambdaComputeProps` to override it (e.g. `Architecture.X86_64` when bundling an x86-only native addon).
+The framework's own Building Blocks are pure-JavaScript esbuild bundles with no architecture-specific native code, so the switch is transparent for them. A backend `entry` bundle is the customer's own handler plus whatever they import, though — so an app that bundles an **x86-only native dependency** into its backend should be aware of the default. There is no per-app override yet; a customer-facing way to pin the architecture (`architecture` on `LambdaComputeProps` is present but internal for now) will be exposed alongside the public compute-configuration surface.
 
 > **Behavior change on next deploy:** the handler function's architecture is `arm64` (an in-place update on an existing function).
--- a/packages/bb-lambda-compute/src/types.ts
+++ b/packages/bb-lambda-compute/src/types.ts
@@ -1,10 +1,26 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import type { Architecture } from 'aws-cdk-lib/aws-lambda';
+
 /**
- * Options for constructing a `LambdaCompute`. Empty for now — a placeholder for
- * future options (e.g. memory, timeout), matching how other Building Blocks
- * carry an options bag.
+ * Options for constructing a `LambdaCompute`.
  */
-// biome-ignore lint/complexity/noBannedTypes: intentional placeholder options bag
-export type LambdaComputeProps = {};
+export interface LambdaComputeProps {
+	/**
+	 * The instruction-set architecture for the compute's Lambda function.
+	 * Defaults to **`Architecture.ARM_64`** (AWS Graviton): arm64 Lambda is ~20%
+	 * cheaper per GB-second than x86_64 at the same performance, and the Blocks
+	 * backend is a pure-JavaScript esbuild bundle with no architecture-specific
+	 * native dependencies, so the switch is free.
+	 *
+	 * Override to `Architecture.X86_64` if you bundle an x86-only native addon
+	 * into your backend:
+	 *
+	 * ```ts
+	 * import { Architecture } from 'aws-cdk-lib/aws-lambda';
+	 * new LambdaCompute(scope, 'compute', { architecture: Architecture.X86_64 });
+	 * ```
+	 */
+	architecture?: Architecture;
+}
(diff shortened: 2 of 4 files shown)


packages/bb-lambda-compute/src/types.ts
```diff
@@ -1,10 +1,26 @@
...
+	 * cheaper per GB-second than x86_64 at the same performance, and the Blocks
+	 * backend is a pure-JavaScript esbuild bundle with no architecture-specific
+	 * native dependencies, so the switch is free.
+	 *
+	 * Override to `Architecture.X86_64` if you bundle an x86-only native addon
+	 * into your backend:
+	 *
+	 * ```ts
+	 * import { Architecture } from 'aws-cdk-lib/aws-lambda';
+	 * new LambdaCompute(scope, 'compute', { architecture: Architecture.X86_64 });
```
soberm: This override isn't reachable by anyone today. The umbrella that builds the real default compute (`packages/blocks/src/index.cdk.ts`, outside this diff) does `new LambdaCompute(root as never, 'DefaultCompute')` with no third argument, and there's no field on `BlocksStackProps`/`BlocksBackendProps` that could carry `architecture` down to it. `LambdaCompute` is also still `@internal`, and the package README says not to depend on it directly, so this example is pointing people at the one thing the README tells them not to do.

Meanwhile the default change itself does apply to every existing app (`umbrella-compute.test.ts` confirms this is the real production default compute), so there's currently no way for an app with an architecture-specific dependency in its own handler bundle to fall back to x86_64 if this bites them. Would wire this through BlocksStackProps/BlocksBackendProps before merging, or at least scope the docs to "internal use only for now" and drop the customer-facing example until that wiring exists.

packages/bb-lambda-compute/src/types.ts
```diff
@@ -1,10 +1,26 @@
...
  */
-// biome-ignore lint/complexity/noBannedTypes: intentional placeholder options bag
-export type LambdaComputeProps = {};
+export interface LambdaComputeProps {
+	/**
+	 * The instruction-set architecture for the compute's Lambda function.
+	 * Defaults to **`Architecture.ARM_64`** (AWS Graviton): arm64 Lambda is ~20%
+	 * cheaper per GB-second than x86_64 at the same performance, and the Blocks
+	 * backend is a pure-JavaScript esbuild bundle with no architecture-specific
+	 * native dependencies, so the switch is free.
```
soberm: nit: "so the switch is free" reads a little informal next to the rest of this doc block. Optional.

.changeset/lambda-compute-arm64.md
```diff
@@ -0,0 +1,12 @@
+---
+"@aws-blocks/bb-lambda-compute": minor
+"@aws-blocks/blocks": patch
+---
+
+`LambdaCompute`: default the function to **arm64 (AWS Graviton)**, with an `architecture` override.
+
+The compute's Lambda now defaults to `Architecture.ARM_64` instead of CDK's `x86_64` default. arm64 Lambda is ~20% cheaper per GB-second than x86_64 at equivalent performance, and the Blocks backend is a pure-JavaScript esbuild bundle with no architecture-specific native dependencies, so the switch carries no runtime risk. Because `LambdaCompute` is the default compute for every `@aws-blocks/blocks` app, the shared handler runs on Graviton out of the box.
```
soberm: This is true for the framework's own bundle, but `entry` here is the customer's own handler module plus whatever they import into it, not just Blocks code, so "no architecture-specific native dependencies" isn't something the framework can actually guarantee for every app. Might be worth qualifying, something like "no risk for the framework's own building blocks; apps with architecture-specific native deps in their own backend should verify before upgrading," especially since there's no override path today if someone does hit it (see the types.ts comment).

soberm (changes requested):
This defaults LambdaCompute's NodejsFunction to arm64 (Graviton) instead of CDK's x86_64 default, and adds an `architecture` override on `LambdaComputeProps` to opt back out. Small, focused diff, and the two new synth tests cover both branches (default + override) properly.

One thing I think is worth resolving before this lands: I traced how the default compute actually gets built in production (`packages/blocks/src/index.cdk.ts`, outside this diff) and the umbrella wires it up as `new LambdaCompute(root as never, 'DefaultCompute')` with no options passed. Neither `BlocksStackProps` nor `BlocksBackendProps` has a field that reaches down to `architecture`, and `LambdaCompute` is still marked internal (README says not to depend on it directly). So the override this PR documents isn't actually usable by anyone right now, while the default switch itself already applies to every existing app, since `umbrella-compute.test.ts` confirms this is the real production default compute today. If an app has an architecture-specific dependency bundled into its own handler, there's no lever to pull if this breaks it. Left details on the types.ts comment.

Also left an inline note on the changeset: the "no runtime risk" framing covers the framework's own bundle, but the Lambda's entry point is the customer's own handler code plus whatever they import, so it's a narrower guarantee than it reads.

Two things unrelated to the diff itself, just flagging since I was in the area: the board card says "all Lambdas pinned to x86_64" but this only touches LambdaCompute; bb-distributed-data's migration Lambda and hosting's kv_keys_handler also create NodejsFunctions without setting `architecture`, so they'll still default to x86_64 after this merges, worth checking if that's a separate tracked follow-up. And bb-lambda-compute's README currently says "nothing in the default app path constructs it," which doesn't match the umbrella wiring mentioned above; not this PR's fault, just noticed it while I was in there.

Everything else here is solid: the JSDoc is thorough with a runnable example, `import type` vs `import` is used correctly in both files, and dropping the stale `biome-ignore` comment now that the props type is an actual interface instead of `{}` is a nice bit of cleanup.

Requesting changes mainly to get the override actually wired through the public props (or have the risk explicitly signed off on) before this reaches every existing deployment.

packages/bb-lambda-compute/src/types.ts
osama-rizk: Good call — you're right that the override isn't reachable and the example pointed at the `@internal` class. Rather than add a public compute-config prop now (which would front-run #341's deliberate decision to keep compute configuration internal — "no public option yet"), I've scoped this down (61dc6e40):

- Kept arm64 as the default (the actual win here).
- Re-documented `LambdaComputeProps.architecture` as `@internal` and dropped the customer-facing `new LambdaCompute(...)` example.
- Noted that the customer-facing override (pin architecture for a backend with an x86-only native addon) will be exposed **alongside the public compute-configuration surface** #341 introduces, where it sits naturally next to memory/timeout — rather than through the internal class.

So this PR ships just the default; the escape hatch lands with the surface that's meant to carry it. Does that split work for you, or would you rather hold the default flip until the override ships too?

packages/bb-lambda-compute/src/types.ts
osama-rizk: Fixed — dropped "so the switch is free"; the doc now just states the ~20% cost benefit.

.changeset/lambda-compute-arm64.md
osama-rizk: Agreed — qualified in 61dc6e40. The changeset now says the pure-JS/no-native-deps property holds for the framework's own Building Blocks, and explicitly calls out that a backend `entry` bundle is the customer's own code, so an app importing an x86-only native dep should be aware — with a note that a per-app override is coming with the public compute surface.

packages/bb-lambda-compute/src/types.ts
```diff
@@ -1,10 +1,23 @@
...
+	 * The instruction-set architecture for the compute's Lambda function.
+	 * Defaults to **`Architecture.ARM_64`** (AWS Graviton), which is ~20% cheaper
+	 * per GB-second than x86_64 at equivalent performance.
+	 *
+	 * @internal Not yet reachable by customers. `LambdaCompute` is internal and
+	 * the umbrella constructs the default compute with no options, so today this
+	 * only takes effect through the arm64 default. A customer-facing override
+	 * (e.g. for a backend that bundles an x86-only native addon) will be exposed
+	 * alongside the public compute-configuration surface, not through this
+	 * internal class directly.
```
Simone319: This is not accurate. The customers will be able to configure via this interface after launch.

osama-rizk: You're right — fixed in 139663d2. Dropped the "not through this internal class" framing. The doc now states this **is** the interface customers will configure the compute through once it's public (set `Architecture.X86_64` here for an x86-only native addon); it's just not wired to a customer entry point yet — pre-launch the umbrella builds the default compute with no options, so today only the arm64 default takes effect.

soberm (approved):

Simone319 (approved):

PR #442, opened 2026-08-25
feat(bb-async-job): auto-chunk submitBatch beyond SQS limits

sharonyajain:
## Problem

`AsyncJob.submitBatch` was hard-capped at 10 payloads and threw `BatchTooLarge` above that, so any bulk-upload caller with more than 10 items had to reimplement SQS's chunking rules by hand — the pain behind the "bulk-upload breaks end-to-end" reports.

Investigation note: the two specifics in the original report did **not** reproduce on `main`. The id `submit`/`submitBatch` returns is already the SQS `MessageId` and is exactly the key `getStatus()`/`waitUntilComplete()` use (no "locally generated" status keys), and the >10 case **threw** rather than silently truncating. The one real gap was the inability to submit more than 10 in a single call, addressed here. Regression tests pin the id==status-key invariant so it can't silently break later.

**Issue #, if available:** N/A (internal bug report, block `bb-async-job`)

## Changes

- **`submitBatch` auto-chunks** (`index.aws.ts`): new `chunkBatch()` packs payloads into `SendMessageBatch` requests bounded by *both* SQS per-request limits — ≤10 entries and ≤256 KB aggregate message body. Each batch entry `Id` is the payload's original index, so returned `jobIds` stay in input order and each is the real SQS `MessageId`.
- **Not atomic across chunks, throw preserved**: a multi-chunk submit issues several requests, so an earlier chunk can land before a later one fails. It still throws `AsyncJobErrors.BatchSubmitFailed` on any failure, but the error now carries the partial reality — `.jobIds` has the real MessageId for every enqueued entry (`null` at failed indexes) and `.failed[]` lists every failure across chunks, sorted by index — so a caller can retry only the failed indexes.
- **Mock runtime** (`index.mock.ts`): cap removed; now validates every payload before enqueuing any, matching the AWS runtime's fail-whole-batch behavior.
- **`BatchTooLarge`** retained but `@deprecated` (never thrown now) to avoid a breaking API removal.
- Docs: `README.md`, `DESIGN.md` (new decision **D-AJ-4a** on chunking + non-atomic partial failure), and a changeset (`@aws-blocks/bb-async-job` + `@aws-blocks/blocks`, patch).

## Validation

- `tsc --build` clean; **69/69** unit tests pass (was 42).
- New `index.aws.test.ts` (stubbed SQS client): 25→`[10,10,5]` chunking, unique batch-entry Ids per request, 256 KB byte-limit split (two ~200 KB payloads → 2 requests), and cross-chunk partial failure asserting `.jobIds`/`.failed`.
- Mock suite: replaced the old `BatchTooLarge` reject test with an auto-chunk test; added two `jobId == getStatus() key` regression tests for `submit` and `submitBatch`.
- `biome lint` clean on changed files (0 errors).
- Not deploy-verified against live SQS.

## Checklist

- [x] PR description included

---

As part of: https://github.com/orgs/aws-amplify/projects/141/views/15?pane=issue&itemId=195525521


packages/bb-async-job/src/index.aws.ts
```diff
@@ -227,41 +256,48 @@ export class AsyncJob<T = unknown> extends Scope {
...
-		);
+		// SQS caps a SendMessageBatch at 10 entries and 256 KB, so a batch of any
+		// size is split across as many requests as those limits require. The SQS
+		// batch-entry `Id` carries the payload's original index, which is how each
+		// returned MessageId (or failure) is mapped back into input order below.
+		const chunks = this.chunkBatch(messageBodies);
+		const jobIds: Array<string | null> = new Array(payloads.length).fill(null);
+		const failed: BatchSubmitResult['failed'] = [];
+
+		for (const indices of chunks) {
```
soberm: [major] This sends chunks one at a time with a sequential `await` in the loop. For a batch big enough to need several chunks (say 1000 payloads → 100 chunks), that's 100 sequential SQS round trips instead of overlapping them. That's a real throughput hit for exactly the bulk-upload case this PR is meant to fix. Might be worth firing these with bounded concurrency (a few in flight at once) rather than fully serial. If serial is intentional (e.g. to keep the SQS client's in-flight request count predictable), a comment here would help — right now it reads like it just happened this way.

soberm: [question] If an early chunk fails, do we still want to keep firing the rest? Looks like yes here (matches the "not atomic, throw at the end" design), which is fine for per-message failures, but if the SQS client itself starts erroring at the connection level, this would hammer the same broken endpoint for every remaining chunk before finally throwing. Not necessarily a blocker, just wondering if that's been thought through or if it's worth a short-circuit on transport-level errors.

packages/bb-async-job/src/index.aws.ts
```diff
@@ -216,6 +216,35 @@ export class AsyncJob<T = unknown> extends Scope {
...
 
+	/**
+	 * Group message indices into `SendMessageBatch` requests bounded by both limits
+	 * SQS enforces per request: at most {@link MAX_BATCH_SIZE} entries, and at most
+	 * {@link MAX_PAYLOAD_BYTES} of aggregate message body. `validatePayload` has
+	 * already rejected any single body over that byte limit, so a message that
+	 * would overflow the running chunk simply starts the next one — it can always
+	 * fit in a chunk of its own.
+	 */
+	private chunkBatch(bodies: string[]): number[][] {
```
soberm, line 235: [minor] Now that the old 10-item cap is gone, there's no upper bound on `payloads.length` at all. Combined with the sequential-chunk-send comment below, someone passing a very large array (thousands of items) will kick off a correspondingly large number of sequential SQS requests with no limit, timeout, or backpressure. Even a generous soft cap with a clear error would be safer than fully unbounded.

packages/bb-async-job/src/status.ts
```diff
@@ -260,8 +260,9 @@ export class JobStatusTracker {
 	 *
 	 * Deliberately individual conditional writes rather than one `putBatch`:
 	 * DynamoDB's `BatchWriteItem` cannot carry a condition expression, so a batch
-	 * write would reintroduce the clobber described on {@link recordQueued}. A
-	 * batch is at most 10 items, so the puts are issued in parallel.
+	 * write would reintroduce the clobber described on {@link recordQueued}. The
+	 * puts are issued in parallel; since `submitBatch` auto-chunks there is no
+	 * 10-item cap on how many jobs arrive here.
 	 */
 	async recordQueuedBatch(jobs: Array<{ jobId: string; submittedAt: string }>): Promise<void> {
```
soberm, line 279: [major] This is the flip side of removing the 10-item batch cap: `recordQueuedBatch` used to fan out at most 10 concurrent conditional `PutItem` calls, and now it can fan out to however many jobs a single (now-unbounded) `submitBatch` call enqueues. `Promise.all` over all of them at once risks throttling (`ProvisionedThroughputExceededException`) on tables without a lot of burst capacity, right when a caller is doing exactly the kind of bulk write this PR was meant to unblock. Worth chunking these writes into fixed-size groups (25 at a time would mirror DynamoDB's own `BatchWriteItem` limit) instead of firing everything in parallel.

soberm (commented):
Solid fix for the real gap here (10-item cap forcing callers to hand-roll SQS chunking), and the test coverage on the tricky parts (byte-limit splits, cross-chunk partial failure, id-to-status invariants) is thorough. Two throughput/throttling concerns worth a look before this goes out at scale, plus a couple of smaller notes below.

packages/bb-async-job/src/index.mock.ts
```diff
@@ -225,16 +224,17 @@ export class AsyncJob<T = unknown> extends Scope {
 	 * Enqueue multiple jobs in a single call.
 	 *
 	 * In AWS, uses SQS's native `SendMessageBatch` for better throughput than
```
soberm: [nit] Minor wording nit — this doc comment leads with "In AWS, uses..." inside the mock implementation's own JSDoc, which reads a little oddly for a comment describing the mock's contract. Not blocking, just a readability thing.

packages/bb-async-job/src/index.aws.test.ts
```diff
@@ -0,0 +1,133 @@
...
+	DelaySeconds?: number;
+}
+
+/**
+ * Build an AsyncJob wired to a stub SQS client. `registerSdkIdentifiers` seeds the
+ * queue URL keyed by the block's fullId so `ensureQueueUrl()` passes without a
+ * deployed stack; the stub replaces the real client so nothing hits the network.
+ */
+function makeJob(failIds: Set<string> = new Set()) {
+	const scope = new Scope(`aj-${Math.random().toString(36).slice(2)}`);
```
soberm, line 31: [nit] Using `Math.random()` for the scope suffix is a bit unusual for test isolation but harmless here since it's test-only and just needs uniqueness, not cryptographic randomness.

bobbor (commented):
Design and tests look solid — chunking by both SQS limits is correct (including the exactly-256KB-fits-alone edge), the id==MessageId invariant regression tests are a good catch, and D-AJ-4a documents the non-atomicity honestly. CI is green. Two gaps in the new partial-failure contract below, complementing @soberm's comments (which I agree with, especially bounded concurrency in `recordQueuedBatch`).

packages/bb-async-job/src/index.aws.ts
```diff
@@ -227,41 +256,48 @@ export class AsyncJob<T = unknown> extends Scope {
...
+		// SQS caps a SendMessageBatch at 10 entries and 256 KB, so a batch of any
+		// size is split across as many requests as those limits require. The SQS
+		// batch-entry `Id` carries the payload's original index, which is how each
+		// returned MessageId (or failure) is mapped back into input order below.
+		const chunks = this.chunkBatch(messageBodies);
+		const jobIds: Array<string | null> = new Array(payloads.length).fill(null);
+		const failed: BatchSubmitResult['failed'] = [];
+
+		for (const indices of chunks) {
+			const result = await this._sqsClient.send(new SendMessageBatchCommand({
```
bobbor: [major] The partial-failure contract only survives *entry-level* failures. If `send` itself rejects (throttling, connection error, access denied) on chunk 2+, the raw SDK error propagates with no `.jobIds` / `.failed` attached — the caller can't tell that earlier chunks already landed, and the documented recovery ("retry only the failed indexes") becomes impossible. Re-submitting the whole batch double-enqueues everything that succeeded, which is exactly the failure mode this PR set out to make recoverable.

Suggest wrapping the send in try/catch: on a rejected send, push every index of that chunk into `failed` with the transport error's code/message, then continue (or short-circuit — this is also the natural place to resolve @soberm's question: mark all *remaining* indices failed and stop hammering a broken endpoint), and throw `BatchSubmitFailed` at the end carrying the real partial `jobIds`. A test with a stub whose `send` rejects on the second chunk would pin it.

packages/bb-async-job/src/index.aws.ts
```diff
@@ -227,41 +256,48 @@ export class AsyncJob<T = unknown> extends Scope {
...
+			for (const f of result.Failed ?? []) {
+				failed.push({
+					index: parseInt(f.Id!, 10),
+					code: f.Code ?? 'UnknownError',
+					message: f.Message ?? 'Unknown error',
+				});
+			}
+		}
 
 		if (failed.length > 0) {
```
bobbor, line 368: [minor] Related, just below this hunk: on the full-success path, `await this._status.recordQueuedBatch(...)` runs *after* every message is already enqueued. Pre-existing, but this PR amplifies it — with the 10-item cap gone the fan-out is unbounded (see @soberm's comment on status.ts), so a throttling throw there is much more likely, and it makes `submitBatch` reject with a generic error carrying no `jobIds` even though the enqueue fully succeeded. A caller treating that as a submit failure and retrying double-enqueues the entire batch.

Since the handler already backfills `queued` records when it first sees a job — the partial-failure path in D-AJ-4a explicitly relies on this — the status write could be best-effort here (log and still return `{ jobIds, failed: [] }`), or at minimum the thrown error should carry `jobIds`.

packages/bb-async-job/src/index.aws.ts
```diff
@@ -227,41 +256,48 @@ export class AsyncJob<T = unknown> extends Scope {
...
+					Id: String(i),
+					MessageBody: messageBodies[i],
+					DelaySeconds: options?.delaySeconds ?? 0,
+				})),
+			}));
+
+			for (const s of result.Successful ?? []) {
+				jobIds[parseInt(s.Id!, 10)] = s.MessageId!;
+			}
+			for (const f of result.Failed ?? []) {
```
bobbor: [minor] If SQS ever returns an entry in *neither* `Successful` nor `Failed` (or with an `Id` that doesn't parse back to an index), that index silently stays `null` in `jobIds` while `failed` stays empty — the call returns as a full success containing a `null` id, which violates the documented contract (`null` only at failed indexes, and only on the thrown error). Cheap defense: after the loop, any index still `null` and not in `failed` gets a synthetic `failed` entry (`code: 'MissingResult'`).

packages/bb-async-job/src/index.aws.ts
sharonyajain: Fixed in 3b40b21 — chunks now send with bounded concurrency (`MAX_BATCH_CONCURRENCY = 5`) via a small worker pool rather than a fully serial loop, so a 1,000-payload batch overlaps requests instead of doing 100 sequential round trips. The limit keeps the SQS client's in-flight count predictable (and bounds the blast radius on an unhealthy endpoint — see the transport short-circuit below). Covered by a new test asserting peak concurrency is >1 and ≤5.

sharonyajain: Good call, handled in 3b40b21. A *transport-level* `send()` rejection (throttling, connection, auth) now sets an `aborted` flag: chunks already in flight settle, but chunks not yet started are short-circuited (marked failed with `code: 'BatchSubmitAborted'`) instead of hammering the broken endpoint. *Entry-level* failures still don't short-circuit — they're per-message and don't implicate the rest of the batch. Both feed the thrown error's partial `jobIds`/`failed`.

packages/bb-async-job/src/index.aws.ts
sharonyajain: Added in 3b40b21 — a generous soft cap `MAX_BATCH_PAYLOADS = 10,000`. Over the cap throws `BatchTooLarge` before any validation or send, as a guardrail against unbounded fan-out. (`BatchTooLarge` is repurposed for this rather than removed.) Mirrored in the mock runtime, with a parity test on both.

packages/bb-async-job/src/status.ts
sharonyajain: Fixed in 3b40b21 — `recordQueuedBatch` no longer does one `Promise.all` over the whole batch. It now issues the conditional writes in fixed groups of `STATUS_WRITE_GROUP = 25` (mirroring DynamoDB's `BatchWriteItem` limit), so a large `submitBatch` fan-out can't fire thousands of `PutItem`s at once and throttle the table.

packages/bb-async-job/src/index.mock.ts
sharonyajain: Reworded in 3b40b21 — the mock's JSDoc now leads with the mock's own behavior ("enqueues each payload via `submit()` in turn") and mentions the AWS `SendMessageBatch` path second.

packages/bb-async-job/src/index.aws.test.ts
sharonyajain: Acknowledged — left as-is. It's test-only and just needs per-run uniqueness for the scope id, not randomness quality. Happy to switch it to `randomUUID()` if you'd prefer the consistency.

packages/bb-async-job/src/index.aws.ts
sharonyajain: Fixed in 3b40b21 — this was the most important one. Each chunk send is now wrapped in try/catch: on a rejected `send()` every index in that chunk is pushed into `failed` with the transport error's `name`/`message`, the remaining not-yet-started chunks are short-circuited (`BatchSubmitAborted`), and the call throws `BatchSubmitFailedError` carrying the real partial `jobIds` (with the landed chunks' MessageIds intact). So "retry only the failed indexes" now holds even for a transport failure on chunk 2+. Pinned by a new test whose stub rejects on the second chunk and asserts chunk 1's ids survive, chunk 2 carries the throttling code, and later chunks are aborted.

packages/bb-async-job/src/index.aws.ts
sharonyajain: Fixed in 3b40b21 — the success-path `recordQueuedBatch` is now best-effort: wrapped in try/catch and logged, not thrown. Since the handler backfills a `queued` record when it first sees a job (the same mechanism D-AJ-4a's partial-failure path relies on), a throttled status write no longer turns a fully-successful enqueue into a reject that would make the caller re-submit and double-enqueue. New test injects a tracker whose `recordQueuedBatch` rejects and asserts `submitBatch` still resolves with the jobIds.

packages/bb-async-job/src/index.aws.ts
sharonyajain: Fixed in 3b40b21 — after the send loop, any index still `null` that isn't already in `failed` gets a synthetic `{ code: 'MissingResult' }` failure, so an entry SQS returns in neither list (or with an unparseable Id) surfaces as a failure instead of a "success" containing a `null` id. New test drops an entry from both result lists and asserts it becomes a MissingResult failure.

bobbor:
Re-reviewed 3b40b21 — all three of my comments are properly resolved, and the fixes go beyond band-aids:

- **Transport failures**: `sendChunksBounded` now catches a rejected `send()`, fails that chunk's indexes with the error's name/message, and short-circuits not-yet-started chunks (`BatchSubmitAborted`) while letting in-flight ones settle — neatly answering @soberm's short-circuit question in the same mechanism. The dedicated `BatchSubmitFailedError` class (exported from all three runtime entry points, browser included) is a nicer contract than the ad-hoc `(err as any)` property stuffing it replaces.
- **Best-effort status write**: success-path `recordQueuedBatch` is try/caught and logged via the block's logger; test injects a rejecting tracker and asserts the submit still resolves. ✅
- **MissingResult defense**: post-loop sweep with test coverage. ✅

Also verified the responses to @soberm's threads: worker-pool concurrency capped at `MAX_BATCH_CONCURRENCY = 5` (single shared `next` cursor — safe in single-threaded JS; `failed` push order is nondeterministic across workers but the final sort restores index order), `MAX_BATCH_PAYLOADS = 10,000` soft cap with the `BatchTooLarge` constant repurposed and its stale `@deprecated` JSDoc correctly removed, `STATUS_WRITE_GROUP = 25` grouping in `recordQueuedBatch`, and README/DESIGN/changeset all updated consistently (README re-lists `BatchTooLarge` with its new meaning).

**One blocker — CI is red on this revision.** `Build, Unit Tests, E2E Local` fails at the API-report check:

```
❌ API reports are out of date:
   • packages/bb-async-job/API.md
Run 'npm run update:api' and commit the updated API.md files.
```

Expected, since `BatchSubmitFailedError` is a new public export. Please run `npm run update:api` and commit the regenerated `API.md` (the `Build and Test` failure is just the aggregator mirroring it). After that this looks ready from my side.

sharonyajain:
@bobbor thanks for the thorough re-review  and good catch on the API report.

Done in 5e3dd7b: ran `npm run update:api` and committed the regenerated `packages/bb-async-job/API.md` (it now carries the new `BatchSubmitFailedError` export). CI is green on that revision — `Build, Unit Tests, E2E Local` (the `check:api` step included) and the `Build and Test` aggregator. Should be ready from your side now.

bobbor (approved):
thanks. looks good to me

soberm (approved):

PR #448, opened 2026-08-27
fix(core): stop leaking internal exception details in RPC error responses

### LINKED ISSUE #227: RPC handler leaks raw backend exception class names and messages to clients
## Summary

`errorResponseFromCatch` in `packages/core/src/rpc.ts` forwards the raw `error.message` and `error.name` from any non-`ApiError` throw straight into the JSON-RPC error response returned to the client. When the underlying failure comes from a driver or SDK (Postgres, DynamoDB, etc.), the internal exception class name and raw message end up on the wire. This is both an information-disclosure concern and an error-quality problem.

## Where

- `packages/core/src/rpc.ts` — `errorResponseFromCatch()` sets `message = error.message` for any `Error`, and copies `error.name` into `data.name` whenever it isn't the generic `Error`.
- Emitted to clients by:
  - `packages/core/src/lambda-handler.ts` (RPC error branch)
  - `packages/core/src/scripts/dev-server.ts` (dev-server RPC branch)

## Impact

- Non-`ApiError` throws (the common case for driver/SDK failures) return an HTTP 200 JSON-RPC error whose `message` is the raw driver message and whose `data.name` is the raw exception class name.
- Leaks internal implementation details to any caller and produces unstable, non-actionable error shapes for clients.

## Expected

Only `ApiError` (or an explicitly allowed shape) should reach the client verbatim. Everything else should collapse to a stable, generic error (e.g. code `500`, `message: "Internal error"`), with the detailed error logged server-side only.

## Suggested fix

Sanitize the non-`ApiError` branch of `errorResponseFromCatch` so unknown errors map to a generic message/code and never copy `error.name` or the raw `error.message` onto the wire.

osama-rizk:
Fixes #227.

## Problem

`errorResponseFromCatch` (`packages/core/src/rpc.ts`) forwarded the raw `error.message` — and, for non-generic names, the exception **class name** as `data.name` — of *any* thrown value into the JSON-RPC error response. For the common case (a driver/SDK failure rather than an `ApiError`), that put internal implementation details on the wire to every caller: table names, SQL fragments, ARNs, `$metadata`, and the internal exception class name. It's both an information-disclosure issue (AGENTS core rules 5–6) and an error-quality one — clients got unstable, non-actionable shapes.

## Fix

Only `ApiError` reaches the client verbatim — its `status`, `message`, `name`, and `retriable` are the deliberate, `isBlocksError`-matchable public contract. Every other throw now collapses to a stable generic `{ code: 500, message: "Internal error" }` with no `data`. The full error is still logged server-side (`lambda-handler.ts:558` `console.error('Lambda Error:', …)` and `dev-server.ts` `[rpc-err]`), so debuggability is unchanged.

## Test (red → green)

Added two cases to `rpc.test.ts` (both fail on the old code):
- a non-`ApiError` throw encodes as `500` / `"Internal error"` with no `data` — previously the raw message (`"plain"`) leaked through unchecked;
- a driver-shaped error (`name = "ResourceNotFoundException"`, message embedding `Table: app-notes-prod`) never puts the class name or raw message on the wire.

`ApiError` round-tripping (status ↔ code, `name`/`retriable` in `data`, client re-hydration via `decodeRpcResponse`/`isBlocksError`) is unchanged and still covered.

## Verification

- `packages/core` unit suite: **699 pass / 0 fail**; `rpc.test.ts` 24/24.
- `npm run lint` 0 errors; umbrella-changeset guard ✅.

## Changeset

`@aws-blocks/core` patch + `@aws-blocks/blocks` patch (umbrella re-export).

> **Behavior change:** clients that previously read the raw `message`/`data.name` of a *non-`ApiError`* failure now receive `"Internal error"` / code `500`. To send a specific, client-safe message, throw an `ApiError` (that path is unchanged). This is the intended security fix, not a type/API-signature change.

### DIFF
--- a/.changeset/rpc-sanitize-internal-errors.md
+++ b/.changeset/rpc-sanitize-internal-errors.md
@@ -1,0 +1,12 @@
+---
+"@aws-blocks/core": patch
+"@aws-blocks/blocks": patch
+---
+
+`ApiNamespace` RPC: stop leaking internal exception details to clients.
+
+`errorResponseFromCatch` previously forwarded the raw `error.message` (and, for non-generic names, the exception class name in `data.name`) of **any** thrown value into the JSON-RPC error response. For the common case — a driver/SDK failure (Postgres, DynamoDB, …) rather than an `ApiError` — that put internal implementation details (table names, SQL fragments, ARNs, exception class names) on the wire to every caller.
+
+Now only `ApiError` reaches the client verbatim (its status, message, `name`, and `retriable` flag are the public, `isBlocksError`-matchable contract). Every other throw collapses to a stable generic `{ code: 500, message: "Internal error" }` with no `data`. The full error is still logged server-side by the Lambda handler and the dev server, so debuggability is unchanged.
+
+> Behavior change: clients that previously read the raw `message`/`data.name` of a non-`ApiError` failure now receive `"Internal error"` / code `500`. Throw an `ApiError` to send a specific, client-safe message.
--- a/packages/core/src/rpc.test.ts
+++ b/packages/core/src/rpc.test.ts
@@ -158,9 +158,28 @@ describe('ApiError status ↔ JSON-RPC error code', () => {
     assert.strictEqual(parsed.error.data.retriable, true);
   });
 
-  it('encodes a non-ApiError throw as code 500 with no data.name', () => {
+  it('collapses a non-ApiError throw to a generic 500 with no message or name leak', () => {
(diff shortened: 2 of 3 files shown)


osama-rizk:
Closing in favor of https://github.com/aws-devtools-labs/aws-blocks/pull/231

osama-rizk:
Closed as a duplicate of #231 (@pranavosu), which already implements the same fix. I missed it before opening this. #231 is currently CONFLICTING and needs a rebase — happy to help land it rather than duplicate.

PR #449, opened 2026-08-27
fix(bb-knowledge-base): validate non-integer retrieve maxResults

### LINKED ISSUE #430: bug(bb-knowledge-base): validate non-integer retrieve maxResults
## Problem

`RetrieveOptions.maxResults` is documented as clamped to the 1-100 range, but the current normalization preserves non-integer and non-finite values:

```ts
const maxResults = Math.min(Math.max(options?.maxResults ?? 10, 1), 100);
```

This makes the mock and AWS runtimes diverge for values that can occur when an option is derived from user input or a calculation.

## Reproduction

Using the current source:

- `maxResults: 1.5`
  - The mock passes `1.5` into its local search; array slicing truncates it and returns one result in the fixture used for verification.
  - The AWS runtime constructs a `RetrieveCommand` with `numberOfResults: 1.5`.
- `maxResults: NaN`
  - The mock search produces an empty result set because the search limit becomes `NaN`.
  - The AWS runtime constructs a `RetrieveCommand` with `numberOfResults: NaN`.

Amazon Bedrock defines `numberOfResults` as an integer with a valid range of 1-100, so neither value is a valid production request:
https://docs.aws.amazon.com/goto/SdkForJavaScriptV3/bedrock-agent-runtime-2023-07-26/Retrieve

## Expected behavior

Both runtimes should handle invalid `maxResults` values consistently before searching or sending a Bedrock request.

The existing behavior for finite integer values outside the range can remain clamping to 1-100. For fractional or non-finite values, I recommend rejecting the call with the existing `KnowledgeBaseErrors.ValidationError`, rather than silently rounding or producing a mock-only result.

## Proposed scope

- Add a shared normalization/validation helper used by `index.mock.ts` and `index.aws.ts`.
- Preserve the documented clamp for finite integers.
- Reject fractional and non-finite values with `ValidationError`.
- Add mock/AWS-command parity tests and clarify the option documentation.

osama-rizk:
Fixes #430.

## Problem

`KnowledgeBase.retrieve` normalized `maxResults` with `Math.min(Math.max(options?.maxResults ?? 10, 1), 100)` in **both** `index.mock.ts` and `index.aws.ts`. That clamp silently preserves fractional and non-finite values:

- `maxResults: 1.5` → the mock slices its result array by `1.5`; the AWS runtime sends `numberOfResults: 1.5`.
- `maxResults: NaN` → the mock returns an empty set; the AWS runtime sends `numberOfResults: NaN`.

Bedrock defines `numberOfResults` as an **integer** in 1–100, so neither is a valid request — and the mock and AWS runtimes diverge on exactly these inputs (which can arise when the option comes from user input or a calculation).

## Fix

A shared `normalizeMaxResults` helper (new `normalize.ts`), used by both runtimes:

- `undefined` → default `10`.
- Finite integer → clamped to 1–100 (documented behavior preserved; `0`/`-5` → `1`, `200` → `100`).
- Fractional or non-finite (`1.5`, `NaN`, `±Infinity`) → rejected with `KnowledgeBaseErrors.ValidationError`, **before** any local search or Bedrock request.

Also hoisted the `blocksError` helper (duplicated verbatim in both entries) into `errors.ts` as the single source, and clarified the `RetrieveOptions.maxResults` doc.

## Tests (red → green)

- `normalize.test.ts` — default, pass-through, clamp bounds, and rejection of `1.5`/`NaN`/`±Infinity`.
- `index.mock.test.ts` — `retrieve` rejects `1.5`/`NaN` with `ValidationError` (parity with AWS).
- `index.aws.test.ts` — `retrieve` rejects `1.5`/`NaN` with `ValidationError` and sends **no** `RetrieveCommand`.

## Verification

- `@aws-blocks/bb-knowledge-base` suite: **139 pass / 0 fail**.
- Conditional-export parity 21/21; `check:api` ✅ (no public-API change — helper stays internal); `npm run lint` 0 errors; umbrella-changeset guard ✅.

## Changeset

`@aws-blocks/bb-knowledge-base` patch + `@aws-blocks/blocks` patch (umbrella re-export).

### DIFF
--- a/packages/bb-knowledge-base/src/errors.ts
+++ b/packages/bb-knowledge-base/src/errors.ts
@@ -34,11 +34,14 @@ export const KnowledgeBaseErrors = {
 	Timeout: 'KnowledgeBaseTimeoutException',
 } as const;
 
+/** One of the `name` values in {@link KnowledgeBaseErrors}. */
+export type KnowledgeBaseErrorName = (typeof KnowledgeBaseErrors)[keyof typeof KnowledgeBaseErrors];
+
 /**
  * Build a typed `Error` whose `name` is one of {@link KnowledgeBaseErrors}, so it
  * crosses the RPC wire by `name` and is matchable with `isBlocksError()`.
  */
-export function blocksError(name: string, message: string): Error {
+export function blocksError(name: KnowledgeBaseErrorName, message: string): Error {
 	const err = new Error(`${name}: ${message}`);
 	err.name = name;
 	return err;
--- a/packages/bb-knowledge-base/src/errors.ts
+++ b/packages/bb-knowledge-base/src/errors.ts
@@ -33,3 +33,13 @@ export const KnowledgeBaseErrors = {
 	/** `waitUntilSynced()` exceeded its timeout before the knowledge base finished ingesting. */
 	Timeout: 'KnowledgeBaseTimeoutException',
 } as const;
+
+/**
+ * Build a typed `Error` whose `name` is one of {@link KnowledgeBaseErrors}, so it
+ * crosses the RPC wire by `name` and is matchable with `isBlocksError()`.
+ */
+export function blocksError(name: string, message: string): Error {
+	const err = new Error(`${name}: ${message}`);
+	err.name = name;
+	return err;
+}
--- a/packages/bb-knowledge-base/src/index.aws.test.ts
+++ b/packages/bb-knowledge-base/src/index.aws.test.ts
@@ -5,6 +5,7 @@ import { test, describe, mock, afterEach } from 'node:test';
 import assert from 'node:assert';
 import { BedrockAgentRuntimeClient } from '@aws-sdk/client-bedrock-agent-runtime';
 import { BedrockAgentClient } from '@aws-sdk/client-bedrock-agent';
+import { isBlocksError } from '@aws-blocks/core';
 import { KnowledgeBaseErrors, KnowledgeBase } from './index.aws.js';
 
 // ── SDK mock helpers ───────────────────────────────────────────────────────
@@ -365,7 +366,7 @@ describe('retrieve (SDK-mocked)', () => {
 			for (const bad of [1.5, Number.NaN]) {
 				await assert.rejects(
 					() => kb.retrieve('query', { maxResults: bad }),
-					(e: unknown) => (e as Error).name === KnowledgeBaseErrors.ValidationError,
+					(e: unknown) => isBlocksError(e, KnowledgeBaseErrors.ValidationError),
 					`maxResults=${bad} should reject with ValidationError`,
 				);
 			}
--- a/packages/bb-knowledge-base/src/index.aws.test.ts
+++ b/packages/bb-knowledge-base/src/index.aws.test.ts
@@ -351,6 +351,30 @@ describe('retrieve (SDK-mocked)', () => {
 		}
 	});
 
+	test('rejects a fractional or non-finite maxResults before sending a Bedrock request', async () => {
+		const cleanup = setKbEnv('TEST', 'R9b');
+		let sent = false;
+		mockRuntimeSend(() => {
+			sent = true;
+			return { retrievalResults: [] };
+		});
+
+		try {
+			const kb = new KnowledgeBase({ id: 'test' }, 'r9b', { source: './knowledge' });
+
+			for (const bad of [1.5, Number.NaN]) {
+				await assert.rejects(
+					() => kb.retrieve('query', { maxResults: bad }),
+					(e: unknown) => (e as Error).name === KnowledgeBaseErrors.ValidationError,
+					`maxResults=${bad} should reject with ValidationError`,
+				);
+			}
+			assert.strictEqual(sent, false, 'no RetrieveCommand should be sent for an invalid maxResults');
+		} finally {
+			cleanup();
+		}
+	});
+
 	test('handles empty retrievalResults', async () => {
 		const cleanup = setKbEnv('TEST', 'R10');
 		mockRuntimeSend(() => ({ retrievalResults: [] }));
--- a/packages/bb-knowledge-base/src/index.aws.ts
+++ b/packages/bb-knowledge-base/src/index.aws.ts
@@ -279,7 +279,7 @@ export class KnowledgeBase extends Scope {
 	 * @param query - Natural language search query. Must be non-empty.
 	 * @param {RetrieveOptions} options - Optional retrieval parameters (maxResults, filter).
 	 * @returns Chunks ranked by relevance score (highest first). Empty array if no matches.
-	 * @throws {KnowledgeBaseValidationError} If query is empty or whitespace-only.
+	 * @throws {KnowledgeBaseValidationError} If query is empty or whitespace-only, or `maxResults` is a non-integer.
 	 * @throws {KnowledgeBaseNotReadyException} If the KB has not been created/deployed.
 	 * @throws {InvalidFilterException} If the filter keys are invalid for the Bedrock query.
 	 * @throws {RetrievalFailedException} For other Bedrock retrieval errors (network, service).
@@ -297,8 +297,7 @@ export class KnowledgeBase extends Scope {
 			throw blocksError(KnowledgeBaseErrors.ValidationError, 'Query must be a non-empty string.');
 		}
 
-		// Bedrock API limits numberOfResults to an integer in 1–100 (well within
-		// Lambda's 6 MB response payload); reject fractional/non-finite inputs.
+		// 1–100 results stays well within Lambda's 6 MB response payload.
 		const maxResults = normalizeMaxResults(options?.maxResults);
 		const filter = buildFilter(options?.filter);
 		const knowledgeBaseId = this.ensureKbId();
--- a/packages/bb-knowledge-base/src/index.aws.ts
+++ b/packages/bb-knowledge-base/src/index.aws.ts
@@ -22,7 +22,8 @@ import type {
 	MetadataFilter,
 	WaitUntilSyncedOptions,
 } from './types.js';
-import { KnowledgeBaseErrors } from './errors.js';
+import { blocksError, KnowledgeBaseErrors } from './errors.js';
+import { normalizeMaxResults } from './normalize.js';
 import { BB_NAME, BB_VERSION } from './version.js';
 import { Logger } from '@aws-blocks/bb-logger';
 import type { ChildLogger } from '@aws-blocks/bb-logger';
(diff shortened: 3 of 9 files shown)


svidgen (dismissed):

soberm (commented):
This fixes #430 cleanly. `KnowledgeBase.retrieve()`'s `maxResults` was clamped with `Math.min(Math.max(v ?? 10, 1), 100)`, duplicated in both the mock and AWS runtimes, which silently let non-integer values (`1.5`, `NaN`, `±Infinity`) through to code that can't handle them. Routing both runtimes through one shared `normalizeMaxResults()` fixes the actual bug and, more importantly, removes the structural reason it could happen again in the first place (two people keeping two copies of a clamp expression in sync). Nice bonus cleanup hoisting the duplicated `blocksError()` into errors.ts while in there too. Left a few inline notes below, nothing blocking.

One thing worth checking before merging: this PR currently shows 3 failing checks (E2E Production, Telemetry E2E, and the Build and Test gate that depends on them). E2E Production fails on a `greaterThan` DynamoDB GSI assertion in `distributed-table.test.ts`, and Telemetry E2E fails on an `UnresolvedAccount`/telemetry-delivery timeout in `telemetry-e2e.test.ts` — neither touches bb-knowledge-base or maxResults, so this looks like pre-existing E2E flakiness rather than something this change caused, but worth confirming (e.g. whether main is currently green) before merging.

packages/bb-knowledge-base/src/index.aws.ts
```diff
@@ -302,8 +297,9 @@ export class KnowledgeBase extends Scope {
 			throw blocksError(KnowledgeBaseErrors.ValidationError, 'Query must be a non-empty string.');
 		}
 
-		// Bedrock API limits numberOfResults to 1–100. Well within Lambda's 6 MB response payload.
-		const maxResults = Math.min(Math.max(options?.maxResults ?? 10, 1), 100);
+		// Bedrock API limits numberOfResults to an integer in 1–100 (well within
+		// Lambda's 6 MB response payload); reject fractional/non-finite inputs.
+		const maxResults = normalizeMaxResults(options?.maxResults);
```
soberm, line 301: [suggestion] The `@throws {KnowledgeBaseValidationError}` line in this method's JSDoc (line 282) still just says "If query is empty or whitespace-only." Worth adding that it now also covers a non-integer `maxResults`, to match the `RetrieveOptions.maxResults` doc update in types.ts.

packages/bb-knowledge-base/src/index.mock.ts
```diff
@@ -220,7 +215,7 @@ export class KnowledgeBase extends Scope {
 			throw blocksError(KnowledgeBaseErrors.ValidationError, 'Query must be a non-empty string.');
 		}
 
-		const maxResults = Math.min(Math.max(options?.maxResults ?? 10, 1), 100);
+		const maxResults = normalizeMaxResults(options?.maxResults);
```
soberm, line 218: [suggestion] Same as the AWS runtime — the `@throws {KnowledgeBaseValidationError}` line in this method's JSDoc (line 199) wasn't updated to mention the new maxResults check, only the empty-query case.

packages/bb-knowledge-base/src/normalize.ts
```diff
@@ -0,0 +1,34 @@
...
+ * is derived from user input or a calculation — is rejected with
+ * {@link KnowledgeBaseErrors.ValidationError}, so the mock and AWS runtimes never
+ * diverge on an input Bedrock would reject.
+ *
+ * @param maxResults - The caller-supplied value, or `undefined` for the default.
+ * @returns An integer in the range 1–100.
+ * @throws {KnowledgeBaseValidationError} If `maxResults` is defined but not an integer.
+ */
+export function normalizeMaxResults(maxResults: number | undefined): number {
+	if (maxResults === undefined) return DEFAULT_RESULTS;
```
soberm: [suggestion] Before this, `options?.maxResults ?? 10` meant an explicit `maxResults: null` nullish-coalesced to the default of 10. Now that the check is `=== undefined` instead of `== null`, an explicit `null` falls through to `Number.isInteger(null)` → false → a thrown ValidationError. That's a behavior change for an input that used to just work, and it's not unlikely here specifically — this is the kind of API an agent/tool-calling layer often drives from parsed JSON where "unset" comes through as `null` rather than omitted. Not covered by a test either way. If `null` should still mean "unset", `maxResults == null` would preserve the old default; if rejecting it is intentional, probably worth a test pinning that down.

packages/bb-knowledge-base/src/index.aws.ts
```diff
@@ -302,8 +297,9 @@ export class KnowledgeBase extends Scope {
 			throw blocksError(KnowledgeBaseErrors.ValidationError, 'Query must be a non-empty string.');
 		}
 
-		// Bedrock API limits numberOfResults to 1–100. Well within Lambda's 6 MB response payload.
-		const maxResults = Math.min(Math.max(options?.maxResults ?? 10, 1), 100);
+		// Bedrock API limits numberOfResults to an integer in 1–100 (well within
+		// Lambda's 6 MB response payload); reject fractional/non-finite inputs.
```
soberm: [nit] This comment mostly restates what normalizeMaxResults's own JSDoc already says in more detail. Could probably trim it down to just the Lambda-payload aside, since that's the one bit that's specific to this call site.

packages/bb-knowledge-base/src/errors.ts
```diff
@@ -33,3 +33,13 @@ export const KnowledgeBaseErrors = {
 	/** `waitUntilSynced()` exceeded its timeout before the knowledge base finished ingesting. */
 	Timeout: 'KnowledgeBaseTimeoutException',
 } as const;
+
+/**
+ * Build a typed `Error` whose `name` is one of {@link KnowledgeBaseErrors}, so it
+ * crosses the RPC wire by `name` and is matchable with `isBlocksError()`.
+ */
+export function blocksError(name: string, message: string): Error {
```
soberm: [nit] Now that this is the one shared copy instead of two duplicated ones, typing `name` as `(typeof KnowledgeBaseErrors)[keyof typeof KnowledgeBaseErrors]` instead of a bare `string` would catch a typo'd error constant at every call site for free. Not a regression, just a nice tightening opportunity.

packages/bb-knowledge-base/src/index.aws.test.ts
```diff
@@ -351,6 +351,30 @@ describe('retrieve (SDK-mocked)', () => {
...
+			return { retrievalResults: [] };
+		});
+
+		try {
+			const kb = new KnowledgeBase({ id: 'test' }, 'r9b', { source: './knowledge' });
+
+			for (const bad of [1.5, Number.NaN]) {
+				await assert.rejects(
+					() => kb.retrieve('query', { maxResults: bad }),
+					(e: unknown) => (e as Error).name === KnowledgeBaseErrors.ValidationError,
```
soberm: [nit] The normalize.test.ts unit tests use `isBlocksError(e, KnowledgeBaseErrors.ValidationError)` for this same check. Could use that here too instead of the manual cast — though this matches the style already used elsewhere in this file, so optional.

packages/bb-knowledge-base/src/index.mock.test.ts
```diff
@@ -113,6 +113,18 @@ test('maxResults is clamped to [1, 100]', async () => {
 	assert.ok(resultsNeg.length >= 1, 'negative maxResults should clamp to 1');
 });
 
+test('rejects a fractional or non-finite maxResults with ValidationError (parity with AWS)', async () => {
+	const kb = new KnowledgeBase({ id: 'test' }, 'kb', { source: 'test-knowledge-tmp' });
+
+	for (const bad of [1.5, Number.NaN]) {
+		await assert.rejects(
+			() => kb.retrieve('password', { maxResults: bad }),
+			(e: unknown) => (e as Error).name === KnowledgeBaseErrors.ValidationError,
```
soberm: [nit] Same note as the aws.test.ts version of this test — `isBlocksError()` would be a slightly safer check than the manual `(e as Error).name ===` cast, though this matches the file's existing convention too.

packages/bb-knowledge-base/src/index.aws.ts
osama-rizk: Done (d5cee56c) — the aws `retrieve()` `@throws` now reads "If query is empty or whitespace-only, or `maxResults` is a non-integer."

packages/bb-knowledge-base/src/index.mock.ts
osama-rizk: Done (d5cee56c) — same update to the mock `retrieve()` `@throws` line.

packages/bb-knowledge-base/src/normalize.ts
osama-rizk: Good catch — that was an unintended behavior change. Restored the old "null = unset" semantics: the guard is now `maxResults === undefined || maxResults === null` (and the param type is `number | undefined | null`), so an explicit `null` defaults to 10 as it did under `?? 10`. Added a `normalize.test.ts` case pinning `normalizeMaxResults(null) === 10`, with a comment noting the parsed-JSON/agent-layer reason.

packages/bb-knowledge-base/src/index.aws.ts
osama-rizk: Done — trimmed to just the call-site-specific aside: `// 1–100 results stays well within Lambda's 6 MB response payload.` The validation rationale lives in `normalizeMaxResults`'s JSDoc.

packages/bb-knowledge-base/src/errors.ts
osama-rizk: Done — added `export type KnowledgeBaseErrorName = (typeof KnowledgeBaseErrors)[keyof typeof KnowledgeBaseErrors]` and typed the `name` param with it. Confirmed all call sites already pass a `KnowledgeBaseErrors.*` constant, so it compiles clean and now catches a typo'd name for free.

packages/bb-knowledge-base/src/index.aws.test.ts
osama-rizk: Done — switched this test to `isBlocksError(e, KnowledgeBaseErrors.ValidationError)` (imported from `@aws-blocks/core`), matching `normalize.test.ts`.

packages/bb-knowledge-base/src/index.mock.test.ts
osama-rizk: Done — same switch to `isBlocksError()` here.

sarayev (approved):

soberm (approved):

Issue #450, opened 2026-08-28
bb-data: migration Lambda does not retry `DatabaseResumingException`, so any deploy against a scale-to-zero (`minCapacity: 0`) cluster fails


fossamagna:
### Before opening, please confirm

- [x] I have [searched for duplicate or closed issues](https://github.com/aws-devtools-labs/aws-blocks/issues?q=is%3Aissue).
- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Which package(s) are affected?

Building Block: data

### Describe the bug

Two supported options of the same Building Block do not compose: a `Database` configured with
`minCapacity: 0` (Aurora Serverless v2 scale-to-zero) auto-pauses after 5 minutes of inactivity,
and the migration custom resource fails on the first deploy that lands while the cluster is
paused. The stack rolls back.

The migration Lambda already has a retry helper for exactly this class of problem, but its
predicate does not match the auto-pause error. In `aws-blocks/packages/bb-data/src/migration-lambda.ts`:

```ts
const isBadRequest = e?.name === 'BadRequestException' ||
        e?.message?.includes('Communications link failure');
      if (!isBadRequest || attempt === MAX_RETRIES) throw e;
```

The RDS Data API raises the auto-pause resume error as its own error type,
`DatabaseResumingException` — not `BadRequestException` — so `withRetry` rethrows on the first
attempt instead of backing off. The helper's own doc comment anticipates the neighbouring case
("Aurora Serverless v2 can take time to become available after cluster creation; the writer
instance may not be ready when the migration Lambda first fires"), which suggests the auto-pause
variant was simply not considered when scale-to-zero became configurable.

The backoff ladder itself looks sufficient — `MAX_RETRIES = 8` with 1s → 30s capped delays is
roughly two minutes, and a resume typically completes in well under a minute — so the retry
would very likely succeed if the error were classified as retryable.

This is deterministic rather than a race: with `minCapacity: 0`, any deploy made more than
~5 minutes after the last database activity hits it.

### Expected behavior

`npm run deploy` applies migrations to a scale-to-zero cluster, waking it if it is paused,
rather than failing the custom resource and rolling back the stack.

Adding the error name to the existing predicate looks like it would be enough, e.g.:

```ts
const isBadRequest = e?.name === 'BadRequestException' ||
        e?.name === 'DatabaseResumingException' ||
        e?.message?.includes('Communications link failure');
```

### Reproduction steps

[...]

PR #451, opened 2026-08-28
fix(bb-data): retry migrations while a scale-to-zero cluster resumes

fossamagna:
## Problem

A `Database` configured with `minCapacity: 0` (Aurora Serverless v2 scale-to-zero) auto-pauses after ~5 minutes idle. The first Data API call of a deploy wakes the cluster and fails with `DatabaseResumingException` while it resumes, so the migration custom resource failed and the stack rolled back. Deterministic rather than a race: any deploy made more than ~5 minutes after the last database activity hit it.

**Issue #, if available:** #450

## Changes

The migration Lambda already backed off (1s → 30s, 8 attempts) while Aurora was coming up, but two layers kept the resume error out of that path, so a name-only fix in the Lambda would not have worked:

- **`DataApiEngine` classified it as `QueryFailed`.** Errors carrying no SQLState are classified by SDK exception name, and `DatabaseResumingException` was not in that list. It now maps to `ConnectionFailed`, alongside `ServiceUnavailableException` and `InternalServerErrorException`. Application code that queries a paused cluster sees the same, more accurate name — transient and safe to retry, unlike `QueryFailed`.
- **The Lambda's predicate only matched raw SDK error names**, which the engine has already rewritten by the time the error arrives. It now retries on `ConnectionFailed`. The raw SDK names are kept for errors raised outside the engine, and the retry log line now names the error, since its absence was the reporter's diagnostic signal.

Worth calling out for review:

- Retrying on `ConnectionFailed` also covers SQLState `08xxx` connection failures, which were previously not retried. In a migration Lambda that seems right — a transient connection failure is exactly what the backoff is for — but the cost is that a genuinely unreachable database now takes ~2 minutes to fail the deploy instead of failing immediately.
- Reclassifying `DatabaseResumingException` is a behavior change for application code: a query against a paused cluster now throws `ConnectionFailedException` instead of `QueryFailedException`. No types change, and the README documents the new meaning.
- Backoff limits are unchanged (~2 minutes total). The verification below used 3 of the 8 attempts.
- `withRetry` and its predicate are now exported for testing. `migration-lambda.ts` is a Lambda entry point, not part of the package's `exports` map, so this does not widen the public API surface (`npm run check:api` confirms the reports are unchanged).

## Validation

Unit tests, written to fail first:

- `data-api-engine.test.ts` — the resume error maps to `ConnectionFailed`.
- `migration-lambda.test.ts` — three predicate cases plus a `withRetry` loop test.

The Lambda-side tests drive the error through a stub Data API client into the real `DataApiEngine.execute`, rather than asserting on a hand-built SDK error. That matters here: a raw-error test passes without the engine-side change while the bug survives.

No e2e test covers this combination. Local e2e runs on PGlite, so it cannot reach the Data API path at all, and a sandbox repro requires a `minCapacity: 0` cluster idled past its auto-pause window. Instead it was verified manually against a real deployed stack (ap-northeast-1, Aurora PostgreSQL 16.13, `minCapacity: 0`) whose cluster had been paused for 30+ minutes (`ServerlessDatabaseCapacity` at 0), by adding a no-op migration to change `migrationsHash`:

```
06:05:51  [migration-lambda] Event: {"RequestType":"Update","migrationsHash":"57c5906d2974d89f"}
06:05:54  [migration-lambda] Aurora not ready (ConnectionFailedException), retry 1/8 in 1000ms
06:05:59  [migration-lambda] Aurora not ready (ConnectionFailedException), retry 2/8 in 2000ms
06:06:04  [migration-lambda] Aurora not ready (ConnectionFailedException), retry 3/8 in 4000ms
06:06:08  [migrations] Applied: 003_verify_450_noop.sql
```

The stack reached `UPDATE_COMPLETE` instead of rolling back. Note the error surfaces as `ConnectionFailedException`, confirming the engine-side classification is the part that carries the signal.

Local gates: `npm run build`, `npm run lint`, `npm run lint:deps`, `npm run check:api`, `npm test`, `npm run test:e2e:local` all pass.

## Checklist

- [x] PR description included

---


pranavosu (dismissed):

packages/bb-data/src/migration-lambda.test.ts
```diff
@@ -19,3 +21,59 @@ test('migration handler returns early on Delete event', async () => {
...
+  const e = await errorFromEngine('DatabaseResumingException', RESUMING_MESSAGE);
+  let attempts = 0;
+  const result = await withRetry(async () => {
+    attempts++;
+    if (attempts === 1) throw e;
+    return 'migrated';
+  });
+  assert.strictEqual(result, 'migrated');
+  assert.strictEqual(attempts, 2);
+});
```
sharonyajain: 🟡 **withRetry's exhaustion branch (retryable error that never recovers) has no test, leaving the give-up path unguarded**

The withRetry loop test only covers recover-after-one-retry. The exhaustion branch - `if (!isRetryableMigrationError(e) || attempt === MAX_RETRIES) throw e;` after all 8 attempts fail - is never exercised, and neither is the immediate-throw path for a non-retryable error passed to withRetry (only isRetryableMigrationError is tested in isolation for that).

```typescript
test('withRetry retries an auto-pause resume error until the cluster is awake', async () => {
  const e = await errorFromEngine('DatabaseResumingException', RESUMING_MESSAGE);
  let attempts = 0;
  const result = await withRetry(async () => {
    attempts++;
    if (attempts === 1) throw e;
    return 'migrated';
  });
  assert.strictEqual(result, 'migrated');
  assert.strictEqual(attempts, 2);
});
```

**Why it matters:** A future refactor of the loop bound or termination condition (an off-by-one, or making the `throw new Error('unreachable')` line reachable) would regress the give-up behavior with no test to catch it -> on a genuinely unreachable database a deploy could surface a misleading 'unreachable' error instead of the real DB error, costing on-call debugging time during an already-failing deploy.

**Suggestion:** Add a withRetry test where the wrapped fn always throws a retryable error and assert it rejects with that same error after exactly MAX_RETRIES+1 attempts (stub the delay/timer to keep it fast), and optionally one where a non-retryable error rejects immediately after a single attempt.

_[code-review-sage]_

packages/bb-data/src/migration-lambda.ts
```diff
@@ -13,20 +14,40 @@ const INITIAL_DELAY_MS = 1000;
...
+
+/**
+ * Execute a function with exponential backoff while Aurora is unreachable.
+ * Aurora Serverless v2 can take time to become available after cluster creation
+ * (the writer instance may not be ready when the migration Lambda first fires),
+ * and a scale-to-zero cluster needs to resume from auto-pause before it accepts
+ * the first statement of a deploy.
  */
-const withRetry = async <T>(fn: () => Promise<T>): Promise<T> => {
+export const withRetry = async <T>(fn: () => Promise<T>): Promise<T> => {
```
sarayev, line 47: Nice catch on the two-layer fix, keying the predicate off `ConnectionFailed` instead of the raw name is the right call since the engine has already renamed it by the time we get here.

One thing I would like before this goes in: a test that `withRetry` throws a non-retryable error straight away without retrying. We have the retries-then-succeeds case and the predicate returning false for a syntax error, but nothing actually asserts the helper bails immediately on something like a `QueryFailed` (attempts === 1). It is cheap to add and it locks in the half of the contract that keeps us from accidentally retrying real SQL errors later.

packages/bb-data/src/migration-lambda.ts
```diff
@@ -13,20 +14,40 @@ const INITIAL_DELAY_MS = 1000;
...
+ *
+ * `DataApiEngine` rewrites `error.name` to a `DatabaseErrors` name before it
+ * reaches here, so `ConnectionFailed` is the check that actually fires in the
+ * Lambda; the raw SDK names are kept for errors raised outside the engine.
+ */
+export const isRetryableMigrationError = (e: unknown): boolean =>
+  e instanceof Error && (
+    e.name === DatabaseErrors.ConnectionFailed ||
+    e.name === 'BadRequestException' ||
+    e.name === 'DatabaseResumingException' ||
```
sarayev: Small heads-up more than a request: retrying on any `ConnectionFailed` now covers real connection failures too (bad secret, SG blocking the cluster, 08xxx SQLStates), so those go from failing fast to grinding through all 8 attempts (~2 min) before the deploy surfaces the actual error. I think that is a fine trade for the migration path since the resume case is what we care about and the real error still comes out the end, but wanted to confirm the slower-failure-on-misconfig behavior is intended.

Also, the raw `DatabaseResumingException` and `BadRequestException` branches are effectively dead on this path (everything flows through the engine, which rewrites the name), and no test hits them. Totally fine to keep as defensive belt-and-suspenders, just flagging they are untested if you would rather trim them.

sarayev:
Overall LGTM, two minor comments that are not blockers. Approving.

sarayev (dismissed):

packages/bb-data/src/migration-lambda.ts
fossamagna: Added in 6fd055b — `withRetry` now has a case asserting `attempts === 1` on a non-retryable error, and it asserts identity (`err === e`) so the original error has to reach the caller unwrapped, which is what puts the real SQL error in the CloudFormation message.

I checked the test actually holds the contract by mutating the loop to drop the retryability check: only this new case fails, the other five pass. So you were right that nothing covered this half.

packages/bb-data/src/migration-lambda.ts
fossamagna: Confirming the slower-failure-on-misconfig behavior is intended, with one narrowing: only errors the engine classifies as `ConnectionFailed` are retried, i.e. SQLState `08xxx`, `ServiceUnavailableException`, `InternalServerErrorException`, and now `DatabaseResumingException`. A bad secret surfaces as `BadRequestException`, which the engine rewrites to `QueryFailed`, so that still fails fast. Security-group misconfiguration doesn't reach this classification either, since Data API calls go to an AWS-managed endpoint rather than through the customer's SG. So the ~2 minute window applies to `08xxx` and service-side transients, which is what the backoff is for. 6fd055b makes that explicit with a test on the `08xxx` path, comment included.

On the raw name branches: kept, and now covered by a test that calls the predicate without the engine. Two reasons for keeping rather than trimming. The bug being fixed here was a classification gap, so a name-level fallback is a cheap safety net if the engine's mapping shifts again. And the `BadRequestException` branch predates this PR, so dropping it inside a bugfix would be an unrelated behavior removal. The test documents that nothing routes around the engine today.

fossamagna:
@sarayev @pranavosu  Sorry for the extra round — your approval got dismissed by a follow-up push. Could you re-approve when you get a chance?

sarayev (dismissed):
LGTM 👍 

Thanks for addressing the comments!

packages/bb-data/src/engines/data-api-engine.ts
```diff
@@ -110,7 +111,14 @@ function translateError(e: unknown): never {
...
-    } else if (e.name === 'ServiceUnavailableException' || e.name === 'InternalServerErrorException') {
+    } else if (
+      e.name === 'ServiceUnavailableException' ||
+      e.name === 'InternalServerErrorException' ||
+      // A scale-to-zero cluster (minCapacity: 0) auto-pauses after ~5 minutes
+      // idle; the call that wakes it fails while it resumes. Transient, and
+      // retryable — not a bad statement.
+      e.name === 'DatabaseResumingException'
+    ) {
       e.name = DatabaseErrors.ConnectionFailed;
```
soberm: **Suggestion**

The whole fix depends on these three SDK exception-name literals (`ServiceUnavailableException`, `InternalServerErrorException`, `DatabaseResumingException`) matching `error.name` exactly, and `'DatabaseResumingException'` is duplicated in `migration-lambda.ts` L31. If AWS renames one or a refactor drifts a copy, classification silently falls through to `QueryFailed` and issue #450 regresses at deploy time — and no unit test would catch it, since the tests construct the same literals.

Per `A0-API-DESIGN.md` § G6, typed constants exist precisely to prevent this class of string-matching drift.

What do you think about hoisting these into one shared constant that both the engine classifier and the Lambda fallback import?

```ts
export const RESUMABLE_DATA_API_ERROR_NAMES = new Set([
  'ServiceUnavailableException',
  'InternalServerErrorException',
  'DatabaseResumingException',
]);
```

packages/bb-data/src/migration-lambda.ts
```diff
@@ -13,20 +14,40 @@ const INITIAL_DELAY_MS = 1000;
...
+    e.message.includes('Communications link failure')
```
soberm, line 34: **Nit**

`e.message.includes('Communications link failure')` is a MySQL Connector/J–style message, but this BB is Aurora PostgreSQL via the Data API. It predates this PR and is currently load-bearing for the writer-not-ready case (the engine rewrites the name to `QueryFailed` but preserves the message, so this is what catches it — see the test at L59-62), so it should not just be deleted.

Consider a one-line comment noting where the string comes from and the exact scenario it covers, so a future reader doesn't remove it as PostgreSQL-irrelevant dead code.

packages/bb-data/src/migration-lambda.ts
```diff
@@ -13,20 +14,40 @@ const INITIAL_DELAY_MS = 1000;
...
-      if (!isBadRequest || attempt === MAX_RETRIES) throw e;
+    } catch (e) {
+      if (!isRetryableMigrationError(e) || attempt === MAX_RETRIES) throw e;
       const delay = Math.min(INITIAL_DELAY_MS * 2 ** attempt, MAX_DELAY_MS);
-      console.log(`[migration-lambda] Aurora not ready, retry ${attempt + 1}/${MAX_RETRIES} in ${delay}ms`);
+      const name = e instanceof Error ? e.name : 'unknown';
+      console.log(`[migration-lambda] Aurora not ready (${name}), retry ${attempt + 1}/${MAX_RETRIES} in ${delay}ms`);
       await new Promise(r => setTimeout(r, delay));
     }
   }
```
soberm, line 58: **Suggestion**

`withRetry` re-throws the last error at `attempt === MAX_RETRIES`, and the trailing `throw new Error('unreachable')` exists only to satisfy the `Promise<T>` return type. A `@throws` line would make the exhausted-retries contract explicit for the next reader.

```ts
/**
 * ...
 * @throws the last underlying error once MAX_RETRIES retryable attempts are exhausted.
 */
```

Separately (no code change needed here — a question for the thread): because the retryable set now includes any `ConnectionFailed`, a mid-run `08xxx` drop will re-invoke `runMigrations` from the top. Is `runMigrations` idempotent across a partial-then-retry, and are all migration files transactional?

packages/bb-data/src/engines/data-api-engine.ts
fossamagna: Hoisted in 0f5abd9 — both sites now read `TRANSIENT_DATA_API_ERROR_NAMES` from `errors.ts`, so the duplicated literal is gone. It stays internal: the entry points re-export `DatabaseErrors` by name, so the API report is unchanged. Named `TRANSIENT_` rather than `RESUMABLE_` since `ServiceUnavailableException` isn't a resume.

One behavior delta: the Lambda's fallback now also accepts raw `ServiceUnavailableException` / `InternalServerErrorException`, since it shares the engine's set — the raw-name test covers all three. `BadRequestException` stays a separate check, as the engine maps it to `QueryFailed`.

Two notes on the reasoning, since they change what this buys us. G6 (in `docs/design/API-DESIGN.md` — `A0-` looks like a stale path) is about exporting `{BlockName}Errors` for customer catch blocks, which this PR already follows via `DatabaseErrors.ConnectionFailed`; the names here are inbound AWS strings, so this is a DRY fix rather than a G6 one. And a shared constant doesn't cover an AWS rename — one copy goes stale instead of two drifting apart. Worth doing for the drift case either way.

packages/bb-data/src/migration-lambda.ts
fossamagna: Added in 3a4e2b4, kept to two lines: JDBC-style but load-bearing, the only check for a not-yet-ready writer, whose name the engine rewrites while leaving the message intact. Left the test name out of the comment so it can't go stale.

packages/bb-data/src/migration-lambda.ts
fossamagna: `@throws` added in 90f1a39.

On idempotency: yes, at file granularity. Each file's statements and its `_migrations` insert share one transaction, so a file is either applied and recorded or rolled back, and a retry re-reads `_migrations` and skips it. Covered today by `runMigrations rolls back failed migration` and `applies only new migrations on second run` (`migrations-integration.test.ts` L77, L60).

Transactional, with one caveat that predates this PR: PostgreSQL DDL is, but a statement that cannot run inside a transaction block (`CREATE INDEX CONCURRENTLY`) errors rather than half-applying.

One edge worth naming: if the rollback in the catch also fails, the orphaned transaction holds its locks until the Data API's 3-minute transaction timeout, which outlasts our ~2-minute backoff, so a retry can block — not corrupt. #450's own failure is on the first `CREATE TABLE IF NOT EXISTS _migrations`, outside any transaction, so no locks are involved there.

soberm (approved):

sarayev (approved):

PR #452, opened 2026-08-28
feat(bb-app-setting): customer-managed KMS key for secrets (kmsKeyArn)

osama-rizk:
## Card

Bug Bash board (Starter Kit, Security P1): **"core: SSM SecureString uses default `alias/aws/ssm` KMS key"** — *"Should use a CMK so cross-account/decrypt scope can be controlled."*
🔗 https://github.com/orgs/aws-amplify/projects/141/views/15?pane=issue&itemId=195530935

The on-`main` manifestation of this card is `bb-app-setting` (its secrets are the SecureString params; the hosting `secret()` CLI isn't on `main` yet).

## Problem

`AppSetting` secrets (`secret: true`) were always encrypted with the default `aws/ssm` AWS-managed key, and the handler got `kms:Decrypt`/`Encrypt` on `*` (scoped only by a `kms:ViaService` condition). There was no way to pin a secret to a **customer-managed key** to control its decrypt/grant scope (cross-account, dedicated key policy, rotation).

## Change

New optional `kmsKeyArn?: string` on `AppSettingOptions` (only valid with `secret: true`):

- **CDK** — the per-stack bulk-init Custom Resource now carries a `keyId` per parameter and passes it as `KeyId` on `PutParameter`; the handler is granted `kms:Decrypt`/`Encrypt`/`GenerateDataKey*` scoped to that **specific key ARN** (no `ViaService` wildcard). Default-key secrets are unchanged.
- **Runtime** — `put()` re-specifies `KeyId` on overwrite. Without this, SSM silently falls back to `aws/ssm` on every write, downgrading the encryption.
- **Mock** — unaffected (no encryption locally).

Because a BYO key isn't owned by AWS Blocks, **the CMK's key policy must also allow the app's execution role** — documented in the README/DESIGN.

## Tests (all new, red→green)

- CDK synth: CMK secret grants KMS on the exact key ARN (with `GenerateDataKey*`, no `ViaService` condition); the bulk CR `Parameters` carries the `keyId`; a default-key secret carries **no** `keyId`; `kmsKeyArn` without `secret` throws `ValidationFailed`.
- AWS runtime (`SSMClient` mock): `put()` sends `KeyId` for a CMK secret, and omits it for a default-key secret and for a non-secret String.

## Verification

- `@aws-blocks/bb-app-setting` suite: **66 pass / 0 fail** (was 59).
- `npm run lint` 0 errors · `lint:deps` ✅ · conditional-export parity 21/21 · umbrella-changeset guard ✅.
- `API.md` regenerated (`update:api`) — adds `kmsKeyArn`. (Local `check:api` shows only a CRLF-vs-LF artifact; `diff -w` is clean — CI on Linux emits LF.)

## Changeset

`@aws-blocks/bb-app-setting` **minor** + `@aws-blocks/blocks` **patch** (umbrella re-export). Additive/opt-in — no behavior change when `kmsKeyArn` is omitted.
## Scope & follow-up (not in this PR)

This PR is the foundational primitive (`AppSetting`-level `kmsKeyArn`), intentionally **not** the whole framework fix. Other SecureString writers still use the default `aws/ssm` key and are left for follow-up work:

- **`bb-auth-cognito` / `bb-auth-oidc` / `bb-auth-basic`** — their session / cookie / JWT signing secrets are created via an internal `new AppSetting(…, { secret: true })` and don't yet forward a key.
- **`core/scripts/ensure-secrets.ts`** — the DB connection-string provisioner `PutParameter`s a SecureString with no `KeyId`.

Recommended follow-up: a single stack-level `secretsKmsKey` that every secret writer consumes, rather than sprinkling `kmsKeyArn` across each API. This PR is the layer that work builds on.

### DIFF
--- a/packages/bb-app-setting/src/index.aws.test.ts
+++ b/packages/bb-app-setting/src/index.aws.test.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import assert from 'node:assert';
-import { describe, it, mock } from 'node:test';
+import { describe, test, mock } from 'node:test';
 import { SSMClient } from '@aws-sdk/client-ssm';
 import { Scope } from '@aws-blocks/core';
 import { AppSetting } from './index.aws.js';
@@ -20,7 +20,7 @@ function captureSend() {
 }
 
 describe('AWS runtime put() KMS key', () => {
-	it('passes the CMK as KeyId when a secret is backed by kmsKeyArn', async () => {
+	test('passes the CMK as KeyId when a secret is backed by kmsKeyArn', async () => {
 		const { inputs, restore } = captureSend();
 		try {
 			const setting = new AppSetting(new Scope('app'), 'cmk', { secret: true, kmsKeyArn: TEST_CMK });
@@ -33,7 +33,7 @@ describe('AWS runtime put() KMS key', () => {
 		}
 	});
 
-	it('omits KeyId for a default-key secret', async () => {
+	test('omits KeyId for a default-key secret', async () => {
 		const { inputs, restore } = captureSend();
 		try {
 			const setting = new AppSetting(new Scope('app'), 'plain', { secret: true });
@@ -45,7 +45,7 @@ describe('AWS runtime put() KMS key', () => {
 		}
 	});
 
-	it('omits KeyId for a non-secret String parameter', async () => {
+	test('omits KeyId for a non-secret String parameter', async () => {
 		const { inputs, restore } = captureSend();
 		try {
 			const setting = new AppSetting<string>(new Scope('app'), 'cfg', { value: 'init' });
--- a/packages/bb-app-setting/src/index.aws.test.ts
+++ b/packages/bb-app-setting/src/index.aws.test.ts
@@ -1,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import assert from 'node:assert';
+import { describe, it, mock } from 'node:test';
+import { SSMClient } from '@aws-sdk/client-ssm';
+import { Scope } from '@aws-blocks/core';
+import { AppSetting } from './index.aws.js';
+
+const TEST_CMK = 'arn:aws:kms:us-east-1:111122223333:key/abcd1234-5678-90ab-cdef-1234567890ab';
+
+/** Capture the input of the next PutParameterCommand sent by the runtime. */
+function captureSend() {
+	const inputs: any[] = [];
+	const m = mock.method(SSMClient.prototype, 'send', async (cmd: any) => {
+		inputs.push(cmd.input);
+		return {};
+	});
+	return { inputs, restore: () => m.mock.restore() };
+}
+
+describe('AWS runtime put() KMS key', () => {
+	it('passes the CMK as KeyId when a secret is backed by kmsKeyArn', async () => {
+		const { inputs, restore } = captureSend();
+		try {
+			const setting = new AppSetting(new Scope('app'), 'cmk', { secret: true, kmsKeyArn: TEST_CMK });
+			await setting.put('rotated-value');
+			assert.strictEqual(inputs.length, 1);
+			assert.strictEqual(inputs[0].Type, 'SecureString');
+			assert.strictEqual(inputs[0].KeyId, TEST_CMK, 'overwrite must re-specify the CMK, not fall back to aws/ssm');
+		} finally {
+			restore();
+		}
+	});
+
+	it('omits KeyId for a default-key secret', async () => {
+		const { inputs, restore } = captureSend();
+		try {
+			const setting = new AppSetting(new Scope('app'), 'plain', { secret: true });
+			await setting.put('v');
+			assert.strictEqual(inputs[0].Type, 'SecureString');
+			assert.strictEqual(inputs[0].KeyId, undefined);
+		} finally {
+			restore();
+		}
+	});
+
+	it('omits KeyId for a non-secret String parameter', async () => {
+		const { inputs, restore } = captureSend();
+		try {
+			const setting = new AppSetting<string>(new Scope('app'), 'cfg', { value: 'init' });
+			await setting.put('next');
+			assert.strictEqual(inputs[0].Type, 'String');
+			assert.strictEqual(inputs[0].KeyId, undefined);
+		} finally {
+			restore();
+		}
+	});
+});
--- a/packages/bb-app-setting/src/index.cdk.ts
+++ b/packages/bb-app-setting/src/index.cdk.ts
@@ -42,7 +42,7 @@ export class AppSetting<T = string> extends Scope {
 	static fromExisting<T = string>(
 		scope: ScopeParent,
 		id: string,
-		options: { name: string; secret?: boolean },
+		options: { name: string; secret?: boolean; kmsKeyArn?: string },
 	): AppSetting<T> {
 		const opts: InternalAppSettingOptions<T> = { ...options, external: true };
 		return new AppSetting<T>(scope, id, opts);
@@ -74,13 +74,23 @@ export class AppSetting<T = string> extends Scope {
 			throw err;
 		}
 
-		if (options.kmsKeyArn && !options.secret) {
-			const err = new Error(
-				`AppSetting '${id}': 'kmsKeyArn' is only valid with 'secret: true'. ` +
-				`Non-secret String parameters are not encrypted.`
-			);
-			err.name = AppSettingErrors.ValidationFailed;
-			throw err;
+		if (options.kmsKeyArn !== undefined) {
+			if (!options.secret) {
+				const err = new Error(
+					`AppSetting '${id}': 'kmsKeyArn' is only valid with 'secret: true'. ` +
+					`Non-secret String parameters are not encrypted.`
+				);
+				err.name = AppSettingErrors.ValidationFailed;
+				throw err;
+			}
+			if (options.kmsKeyArn.trim() === '') {
+				const err = new Error(
+					`AppSetting '${id}': 'kmsKeyArn' must be a non-empty KMS key ARN. ` +
+					`Omit it to use the default aws/ssm key.`
+				);
+				err.name = AppSettingErrors.ValidationFailed;
+				throw err;
+			}
 		}
 
 		if (options.secret && options.value !== undefined) {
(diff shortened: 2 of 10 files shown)


packages/bb-app-setting/src/index.cdk.ts
```diff
@@ -201,51 +232,41 @@ function registerSecret(stack: cdk.Stack, parameterName: string): void {
...
-						}
-					}
-					for (const name of removed) {
-						try { await client.send(new DeleteParameterCommand({ Name: name })); } catch {}
+					for (const p of params) { if (!oldNames.includes(p.name)) await putSecret(p, tags); }
+					for (const name of oldNames) {
+						if (!names.includes(name)) { try { await client.send(new DeleteParameterCommand({ Name: name })); } catch {} }
 					}
 					return { PhysicalResourceId: 'bb-secrets-bulk' };
 				}
```
soberm: [blocker] This only diffs by name, so if a secret's name stays the same but its `keyId` changes (e.g. adding `kmsKeyArn` to a secret that was previously on the default key), nothing here re-encrypts it. `putSecret` uses `Overwrite: false` and swallows `ParameterAlreadyExists`, so even if it got called for an existing name it still wouldn't touch the stored value.

Meanwhile the execution role's KMS grant above (around line 150) is recomputed fresh from the current `kmsKeyArn` on every synth, not additively. So right after this kind of stack update finishes, the parameter is still encrypted under the old key but the role can only `kms:Decrypt` the new key ARN. The next `get()` call fails with AccessDenied, and nothing here catches it at deploy time.

I ran this handler against a mocked SSM client to confirm: an Update event with an unchanged name but a new `keyId` produces zero PutParameter/DeleteParameter calls.

Might be worth detecting `oldNames.includes(p.name)` with a changed `keyId` and logging a warning (there's no logging anywhere in this Lambda right now) or failing the custom resource, so it's a visible signal instead of something that only surfaces after someone's already redeployed.

packages/bb-app-setting/src/index.cdk.ts
```diff
@@ -130,21 +141,34 @@ export class AppSetting<T = string> extends Scope {
...
+					resources: ['*'],
+					conditions: {
+						StringEquals: {
+							'kms:ViaService': `ssm.${cdk.Stack.of(this).region}.amazonaws.com`,
+						},
 					},
-				},
-			}));
+				}));
+			}
```
soberm, line 180: [blocker] This grant is fully recomputed from whatever `kmsKeyArn` is set to right now, it's not additive across deploys. Combined with the Update handler not re-encrypting existing secrets when `keyId` changes (see the comment further down on that block), changing `kmsKeyArn` on an already-deployed secret swaps this grant to the new key immediately while the parameter itself stays encrypted under the old one. That's an access break, not just a permissions tweak.

packages/bb-app-setting/src/index.cdk.ts
```diff
@@ -257,18 +278,22 @@ function registerSecret(stack: cdk.Stack, parameterName: string): void {
...
 	}));
 
+	// SSM SecureString encryption goes through KMS via the SSM service. Scoping to
+	// `kms:ViaService = ssm.<region>` covers both the default aws/ssm key and any
+	// customer-managed key used above (the CMK's own key policy must also allow
+	// this role). `GenerateDataKey*` is required when a CMK is in play.
 	secretInitFn.addToRolePolicy(new iam.PolicyStatement({
-		actions: ['kms:Encrypt'],
+		actions: ['kms:Encrypt', 'kms:GenerateDataKey*'],
 		resources: ['*'],
```
soberm, line 331: [major] Standard-tier SecureString parameters (which is all this construct ever creates, nothing here sets `Tier: 'Advanced'`) only need `kms:Encrypt` to write and `kms:Decrypt` to read, per the SSM docs. `kms:GenerateDataKey` is for advanced-tier parameters using envelope encryption. The comment above says it's required for a CMK, but that's not accurate for standard tier, so this (and the same addition on the per-secret grant a bit above) is granting more than's needed. Doesn't cause a functional problem, but `GenerateDataKey*` as a wildcard also pulls in `GenerateDataKeyPair*`, which has nothing to do with SSM parameters at all.

packages/bb-app-setting/src/index.cdk.ts
```diff
@@ -130,21 +141,34 @@ export class AppSetting<T = string> extends Scope {
...
+			if (options.kmsKeyArn) {
+				// Customer-managed key: grant on the specific key ARN. NOTE: the key's
+				// own key policy must also allow this role (we can't edit a BYO key's
+				// policy from here) — see the README.
+				this.executionRole.addToPrincipalPolicy(new iam.PolicyStatement({
+					actions: external
+						? ['kms:Decrypt']
+						: ['kms:Decrypt', 'kms:Encrypt', 'kms:GenerateDataKey*'],
+					resources: [options.kmsKeyArn],
+				}));
```
soberm, line 168: [major] Same `GenerateDataKey*` addition here — see the note further down on the shared `secretInitFn` grant, standard-tier SecureString parameters don't look like they need this action.

packages/bb-app-setting/src/types.ts
```diff
@@ -27,8 +27,23 @@ export interface AppSettingOptions<T = string> {
...
+	 *
+	 * Only valid together with `secret: true`. The CDK layer grants the shared
+	 * handler `kms:Decrypt` (plus `kms:Encrypt`/`kms:GenerateDataKey*` for
+	 * stack-managed secrets it writes) on this key, and the runtime `put()` passes
+	 * the key so an overwrite does not silently fall back to the default key.
+	 *
+	 * @example
+	 * new AppSetting(scope, 'apiKey', { secret: true, kmsKeyArn: myKey.keyArn });
+	 */
+	kmsKeyArn?: string;
```
soberm, line 47: [major] `fromExisting()` (on all three variants: cdk/aws/mock) only accepts `{ name, secret? }`, there's no way to pass `kmsKeyArn` through it. So an externally-owned secret that's actually encrypted with a CMK never gets the tighter, key-scoped read grant this PR adds, it falls back to the existing wildcard `kms:Decrypt` grant. Not broken today since that wildcard covers any key via the ViaService condition, but if the goal is consistent key-scoped grants it might be worth extending `fromExisting` too at some point.

soberm (commented):
**Needs work.** Solid feature overall, but there's a real gap around what happens when `kmsKeyArn` changes on an already-deployed secret (flagged as an inline comment on the Update handler below), plus a couple of smaller suggestions.

This adds an optional `kmsKeyArn` to `AppSettingOptions` so `secret: true` parameters can be encrypted with a customer-managed key instead of the default `aws/ssm` key, threaded through CDK synth, the runtime `put()`, and the docs. I pulled the actual Custom Resource Lambda source out of the diff and ran it against a mocked SSM client to exercise the Create/Update/Delete/collision paths directly. The primary path (a brand new secret created with `kmsKeyArn` from the start) works correctly and is well covered by the new tests.

A few things worth calling out positively: the new validation guard (`kmsKeyArn` requires `secret: true`) follows the exact same `Error`/`err.name`/`throw` pattern as every other constructor check in this file, and it's tested. The comments are doing real work too, `index.aws.ts:174-176` on why `put()` has to re-specify the key on overwrite, and `index.cdk.ts:151-153` on the BYO-key-policy caveat, rather than restating the code. Docs (README/DESIGN/API.md) match the implementation, including the mock correctly documenting that it ignores `kmsKeyArn`. I also checked that the `ParameterNames` → `Parameters` custom resource property rename is safe for already-deployed stacks upgrading to this version (`Overwrite: false` plus the existing-parameter catch degrades it to a harmless re-tag, not a regeneration). Good scoping overall, this stays focused on the one primitive and explicitly calls out what's left for follow-up.

Specifics are left as inline comments below.

packages/bb-app-setting/src/index.aws.test.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import assert from 'node:assert';
+import { describe, it, mock } from 'node:test';
```
soberm: [nit] Rest of this package (and siblings like bb-knowledge-base, bb-auth-cognito, bb-data) use `describe`/`test` from node:test. This is the only file using `describe`/`it`. Both work, `it`'s just an alias, but might as well match the convention.

packages/bb-app-setting/src/index.cdk.ts
```diff
@@ -72,6 +74,15 @@ export class AppSetting<T = string> extends Scope {
...
 		}
 
+		if (options.kmsKeyArn && !options.secret) {
+			const err = new Error(
+				`AppSetting '${id}': 'kmsKeyArn' is only valid with 'secret: true'. ` +
+				`Non-secret String parameters are not encrypted.`
+			);
+			err.name = AppSettingErrors.ValidationFailed;
+			throw err;
+		}
```
soberm: [nit] `kmsKeyArn: ''` slips past this check and the `if (options.kmsKeyArn)` one below since both are plain truthy checks, so it silently behaves as if the option weren't set instead of throwing `ValidationFailed`. Pretty unlikely in practice, but an explicit `!== undefined` check would catch it instead of failing quietly.

packages/bb-app-setting/src/index.cdk.ts
osama-rizk: Fixed in 7f83be52 — the Update handler now re-encrypts on a key change. For a name present in both old and new whose `keyId` differs, it does `GetParameter` (WithDecryption) → `PutParameter` (Overwrite:true, new KeyId, same value), so the value survives the rotation and ends up under the same key the IAM grant points at. Also added a `readOld()` fallback that maps the legacy `ParameterNames` string[] shape from `OldResourceProperties`, so the very first upgrade deploy diffs correctly instead of treating everything as new. Bulk-init role gained `ssm:GetParameter` + `kms:Decrypt` for the re-key read.

packages/bb-app-setting/src/index.cdk.ts
osama-rizk: Addressed together with the re-encrypt fix (see the Update-handler thread). The grant is still recomputed per synth, but the parameter is now re-encrypted to the new key in the same deploy, so the end state is consistent (value + grant both on the new key). The transient only exists mid-deploy, before the app serves the new config.

packages/bb-app-setting/src/index.cdk.ts
osama-rizk: Good catch — dropped `kms:GenerateDataKey*`. Standard-tier SecureStrings (all this construct creates) only use `kms:Encrypt`/`kms:Decrypt`; `GenerateDataKey` is advanced-tier envelope encryption, and the `*` wildcard needlessly pulled in `GenerateDataKeyPair*`. Both the per-secret handler grant and the shared bulk-init grant are now just Encrypt/Decrypt (the bulk-init one adds Decrypt precisely for the re-key read).

packages/bb-app-setting/src/index.cdk.ts
osama-rizk: Done — removed here too (see the shared `secretInitFn` grant thread).

packages/bb-app-setting/src/types.ts
osama-rizk: Extended — `fromExisting()` on all three variants (cdk/aws/mock) now accepts `kmsKeyArn`, so an externally-owned CMK secret gets the tight, key-scoped `kms:Decrypt` grant (read-only, no Encrypt) instead of the `ViaService` wildcard. Added a CDK test asserting that.

packages/bb-app-setting/src/index.aws.test.ts
osama-rizk: Done — switched `index.aws.test.ts` to `describe`/`test` to match the rest of the package.

packages/bb-app-setting/src/index.cdk.ts
osama-rizk: Fixed — validation is now `kmsKeyArn !== undefined`, and an empty/whitespace ARN throws `ValidationFailed` ("must be a non-empty KMS key ARN") rather than silently behaving as unset.

soberm (approved):

Simone319 (approved):

PR #455, opened 2026-08-28
feat(bb-agent): cap model and tool calls per turn to bound runaway cost

pjkroker:
## What

Adds two per-turn safety caps to `AgentConfig`, **both defaulting to `20`**, that stop an agent from looping its reason→act cycle indefinitely and running up unexpected Bedrock cost:

- **`maxLLMCalls?: number | false`** — max model (Bedrock) invocations per turn. Model calls are the Bedrock billing unit, so this is the most direct spend guard; because every tool round needs a model call, it transitively bounds tool loops too. (≈ the conventional "maxTurns".)
- **`maxToolIterations?: number | false`** — max tool calls per turn (parallel tool batches count each call).

Set a cap to **`false`** to disable it entirely.

## Why

A misbehaving agent — or a prompt that induces one — can loop far longer than intended. Strands TS ships no built-in `maxTurns`, and on-demand `stop()`/cancel is a separate, whole-team API-design problem (compute-agnostic cancellation). These self-limiting caps need no cross-process signaling, so they work identically on every compute target. Jan raised the runaway-spend risk; this is the minimal, always-on backstop.

**Framing:** the README is explicit that these are a **safety backstop, not a way to steer the agent**. The defaults only stop runaways — they are not tuned for any given agent, and customers should set them deliberately (based on how many steps/tool calls a healthy turn takes) so a normal turn always completes.

## How

- Enforced by counting Strands' `BeforeModelCallEvent` / `BeforeToolCallEvent` hooks; past the cap → `strandsAgent.cancel()` (the `event.agent` `LocalAgent` type has no `cancel()`, so the held instance is used).
- A trip ends the stream cleanly (`stopReason: 'cancelled'`, no throw), so `runAgent` surfaces it as an `error` chunk (so `complete()` rejects) and skips the final persist + `done`.
- Hooks are registered in `runAgent` (not `createStrandsAgent`) so the per-turn counters + `capExceeded` reason stay in scope for the result handling; they coexist with the HITL `BeforeToolCallEvent` hook.

## Tests

4 new unit tests (via the canned provider), all green:
- `maxLLMCalls: 1` trips on the follow-up model call → `error` chunk.
- `maxToolIterations: 1` trips on a 2-tool fan-out → `error` chunk.
- a normal single-tool turn under the defaults completes with `done`.
- `maxToolIterations: false` lets the same fan-out that trips at `1` complete → proves disable works.

## Verification

- `npm run build` · `npm run lint` · `npm run lint:deps` ✅
- bb-agent unit: **82 pass / 0 fail** · blocks conditional-export parity: **42 pass / 0 fail** ✅
- `API.md` regenerated (`npm run update:api`); changeset added (`@aws-blocks/bb-agent` + `@aws-blocks/blocks`, `patch`).


packages/bb-agent/src/agent.ts
```diff
@@ -36,6 +36,13 @@ const jobPayloadSchema = z.object({
 /** Key under which the per-call tool context is threaded through Strands `invocationState`. */
 const TOOL_CONTEXT_KEY = '__bbAgentToolContext';
 
+/**
+ * Default runaway-protection caps applied per turn when `AgentConfig` does not
+ * override them. See `AgentConfig.maxLLMCalls` / `maxToolIterations`.
+ */
+const DEFAULT_MAX_LLM_CALLS = 20;
```
sarayev, line 51: Is there any rationale to chose those as default values?

packages/bb-agent/src/agent.ts
```diff
@@ -301,6 +337,16 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
 		const latencyMs = Date.now() - startTime;
 		this.log.info('runAgent done', { textLength: fullText.length, latencyMs, interrupted });
 
+		// A runaway-protection cap (maxLLMCalls / maxToolIterations) cancelled the turn.
+		// Cancellation ends the stream normally (no throw, stopReason 'cancelled'), and
+		// `capExceeded` is only ever set by the cap hooks above — so surface it as an
+		// error chunk and skip the final persist + 'done' below.
+		if (capExceeded) {
```
sarayev, line 415: Nice, clean stop. One thing I want to be sure of before this goes in: what does the conversation look like on the *next* turn after a cap trips?

When `maxToolIterations` fires, the tool that trips it gets cancelled, so it never produces a `tool-result`, but we've already persisted its `tool-call` earlier in the loop. That leaves a `tool_use` with nothing answering it, and Bedrock tends to reject a history like that on the following turn ("tool_use ids must have corresponding tool_result"). Strands' own `SessionManager` worries me for the same reason if `cancel()` flushes a half-finished turn.

Have you tried running a second turn after a trip on a real agent? If it replays fine, ignore me. If not, I think we'd need to patch up the history on the way out here (drop the dangling call, or write a synthetic error result) before returning.

packages/bb-agent/src/agent.ts
```diff
@@ -213,10 +220,39 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
...
+		// Enforced by counting Strands' per-turn hooks and cancelling the agent once
+		// a cap is exceeded — cancellation surfaces below as stopReason 'cancelled'.
+		// Registered here (not in createStrandsAgent) so the per-turn counters and the
+		// `capExceeded` reason stay in scope for the AgentResultEvent handler; Strands
+		// runs multiple hooks per event, so this coexists with the HITL hook.
+		let modelCallCount = 0;
+		let toolCallCount = 0;
+		let capExceeded: string | undefined;
+		// `false` disables a cap (→ Infinity, never trips); undefined uses the default.
+		const maxLLMConfig = this.config.maxLLMCalls ?? DEFAULT_MAX_LLM_CALLS;
```
sarayev: What happens if someone passes `maxLLMCalls: 0`? Looks like `++count > 0` trips on the very first model call, so the agent dies every turn. A negative value does the same, and `NaN` quietly turns the cap off. Someone reaching for `0` to mean "unlimited" (or "off") would be pretty confused.

Could we reject anything that isn't a positive integer and point people at `false` for disabling? Roughly:

```ts
if (v !== false && v !== undefined && (!Number.isInteger(v) || v < 1)) {
    throw blocksAgentError(AgentErrors.InvalidModelConfig, `cap must be a positive integer or false, got ${v}`);
}
```

packages/bb-agent/src/agent.ts
```diff
@@ -213,10 +220,39 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
...
+		const maxToolsConfig = this.config.maxToolIterations ?? DEFAULT_MAX_TOOL_ITERATIONS;
+		const maxLLM = maxLLMConfig === false ? Number.POSITIVE_INFINITY : maxLLMConfig;
+		const maxTools = maxToolsConfig === false ? Number.POSITIVE_INFINITY : maxToolsConfig;
+		strandsAgent.addHook(BeforeModelCallEvent, (event) => {
+			if (++modelCallCount > maxLLM) {
+				capExceeded ??= `exceeded maxLLMCalls (${maxLLM})`;
+				event.cancel = 'maxLLMCalls exceeded';
```
sarayev: Small thing: are both of these pulling weight? My read is `event.cancel` is what actually stops us paying for the over-limit call, while `strandsAgent.cancel()` unwinds the loop, in which case great, maybe just a one-line comment saying so. But if `cancel()` alone already blocks the in-flight call, then `event.cancel` is redundant (or the other way round). Which is it?

packages/bb-agent/src/index.test.ts
```diff
@@ -541,6 +541,80 @@ describe('CannedProvider', () => {
 	});
 });
 
+// ── runaway protection caps ──────────────────────────────────────────────────
+
+describe('runaway protection caps', () => {
```
sarayev, line 574: These four cover the counting logic well. The gap is the cancel path itself: the canned provider can't tell us whether `cancel()` leaves a replayable session behind, which is exactly the thing I'm nervous about in my comment on the `capExceeded` return. A sandbox e2e that caps a turn *and then sends a follow-up* would put that to bed.

packages/bb-agent/src/agent.ts
```diff
@@ -213,10 +220,39 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
 	 * TODO add comments for args
```
sarayev: nit, and not yours: there's a `TODO add comments for args` sitting right above the new block. Easy to knock out while you're in the file if you feel like it.

packages/bb-agent/API.md
```diff
@@ -29,6 +29,8 @@ export interface AgentConfig<TContext = DefaultToolContext> {
     description?: string;
     inferenceOnly?: boolean;
     logger?: ChildLogger;
+    maxLLMCalls?: number | false;
```
jvh-aws: Shouldn't this be `maxLlmCalls`?

packages/bb-agent/src/agent.ts
```diff
@@ -213,10 +220,39 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
...
 	 */
 	private async runAgent(message: string, conversationId: string | undefined, channelId: string, userId: string, interruptResponses?: Array<{ interruptId: string; response: string }>, context?: TContext): Promise<void> {
-		const { InterruptResponseContent, ModelStreamUpdateEvent, BeforeToolCallEvent, AfterToolCallEvent, AgentResultEvent } = await loadStrands();
+		const { InterruptResponseContent, ModelStreamUpdateEvent, BeforeModelCallEvent, BeforeToolCallEvent, AfterToolCallEvent, AgentResultEvent } = await loadStrands();
 		const strandsAgent = await this.createStrandsAgent(conversationId, context);
 		const startTime = Date.now();
 
+		// Runaway-protection caps (see AgentConfig.maxLLMCalls / maxToolIterations).
```
jvh-aws: Those counters are locals in `runAgent`, so `resume()` resets them to 0 every time. Isn't this per-segment rather than per-turn? tested it on real bedrock: `needsApproval` tool + `maxToolIterations: 3` + auto-approving the `resume()` → 10 tool calls and the cap never fired. isn't that basically the #498 runaway (a loop that re-enters itself), which is the thing this is meant to stop? fine for a human clicking approve, but wouldn't auto-approve / trusted tools / any non-HITL resume slip right through? should we persist the counts across `resume()` so it's a real per-turn cap?

jvh-aws (dismissed):

packages/bb-agent/src/agent.ts
```diff
@@ -213,10 +220,39 @@ export class AgentBase<TContext = DefaultToolContext> extends Scope {
...
+		const maxLLM = maxLLMConfig === false ? Number.POSITIVE_INFINITY : maxLLMConfig;
+		const maxTools = maxToolsConfig === false ? Number.POSITIVE_INFINITY : maxToolsConfig;
+		strandsAgent.addHook(BeforeModelCallEvent, (event) => {
+			if (++modelCallCount > maxLLM) {
+				capExceeded ??= `exceeded maxLLMCalls (${maxLLM})`;
+				event.cancel = 'maxLLMCalls exceeded';
+				strandsAgent.cancel();
+			}
+		});
+		strandsAgent.addHook(BeforeToolCallEvent, (event) => {
```
soberm, line 319: [blocker] Think this leaves a dangling `tool_use` in history when this trips mid-batch. The loop above always persists a `'tool-call'` message on `BeforeToolCallEvent` before it's known whether the call gets cancelled, and when this cap fires there's no matching `AfterToolCallEvent`/`tool-result` — compare to the deny path in `createStrandsAgent` (~line 456), which sets `event.cancel` *alone*, with no `agent.cancel()`, and lets the turn continue (implying Strands backfills a result there so history stays paired). Here we cancel the call *and* kill the whole turn via `strandsAgent.cancel()`, which is exactly the case where that backfill might not happen. If the session snapshot persists that, the next ordinary message on this conversation — not just a HITL resume — could replay a broken tool_use/tool_result pairing into Bedrock and get rejected outright. Worth verifying against a real session round trip before this ships; none of the 4 new tests send a follow-up turn after a trip.

packages/bb-agent/src/agent.ts
soberm: [blocker] `maxLLMCalls: 0` trips on the very first call (`1 > 0`) instead of reading as "unlimited" — same for negatives. Worse, `NaN` makes `n > NaN` always `false`, so the cap silently never fires, forever, no warning. For a safety backstop that's the bad-case failure mode. Can we validate here (or in the constructor) the same way `resolveContext`/the `needsApproval`+`interrupt` check do — throw `blocksAgentError(AgentErrors.InvalidModelConfig, ...)` for anything that isn't a positive integer or `false`?

packages/bb-agent/src/agent.ts
soberm: [question] Following up on the open thread here — pretty sure `event.cancel` and `strandsAgent.cancel()` are both doing real work, not redundant: the deny branch below (line 456) uses `event.cancel` alone to skip one call while letting the turn continue, so `strandsAgent.cancel()` here is what stops everything *after* this call. If that's right, the specific over-cap model call should never actually reach the provider. None of the new tests assert that directly though (they only check the client sees an `error`) — a call-counter on the canned provider would nail it down instead of us inferring it from the deny path.

packages/bb-agent/src/agent.ts
soberm, line 415: [minor] This branch only publishes to Realtime — nothing gets written to `this.messages`. Compare to the AsyncJob catch handler in the constructor, which persists an assistant message with the error in `metadata` in addition to the error chunk. Right now reloading a conversation after a cap trip shows a dangling tool-call (see the other comment) and then nothing — no record of why it stopped.

packages/bb-agent/src/index.test.ts
soberm, line 574: [minor] These cover the counting logic well but not the interaction with the HITL hook in `createStrandsAgent`, which also attaches to `BeforeToolCallEvent`. The "resets per segment on resume()" behavior seems right from reading the code, but it's subtle enough (two independent hooks on the same event) that a test pairing a `needsApproval` tool with a cap near its limit would be worth adding.

packages/bb-agent/README.md
```diff
@@ -359,6 +361,38 @@ const agent = new Agent(scope, 'support', {
 });
 ```
 
+### Limiting runaway cost
+
+An agent runs a reason→act loop: each iteration is one **model call**, optionally followed by tool calls, and a model call that requests no tools ends the turn. A misbehaving agent — or a prompt that induces one — can loop this cycle far longer than intended, an unbounded loop can run up unexpected cost.
```
soberm: [nit] "...than intended, an unbounded loop can run up unexpected cost." — comma splice, reads cleaner as two sentences or with a semicolon.

packages/bb-agent/README.md
```diff
@@ -359,6 +361,38 @@ const agent = new Agent(scope, 'support', {
...
+  systemPrompt: '...',
+  maxLLMCalls: 40,          // agent legitimately reasons over many steps
+  maxToolIterations: 60,    // ...and chains many tools per turn
+});
+
+// Or disable a cap entirely with `false`:
+const unbounded = new Agent(scope, 'batch', {
+  systemPrompt: '...',
+  maxLLMCalls: false,       // no per-turn model-call limit
+  maxToolIterations: false,
```
soberm: [nit] Trailing whitespace after `maxToolIterations: false,`.

soberm (changes requested):
[minor] (couldn't attach inline — `packages/bb-agent/src/errors.ts` isn't part of this PR's diff) `complete()` maps every `error` chunk to `StreamFailedException`, so a cap trip and a genuine crash look identical to callers unless they regex the message. Given `InterruptRequired` already gets its own code/class there, might be worth a dedicated `CapExceeded` code so client code can branch without string-matching.

packages/bb-agent/src/agent.ts
soberm: Documented the reasoning on the DEFAULT_MAX_* constants in 220be2c. Short version: 20 is a backstop, not a tuned budget — a healthy interactive turn settles at a handful of model/tool calls (plan step, one or two tool rounds, final answer), so 20 leaves roughly an order of magnitude of headroom while still cutting a runaway loop off well before it shows up on the bill. Anything that legitimately needs more is expected to raise it (or set `false`) deliberately.

packages/bb-agent/src/agent.ts
soberm: Fixed in 220be2c. Both caps are now validated at construction — positive integer or `false`, anything else throws `InvalidModelConfigException` naming the option and the bad value. That covers `0`, negatives, non-integers and `NaN` (the `NaN` case was the worst one: `n > NaN` is always false, so the cap silently never fired). New test asserts `0, -1, 1.5, NaN` are rejected for both options and that valid values still construct.

packages/bb-agent/API.md
soberm: Good catch — renamed to `maxLlmCalls` in 220be2c. Checked the rest of the repo: every other option lowercases acronyms (`userId`, `tokenUrl`, `secretArn`, `sessionTtlSeconds`), and this was the only one with an all-caps one, so it was the odd one out. Updated types, README, DESIGN, API.md, changeset and tests.

packages/bb-agent/src/agent.ts
soberm: Done in 220be2c — replaced the TODO with real `@param` docs for all six `runAgent` args.

packages/bb-agent/src/agent.ts
soberm: Both are pulling weight, and I added a comment saying so. `event.cancel` skips this specific over-cap call so it never reaches the provider (same mechanism the deny branch below uses on its own, which is why that one can let the turn continue); `agent.cancel()` is what additionally unwinds the loop so nothing after this call runs. Neither is redundant: without `event.cancel` the over-limit call would still go out, and without `agent.cancel()` the turn would carry on past it.

packages/bb-agent/src/agent.ts
soberm: You were right, and it was worse than per-segment in the auto-approve case you tested. Fixed in 220be2c: the counters now live in the agent`s `appState`, which the SessionManager persists with the session snapshot (same mechanism as the `trusted:<tool>` flags), and `runAgent` only resets them on the initial path — a resume carries the existing budget forward, so the cap bounds a whole logical turn. One subtlety that fell out of it: a tool call paused on an interrupt re-emits `BeforeToolCallEvent` when the turn resumes, so calls are deduped by `toolUseId`, otherwise an approved call would be charged twice and a cap of N would trip at N/2 approvals. Docs (types JSDoc, README, DESIGN, changeset) rewritten from "per segment" to "whole turn". I have not re-run your real-Bedrock repro against this — worth redoing that exact needsApproval + maxToolIterations: 3 + auto-approve loop to confirm it now trips at 3.

packages/bb-agent/src/agent.ts
soberm: Checked this, and the history stays paired — but not for the reason I assumed either. Strands still emits `AfterToolCallEvent` for a call cancelled via `event.cancel`, carrying the cancellation as the result, so the `tool-call` row we persisted on the way in does get its `tool-result` partner and there is no dangling `tool_use`. I actually implemented your synthetic-error-result suggestion first and the new test caught it as a duplicate (3 results for 2 calls), which is how I found out. So no patch-up needed; 220be2c documents the behaviour in DESIGN.md and adds a test asserting tool-call/tool-result counts match after a trip. It also persists an assistant message with the stop reason, so a reloaded conversation says why it ended instead of just stopping. Caveat: that is the local message table plus the canned provider, not a real Bedrock replay — a follow-up turn against real Bedrock after a trip is still unverified, flagged below.

packages/bb-agent/src/index.test.ts
soberm: Agreed, and this one is not covered — flagging it for follow-up rather than claiming it. 220be2c adds a test that a cap trip leaves paired tool-call/tool-result history (which is what caught that Strands does emit `AfterToolCallEvent` for a cancelled call, so no synthetic result is needed), but that is the local message table with the canned provider, not a session replay. A sandbox e2e that caps a turn and then sends a follow-up is still needed to close it out. Same for pairing a `needsApproval` tool with a cap near its limit to exercise the two `BeforeToolCallEvent` hooks together — the resume path now dedupes by `toolUseId`, and that is currently verified by reading the code, not by a test.

soberm (approved):

sarayev (approved):

PR #456, opened 2026-08-28
fix(core): fail fast on an invalid stack name before CDK provisions

osama-rizk:
## Card

Bug Bash board (Starter Kit, DX P2): **"Validate stack name before CDK runs (reject invalid chars)"** — *blocked 3/4 deploys in Round 4.*
🔗 https://github.com/orgs/aws-amplify/projects/141/views/15?pane=issue&itemId=195530047

## Problem

A stack name containing invalid characters (e.g. underscores) **passes `cdk synth`** and only fails with an opaque `StackNameInvalidFormat` **minutes into CloudFormation provisioning** — after resources have started being created. There's no pre-flight check, so the feedback loop is slow and the error doesn't name the real cause.

This is the same fail-fast category as the credential pre-check (#424): catch a deploy-killing input up front with an actionable message instead of waiting through provisioning for an opaque failure.

## Fix

A single shared **`assertValidStackName`** at the naming chokepoint in `core/common`, called at the top of `BlocksStack.create` (before any construct is built). It validates the stack name against the CloudFormation contract:

- must start with a letter,
- only letters, digits, and hyphens (no underscores/spaces/other),
- ≤ 128 characters,

and throws immediately with a clear message **and a suggested valid name**, e.g.:

> Invalid AWS Blocks stack name "my_app": it must start with a letter and contain only letters, digits, and hyphens (no underscores, spaces, or other characters). CloudFormation rejects it, but only after several minutes of provisioning. Rename the stack — for example "my-app".

### Why here (structural, not tactical)

Every stack and Building Block resource name derives from the stack name via `computeScopeFullId` → `fullId` (`fullId.substring(0, 255)`). That single derivation point is where the name contract belongs, so this is deliberately **one shared validator** — the home other name checks can consolidate onto — rather than an ad-hoc throw. The validator is internal (not added to the public API surface).

## Testing

- New unit tests for `assertValidStackName`: accepts valid names (incl. 128-char boundary); rejects underscores, spaces, leading digit/hyphen, punctuation, and >128 chars; asserts the error names the offending value and offers a sanitized suggestion.
- No behavior change for valid stack names — full `@aws-blocks/core` suite **710 pass / 0 fail**, CDK `BlocksStack` tests 41/41. Scanned the repo/test-apps: no existing stack name uses an invalid character.

## Verification

`npm run lint` 0 errors · `check:api` ✅ (validator is internal — no public-API change) · umbrella-changeset guard ✅.

## Changeset

`@aws-blocks/core` **patch** + `@aws-blocks/blocks` **patch** (umbrella re-export). No valid app is affected; an invalid name never deployed successfully — it now just fails in seconds instead of minutes.


packages/core/src/cdk/index.ts
```diff
@@ -64,6 +65,10 @@ export class BlocksStack extends cdk.Stack implements BaseBlocksStack {
   static async create(scope: Construct, id: string, props: BlocksStackProps) {
     assertCdkConditionActive();
 
+    // Fail fast: an invalid stack name otherwise passes synth and only fails
+    // minutes into CloudFormation provisioning with an opaque error.
+    assertValidStackName(id);
```
soberm: [blocker] This validates `id`, not the actual stack name CloudFormation gets. Two ways that bites: `props.stackName` overrides (used elsewhere in this repo, see `packages/pipeline/src/types.ts`) go straight into `super(scope, id, props)` and skip this check entirely. And when `create()` runs inside a `cdk.Stage` — which is exactly how every Pipeline `stageFactory` example here works, always with a fixed id like `'App'` — CDK prepends the stage name to build the real stack name. Checked it directly: `new cdk.Stack(new cdk.Stage(app, 'qa_east'), 'App').stackName` comes out `'qa_east-App'`. `validateStageName()` in `pipeline-construct.ts` currently allows underscores in stage names, so that's a genuinely invalid name slipping past both checks, since this one only ever sees the constant `'App'`. Would validating `stack.stackName` after construction, instead of the raw `id`, cover both cases?

packages/core/src/common/index.ts
```diff
@@ -323,6 +323,56 @@ export function computeScopeFullId(scope: { id: string; parent?: any }) {
...
+    .replace(/-+/g, '-')
+    .replace(/^-|-$/g, '')
+    .slice(0, STACK_NAME_MAX_LENGTH);
+  return cleaned || 'my-app';
+}
+
+/**
+ * Fail fast at synth if a stack name would be rejected by CloudFormation.
+ *
+ * Without this, an invalid name (e.g. one containing underscores) passes `cdk
```
soberm: [question] Did you check this against the CDK version pinned here? `aws-cdk-lib` is `^2.257.0`, and running `new cdk.Stack(new cdk.App(), 'my_bad_name')` against what's actually in `node_modules` throws synchronously — `StackNameInvalidFormat`, right when the Stack gets constructed, not during provisioning. CDK's own constructor already runs this same regex plus the 128-char cap. If that's accurate, this mostly buys a nicer message and a suggested name rather than shortening the feedback loop, and the real bug-bash repro might be coming from somewhere this doesn't touch (a Stage-prefixed pipeline stack, or `BlocksBackend`/`NestedStack`, where `stackName` is a token and CDK's check gets bypassed).

packages/core/src/common/validate-stack-name.test.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+import { assertValidStackName } from './index.js';
+
+describe('assertValidStackName', () => {
```
soberm: [major] All of this tests `assertValidStackName` in isolation. Given the id-vs-stackName gap above, an end-to-end test through `BlocksStack.create()` — or better, a Pipeline stage with an underscore in the stage name and the id fixed to something valid — would actually prove the "fails before any infrastructure exists" claim. Worth adding either way.

packages/core/src/common/index.ts
```diff
@@ -323,6 +323,56 @@ export function computeScopeFullId(scope: { id: string; parent?: any }) {
...
+  const reason =
+    name.length > STACK_NAME_MAX_LENGTH
+      ? `it is ${name.length} characters (max ${STACK_NAME_MAX_LENGTH})`
+      : 'it must start with a letter and contain only letters, digits, and hyphens (no underscores, spaces, or other characters)';
+  const err = new Error(
+    `Invalid AWS Blocks stack name "${name}": ${reason}. ` +
+      `CloudFormation rejects it, but only after several minutes of provisioning. ` +
+      `Rename the stack — for example "${suggestStackName(name)}".`,
+  );
+  err.name = 'InvalidStackNameError';
```
soberm: [minor] Plain `Error` with `.name` set by hand means callers can only tell this apart from any other error by string-comparing `.name`. A small `InvalidStackNameError` class would let people `instanceof` it, and it'd match the `ApiError` pattern already in this package.

packages/core/src/common/index.ts
```diff
@@ -323,6 +323,56 @@ export function computeScopeFullId(scope: { id: string; parent?: any }) {
...
+ * synth` and only fails with an opaque `StackNameInvalidFormat` **minutes into
+ * provisioning**. This turns that into an immediate, actionable error before any
+ * infrastructure is created.
+ *
+ * @param name - The stack name (the `BlocksStack` id).
+ * @throws {Error} With guidance and a suggested valid name when `name` is invalid.
+ */
+export function assertValidStackName(name: string): void {
+  if (STACK_NAME_PATTERN.test(name) && name.length <= STACK_NAME_MAX_LENGTH) return;
+  const reason =
```
soberm: [minor] When a name is both too long and has bad characters, this only reports the length problem, even though the suggestion below still rewrites the characters too. Not wrong, just a bit incomplete for that combo.

packages/core/src/common/validate-stack-name.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  for (const name of ['my_app', 'my app', '1app', '-app', 'app!', 'a'.repeat(129)]) {
+    it(`rejects the invalid name "${name.slice(0, 12)}${name.length > 12 ? '…' : ''}"`, () => {
+      assert.throws(
+        () => assertValidStackName(name),
+        (e: unknown) => {
+          assert.ok(e instanceof Error);
+          assert.strictEqual(e.name, 'InvalidStackNameError');
+          assert.match(e.message, new RegExp(`"${name.slice(0, 5)}`)); // names the offending value
```
soberm: [nit] `new RegExp(`\"${name.slice(0, 5)}`)` works for the current list, but the first name with a regex-special character in it (a `.` or a `(`) will either throw or match something unexpected. `e.message.includes(name.slice(0, 5))` does the same job without that risk.

packages/core/src/common/index.ts
```diff
@@ -323,6 +323,56 @@ export function computeScopeFullId(scope: { id: string; parent?: any }) {
...
+ * at each call site: it is the one place the naming contract is enforced, so new
+ * name rules and other name-producing sites can consolidate here.
+ */
+const STACK_NAME_PATTERN = /^[A-Za-z][A-Za-z0-9-]*$/;
+const STACK_NAME_MAX_LENGTH = 128;
+
+/** Suggest a valid stack name derived from an invalid one, for the error message. */
+function suggestStackName(name: string): string {
+  const cleaned = name
+    .replace(/[^A-Za-z0-9-]+/g, '-') // non-conforming runs → single hyphen
```
soberm: [nit] These two comments mostly just restate the regex above them. Could probably go, the JSDoc already explains the contract.

soberm (changes requested):
Solid unit tests on the validator itself, and the error message is a real improvement over CDK's raw regex dump. But the check validates `id` instead of the resolved stack name, so it doesn't cover this repo's own Pipeline stage-prefixing pattern or `stackName` overrides — see inline comments. Also worth double-checking against CDK's existing synth-time validation before merging; not convinced this closes the loop the bug-bash ticket describes yet.

packages/core/src/cdk/index.ts
osama-rizk: You're right — verified against 2.257.0. CDK's `Stack` constructor already throws `StackNameInvalidFormat` synchronously at construction for a concrete `stackName` (incl. `props.stackName` overrides and stage-prefixed names like `qa_east-App`), and my check validated the wrong string (`id`, not the effective name). The real unguarded input is the **stage name**: `validateStageName` allowed underscores/leading digits, and the stage name prefixes the derived stack name.

Pivoted: dropped the `BlocksStack.create(id)` check entirely and moved enforcement into `validateStageName` — it now requires `/^[A-Za-z][A-Za-z0-9-]*$/` and fails fast at synth with a suggested valid name (e.g. `qa_east` → `qa-east`). (Token stack names — nested/`BlocksBackend` — remain deploy-time by nature; they're unresolved at synth, so nothing can regex them there.)

packages/core/src/common/index.ts
osama-rizk: Confirmed — great catch. Probe on the pinned 2.257.0: `new cdk.Stack(new cdk.App(), 'my_bad_name')` throws `StackNameInvalidFormat` at **construction**, not provisioning; same for `props.stackName` and `Stage('qa_east')/'App'` (`got 'qa_east-App'`). So my check only duplicated CDK for concrete names. `NestedStack.stackName` is a **token** (`Token.isUnresolved === true`) → CDK skips it → that's the genuine deploy-time case, and it's un-validatable at synth. So the useful fix is the **stage-name** path, which is what this now does.

packages/core/src/common/validate-stack-name.test.ts
osama-rizk: Done — replaced the isolated `assertValidStackName` tests with end-to-end `Pipeline` tests: a stage named `qa_east` throws at synth (asserting the message + the `qa-east` suggestion), and a valid `qa-east` stage constructs cleanly. That exercises the real path (`validateStageName` → derived stack name) rather than a standalone helper.

packages/core/src/common/index.ts
osama-rizk: Resolved by the pivot — the standalone `assertValidStackName`/`InvalidStackNameError` in core is removed (it was redundant with CDK). The pipeline error stays a plain `Error` to match the surrounding `validateStageName`/`validateBakeTime` style in this file. Happy to introduce a typed pipeline error class if you'd prefer consistency there.

packages/core/src/common/index.ts
osama-rizk: Moot now (that validator was removed), but noted for the pipeline message — it reports the single contract clearly ("must start with a letter and contain only letters, numbers, and hyphens") and always includes a sanitized suggestion, so a name that's wrong for multiple reasons still gets one actionable fix.

packages/core/src/common/validate-stack-name.test.ts
osama-rizk: Fixed indirectly — the isolated test with the `new RegExp(...)` matcher is gone. The new pipeline tests use `assert.match(msg, /…/)` with fixed literals (no interpolation of the offending name into a regex), so the regex-special-char risk doesn't apply.

packages/core/src/common/index.ts
osama-rizk: Removed — that whole block (and its restating comments) is gone with the `assertValidStackName` revert. The pipeline change keeps one short comment explaining *why* the contract applies (the stage name prefixes the stack name), which isn't obvious from the regex alone.

soberm (dismissed):

Simone319 (approved):

PR #457, opened 2026-08-30
fix(dart): Pin flutter_secure_storage to ^10.3.1 to unbreak Android builds

cadivus:
## Problem

`blocks_runtime_flutter` depends on `flutter_secure_storage: ^11.0.0`. v11 raised its Android plugin to `compileSdk = 37`, but the current stable Flutter app template (3.47) compiles against SDK 36 with AGP 9.1.0, whose maximum recommended `compileSdk` is 36. Every Flutter app using `blocks_runtime_flutter` therefore fails its Android build:

```
Execution failed for task ':app:checkReleaseAarMetadata'.
> Dependency ':flutter_secure_storage' requires libraries and applications that
  depend on it to compile against version 37 or later of the Android APIs.
  :app is currently compiled against android-36.
```

This is the upstream issue juliansteenbakker/flutter_secure_storage#1224 — https://github.com/juliansteenbakker/flutter_secure_storage/issues/1224.

Raising the app's `compileSdk` to 37 is not a workable fix for consumers, since it also requires an AGP newer than the stable Flutter template ships.

## Change

Constrain `flutter_secure_storage` to `^10.3.1`, whose Android plugin uses `compileSdk = 36`.

No source changes needed: `FlutterSecureStore` only uses `read`/`write`/`delete`, so none of the v11 removals (`encryptedSharedPreferences`, `sharedPreferencesName`, legacy cipher algorithms) apply.

## Verification

Tested with a Flutter app (Flutter 3.47.0 stable) using `dependency_overrides` to this branch:

- `flutter build apk` — succeeds (previously failed at `checkReleaseAarMetadata`)
- `flutter run` on an Android API 36 emulator — app launches, no runtime errors
- `blocks_runtime_flutter` unit tests — 7/7 pass on 10.3.1

### DIFF
--- a/.github/dependency-review-config.yml
+++ b/.github/dependency-review-config.yml
@@ -27,3 +27,18 @@ allow_licenses:
   - "Python-2.0"
   - "WTFPL"
   - "Zlib"
+
+# Temporary exclusions from license checking.
+#
+# GitHub's dependency graph reports Pub (Dart/Flutter) package licenses in
+# lowercase (e.g. "bsd-3-clause", sourced from pub.dev's pana tags), which the
+# dependency-review-action's case-sensitive SPDX parser rejects as invalid —
+# it fails the check before the allowlist above is even consulted.
+#
+# flutter_secure_storage is verifiably BSD-3-Clause:
+# https://pub.dev/packages/flutter_secure_storage/license
+#
+# Remove once the upstream parser normalizes license casing:
+# https://github.com/actions/dependency-review-action
+allow_dependencies_licenses:
+  - "pkg:pub/flutter_secure_storage"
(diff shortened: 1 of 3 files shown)


VarshithaPamisetty (dismissed):

osama-rizk:
The `dependency-review-pr` check is failing on the **license** step (not vulnerabilities):

```
native/dart/packages/blocks_runtime_flutter/pubspec.yaml » flutter_secure_storage@>= 10.3.1, < 11.0.0 – License: bsd-3-clause
Dependency review could not detect the validity of all licenses.
```

`flutter_secure_storage` publishes its license as `bsd-3-clause` (lowercase), which `dependency-review-action` can't match to a canonical SPDX identifier — so it's reported as "validity could not be determined" rather than checked against the allowlist. `BSD-3-Clause` is already in `allow_licenses`, so adding the license won't clear it; the **package** has to be excluded from the license check.

The action's mechanism for that is `allow-dependencies-licenses` (a purl list). Add to `.github/dependency-review-config.yml`:

```yaml
allow-dependencies-licenses:
  - "pkg:pub/flutter_secure_storage"
```

One thing worth double-checking while you're in that file: the existing key is `allow_licenses` (snake_case), but `dependency-review-action@v5` documents its config keys in kebab-case (`allow-licenses` / `allow-dependencies-licenses`). If the snake_case key is being silently ignored, the license allowlist may not be taking effect at all — worth confirming the key name matches what the action expects.

osama-rizk (dismissed):

soberm (approved):

osama-rizk (approved):

PR #458, opened 2026-08-31
fix(bb-cron-job): validate the schedule/timezone at synth, not at deploy

osama-rizk:
## Card

Bug Bash board (Starter Kit, DX P2): **"Cron template `rate(10 seconds)` rejected by EventBridge."**
🔗 https://github.com/orgs/aws-amplify/projects/141/views/15?pane=issue&itemId=195535058

## Problem

`CronJob`'s CDK layer passed `options.schedule`/`options.timezone` straight into the EventBridge `CfnSchedule` **with no validation**. An invalid expression — e.g. `rate(10 seconds)` (EventBridge's minimum is 1 minute) or a malformed `cron(...)` — **passed `cdk synth`** and was only rejected by EventBridge **minutes into provisioning**. Meanwhile the mock already validated the same expressions, so **local dev and deploy diverged**: it worked (failed) locally but the deploy path let it through to a slow, opaque failure.

Same fail-fast category as the credential pre-check (#424) and the stack-name check — catch a deploy-killing input up front instead of after provisioning.

## Fix

Extracted the rate/cron parser and the timezone check out of `index.mock.ts` into a shared **`schedule`** module used by **both** the mock (which parses to fire locally) and the CDK construct (which only needs validation). `CronJob`'s constructor now calls `validateSchedule(...)` / `validateTimezone(...)` up front and throws **`CronJobErrors.InvalidSchedule` / `CronJobErrors.InvalidTimezone` at synth**, before any infrastructure is created.

This also removes ~130 lines of parsing logic from the mock in favor of the shared module — one source of truth for the schedule contract, so mock and deploy can't drift again. The shared module is internal (no public API change).

## Testing

- **`schedule-validate.test.ts`** — direct unit tests for the shared validators: accepts valid `rate()`/`cron()`; rejects `rate(10 seconds)`, `rate(0 minutes)`, plural/singular mismatches, 5-field cron, out-of-range cron, empty string; timezone accept/reject.
- **`index.cdk.test.ts`** — CDK synth tests (stub stack): the `CronJob` construct itself throws `InvalidSchedule` for `rate(10 seconds)` and `InvalidTimezone` for a bad zone, and synthesizes a `CfnSchedule` for a valid schedule.
- Existing `schedule.test.ts` (mock parsing) unchanged and green.
- `@aws-blocks/bb-cron-job` suite **45 pass / 0 fail**.

## Verification

`npm run lint` 0 errors · `check:api` ✅ (regenerated — adds the public `CronJobErrors.ScheduleNotSupported`) · conditional-export parity 21/21 · umbrella-changeset guard ✅.

## Changeset

`@aws-blocks/bb-cron-job` **minor** + `@aws-blocks/blocks` **patch** (umbrella re-export). No behavior change for valid schedules; an invalid schedule never deployed successfully — it now fails in seconds at synth instead of minutes into provisioning.

### DIFF
--- a/.changeset/cron-schedule-synth-validation.md
+++ b/.changeset/cron-schedule-synth-validation.md
@@ -1,0 +1,10 @@
+---
+"@aws-blocks/bb-cron-job": patch
+"@aws-blocks/blocks": patch
+---
+
+`CronJob`: validate the schedule (and timezone) at synth so an invalid expression fails fast instead of after minutes of deploy.
+
+The CDK layer passed `schedule`/`timezone` straight into the EventBridge `CfnSchedule` with no validation, so an invalid expression — e.g. `rate(10 seconds)` (EventBridge's minimum is 1 minute) or a malformed `cron(...)` — passed `cdk synth` and was only rejected by EventBridge minutes into provisioning. The mock already validated these, so local dev and deploy diverged.
+
+The schedule/rate/cron parser and the timezone check are now a shared `schedule` module used by both the mock and the CDK construct. `CronJob`'s constructor validates up front and throws `CronJobErrors.InvalidSchedule` / `CronJobErrors.InvalidTimezone` at synth, before any infrastructure is created. No behavior change for valid schedules.
--- a/.changeset/cron-schedule-synth-validation.md
+++ b/.changeset/cron-schedule-synth-validation.md
@@ -1,5 +1,5 @@
 ---
-"@aws-blocks/bb-cron-job": patch
+"@aws-blocks/bb-cron-job": minor
 "@aws-blocks/blocks": patch
 ---
 
@@ -7,4 +7,6 @@
 
 The CDK layer passed `schedule`/`timezone` straight into the EventBridge `CfnSchedule` with no validation, so an invalid expression — e.g. `rate(10 seconds)` (EventBridge's minimum is 1 minute) or a malformed `cron(...)` — passed `cdk synth` and was only rejected by EventBridge minutes into provisioning. The mock already validated these, so local dev and deploy diverged.
 
-The schedule/rate/cron parser and the timezone check are now a shared `schedule` module used by both the mock and the CDK construct. `CronJob`'s constructor validates up front and throws `CronJobErrors.InvalidSchedule` / `CronJobErrors.InvalidTimezone` at synth, before any infrastructure is created. No behavior change for valid schedules.
+The schedule parser and timezone check are now a shared `schedule` module used by both the mock and the CDK construct. `CronJob`'s constructor validates up front and throws `CronJobErrors.InvalidSchedule` / `CronJobErrors.InvalidTimezone` at synth, before any infrastructure is created.
+
+The synth gate is deliberately lenient for `cron(...)` — it checks the 6-field shape and defers field-level semantics (`L`/`W`/`#`, year fields, named days) to EventBridge — so it never rejects an advanced-but-valid schedule. The local mock, which must actually simulate the schedule, still can't model those forms; it now throws the new `CronJobErrors.ScheduleNotSupported` (rather than `InvalidSchedule`) for them, so local dev doesn't call a deployable schedule "invalid". No behavior change for valid, mock-supported schedules.
--- a/packages/bb-cron-job/src/index.cdk.ts
+++ b/packages/bb-cron-job/src/index.cdk.ts
@@ -10,7 +10,7 @@ import type {
 	CronJobEvent,
 	CronJobOptions,
 } from './types.js';
-import { CronJobErrors } from './errors.js';
+import { validateSchedule, validateTimezone } from './schedule.js';
 
 export { CronJobErrors } from './errors.js';
 export type { CronJobEvent, CronJobOptions } from './types.js';
@@ -20,6 +20,12 @@ export class CronJob<T = void> extends Scope {
 
 	constructor(scope: ScopeParent, id: string, options: CronJobOptions<T>) {
 		super(id, { parent: scope });
+
+		// Fail fast at synth: an invalid schedule/timezone otherwise passes synth
+		// and is only rejected by EventBridge minutes into the deploy. Mirrors the
+		// validation the mock already runs, so local dev and deploy agree.
+		validateSchedule(options.schedule);
+		if (options.timezone !== undefined) validateTimezone(options.timezone);
 
 		const lambdaArn = this.handler.functionArn;
 		const schedulerRole = getOrCreateSchedulerRole(cdk.Stack.of(this), lambdaArn);
--- a/packages/bb-cron-job/src/schedule.ts
+++ b/packages/bb-cron-job/src/schedule.ts
@@ -75,10 +75,21 @@ export function parseScheduleForMock(expr: string): ParsedSchedule {
 	try {
 		return parseSchedule(expr);
 	} catch (err) {
-		// Reclassify ONLY when the parse failure is due to an advanced cron form the
-		// mock doesn't model (not a genuinely invalid value like minute 100 or an
-		// inverted range — those stay InvalidSchedule).
-		if (!usesUnsupportedCronForm(expr)) throw err;
+		// Reclassify as "unsupported in mock" ONLY when the parse failure is
+		// attributable to an advanced cron form (L/W/#, year) AND the rest of the
+		// expression is otherwise well-formed. A genuinely bad field (minute 100,
+		// an inverted range) must still throw InvalidSchedule even when a year or
+		// marker is also present — otherwise we'd claim a deploy-doomed schedule
+		// "will deploy". We test this by normalizing the markers/year to valid
+		// placeholders and re-parsing: if it now parses, the marker was the only
+		// problem; if it still throws, a real field is bad.
+		const normalized = normalizeUnsupportedCron(expr);
+		if (normalized === null) throw err; // no advanced form → the failure is a genuine error
+		try {
+			parseSchedule(normalized);
+		} catch {
+			throw err; // a non-marker field is also invalid → keep InvalidSchedule
+		}
 		const e = new Error(
 			`${CronJobErrors.ScheduleNotSupported}: "${expr}" is valid for EventBridge and will deploy, ` +
 				'but the local mock cannot simulate it (advanced cron forms like L/W/#, or a year field). ' +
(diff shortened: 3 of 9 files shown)


packages/bb-cron-job/src/schedule.ts
```diff
@@ -0,0 +1,185 @@
...
+	dayOfWeek: number[];
+}
+export type ParsedSchedule = RateSchedule | CronSchedule;
+
+const RATE_RE = /^rate\((\d+)\s+(minutes?|hours?|days?)\)$/i;
+const CRON_RE = /^cron\((.+)\)$/;
+
+function scheduleError(expr: string): Error {
+	const err = new Error(`${CronJobErrors.InvalidSchedule}: "${expr}" is not a valid cron or rate expression`);
+	err.name = CronJobErrors.InvalidSchedule;
```
Simone319, line 42: **Major:** The mock's `parseSchedule` path throws an error named `InvalidSchedule` for advanced cron that EventBridge — and this PR's own synth gate — accept, e.g. `cron(0 10 L * ? *)`. In local dev that tells a developer their schedule is invalid when it will actually deploy and run fine. Consider a distinct error name/message (e.g. `ScheduleNotSupportedInMock`) so 'can't simulate locally' reads differently from 'invalid'.

packages/bb-cron-job/src/schedule.ts
```diff
@@ -0,0 +1,185 @@
...
+export interface CronFields {
+	minute: number[];
+	hour: number[];
+	dayOfMonth: number[];
+	month: number[];
+	dayOfWeek: number[];
+}
+export type ParsedSchedule = RateSchedule | CronSchedule;
+
+const RATE_RE = /^rate\((\d+)\s+(minutes?|hours?|days?)\)$/i;
```
Simone319: **Minor:** `RATE_RE` has the `/i` flag but `CRON_RE` does not, so `Rate(5 minutes)` passes the synth gate and then EventBridge rejects it at deploy (EventBridge requires lowercase `rate`/`cron`). Dropping `/i` here keeps the two consistent and fails fast on uppercase `rate` too.

packages/bb-cron-job/src/index.cdk.ts
```diff
@@ -21,6 +21,12 @@ export class CronJob<T = void> extends Scope {
 	constructor(scope: ScopeParent, id: string, options: CronJobOptions<T>) {
 		super(id, { parent: scope });
 
+		// Fail fast at synth: an invalid schedule/timezone otherwise passes synth
+		// and is only rejected by EventBridge minutes into the deploy. Mirrors the
+		// validation the mock already runs, so local dev and deploy agree.
+		validateSchedule(options.schedule);
```
Simone319, line 27: **Minor:** `validateSchedule(options.schedule)` has no `undefined` guard, while `timezone` gets one on the next line. If `schedule` can ever be missing at runtime this throws a raw `TypeError` instead of `InvalidSchedule` — fine if `schedule` is a required field, just worth confirming.

packages/bb-cron-job/src/schedule.ts
```diff
@@ -0,0 +1,185 @@
...
+const CRON_RE = /^cron\((.+)\)$/;
+
+function scheduleError(expr: string): Error {
+	const err = new Error(`${CronJobErrors.InvalidSchedule}: "${expr}" is not a valid cron or rate expression`);
```
Simone319: **Nit:** For the flagship case this PR fixes (`rate(10 seconds)`), the message just says it 'is not a valid cron or rate expression' without naming the constraint. Calling out the actual rule (units are minute/hour/day; EventBridge minimum is 1 minute) would make it actionable.

packages/bb-cron-job/src/schedule.ts
osama-rizk: Agreed — fixed. Added `CronJobErrors.ScheduleNotSupported` ('ScheduleNotSupportedInMock') and a `parseScheduleForMock` the mock uses: when a parse failure is caused by an advanced-but-valid cron form (`L`/`W`/`#` in a day field, or a non-`*` year), it throws `ScheduleNotSupported` ("valid for EventBridge and will deploy, but the mock can't simulate it locally") instead of `InvalidSchedule`. Genuinely bad values (minute 100, inverted ranges, `rate(10 seconds)`) still throw `InvalidSchedule`. Added tests (`cron(0 10 L * ? *)`, `cron(0 10 LW * ? *)` → ScheduleNotSupported; invalids → InvalidSchedule) and documented it in D-CJ-4.

packages/bb-cron-job/src/schedule.ts
osama-rizk: Good catch — dropped the `/i` from `RATE_RE` so it's case-sensitive like `CRON_RE`. `Rate(5 minutes)`/`CRON(...)` now fail fast at the gate instead of being rejected by EventBridge at deploy. Added both to the rejection tests.

packages/bb-cron-job/src/index.cdk.ts
osama-rizk: Confirmed: `schedule` is a required field on `CronJobOptions` (`types.ts:26`, `schedule: string`), so there's no `undefined` path for typed callers — the guard exists on `timezone` only because it's optional (`timezone?: string`). Left `validateSchedule(options.schedule)` unguarded intentionally.

packages/bb-cron-job/src/schedule.ts
osama-rizk: Done — `scheduleError` now names the constraint. A `rate(`-shaped expression reports: `a rate must be \"rate(<n> minute|minutes|hour|hours|day|days)\" with n ≥ 1 (EventBridge's minimum is 1 minute; seconds are not supported)`, so `rate(10 seconds)` is actionable; other shapes get the `rate(...)`-or-6-field-`cron(...)` guidance.

packages/bb-cron-job/src/schedule.ts
```diff
@@ -0,0 +1,232 @@
...
+ * parser doesn't model: `L`/`W`/`#` in a day field, or an explicit (non-`*`/`?`)
+ * year field. Used to tell "can't simulate locally" apart from "invalid".
+ */
+function usesUnsupportedCronForm(expr: string): boolean {
+	const m = expr.match(CRON_RE);
+	if (!m) return false;
+	const fields = m[1].trim().split(/\s+/);
+	if (fields.length !== 6) return false;
+	const [, , dayOfMonth, , dayOfWeek, year] = fields;
+	return /[LW#]/i.test(dayOfMonth) || /[LW#]/i.test(dayOfWeek) || (year !== '*' && year !== '?');
```
soberm: [blocker] This returns `true` the moment a year field is present (or L/W/# shows up), regardless of whether the rest of the expression is actually valid. So something like `cron(100 9 * * ? 2025)` — minute 100, plus a year — fails `parseSchedule` correctly, but then this says "unsupported form" purely because of the year, and `parseScheduleForMock` rethrows it as `ScheduleNotSupported` claiming it "is valid for EventBridge and will deploy." It isn't and it won't — EventBridge rejects minute 100, and `validateSchedule` only checks field count so the synth gate lets it through too. That's the exact "looks fine, fails at deploy" gap this PR is closing, reopened through the year-field branch. Also contradicts the DESIGN.md D-CJ-4 note (~line 80) that genuinely bad values still throw `InvalidSchedule` in both layers, and it's untested — the existing cases all use year `*`, which never touches this branch. Can we confirm the other fields are well-formed (marker normalized out, re-run through the range checks) before deciding this is "just unsupported"? Would add `cron(100 9 * * ? 2025)` and an inverted range with a year as cases, both asserting `InvalidSchedule`.

.changeset/cron-schedule-synth-validation.md
```diff
@@ -0,0 +1,12 @@
+---
+"@aws-blocks/bb-cron-job": minor
```
soberm, line 2: [nit] This says `minor`, which is correct now that `ScheduleNotSupported` is a public addition, but the PR description's Changeset section still says `patch` (and the Fix section still says "no public API change") from before this got bumped. Worth a quick pass on the description so it matches what's actually shipping.

soberm (changes requested):
Went through this again against the latest push (3da30af → 27cf697 → 4f6fa80). The synth-gate/shared-module split is solid and `parseScheduleForMock` vs `validateSchedule` is the right shape conceptually, but there's a real bug in how the mock decides "unsupported" vs "actually invalid" (inline below), plus a couple loose ends from the last two commits worth tidying in the same pass:

- `index.browser.ts`'s `CronJobErrors` wasn't updated with `ScheduleNotSupported` — still just the original two members. Not inline-commentable since this file isn't touched by the diff, but it's a real drift: anyone on the browser build referencing `CronJobErrors.ScheduleNotSupported` gets `undefined`. The conditional-exports parity test won't catch it — it only diffs `Object.keys()` of the module namespace, not what's inside this const, and the browser condition isn't even in the compared set for BB packages. Cleanest fix is probably re-exporting from `./errors.js` here like the mock/cdk/aws entries do instead of hand-maintaining a second copy.
- `README.md`'s "Error Constants" section (~line 101) still lists only two constants and says "Both errors are thrown at construction time" — stale now that there's a third one that's mock-only. Worth adding `ScheduleNotSupported` there with a note that it's thrown locally, not at synth/deploy, and fixing "Both."

The `schedule.ts` one is the blocker, the rest is cheap cleanup in the same revision.

Separate from the code: the three failing checks (Build and Test, E2E Production, Telemetry E2E) trace back to an EC2 VPC quota limit and an API Gateway 429 during stack teardown on unrelated resources in the shared test account, not this diff. Worth a re-run, wouldn't hold the PR on it.

packages/bb-cron-job/src/schedule.ts
osama-rizk: Great catch — real gap, fixed in f7d468bf. The classifier no longer trusts "a marker/year is present"; it now **normalizes the markers/year to valid placeholders and re-parses**. Only if the normalized expression parses do we call it `ScheduleNotSupported`; if it still throws, a real field is bad and we keep `InvalidSchedule`. So `cron(100 9 * * ? 2025)` normalizes to `cron(100 9 * * ? *)`, which still throws on minute 100 → `InvalidSchedule` (no false "will deploy"). Added both cases you asked for — `cron(100 9 * * ? 2025)` and `cron(0 9 30-10 * ? 2025)` (inverted range + year) — asserting `InvalidSchedule`. This also restores agreement with the D-CJ-4 note that genuinely bad values throw `InvalidSchedule` in both layers.

.changeset/cron-schedule-synth-validation.md
osama-rizk: Fixed the description — the Changeset section now says `@aws-blocks/bb-cron-job` **minor**, and the verification line reflects that `check:api` was regenerated to add the public `CronJobErrors.ScheduleNotSupported` (dropped the stale "no public-API change" wording).

soberm (approved):

Simone319 (approved):

PR #459, opened 2026-08-31
[bb-async-job + 2 more] Route event-block resources to their resolved compute

Simone319:
## Problem
Event Building Blocks (AsyncJob, CronJob, Realtime) wired their native AWS resources — SQS event source, EventBridge scheduler target, WebSocket integration — directly to the stack's shared Lambda handler. For multi-compute, each event block must instead target the compute that owns it, so its resources follow the resolved compute rather than assuming the single default handler.

## Changes
- **AsyncJob** (`bb-async-job`): attach the SQS event source to `this.compute`'s function (guarded on `LambdaCompute`) instead of the shared handler; partial-batch-failure options preserved.
- **Synth guard (new, all three blocks):** AsyncJob, CronJob, and Realtime each gain a `LambdaCompute` guard in this change (none had one before) that throws a typed `UnsupportedCompute` error on any other compute type — assertable via `isBlocksError` and covered by a throw-path test per block.
- **CronJob** (`bb-cron-job`): resolve the scheduler target from `this.compute`.
- **Realtime** (`bb-realtime`): wire the shared WebSocket infra from `this.compute`'s function.
- Each event block declares a `@aws-blocks/bb-lambda-compute` dependency and imports the CDK-typed `LambdaCompute` via its new `/cdk` subpath (no coupling to core internals).
- No public `{ compute }` option is added here — blocks reach `this.compute`, which resolves to the default compute for any current app. The customer-facing assignment surface is a later track.

## Validation
- CDK tests updated to the `BlocksStack.create` harness and assert the event source / target / integration lands on the resolved compute's function (default and ancestor-`_compute` cases).
- Green locally: `bb-async-job` 77, `bb-cron-job` 29, `bb-realtime` 72, `core` 729; full build + lint pass.

## Checklist
- [x] PR description included
- [x] Relevant documentation is changed or added (changeset `event-blocks-compute-targeting.md`)
---

### DIFF
--- a/.changeset/event-blocks-compute-targeting.md
+++ b/.changeset/event-blocks-compute-targeting.md
@@ -1,0 +1,35 @@
+---
+"@aws-blocks/bb-async-job": patch
+"@aws-blocks/bb-cron-job": patch
+"@aws-blocks/bb-realtime": patch
+"@aws-blocks/bb-lambda-compute": patch
+"@aws-blocks/bb-agent": patch
+"@aws-blocks/blocks": patch
+---
+
+refactor: route event-block resources to their resolved compute
+
+AsyncJob, CronJob, and Realtime now attach their compute-bound resources to the
+compute resolved via `this.compute` instead of the stack's shared handler:
+
+- AsyncJob attaches its SQS event source to the compute's function;
+- CronJob points its EventBridge Scheduler target at the compute's function;
+- Realtime wires its shared WebSocket API integrations to the compute's function.
+
+All three require a `LambdaCompute` today and fail at synth with a clear error
+on any other compute type — AsyncJob now guards this the same way CronJob and
+Realtime do, instead of silently skipping its event source (which would have
+left submitted jobs with nothing to consume them).
+
+On the default single-Lambda setup `this.compute` resolves to the stack's default
+compute, whose function is the shared handler — so this is non-breaking with no
+change to synthesized infrastructure. AsyncJob also grants SQS send to the shared
+execution role rather than the handler directly.
+
+`@aws-blocks/bb-lambda-compute` adds a `./cdk` subpath that exposes the CDK-typed
+`LambdaCompute`, letting framework blocks reference its `fn` at synth time.
+
+`@aws-blocks/bb-agent` builds AsyncJob and Realtime internally, so its CDK test
+is moved onto the `BlocksStack.create` harness (which initializes the default
+compute the blocks now resolve) instead of a handler-only stub. Test-only
+change — no runtime behavior change to the Agent.
--- a/.changeset/event-blocks-compute-targeting.md
+++ b/.changeset/event-blocks-compute-targeting.md
@@ -1,35 +1,45 @@
 ---
-"@aws-blocks/bb-async-job": patch
-"@aws-blocks/bb-cron-job": patch
-"@aws-blocks/bb-realtime": patch
-"@aws-blocks/bb-lambda-compute": patch
+"@aws-blocks/core": minor
+"@aws-blocks/bb-async-job": minor
+"@aws-blocks/bb-cron-job": minor
+"@aws-blocks/bb-realtime": minor
+"@aws-blocks/bb-lambda-compute": minor
 "@aws-blocks/bb-agent": patch
-"@aws-blocks/blocks": patch
+"@aws-blocks/blocks": minor
 ---
 
-refactor: route event-block resources to their resolved compute
+feat: route event-block resources to their resolved compute
 
-AsyncJob, CronJob, and Realtime now attach their compute-bound resources to the
-compute resolved via `this.compute` instead of the stack's shared handler:
+AsyncJob, CronJob, and Realtime now attach their compute-bound resources to a
+compute resolved at synth rather than the stack's shared handler:
 
-- AsyncJob attaches its SQS event source to the compute's function;
-- CronJob points its EventBridge Scheduler target at the compute's function;
-- Realtime wires its shared WebSocket API integrations to the compute's function.
+- AsyncJob attaches its SQS event source to the resolved compute's function;
+- CronJob points its EventBridge Scheduler target at the resolved compute's function;
+- Realtime binds its shared WebSocket API integrations to the stack's **default**
+  compute (its routes are a stack-level singleton) and grants `postToConnection`
+  to the shared execution role, so `publish()` works from any compute.
 
-All three require a `LambdaCompute` today and fail at synth with a clear error
-on any other compute type — AsyncJob now guards this the same way CronJob and
-Realtime do, instead of silently skipping its event source (which would have
-left submitted jobs with nothing to consume them).
+Each block requires a Lambda compute today and fails at synth with a typed
+`UnsupportedCompute` error (assertable via `isBlocksError`) on any other type.
+The check uses a duplicate-copy-safe brand (`LambdaCompute.isLambdaCompute`,
+backed by a `Symbol.for` marker) instead of `instanceof`, so it does not misfire
+when two copies of `bb-lambda-compute` resolve in one dependency tree.
 
-On the default single-Lambda setup `this.compute` resolves to the stack's default
-compute, whose function is the shared handler — so this is non-breaking with no
+On the default single-Lambda setup the resolved/default compute is the stack's
+default, whose function is the shared handler — so this is non-breaking with no
 change to synthesized infrastructure. AsyncJob also grants SQS send to the shared
 execution role rather than the handler directly.
 
-`@aws-blocks/bb-lambda-compute` adds a `./cdk` subpath that exposes the CDK-typed
-`LambdaCompute`, letting framework blocks reference its `fn` at synth time.
+New public surface (hence `minor`):
+
+- `@aws-blocks/core` exports `blocksError(name, message)` (the producer half of
+  the `isBlocksError` contract) and `sanitizeConfigKey(id)` from `./bb-utils`
+  (the single env-var-key sanitizer both config writers and runtime readers use).
+- `@aws-blocks/bb-lambda-compute` adds a `./cdk` subpath exposing the CDK-typed
+  `LambdaCompute` and its `LambdaCompute.isLambdaCompute` guard.
+- `bb-async-job` / `bb-cron-job` / `bb-realtime` add an `UnsupportedCompute`
+  error member.
 
 `@aws-blocks/bb-agent` builds AsyncJob and Realtime internally, so its CDK test
-is moved onto the `BlocksStack.create` harness (which initializes the default
-compute the blocks now resolve) instead of a handler-only stub. Test-only
-change — no runtime behavior change to the Agent.
+moves onto the `BlocksStack.create` harness instead of a handler-only stub.
+Test-only change — no runtime behavior change to the Agent.
--- a/packages/bb-async-job/src/index.cdk.test.ts
+++ b/packages/bb-async-job/src/index.cdk.test.ts
@@ -2,50 +2,106 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /**
- * CDK-side regression tests for AsyncJob's SQS event source mapping.
+ * CDK-side tests for AsyncJob's SQS event source.
  *
- * History: the event source shipped with BatchSize 1 and no partial-batch
- * reporting, so every message cost a full Lambda invocation. Raising the batch
- * size alone is unsafe — without FunctionResponseTypes containing
- * ReportBatchItemFailures, SQS deletes an entire batch when any single record
- * fails. These tests pin both halves of the fix together.
+ * Compute targeting: AsyncJob attaches its SQS event source to `this.compute`'s
+ * function rather than the stack's shared handler. By default `this.compute`
+ * resolves to the stack's default LambdaCompute, so the event source lands on
+ * the same function as before. When a nearer scope assigns a different compute
+ * (internal `_compute`, the seam a future customer-facing option will drive),
+ * the event source follows it to that compute's function.
+ *
+ * Mapping regressions: the event source shipped with BatchSize 1 and no
+ * partial-batch reporting, so every message cost a full Lambda invocation.
+ * Raising the batch size alone is unsafe — without FunctionResponseTypes
+ * containing ReportBatchItemFailures, SQS deletes an entire batch when any
+ * single record fails. These tests pin both halves of the fix together.
+ * (Ported onto the BlocksStack harness: under compute targeting, AsyncJob
+ * resolves `this.compute`, so a handler-only stub stack can no longer
+ * synthesize it.)
  */
-import { test, afterEach } from 'node:test';
+
+import { test, describe, before, after } from 'node:test';
 import assert from 'node:assert';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { dirname, join } from 'node:path';
 import * as cdk from 'aws-cdk-lib';
-import type { Construct } from 'constructs';
 import { Template } from 'aws-cdk-lib/assertions';
-import { Scope, DEFAULT_NODE_RUNTIME } from '@aws-blocks/core/cdk';
+import { BlocksStack, BlocksPresets, Scope } from '@aws-blocks/core/cdk';
+import type { DefaultComputeFactory } from '@aws-blocks/core/cdk/internal';
+import { LambdaCompute } from '@aws-blocks/bb-lambda-compute/cdk';
 import { AsyncJob, AsyncJobErrors } from './index.cdk.js';
 
-class StubBlocksStack extends cdk.Stack {
-	public readonly handler: cdk.aws_lambda.Function;
-	public readonly id: string;
-	constructor(scope: Construct, id: string) {
-		super(scope, id);
-		this.id = id;
-		(globalThis as any).CURRENT_BLOCKS_STACK = this;
-		this.handler = new cdk.aws_lambda.Function(this, 'StubHandler', {
-			runtime: DEFAULT_NODE_RUNTIME,
-			handler: 'index.handler',
-			code: cdk.aws_lambda.Code.fromInline('exports.handler = async () => {};'),
-		});
-	}
-}
+const lambdaFactory: DefaultComputeFactory = (root) => new LambdaCompute(root as never, 'DefaultCompute');
 
-// setup() installs its own stack as the ambient CURRENT_BLOCKS_STACK; clear it
-// afterwards so no test observes a stack left behind by the previous one (node
-// --test isolates files by default, so this is hygiene against future sharing).
-afterEach(() => {
-	delete (globalThis as any).CURRENT_BLOCKS_STACK;
+const __dirname = dirname(fileURLToPath(import.meta.url));
+let handlerPath: string;
+let backendPath: string;
+let tmpDir: string;
+
+before(() => {
+	process.env.NODE_OPTIONS = `${process.env.NODE_OPTIONS ?? ''} --conditions=cdk`;
+	tmpDir = mkdtempSync(join(__dirname, 'tmp-async-cdk-'));
+	handlerPath = join(tmpDir, 'handler.mjs');
+	writeFileSync(handlerPath, "export const handler = async () => ({ statusCode: 200, body: '{}' });\n");
+	backendPath = join(tmpDir, 'backend.mjs');
+	writeFileSync(backendPath, 'export default () => {};\n');
 });
 
-function setup(): { stack: StubBlocksStack; parent: Scope } {
+after(() => {
+	rmSync(tmpDir, { recursive: true, force: true });
+});
+
+async function makeStack(id: string): Promise<BlocksStack> {
 	const app = new cdk.App();
-	const stack = new StubBlocksStack(app, 'teststack');
-	const parent = new Scope('app');
-	return { stack, parent };
+	return BlocksStack.create(app, id, {
+		backendHandlerPath: handlerPath,
+		backendCDKPath: backendPath,
+		defaults: BlocksPresets.production,
+		defaultComputeFactory: lambdaFactory,
+	});
 }
+
+function fnLogicalId(stack: BlocksStack, compute: LambdaCompute): string {
+	return stack.getLogicalId(compute.fn.node.defaultChild as cdk.CfnElement);
+}
+
+function eventSourceTargets(template: Template, logicalId: string): boolean {
+	return JSON.stringify(template.findResources('AWS::Lambda::EventSourceMapping')).includes(logicalId);
+}
+
+describe('AsyncJob compute targeting', () => {
+	test('default: SQS event source attaches to the stack default compute function', async () => {
+		const stack = await makeStack('AsyncDefault');
+		new AsyncJob(stack, 'jobs', { handler: async () => {}, trackStatus: false });
+
+		const template = Template.fromStack(stack);
+		template.resourceCountIs('AWS::Lambda::EventSourceMapping', 1);
+		assert.ok(
+			eventSourceTargets(template, fnLogicalId(stack, stack._defaultCompute as LambdaCompute)),
+			'event source targets the default compute function',
+		);
+	});
+
+	test('targeted: event source follows an ancestor scope _compute to that function', async () => {
+		const stack = await makeStack('AsyncTargeted');
+		const lambdaB = new LambdaCompute(stack, 'LambdaB');
+		const scoped = new Scope('scoped', { parent: stack });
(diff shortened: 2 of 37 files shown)


packages/bb-async-job/src/index.cdk.ts
```diff
@@ -112,18 +124,22 @@ export class AsyncJob<T = unknown> extends Scope {
...
 		});
 
-		this.queue.grantSendMessages(this.handler);
+		this.queue.grantSendMessages(this.executionRole);
 		registerConfig(
 			this,
 			`BLOCKS_QUEUE_URL_${this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')}`,
 			this.queue.queueUrl
 		);
+		registerConfig(this, `BLOCKS_HANDLER_OWNER_${this.fullId}`, compute.fullId);
```
soberm: **[HIGH] Unsanitized `BLOCKS_HANDLER_OWNER_` config key.** The adjacent QUEUE_URL key sanitizes the id (`this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')`) precisely because the runtime reconstructs the same string byte-for-byte. This key uses the raw `this.fullId`, which carries `-` and mixed case (e.g. `BLOCKS_HANDLER_OWNER_MyStack-jobs`). Two problems: (a) it's **dead config** — there is no runtime reader of this key anywhere in the repo today; and (b) it's a **trap** — whoever adds the reader will follow the QUEUE_URL precedent and reconstruct a *sanitized* key that never matches this raw one, causing a silent lookup miss.

Suggest either dropping the line until a consumer exists, or sanitizing identically:
```ts
registerConfig(this, `BLOCKS_HANDLER_OWNER_${this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')}`, compute.fullId);
```

soberm (commented):
Code review — the compute-targeting refactor is sound (IAM swap is correct, `./cdk` export is well-formed, no circular dependency, dependency placement is right). Most findings are posted inline. Two items that don't anchor to a changed line:

**[LOW] Extraneous nested `@aws-blocks/core@0.2.0` in `package-lock.json`** (marked `"extraneous": true`, under `bb-agent/node_modules/.../bb-lambda-compute/node_modules/@aws-blocks/core`). Hygiene noise, but relevant to `instanceof LambdaCompute`: if a block ever resolves a different copy of `bb-lambda-compute` than the default-compute factory used, the guard misfires. `^0.3.0` is consistent across manifests so dedup should hold — please regenerate the lockfile cleanly to drop the extraneous entry.

**[LOW] Test-only type assertions** (e.g. `_defaultCompute as LambdaCompute`, `defaultChild as cdk.CfnElement`) brush against the AGENTS.md no-casts rule. They're narrowing casts (not `as any`) and test-only, so low severity — a small typed helper would keep them out of the assertion sites.

Minor nit on the PR description: it says AsyncJob "now guards this the same way CronJob and Realtime do" — the diff shows the guard added to all three simultaneously; CronJob/Realtime had no prior guard.

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -82,6 +83,17 @@ export class AsyncJob<T = unknown> extends Scope {
...
 		validateEventSourceOptions(this.fullId, batchSize, maxBatchingWindowSeconds);
 
+		// The queue is consumed by an SQS event source on the compute's own
+		// Lambda, so a AsyncJob currently requires a Lambda compute. Other
+		// compute types need a different consumption path (e.g. runtime polling)
+		// that does not exist yet — fail loud at synth rather than provision a
+		// queue nothing consumes (submitted jobs would silently pile up). Matches
+		// the CronJob / Realtime guards.
+		const compute = this.compute;
+		if (!(compute instanceof LambdaCompute)) {
```
soberm: **[MEDIUM] Guard throws a bare `Error` instead of the typed error family.** AsyncJob exports `AsyncJobErrors`, but this guard throws a plain `Error` that isn't assertable via `isBlocksError`. Consider adding an `UnsupportedCompute`-style member and throwing through it. (The message text itself is good and consistent across all three blocks.)

**[MEDIUM] This throw path has no test coverage.** The PR's headline behavior — "fail at synth with a clear error on any other compute type" — is never exercised: no test constructs a non-`LambdaCompute` and asserts the throw. This is the path most likely to regress; it pairs naturally with the typed-error change (`assert.throws(..., e => isBlocksError(e, AsyncJobErrors.UnsupportedCompute))`). Same applies to the cron-job and realtime guards.

packages/bb-cron-job/src/index.cdk.ts
```diff
@@ -27,7 +28,14 @@ export class CronJob<T = void> extends Scope {
 		validateSchedule(options.schedule);
 		if (options.timezone !== undefined) validateTimezone(options.timezone);
 
-		const lambdaArn = this.handler.functionArn;
+		// The scheduler invokes the compute's function directly, so a CronJob
+		// currently requires a Lambda compute. Other compute types need a
+		// different scheduler target (e.g. EventBridge → ECS) — not yet supported.
+		const compute = this.compute;
+		if (!(compute instanceof LambdaCompute)) {
```
soberm: **[MEDIUM] Bare `Error` instead of the typed error family.** CronJob already models synth-time failures as typed errors (`CronJobErrors.InvalidSchedule` / `InvalidTimezone`, asserted via `isBlocksError`). This guard throws a plain `Error`, inconsistent with that taxonomy. Suggest adding an `UnsupportedCompute` member to `CronJobErrors` and throwing through it, plus a test asserting the throw.

packages/bb-realtime/src/index.cdk.ts
```diff
@@ -163,7 +164,14 @@ function getOrCreateSharedInfra(stack: cdk.Stack, handler: cdk.aws_lambda.IFunct
 export class Realtime extends Scope {
 	constructor(scope: ScopeParent, id: string, options: RealtimeOptions<NamespaceDefs>) {
 		super(id, { parent: scope });
-		getOrCreateSharedInfra(cdk.Stack.of(this), this.handler, this);
+		// The WebSocket routes integrate directly with the compute's function, so
+		// Realtime currently requires a Lambda compute; other compute types need a
+		// different WebSocket integration — not yet supported.
+		const compute = this.compute;
+		if (!(compute instanceof LambdaCompute)) {
```
soberm: **[MEDIUM] Bare `Error` instead of a typed error family.** For consistency with CronJob's typed-error taxonomy, throw a typed `UnsupportedCompute`-style error here (assertable via `isBlocksError`) rather than a plain `Error`, and add a test covering the throw path.

packages/bb-async-job/src/index.cdk.ts
Simone319: You're right to flag it. These keys aren't inert JSON — the runtime loads blocks-config.json into process.env at cold start (loadConfigToProcessEnv), so this key becomes an actual Lambda env var. That's exactly why the adjacent QUEUE_URL key sanitizes to [A-Z0-9_]: a raw BLOCKS_HANDLER_OWNER_MyStack-jobs is a malformed env var name (Node tolerates the in-process assignment and there's no reader yet, so nothing breaks today — but it's latent). Fixed in f5ba09dd: kept the write (it's the intentional Phase-2 owner-match seam) and sanitized it identically to QUEUE_URL, with a comment that the Phase-2 reader must reconstruct the same string.

packages/bb-async-job/src/index.cdk.ts
Simone319: Done in f5ba09dd. Added AsyncJobErrors.UnsupportedCompute and throw it via the existing blocksError() helper (assertable with isBlocksError, matching how InvalidOption models synth-time failures), plus a test that assigns a non-LambdaCompute to a scope and asserts isBlocksError(e, AsyncJobErrors.UnsupportedCompute). Same treatment applied to the cron-job and realtime guards.

packages/bb-cron-job/src/index.cdk.ts
Simone319: Done in f5ba09dd — added CronJobErrors.UnsupportedCompute and throw a typed error consistent with the InvalidSchedule/InvalidTimezone pattern (new Error + err.name), plus a test asserting the throw on a non-Lambda compute. (Also added it to the index.browser.ts CronJobErrors stub for parity.)

packages/bb-realtime/src/index.cdk.ts
Simone319: Done in f5ba09dd — added RealtimeErrors.UnsupportedCompute with a throw-path test. I used the inline new Error + err.name pattern here rather than the utils.js blocksError helper: utils.js carries runtime-only deps (token minting/crypto) that I didn't want to pull into the CDK synth bundle. Still assertable via isBlocksError.

Simone319:
Thanks for the thorough pass, @soberm. Addressed the inline findings in f5ba09dd; on the three summary items:

- **Extraneous nested `@aws-blocks/core@0.2.0`:** removed in f5ba09dd. It was a stale lockfile-only entry (no backing dir on disk, not required by the tree) that `npm install` / `--package-lock-only` wouldn't prune, so I dropped the block and re-validated with `npm install --package-lock-only` (npm did not re-add it). `^0.3.0` is consistent across manifests, so the dedup holds and the `instanceof LambdaCompute` guard resolves a single copy.
- **Test-only casts:** leaving these. They are test-internal assertions (`_defaultCompute` is an internal field; `CfnElement` is CDK-tree introspection), not customer-facing API usage, so they fall under the test-plumbing exception to the no-casts rule rather than the customer-DX surface it protects. Happy to add a small typed helper if you feel strongly.
- **Description wording:** the "matches the CronJob / Realtime guards" phrasing was in the AsyncJob code comment (the description didn't actually claim a prior guard); corrected it in 72cb60cd, and made the PR description explicit that all three guards are added together in this change.

soberm (dismissed):

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -82,6 +83,20 @@ export class AsyncJob<T = unknown> extends Scope {
...
 		validateEventSourceOptions(this.fullId, batchSize, maxBatchingWindowSeconds);
 
+		// The queue is consumed by an SQS event source on the compute's own
+		// Lambda, so a AsyncJob currently requires a Lambda compute. Other
+		// compute types need a different consumption path (e.g. runtime polling)
+		// that does not exist yet — fail loud at synth rather than provision a
+		// queue nothing consumes (submitted jobs would silently pile up). The
+		// CronJob and Realtime blocks add the same guard in this change.
+		const compute = this.compute;
+		if (!(compute instanceof LambdaCompute)) {
```
sarayev: This relies on `instanceof LambdaCompute`, which breaks if two copies of `bb-lambda-compute` end up in the tree. The dep is pinned `^0.3.0`, so once this package gets a minor bump to 0.4.0, an app on `bb-lambda-compute@0.4.0` plus a block still on `^0.3.0` resolves two copies, and the app's real default compute fails `instanceof` against our class. We'd then throw `UnsupportedCompute` on a valid Lambda app. Since these packages version independently via changesets, could we use a brand/capability check instead (e.g. a static `LambdaCompute.isLambdaCompute(x)`, or a duck-check on `compute.fn`), the way `fromExisting()` uses `__brand`? That survives duplicate copies.

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -112,18 +127,30 @@ export class AsyncJob<T = unknown> extends Scope {
 			enforceSSL: true,
 		});
 
-		this.queue.grantSendMessages(this.handler);
+		this.queue.grantSendMessages(this.executionRole);
```
sarayev, line 133: Consumer perms follow `compute.fn` via `addEventSource`, so this send grant is for the producer side (`submit()`), which is correct today since `submit()` runs on the shared execution role. Just noting for later: once compute assignment is customer-facing, `submit()` could run on a different compute than `this.executionRole`, so this grant is the half that isn't compute-aware yet. Probably a Phase-2 item, not this PR.

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -112,18 +127,30 @@ export class AsyncJob<T = unknown> extends Scope {
...
 		registerConfig(
 			this,
 			`BLOCKS_QUEUE_URL_${this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')}`,
```
sarayev: `this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')` is now written out on both this key and the HANDLER_OWNER key below, and the comment notes a Phase-2 reader has to reconstruct the exact same string. That's the kind of rule that drifts once it lives in three places. Could we extract a shared `sanitizeEnvKey(fullId)` helper so the producer and the future reader can't disagree?

packages/bb-async-job/src/index.cdk.ts
```diff
@@ -112,18 +127,30 @@ export class AsyncJob<T = unknown> extends Scope {
...
 			this.queue.queueUrl
 		);
+		// Sanitize the id identically to the QUEUE_URL key above: config entries
+		// are loaded into `process.env` at runtime (loadConfigToProcessEnv), so
+		// the key must be a valid env var name, and the Phase-2 owner-match reader
+		// must reconstruct the same sanitized string byte-for-byte.
+		registerConfig(
+			this,
+			`BLOCKS_HANDLER_OWNER_${this.fullId.toUpperCase().replace(/[^A-Z0-9]/g, '_')}`,
```
sarayev: Nothing reads `BLOCKS_HANDLER_OWNER_*` in this PR (it's for Phase-2 owner matching), but it ships now as a second env var per job. Lambda env is capped around 4 KB, so apps with several jobs burn that budget on a value nothing consumes yet. Would it be cleaner to add this in the Phase-2 change that actually reads it, so the write and read land together and stay in sync?

packages/bb-cron-job/src/index.cdk.ts
```diff
@@ -27,7 +29,20 @@ export class CronJob<T = void> extends Scope {
...
```
sarayev: Same `instanceof LambdaCompute` concern as in bb-async-job (duplicate-copy fragility): see the note on async-job's guard. A brand/capability check would be more robust here too.

packages/bb-cron-job/src/index.cdk.ts
```diff
@@ -27,7 +29,20 @@ export class CronJob<T = void> extends Scope {
...
+			// Match the inline typed-error pattern used by schedule.ts (new Error +
+			// err.name) so this is assertable via isBlocksError.
+			const err = new Error(
```
sarayev: async-job builds this same error via its local `blocksError()` helper, while here (and in Realtime) it's hand-rolled with `new Error()` + `err.name`. The comment says `blocksError` carries runtime deps, but async-job's `blocksError` is a local function in its own `errors.ts` that just sets `name`, so it's CDK-safe and does exactly this. For three guards meant to look identical, could they share one small helper? The message format also differs here: the exception name is prefixed into the message, but async-job's isn't.

packages/bb-realtime/src/index.cdk.ts
```diff
@@ -163,7 +165,21 @@ function getOrCreateSharedInfra(stack: cdk.Stack, handler: cdk.aws_lambda.IFunct
...
```
sarayev: Same duplicate-copy `instanceof` concern as async-job/cron: a brand check would be safer than `instanceof LambdaCompute` here.

packages/bb-realtime/src/index.cdk.ts
```diff
@@ -163,7 +165,21 @@ function getOrCreateSharedInfra(stack: cdk.Stack, handler: cdk.aws_lambda.IFunct
...
+			// Inline typed error (not the utils.js blocksError, which carries
+			// runtime-only deps) so this stays assertable via isBlocksError without
+			// pulling those into the CDK synth bundle.
+			const err = new Error(
+				`${RealtimeErrors.UnsupportedCompute}: Realtime "${this.fullId}" currently supports only a Lambda compute.`,
+			);
+			err.name = RealtimeErrors.UnsupportedCompute;
+			throw err;
+		}
+		getOrCreateSharedInfra(cdk.Stack.of(this), compute.fn, this);
```
sarayev, line 188: `getOrCreateSharedInfra` is a per-stack singleton. If two `Realtime` blocks in one stack ever resolve to different computes, doesn't the second reuse the infra already bound to the first compute's function, so its routes land on the wrong compute? Fine today since everything resolves to the default, but since this PR is about per-block targeting, maybe worth a guard that the shared infra's function matches the resolved compute, or at least a follow-up note.

packages/bb-async-job/src/index.cdk.test.ts
```diff
@@ -2,106 +2,202 @@
...
-// --test isolates files by default, so this is hygiene against future sharing).
-afterEach(() => {
-	delete (globalThis as any).CURRENT_BLOCKS_STACK;
+const __dirname = dirname(fileURLToPath(import.meta.url));
+let handlerPath: string;
+let backendPath: string;
+let tmpDir: string;
+
+before(() => {
+	process.env.NODE_OPTIONS = `${process.env.NODE_OPTIONS ?? ''} --conditions=cdk`;
```
sarayev, line 54: Export conditions are fixed at process launch, and the `.cdk.js` modules are already imported before `before()` runs, so setting `NODE_OPTIONS` here doesn't change resolution for this process. Is it doing anything, or is it leftover? If the file needs `--conditions=cdk`, that has to come from how the test is invoked rather than from here.

packages/bb-async-job/src/index.cdk.test.ts
```diff
@@ -2,106 +2,202 @@
...
-	return mappings[keys[0]].Properties;
+function fnLogicalId(stack: BlocksStack, compute: LambdaCompute): string {
+	return stack.getLogicalId(compute.fn.node.defaultChild as cdk.CfnElement);
 }
 
-test('CDK: AsyncJob batches 10 messages per invocation by default', () => {
-	const { stack, parent } = setup();
-	new AsyncJob(parent, 'jobs', { handler: async () => {} });
+function eventSourceTargets(template: Template, logicalId: string): boolean {
+	return JSON.stringify(template.findResources('AWS::Lambda::EventSourceMapping')).includes(logicalId);
```
sarayev, line 81: `JSON.stringify(...).includes(logicalId)` works because logical IDs are unique-ish, but it will also match the id if it appears in any unrelated field. `hasResourceProperties(..., Match.objectLike({ FunctionName: { Ref: logicalId } }))` asserts the actual target field. Nit only: the paired positive/negative asserts already make this safe enough.

.changeset/event-blocks-compute-targeting.md
```diff
@@ -0,0 +1,35 @@
+---
+"@aws-blocks/bb-async-job": patch
+"@aws-blocks/bb-cron-job": patch
+"@aws-blocks/bb-realtime": patch
+"@aws-blocks/bb-lambda-compute": patch
```
sarayev: This PR adds a new public `./cdk` export subpath to bb-lambda-compute (and new public `UnsupportedCompute` error members), which is usually a `minor` rather than a `patch`. Worth confirming before release so the changelog reflects the added API surface.

packages/bb-async-job/src/index.cdk.ts
Simone319: Agreed, and worth fixing now. Done in 61d5d3ce: replaced the three `instanceof LambdaCompute` guards with `LambdaCompute.isLambdaCompute()`, backed by a `Symbol.for('blocks:LambdaCompute')` brand — the same process-global-marker approach core already uses for `API_NAMESPACE_MARKER`, and the `__brand` idea from `fromExisting()`. It survives two copies of bb-lambda-compute in one tree, where `instanceof` misfires once the ranges diverge.

packages/bb-cron-job/src/index.cdk.ts
Simone319: Fixed with the async-job guard — switched to the `Symbol.for`-backed `LambdaCompute.isLambdaCompute()` brand check that survives duplicate copies (61d5d3ce).

packages/bb-realtime/src/index.cdk.ts
Simone319: Same fix (61d5d3ce) — now `LambdaCompute.isLambdaCompute()` (brand check) instead of `instanceof`.

packages/bb-realtime/src/index.cdk.ts
Simone319: Good catch — it steered a real design decision. The WS routes are a stack-level singleton, publishing is compute-agnostic (`postToConnection` is a public IAM call any compute can make), and connection bookkeeping only touches the shared table — so "routes on the wrong compute" is latent today but conceptually wrong under per-block targeting. Rather than a per-compute-match guard (which would forbid the multi-compute case we're preparing for), I made the shared infra explicitly compute-agnostic in 61d5d3ce: bind the WS routes to the stack DEFAULT compute deterministically (new `@internal Scope.defaultCompute`, not "whichever Realtime was first"), and grant `manageConnections` to the shared execution role so publish works from any compute. That removes the divergence entirely — there's no per-compute ownership left to mismatch.

packages/bb-cron-job/src/index.cdk.ts
Simone319: You're right — async-job's `blocksError` is a local, import-free helper that just sets `name`, so it's CDK-safe (my earlier "runtime deps" note was about realtime's `utils.js` one specifically). Consolidated in 61d5d3ce: added a single `blocksError(name, message)` to `@aws-blocks/core` and routed all three guards through it, so they're byte-identical including the name-prefixed message. bb-async-job/errors.ts and bb-realtime/utils.ts now re-export core's instead of defining their own.

packages/bb-async-job/src/index.cdk.ts
Simone319: Done in 61d5d3ce — extracted `sanitizeConfigKey(fullId)` into `@aws-blocks/core/bb-utils`; both the QUEUE_URL and HANDLER_OWNER writers and the runtime reader (index.aws.ts) now go through it, so they can't drift.

.changeset/event-blocks-compute-targeting.md
Simone319: Agreed — the new `./cdk` subpath on bb-lambda-compute and the `UnsupportedCompute` members are additive public surface. Bumped those to `minor` in 61d5d3ce, and also `@aws-blocks/core` (new `blocksError` + `sanitizeConfigKey`) and the `@aws-blocks/blocks` umbrella; bb-agent stays `patch` (test-only). Regenerated the core API report.

packages/bb-async-job/src/index.cdk.ts
Simone319: Right on both counts — the send grant is the producer/`submit()` side, and `this.executionRole` is correct today since `submit()` runs on the shared role. Making it compute-aware when submit runs on a non-shared-role compute is a Phase-2 concern that rides on the general per-compute-role machinery, not this PR. Leaving as-is with that noted.

packages/bb-async-job/src/index.cdk.ts
Simone319: One correction on the cost: `registerConfig` writes to the S3 `blocks-config.json`, not the Lambda env block — `loadConfigToProcessEnv` pulls it into `process.env` at runtime, which is how the config registry sidesteps the 4 KB env limit. So HANDLER_OWNER is ~40 bytes in an S3 JSON, not env-var budget. It's an intentional Phase-1 owner-match seam, and with the shared `sanitizeConfigKey` helper (now added) the write/read can't drift. Keeping it — but happy to defer to the Phase-2 change that reads it if you'd rather they land together; no strong preference.

packages/bb-async-job/src/index.cdk.test.ts
Simone319: You're right it can't affect module resolution — that's handled by the explicit `/cdk` import paths. It's not leftover, though: `BlocksStack.create()` calls `assertCdkConditionActive()`, which reads `process.env.NODE_OPTIONS`, and `before()` runs before `create()`, so this is what satisfies that assertion in-process. I'll add a one-line comment saying so, since the intent clearly wasn't obvious.

packages/bb-async-job/src/index.cdk.test.ts
Simone319: Fair — the paired positive/negative asserts already disambiguate, so I've left it as-is for now. Happy to tighten to `hasResourceProperties(Match.objectLike({ FunctionName: { ... } }))` in a follow-up if you'd prefer a direct field assertion.

sarayev (approved):

soberm (approved):

PR #460, opened 2026-08-31
[core] Record each API namespace on its resolved compute

Simone319:
## Problem
Request routing (a later track) needs to know which compute hosts each API namespace. This records that mapping at synth time — without changing the public `ApiNamespace` signature.

> **Stacked on #391** (config registry / compute registry). Base is the #391 branch; it retargets to `main` automatically once #391 merges. Review only this PR's layer (3 files).

## Changes
- `ApiNamespace` constructor records its namespace on the resolved compute (`recordNamespaceOnCompute` → `compute.namespaces.push(name)`) during CDK synth. Public signature `(scope, name, handler)` is byte-identical; recording is an internal side effect.
- No-op outside CDK synth (mock/runtime bundles have no compute) and when a scope resolves no compute — an API created without a Scope stays unrecorded and routes to the default compute.

## Validation
- Unit tests for recording on a resolvable compute and the scopeless no-op.
- Green locally: `core` 702; full build + lint pass.

## Checklist
- [x] PR description included
- [x] Relevant documentation is changed or added (changeset `core-namespace-compute-recording.md`)
---

### DIFF
--- a/packages/core/src/api.ts
+++ b/packages/core/src/api.ts
@@ -72,23 +72,26 @@ type ComputeResolvingScope = { compute?: { namespaces?: string[] } };
  *
  * CDK-synth concern only: in the CDK bundle `scope.compute` resolves to the
  * namespace's compute (the stack default today), and we append the name to its
- * `namespaces` list. In the mock/runtime bundles — or a scopeless test — there
- * is no `compute`, so this is a silent no-op. The `compute` getter can also
- * throw before the default compute is initialized (outside synth); that too is
- * treated as "nothing to record". The public `ApiNamespace` signature is
- * unchanged; this is a purely internal side effect.
+ * `namespaces` list. In the mock/runtime bundles — or a scopeless test — the
+ * property is absent, so `scope.compute` is `undefined` and this is a silent
+ * no-op. The public `ApiNamespace` signature is unchanged; this is a purely
+ * internal side effect.
+ *
+ * No `try/catch`: `BlocksBackend.create()` initializes the default compute
+ * before it imports the backend module that constructs any `ApiNamespace`, so
+ * the `compute` getter always resolves at this call site. If it ever threw
+ * here it would signal a genuine lifecycle violation (an `ApiNamespace` built
+ * before `create()` resolved), which should surface rather than be swallowed.
  */
 function recordNamespaceOnCompute(scope: ScopeParent | null | undefined, name: string): void {
   // An API created without a Scope stays unrecorded and routes to the
   // default compute.
   if (!scope || typeof scope !== 'object') return;
-  try {
-    const compute = (scope as ComputeResolvingScope).compute;
-    if (compute && Array.isArray(compute.namespaces)) {
-      compute.namespaces.push(name);
-    }
-  } catch {
-    // compute resolution isn't available in this context — nothing to record.
+  const compute = (scope as ComputeResolvingScope).compute;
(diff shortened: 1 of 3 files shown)


packages/core/src/api.ts
```diff
@@ -59,6 +59,39 @@ export const API_NAMESPACE_MARKER = Symbol.for('blocks:ApiNamespace');
...
+ * unchanged; this is a purely internal side effect.
+ */
+function recordNamespaceOnCompute(scope: ScopeParent | null | undefined, name: string): void {
+  // An API created without a Scope stays unrecorded and routes to the
+  // default compute.
+  if (!scope || typeof scope !== 'object') return;
+  try {
+    const compute = (scope as ComputeResolvingScope).compute;
+    if (compute && Array.isArray(compute.namespaces)) {
+      compute.namespaces.push(name);
```
osama-rizk: `compute.namespaces` is a plain `string[]` and this pushes unconditionally, so the same namespace can be recorded more than once — a re-constructed `ApiNamespace`, an HMR/re-import during synth, or two APIs sharing a name on one compute would each append a duplicate. Whether that matters depends entirely on the request-routing consumer this feeds (the stated purpose). Confirm routing tolerates duplicates, or guard here: `if (!compute.namespaces.includes(name)) compute.namespaces.push(name);`.

packages/core/src/api.ts
```diff
@@ -59,6 +59,39 @@ export const API_NAMESPACE_MARKER = Symbol.for('blocks:ApiNamespace');
...
+    }
+  } catch {
```
osama-rizk: Minor: the comment scopes this catch to the pre-init `compute` getter throw, but a bare `catch {}` absorbs *any* error from `.compute` / `.namespaces` — a genuine compute-resolution bug would silently vanish at synth with no signal. Acceptable for a best-effort side effect, but a narrower guard (e.g. check the default-compute-initialized state, or only swallow the known pre-init error) would avoid masking real failures.

packages/core/src/api.ts
Simone319: Good catch — guarded in ae0f860d. `namespaces` is meant to be the set of names hosted on the compute, so a duplicate is never a state we want, and since the routing consumer does not exist yet I would rather record correctly here than depend on it tolerating dupes later. Switched to `if (!compute.namespaces.includes(name)) compute.namespaces.push(name)` (n is tiny, so the linear check is negligible), and added a test locking the at-most-once behavior. Two separate `ApiNamespace`s declared with the *same* name is a distinct user error (they already collide on the handler marker), so I left that to name-collision validation rather than paper over it here.

packages/core/src/api.ts
Simone319: You are right that the `catch` was masking rather than handling — dropped it entirely in ae0f860d. It only guarded the `compute` getter throw ("default compute not initialized"), but that is unreachable here: `BlocksBackend.create()` sets `_defaultCompute` before it imports the module that constructs any `ApiNamespace`, so at this call site the getter always resolves. The mock/runtime no-op now comes from the getter simply being absent (`undefined`), and the scopeless case from the leading guard — neither needed a catch. If `.compute` ever throws here it is a genuine lifecycle violation (an `ApiNamespace` built before `create()` resolved) and should surface rather than vanish.

osama-rizk (approved):

soberm (approved):

PR #465, opened 2026-09-01
fix(ci): stale-stack cleanup empties versioned buckets + fire-and-forget deletes

osama-rizk:
## Problem

The scheduled **Cleanup Stale E2E Stacks** workflow has been `cancelled` on every run for days — it hits `timeout-minutes: 60` and is killed before draining the backlog. As a result E2E stacks and their IAM roles accumulate until the account reaches the **`RolesPerAccount: 1500`** quota, after which hosting/sandbox/amplify deploys fail at role creation across PRs (e.g. `Cannot exceed quota for RolesPerAccount: 1500`). Because the failure surfaced only as a generic `cancelled`, it went unnoticed for days.

## Root causes fixed

**1. Versioned S3 buckets were never emptied.** Hosting/test buckets are versioned. When a stack's CREATE fails at the role quota, its `autoDeleteObjects` custom-resource Lambda never provisions, so CloudFormation can't empty the bucket on teardown → the stack sticks in `DELETE_FAILED` → its ~15 IAM roles leak → the quota worsens (self-amplifying). This adds `empty_stack_buckets()`, which deletes all object **versions + delete markers** (paged in 1000s) before `delete-stack`. It breaks on any reported `.Errors` (legal-hold / retention / denied keys) so an undeletable object can't spin the re-list loop until the timeout, and the `list | jq` substitution is made total (`|| echo ""`) so it can't abort the step under `set -eo pipefail`.

**2. Serial `aws cloudformation wait stack-delete-complete` per stack.** CloudFront-backed hosting stacks take 15–20 min each to delete; waiting on them one-by-one blew past the 60-min job timeout. Switched to **fire-and-forget** — trigger the deletes and let CloudFormation run them asynchronously; anything that doesn't finish is retried on the next scheduled run. Applied to both the e2e-stack step and the Amplify-sandbox step.

## Other changes

- **Dropped `exit 1`** on best-effort failures (cleanup shouldn't hard-fail the job).
- **Restored failure visibility:** after triggering, each step tallies stacks still in `DELETE_FAILED` for its prefixes and emits a `::warning::` + a step-summary line — so a persistent backlog surfaces instead of growing silently (async deletes mean we can't wait for the count, so this is eventually-consistent across runs).
- **Frequency 4h → 2h + a non-overlapping `concurrency` guard.** Now that the job is fast (fire-and-forget), running twice as often roughly halves the worst-case time a dead stack lingers (~6h → ~4h). The 2h age gate is the binding constraint, so 1h would be diminishing returns for double the runs. `concurrency: { group: cleanup-stale-stacks, cancel-in-progress: false }` prevents a slow drain from overlapping the next tick (and never cancels an in-flight run — cancellation is what produced the old `cancelled` noise).

## Dependency

The bucket-emptying needs `s3:DeleteObjectVersion` (+ List/Get) on the E2E bucket prefixes, which the `blocks-github-actions` OIDC role previously lacked. That grant was added in **`AwsBlocksIntegTestAndGithubInfraCDK`** (CR-301551315) and has been **deployed** to the CI account, so the role already has it.

## Not covered here (separate follow-ups)

- VPC-ENI-stuck `bb-test-prod-*` stacks and the Amplify managed-DynamoDB-table blocker (upstream: table-manager Lambda role lacks `dynamodb:DescribeTable`).
- Making sandbox hosting buckets non-versioned at creation (`packages/hosting`) so this class of failure can't recur.
- Alarming on `RolesPerAccount` utilization and on this workflow being cancelled.
- Per-job teardown as the primary reclaimer (makes this scheduled job a backstop).

## Testing

YAML parses; all run scripts pass `bash -euo pipefail -n`. Behavior verified by hand in the CI account: emptying the versioned buckets lets the previously-`DELETE_FAILED` hosting stacks delete, freeing their roles.

### DIFF
--- a/.github/workflows/cleanup-stacks.yml
+++ b/.github/workflows/cleanup-stacks.yml
@@ -2,12 +2,20 @@ name: Cleanup Stale E2E Stacks
 
 on:
   schedule:
-    - cron: '0 */4 * * *' # Every 4 hours
+    - cron: '0 */2 * * *' # Every 2 hours
   workflow_dispatch:
 
 permissions:
   id-token: write
   contents: read
+
+# Never let two cleanup runs overlap (a slow drain + the next scheduled tick
+# would double-trigger deletes and race on the same buckets). Let the in-flight
+# run finish rather than cancelling it — cancellation is what made the old runs
+# show up as "cancelled" and never complete.
+concurrency:
+  group: cleanup-stale-stacks
+  cancel-in-progress: false
 
 jobs:
   cleanup:
--- a/.github/workflows/cleanup-stacks.yml
+++ b/.github/workflows/cleanup-stacks.yml
@@ -25,10 +25,34 @@ jobs:
       - name: Find and delete stale e2e stacks
         run: |
           CUTOFF=$(date -d '2 hours ago' +%s)
-          DELETED=0
-          FAILED=0
-          DELETED_STACKS=""
-          DELETE_FAILURES=""
+          TRIGGERED=0
+          SKIPPED=0
+
+          # Empty every S3 bucket in a stack, including all object versions and
+          # delete markers. Hosting/test buckets are versioned, so CloudFormation
+          # can't remove them on teardown once the stack's autoDeleteObjects Lambda
+          # is gone — which happens when the CREATE failed at the account IAM role
+          # quota (the Lambda never provisioned) or on a partial rollback. The
+          # bucket then blocks delete-stack, the stack sticks in DELETE_FAILED, and
+          # its IAM roles leak — worsening the very quota that caused the failure.
+          # Emptying out-of-band lets the subsequent delete-stack succeed.
+          empty_stack_buckets() {
+            local STACK="$1" B BUCKETS PAYLOAD
+            BUCKETS=$(aws cloudformation describe-stack-resources --stack-name "$STACK" \
+              --query "StackResources[?ResourceType=='AWS::S3::Bucket'].PhysicalResourceId" \
+              --output text 2>/dev/null || echo "")
+            for B in $BUCKETS; do
+              [ -z "$B" ] && continue
+              # Delete in pages of 1000 (the delete-objects limit) until empty.
+              while :; do
+                PAYLOAD=$(aws s3api list-object-versions --bucket "$B" --max-items 1000 --output json 2>/dev/null \
+                  | jq -c '{Objects: [(.Versions // []) + (.DeleteMarkers // [])] | map({Key, VersionId})}' 2>/dev/null)
+                [ -z "$PAYLOAD" ] && break
+                [ "$(echo "$PAYLOAD" | jq '.Objects | length')" = "0" ] && break
+                aws s3api delete-objects --bucket "$B" --delete "$PAYLOAD" >/dev/null 2>&1 || break
+              done
+            done
+          }
 
           # Get all e2e stacks in deletable states across all name patterns
           STACKS=""
@@ -47,6 +71,7 @@ jobs:
 
             if [[ "$PURPOSE" != e2e-* ]]; then
               echo "⏭️  Skipping $STACK (not tagged blocks:purpose=e2e-*)"
+              SKIPPED=$((SKIPPED + 1))
               continue
             fi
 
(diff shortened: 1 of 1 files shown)


sarayev (dismissed):

.github/workflows/cleanup-stacks.yml
```diff
@@ -25,10 +33,46 @@ jobs:
...
+              --output text 2>/dev/null || echo "")
+            for B in $BUCKETS; do
+              [ -z "$B" ] && continue
+              # Delete in pages of 1000 (the delete-objects limit) until empty.
+              # The `|| echo ""` keeps the substitution total: under `set -eo pipefail`
+              # a jq error on empty/null stdin (empty bucket) would otherwise abort the
+              # whole step instead of falling through to the empty-guard below.
+              while :; do
+                PAYLOAD=$(aws s3api list-object-versions --bucket "$B" --max-items 1000 --output json 2>/dev/null \
+                  | jq -c '{Objects: [(.Versions // []) + (.DeleteMarkers // [])] | map({Key, VersionId})}' 2>/dev/null || echo "")
```
soberm, line 60: [blocker] This jq filter throws on every call, even against an empty bucket. jq closes the {...} object literal at the first top-level |, so this parses as {Objects: [...]} | map({Key, VersionId}) — it builds {Objects: [[...]]} and pipes that object into map(), which then tries to read .Key off an array (Cannot index array with string "Key", exit 5). Confirmed with jq 1.8.1 on both real version/delete-marker data and empty {} input. Since this runs under Actions' default bash -eo pipefail and the assignment has no `|| echo ""` fallback (unlike BUCKETS= a few lines up), errexit kills the whole step the first time it hits a stack with any S3 bucket. Fix: keep the pipe inside the value — {Objects: ((.Versions // []) + (.DeleteMarkers // []) | map({Key, VersionId}))}.

.github/workflows/cleanup-stacks.yml
```diff
@@ -25,10 +33,46 @@ jobs:
...
+              --query "StackResources[?ResourceType=='AWS::S3::Bucket'].PhysicalResourceId" \
+              --output text 2>/dev/null || echo "")
+            for B in $BUCKETS; do
+              [ -z "$B" ] && continue
+              # Delete in pages of 1000 (the delete-objects limit) until empty.
+              # The `|| echo ""` keeps the substitution total: under `set -eo pipefail`
+              # a jq error on empty/null stdin (empty bucket) would otherwise abort the
+              # whole step instead of falling through to the empty-guard below.
+              while :; do
+                PAYLOAD=$(aws s3api list-object-versions --bucket "$B" --max-items 1000 --output json 2>/dev/null \
```
soberm, line 59: [major] Even after the jq fix, this call has no `|| echo ""` fallback the way BUCKETS= does a few lines up. Add one so a transient AWS error here doesn't take the whole step down again.

.github/workflows/cleanup-stacks.yml
```diff
@@ -60,61 +105,46 @@ jobs:
...
             fi
 
+            # Empty versioned buckets first so the delete can complete, then trigger
+            # the delete WITHOUT waiting. CloudFront-backed hosting stacks take
+            # 15-20 min each to delete; waiting serially blew past this job's 60-min
+            # timeout, so it was cancelled every run and never drained the backlog.
+            # Fire-and-forget lets CloudFormation delete asynchronously; a stack that
+            # doesn't finish is retried on the next scheduled run.
+            echo "🧹 Emptying buckets for $STACK"
+            empty_stack_buckets "$STACK"
```
soberm, line 120: [minor] Emptying runs inline per-stack, synchronously, before that stack's delete-stack. If the backlog builds up with several large versioned buckets, could this reintroduce the same timeout this PR is trying to eliminate, just moved from CFN wait to S3 pagination? A time-budget check might be worth it.

.github/workflows/cleanup-stacks.yml
```diff
@@ -60,61 +105,46 @@ jobs:
...
-            if aws cloudformation delete-stack --stack-name "$STACK"; then
-              DELETED=$((DELETED + 1))
-              DELETED_STACKS="$DELETED_STACKS $STACK"
-            else
-              echo "❌ Failed to trigger delete for $STACK"
-              FAILED=$((FAILED + 1))
-              DELETE_FAILURES="$DELETE_FAILURES $STACK"
-            fi
+            aws cloudformation delete-stack --stack-name "$STACK" || echo "⚠️  delete-stack call failed for $STACK (will retry next run)"
+            TRIGGERED=$((TRIGGERED + 1))
```
soberm, line 123: [minor] TRIGGERED gets bumped even when delete-stack fails and falls into the || branch, so "Delete triggered: N" is really an attempt count, not a success count. Consider only counting the success path, or relabeling.

.github/workflows/cleanup-stacks.yml
```diff
@@ -164,46 +192,36 @@ jobs:
...
-            else
-              FINAL=$(aws cloudformation describe-stacks --stack-name "$STACK" \
-                --query "Stacks[0].StackStatus" --output text 2>/dev/null || echo "GONE")
-              if [ "$FINAL" = "GONE" ]; then
-                echo "✅ $STACK deleted"; DELETED=$((DELETED + 1))
-              else
-                echo "❌ $STACK final status: $FINAL"; FAILED=$((FAILED + 1)); DELETE_FAILURES="$DELETE_FAILURES $STACK"
-              fi
-            fi
+            TRIGGERED=$((TRIGGERED + 1))
```
soberm, line 206: [minor] Same as the e2e step — TRIGGERED counts attempts, not confirmed triggers, since it's incremented after the || true fallback too.

.github/workflows/cleanup-stacks.yml
```diff
@@ -47,6 +91,7 @@ jobs:
 
             if [[ "$PURPOSE" != e2e-* ]]; then
               echo "⏭️  Skipping $STACK (not tagged blocks:purpose=e2e-*)"
+              SKIPPED=$((SKIPPED + 1))
               continue
```
soberm, line 95: [nit] These two comment blocks (31-38 and 93-98) explain basically the same root cause. Consider trimming the second to a pointer back to the first.

.github/workflows/cleanup-stacks.yml
```diff
@@ -47,6 +91,7 @@ jobs:
 
             if [[ "$PURPOSE" != e2e-* ]]; then
               echo "⏭️  Skipping $STACK (not tagged blocks:purpose=e2e-*)"
+              SKIPPED=$((SKIPPED + 1))
```
soberm, line 94: [nit] SKIPPED lumps "not tagged e2e" and "too new" into one counter. The per-stack logs already distinguish them — splitting the counter would make the run summary more useful for triage.

.github/workflows/cleanup-stacks.yml
```diff
@@ -60,61 +105,46 @@ jobs:
               CREATED_TS=$(date -d "$CREATION_TIME" +%s 2>/dev/null || echo "0")
               if [ "$CREATED_TS" -gt "$CUTOFF" ]; then
                 echo "⏭️  Skipping $STACK (created $CREATION_TIME, less than 2 hours old)"
+                SKIPPED=$((SKIPPED + 1))
```
soberm, line 108: [question] Now that the exit 1/::error::/::warning:: annotations are gone, is there any remaining signal for stacks stuck across many runs (e.g. the known Amplify DynamoDB table-manager bug)? Maybe a non-blocking ::warning:: once a stack's been retried N times.

soberm (changes requested):
The new empty_stack_buckets() bucket-emptying logic doesn't work: the jq filter at lines 48-49 throws on every invocation (even on empty buckets), and under Actions' default bash -eo pipefail with no || fallback it aborts the whole cleanup step on the first bucket-owning stack. Net effect: buckets are never emptied (so the DELETE_FAILED / IAM-role-leak problem is not fixed) and the workflow trades a silent timeout for a louder, more frequent hard failure. One-line fix below. The rest of the PR (fire-and-forget redesign, annotation cleanup, tag/age gating) is solid.

osama-rizk:
Follow-up **#466** addresses all the review feedback here (the jq no-op you flagged as a blocker, plus a second cause found while validating — `describe-stack-resources`'s 100-resource cap missing the storage bucket — and the fallback/counter/comment nits). Details + item-by-item responses on #466.

.github/workflows/cleanup-stacks.yml
osama-rizk: Confirmed — good catch, this was a hard no-op. Fixed in #466 using your fully-parenthesized form: `{Objects: ((.Versions // []) + (.DeleteMarkers // []) | map({Key, VersionId}))}`. Verified on real version/marker data → `{Objects:[...]}` and on empty `{}` → `{Objects:[]}` (len 0 → clean break, no throw). Validating in-account also turned up a *second* cause: `describe-stack-resources` caps at 100 resources, so on a large OpenNext stack it omitted the storage bucket that blocks the delete — #466 also switches to paginated `list-stack-resources`.

.github/workflows/cleanup-stacks.yml
osama-rizk: Added in #466 — the `list-object-versions | jq` substitution now ends in `|| echo ""`, matching `BUCKETS=`, so a transient error can't take the step down under `bash -eo pipefail`.

.github/workflows/cleanup-stacks.yml
osama-rizk: Good call. #466 adds a 40-min wall-clock budget (`DEADLINE=$((SECONDS + 2400))`); the loop breaks when it's reached and the remaining stacks are picked up on the next scheduled run, so a large versioned-bucket backlog can't reintroduce the timeout.

.github/workflows/cleanup-stacks.yml
osama-rizk: Fixed in #466 — `TRIGGERED` now increments only on a successful `delete-stack`; failed calls go to a separate `TRIGGER_FAILED` counter shown in the summary.

.github/workflows/cleanup-stacks.yml
osama-rizk: Same fix applied to the Amplify step in #466 (success-only `TRIGGERED` + `TRIGGER_FAILED`).

.github/workflows/cleanup-stacks.yml
osama-rizk: Trimmed in #466 — the second block is now a one-line pointer back to `empty_stack_buckets`.

.github/workflows/cleanup-stacks.yml
osama-rizk: Split in #466 into `SKIPPED_NOT_E2E` and `SKIPPED_TOO_NEW`, each its own summary line.

.github/workflows/cleanup-stacks.yml
osama-rizk: #466 restores a signal: each step tallies stacks still in `DELETE_FAILED` after the run and emits a non-blocking `::warning::` + a summary line. True per-stack "stuck for N runs" tracking needs state this stateless job doesn't have — glad to add it as a follow-up via a retry-count tag or a CloudWatch metric if you think it's worth it.

PR #466, opened 2026-09-01
fix(ci): correct jq so cleanup actually empties versioned buckets

osama-rizk:
## Problem

Follow-up to #465. In-account validation after that PR merged showed the bucket-emptying **never ran** — hosting stacks stayed `DELETE_FAILED` with `The bucket you tried to delete is not empty` (17 stuck), even though the workflow reported success (fire-and-forget) and freed ~100 roles from the *non-bucket-blocked* stacks.

## Root cause

The jq in `empty_stack_buckets` wrapped the concatenation in an extra array:

```
[(.Versions // []) + (.DeleteMarkers // [])] | map({Key, VersionId})
```

`[ ... ]` makes a **single-element** array containing the concatenated array, so `map({Key,VersionId})` indexes an array with `.Key` → jq errors (`Cannot index array with string "Key"`). The `2>/dev/null || echo ""` then swallowed the error → `PAYLOAD=""` → the loop hit `[ -z "$PAYLOAD" ] && break` and exited **before deleting anything**. Every bucket-empty was a no-op.

## Fix

Drop the extra brackets so `map` runs over the concatenated array:

```
((.Versions // []) + (.DeleteMarkers // [])) | map({Key, VersionId})
```

## Testing

- `bash -euo pipefail -n` passes; YAML parses.
- jq verified: populated input → `{Objects:[{Key,VersionId}...]}`; empty bucket (`Versions/DeleteMarkers: null`) → `{Objects:[]}` (length 0 → loop breaks cleanly, no error).
- Will confirm end-to-end against the 17 currently-stuck hosting stacks.

### DIFF
--- a/.github/workflows/cleanup-stacks.yml
+++ b/.github/workflows/cleanup-stacks.yml
@@ -46,8 +46,12 @@ jobs:
           # Emptying out-of-band lets the subsequent delete-stack succeed.
           empty_stack_buckets() {
             local STACK="$1" B BUCKETS PAYLOAD
-            BUCKETS=$(aws cloudformation describe-stack-resources --stack-name "$STACK" \
-              --query "StackResources[?ResourceType=='AWS::S3::Bucket'].PhysicalResourceId" \
+            # Use list-stack-resources (paginated), NOT describe-stack-resources —
+            # the latter caps at 100 resources and on a large OpenNext SSR stack omits
+            # the very HostingStorageHostingBucket that blocks the delete, so it never
+            # gets emptied.
+            BUCKETS=$(aws cloudformation list-stack-resources --stack-name "$STACK" \
+              --query "StackResourceSummaries[?ResourceType=='AWS::S3::Bucket'].PhysicalResourceId" \
               --output text 2>/dev/null || echo "")
             for B in $BUCKETS; do
               [ -z "$B" ] && continue
--- a/.github/workflows/cleanup-stacks.yml
+++ b/.github/workflows/cleanup-stacks.yml
@@ -33,8 +33,15 @@ jobs:
       - name: Find and delete stale e2e stacks
         run: |
           CUTOFF=$(date -d '2 hours ago' +%s)
+          # Wall-clock budget for the inline empty+trigger loop. Emptying large
+          # versioned buckets is synchronous, so cap total work well under the
+          # job's 60-min timeout and defer the rest to the next run — rather than
+          # risk re-introducing the timeout this redesign removed.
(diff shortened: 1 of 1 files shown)


osama-rizk:
@soberm thanks for the thorough review on #465 — you were right that the bucket-emptying didn't work. #465 was already merged, so I've addressed all of it here in #466 (which branches off merged `main`). Item by item:

**[blocker] jq throws on every call (l60)** — Fixed. Adopted your fully-parenthesized form so the pipe is unambiguously inside the object value:
`{Objects: ((.Versions // []) + (.DeleteMarkers // []) | map({Key, VersionId}))}`
Verified with jq on real version/marker data → `{Objects:[...]}` and on empty `{}` → `{Objects:[]}` (len 0 → loop breaks, no throw).

Validating in-account also surfaced a **second** no-op cause beyond the jq bug: `describe-stack-resources` caps at 100 resources and doesn't paginate, so on a large OpenNext SSR stack it omitted the very `HostingStorageHostingBucket` that blocks the delete. Switched to paginated `list-stack-resources`. End-to-end proof: a stuck stack's storage bucket went 46 → 0 versions and all 17 previously-`DELETE_FAILED` hosting stacks then left `DELETE_FAILED`.

**[major] no `|| echo ""` fallback on the list-object-versions line (l59)** — Added; the substitution is now total so a transient AWS/jq error can't take the step down under `bash -eo pipefail`.

**[minor] inline sync emptying could reintroduce the timeout (l120)** — Added a 40-min wall-clock budget (`DEADLINE=$((SECONDS + 2400))`); the loop breaks when it's hit and remaining stacks are picked up next run.

**[minor] TRIGGERED is an attempt count, not success (l123) + [minor] same in Amplify step (l206)** — Both now increment only on a successful `delete-stack`; failed calls go to a separate `TRIGGER_FAILED` counter, surfaced in each summary.

**[nit] duplicate root-cause comment blocks (l95)** — Trimmed the second block to a pointer back to `empty_stack_buckets`.

**[nit] SKIPPED lumps two reasons (l94)** — Split into `SKIPPED_NOT_E2E` and `SKIPPED_TOO_NEW` in the summary.

**[question] signal for stacks stuck across many runs (l108)** — Each step now tallies stacks still in `DELETE_FAILED` after the run and emits a non-blocking `::warning::` + a summary line, so the backlog can't grow silently again. True per-stack "stuck for N runs" tracking needs state this stateless job doesn't have — happy to do it as a follow-up via a retry-count tag or a CloudWatch metric if you'd like.

PTAL when you get a chance.

soberm (approved):

sarayev (approved):

PR #467, opened 2026-09-01
fix: reliable E2E teardown — retain BucketDeployment CRs, empty versioned buckets, cover telemetry/vpc-smoke

osama-rizk:
Three teardown gaps that leak E2E stacks — and, critically, **CloudFront distributions** — until an account hits a service quota (`RolesPerAccount: 1500`) or accumulates dangling distributions. Follow-up to #465 / #466.

## 1. Retain BucketDeployment CRs so they can't wedge stack teardown (`packages/hosting`)
**This is the fix for the dangling-CloudFront leak** on the high-volume Amplify SSR-adapter e2e (`deployment-type=standalone`, which re-exports this construct). A CDK `BucketDeployment` custom resource runs a **delete-time handler** (object cleanup / CloudFront invalidation — aws-cdk#15891, aws-cdk#23708). When it fails, the custom-resource delete wedges the whole stack in `DELETE_FAILED`, which **orphans the CloudFront distribution** (the distribution can't be deleted while its stack is stuck). Fix: `applyRemovalPolicy(RETAIN)` on every `Custom::CDKBucketDeployment` in `HostingConstruct`, removing them from the teardown path so the stack — and its distribution — delete cleanly. Object cleanup is still handled by the bucket's `autoDeleteObjects` / the sandbox teardown. Synth test asserts every CR is `DeletionPolicy: Retain` (SPA + SSR).

## 2. Per-job teardown empties versioned buckets (`packages/core`)
`destroySandbox` (used by `npm run destroy` and each E2E job's cleanup) relies on `cdk destroy`, whose `autoDeleteObjects` Lambda never provisions if the CREATE failed — so a **versioned** bucket blocks the delete → the stack sticks in `DELETE_FAILED` → its roles leak (the workflow's `|| true` hid it). On a destroy failure we now resolve the app's stacks (`cdk ls`) and empty their versioned buckets — all versions + delete markers, paginated via `list-stack-resources` (not the 100-capped `describe-stack-resources`) + S3 `DeleteObjects` — before retrying, alongside the existing VPC-ENI backoff. Adds `@aws-sdk/client-cloudformation`; unit tests for `toDeleteObjects`, pagination, and the stop-on-`Errors` anti-infinite-loop.

## 3. Scheduled cleanup covers telemetry / vpc-smoke (`cleanup-stacks.yml`)
The janitor never touched `bb-telemetry-*` / `bb-vpc-smoke-*` (missing from the prefix discovery + tally, and the tag gate only accepted `e2e-*` while those carry `telemetry-e2e` / `vpc-smoke-e2e`). Added both prefixes and broadened the gate to accept `e2e-*` **and** `*-e2e`.

## IAM (deployed, separate repo)
Bucket-emptying + widened discovery need `s3:DeleteObjectVersion`, `cloudformation:ListStackResources`, and CFN/S3 on the new prefixes — the `blocks-github-actions` role lacked `ListStackResources` entirely (so the janitor's emptying was AccessDenied → silent no-op). Fixed + deployed via `AwsBlocksIntegTestAndGithubInfraCDK` (CR-301551315, CR-301560186, CR-301593438).

## Testing
- Hosting: 117/117 pass incl. the new RETAIN assertions; snapshot + standalone suites green.
- Core: `toDeleteObjects` / pagination / stop-on-Errors unit tests pass.
- Workflow: YAML parses; `bash -euo pipefail -n` clean; biome lint clean.
- Verified in-account: emptying versioned buckets + retaining CRs lets previously-`DELETE_FAILED` hosting stacks delete and frees their roles.

## Not in scope (follow-ups)
- Amplify-backend: consume the new `@aws-blocks/hosting` (version bump PR) + a backstop dangling-distribution reaper in its e2e cleanup (bucket-based reaper misses a distribution whose stack wedged).
- Level-1 non-versioned sandbox buckets at creation; `RolesPerAccount` quota increase + utilization alarms.


packages/core/src/scripts/sandbox.ts
```diff
@@ -321,6 +323,107 @@ export async function startSandbox(options: SandboxOptions) {
...
+			'npm',
+			['exec', 'cdk', '--', 'ls', '--context', 'sandboxMode=true', '--app', `npm exec tsx -- -C cdk ${backendPath}`],
+			{ env: cdkEnv, encoding: 'utf-8', stdio: ['ignore', 'pipe', 'ignore'] },
+		);
+		return out
+			.split('\n')
+			.map((s) => s.trim())
+			.filter(Boolean);
+	} catch {
+		return [];
```
soberm, line 445: `listSandboxStackNames` swallows `cdk ls` failures with no logging (stderr discarded, catch returns `[]`). A genuine `cdk ls`/synth failure is then indistinguishable from "app has zero stacks", so `destroySandbox`'s retry silently skips bucket-emptying — the same class of silent failure this PR sets out to eliminate. Suggest adding a `console.warn` in the catch block, mirroring the warn style used elsewhere in this PR.

packages/core/src/scripts/sandbox.ts
```diff
@@ -321,6 +323,107 @@ export async function startSandbox(options: SandboxOptions) {
...
+ * a partial/aborted deploy), so the versioned bucket blocks the delete and the
+ * stack (and its IAM roles) leak. Emptying out-of-band before the retry lets
+ * the delete complete. Best-effort: any per-stack/per-bucket error is logged
+ * and skipped so teardown still proceeds.
+ */
+async function emptySandboxBuckets(stackNames: string[]): Promise<void> {
+	try {
+		const { CloudFormationClient } = await import('@aws-sdk/client-cloudformation');
+		const { S3Client } = await import('@aws-sdk/client-s3');
+		const cfn = new CloudFormationClient({});
```
soberm, line 397: `emptySandboxBuckets` constructs `CloudFormationClient`/`S3Client` with no region override. `buildSandboxDeployArgs` notes a Lambda@Edge app synthesizes a second stack forced into `us-east-1`. If `cdk ls` returns cross-region stacks, `listStackBucketNames` targets the wrong region and is silently skipped. Low practical impact (edge-lambda stacks rarely own buckets) but worth a comment or follow-up.

packages/core/src/scripts/sandbox.ts
```diff
@@ -321,6 +323,107 @@ export async function startSandbox(options: SandboxOptions) {
...
+}): Array<{ Key: string; VersionId?: string }> {
+	return [...(listed.Versions ?? []), ...(listed.DeleteMarkers ?? [])]
+		.filter((v): v is { Key: string; VersionId?: string } => v.Key !== undefined)
+		.map((v) => ({ Key: v.Key, VersionId: v.VersionId }));
+}
+
+/** Collect a stack's S3 bucket physical names (paginated — a large stack has
+ * >100 resources, so a single page can miss the hosting bucket that blocks the
+ * delete). Returns [] if the stack is already gone / not accessible. */
+async function listStackBucketNames(cfn: CloudFormationClient, stackName: string): Promise<string[]> {
```
soberm: Test coverage gap: only the pure `toDeleteObjects` helper is unit-tested. `listStackBucketNames` (NextToken pagination) and `emptyBucket` (re-list-and-delete loop + stop-on-`Errors` guard) have the trickiest control flow and take the client as a param already, so they're testable. Consider exporting and covering at least the "stop after Errors" and multi-page NextToken branches so a refactor can't silently break them.

packages/core/src/scripts/sandbox.ts
```diff
@@ -343,6 +446,7 @@ export async function destroySandbox(backendPath: string) {
     // because the subnets still have attached ENIs; retrying after the cleanup
     // window lets the VPC delete succeed.
     const retryDelays = [60_000, 120_000]; // 1min, then 2min
+    let stackNames: string[] | undefined;
```
soberm: Nitpick: `stackNames` is memoized lazily and reused across retries — reasonable since topology won't change between attempts, but a one-line comment stating that assumption would help the next reader.

soberm (commented):
Well-scoped, low-risk fix for E2E teardown. No blockers. A few suggestions below, mainly one silent-failure path that partially undercuts the PR's own goal, plus an unverifiable IAM assumption.

**General: IAM coverage** — widening stack discovery to `bb-telemetry-*`/`bb-vpc-smoke-*` also exercises `DescribeStacks`/`ListStackResources` and S3 `ListBucketVersions`/`DeleteObject`, not just `DeleteStack` (granted in CR-301560186, a different repo). Please confirm the role covers all those APIs for the new prefixes, otherwise the widened coverage may produce silent `AccessDenied` no-ops.

**Praise:** nice use of lazy dynamic AWS SDK imports to keep client cost off the happy path; correct anti-infinite-loop guard in `emptyBucket` (stop when `DeleteObjects` returns `Errors`); well-layered best-effort error handling; and a focused unit test scoped to the riskiest pure logic.

packages/core/src/scripts/sandbox.ts
osama-rizk: Fixed — `listSandboxStackNames` now captures stderr and `console.warn`s the failure (with the cdk-ls detail) instead of silently returning `[]`, so a real synth/`cdk ls` failure is distinguishable from 'zero stacks' and won't quietly skip bucket-emptying.

packages/core/src/scripts/sandbox.ts
osama-rizk: Added a comment on the client construction noting the ambient-region assumption: a Lambda@Edge app's us-east-1 stack, if surfaced by `cdk ls`, would be queried in the wrong region and the per-stack error is caught + skipped — low impact since edge stacks don't own buckets. Left per-stack-region resolution as a noted follow-up rather than building it now.

packages/core/src/scripts/sandbox.ts
osama-rizk: Done — exported `listStackBucketNames` and `emptyBucket` and added unit tests for the tricky branches: NextToken pagination (and bucket-type filtering), the re-list-until-empty loop, and the stop-on-`Errors` guard (asserts it does NOT re-list after Errors, i.e. the anti-infinite-loop). 7/7 pass.

packages/core/src/scripts/sandbox.ts
osama-rizk: Added a one-line comment stating the assumption (stack topology is stable across destroy attempts, so `stackNames` is resolved once and reused).

osama-rizk:
@soberm on the **IAM coverage** point — you were right, and it caught a genuine latent gap. I checked the live `blocks-github-actions` policy: it had **no `cloudformation:ListStackResources`** at all, and the S3 cleanup grant didn't include the `bb-telemetry-*` / `bb-vpc-smoke-*` prefixes. Since the cleanup workflow (#466) and this PR's teardown both enumerate buckets via `ListStackResources`, that call was `AccessDenied` for **every** prefix → bucket-emptying silently no-op'd under the OIDC role (only masked because my in-account validation used admin creds).

Fixed in **CR-301593438** (AwsBlocksIntegTestAndGithubInfraCDK): adds `cloudformation:ListStackResources` + `DescribeStackResources` to `E2eCloudFormation`, and extends the `E2eBucketCleanup` S3 grant to the telemetry/vpc-smoke prefixes. Already deployed to the CI account (`277731792744`) — verified the role now carries all four APIs on the new prefixes. Thanks for flagging it.

soberm (dismissed):

packages/core/src/scripts/sandbox.ts
```diff
@@ -321,6 +323,119 @@ export async function startSandbox(options: SandboxOptions) {
...
+		console.warn(`  ⚠️  bucket-emptying step skipped: ${(e as Error).message}`);
+	}
+}
+
+/** Resolve the sandbox app's stack names via `cdk ls` (best-effort; [] on error). */
+function listSandboxStackNames(backendPath: string, cdkEnv: NodeJS.ProcessEnv): string[] {
+	try {
+		const out = execFileSync(
+			'npm',
+			['exec', 'cdk', '--', 'ls', '--context', 'sandboxMode=true', '--app', `npm exec tsx -- -C cdk ${backendPath}`],
```
sarayev: [minor] `backendPath` gets interpolated straight into the `--app` string that CDK then re-runs through a shell (`npm exec tsx -- -C cdk ${backendPath}`). If that path ever contains a space it'll split there. It's an internal path so the risk is low, but should we quote it just to be safe?

```ts
'--app', `npm exec tsx -- -C cdk '${backendPath}'`,
```

packages/core/src/scripts/sandbox.ts
```diff
@@ -321,6 +323,119 @@ export async function startSandbox(options: SandboxOptions) {
...
+		const { CloudFormationClient } = await import('@aws-sdk/client-cloudformation');
+		const { S3Client } = await import('@aws-sdk/client-s3');
+		// Clients use the ambient region (AWS_REGION / profile). A Lambda@Edge app
+		// synthesizes a second stack pinned to us-east-1; if `cdk ls` surfaced such a
+		// cross-region stack, listStackBucketNames would query the wrong region and
+		// the per-stack error is caught + skipped below. Low impact in practice —
+		// edge-lambda stacks don't own S3 buckets — but noted for a future
+		// per-stack-region follow-up if that changes.
+		const cfn = new CloudFormationClient({});
+		const s3 = new S3Client({});
```
sarayev: [minor] This `S3Client` uses the ambient region. If a stack ever owns a bucket outside `AWS_REGION`, `ListObjectVersions`/`DeleteObjects` fails with `PermanentRedirect`, the per-bucket `catch` below swallows it, and the bucket is left non-empty, so the destroy still can't finish. The comment notes edge stacks don't own buckets, so this is probably fine today. Can we confirm the standalone hosting e2e never places a bucket outside `AWS_REGION`? This is exactly the silent no-op this PR is trying to remove.

packages/core/src/scripts/sandbox.ts
```diff
@@ -321,6 +323,119 @@ export async function startSandbox(options: SandboxOptions) {
...
+ * pages of 1000 (the DeleteObjects limit). Stops if a page reports per-key
+ * errors (object-lock / retention) so it can't loop forever. */
+export async function emptyBucket(s3: S3Client, bucket: string): Promise<void> {
+	const { ListObjectVersionsCommand, DeleteObjectsCommand } = await import('@aws-sdk/client-s3');
+	while (true) {
+		const listed = await s3.send(new ListObjectVersionsCommand({ Bucket: bucket, MaxKeys: 1000 }));
+		const objects = toDeleteObjects(listed);
+		if (objects.length === 0) return;
+		const res = await s3.send(new DeleteObjectsCommand({ Bucket: bucket, Delete: { Objects: objects, Quiet: true } }));
+		if (res.Errors && res.Errors.length > 0) {
```
sarayev, line 368: [question] Nice guard against the infinite loop here. When it bails on `Errors` (object-lock / AccessDenied) it returns normally, so `destroySandbox` just retries `cdk destroy` and hits the same wall until the retries run out. That's acceptable for best-effort, but should this warning be louder or counted? A silently-skipped bucket is the thing that leaks the stack in the first place.

packages/core/src/scripts/sandbox.ts
```diff
@@ -351,6 +467,19 @@ export async function destroySandbox(backendPath: string) {
...
+          //  1. Non-empty versioned S3 buckets — `cdk destroy` relies on each
+          //     bucket's autoDeleteObjects Lambda, which never provisioned if the
+          //     CREATE failed, so the bucket blocks the delete and the stack (with
+          //     its IAM roles) leaks. Empty them out-of-band here.
+          //  2. VPC ENIs that take 60-120s to detach (the backoff below).
+          // Resolve stack names once and reuse across retries — the app's stack
+          // topology doesn't change between destroy attempts.
+          if (stackNames === undefined) stackNames = listSandboxStackNames(backendPath, cdkEnv);
+          if (stackNames.length > 0) {
+            console.log("\n🧹 Emptying versioned S3 buckets before retry...");
```
sarayev: [minor] The leaf helpers (`toDeleteObjects`, pagination, stop-on-Errors) are well covered, but this wiring, emptying buckets between destroy attempts plus `listSandboxStackNames`, is the actual sev2 fix and has no test. Worth a small one asserting "on a destroy failure, buckets get emptied before the retry" (inject fakes for the two)? Would lock in the behavior this PR exists for.

packages/core/src/scripts/sandbox.ts
```diff
@@ -351,6 +467,19 @@ export async function destroySandbox(backendPath: string) {
...
         if (attempt < retryDelays.length) {
+          // Clear the two known teardown blockers before retrying:
+          //  1. Non-empty versioned S3 buckets — `cdk destroy` relies on each
+          //     bucket's autoDeleteObjects Lambda, which never provisioned if the
+          //     CREATE failed, so the bucket blocks the delete and the stack (with
+          //     its IAM roles) leaks. Empty them out-of-band here.
+          //  2. VPC ENIs that take 60-120s to detach (the backoff below).
+          // Resolve stack names once and reuse across retries — the app's stack
+          // topology doesn't change between destroy attempts.
+          if (stackNames === undefined) stackNames = listSandboxStackNames(backendPath, cdkEnv);
```
sarayev: [nit] If a destroy fails purely because of a non-empty bucket (no VPC/ENI involved), we empty it instantly here but still sleep the full 60s/120s before retrying. Minor, could retry right away when the error doesn't look ENI-related. Not blocking.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -1575,6 +1575,24 @@ export class HostingConstruct extends Construct {
...
+    // resource delete wedges the whole stack in DELETE_FAILED, which orphans the
+    // CloudFront distribution. Retaining the CR removes it from the teardown
+    // path entirely, so the stack (and its distribution) can delete cleanly.
+    // Safe because the bucket's own autoDeleteObjects / the sandbox teardown
+    // empties the objects — the CR isn't needed to clean them up. This was
+    // observed leaking distributions on the high-volume Amplify SSR-adapter e2e
+    // (deployment-type=standalone), which consumes this construct.
+    for (const child of this.node.findAll()) {
+      if (CfnResource.isCfnResource(child) && child.cfnResourceType === 'Custom::CDKBucketDeployment') {
+        child.applyRemovalPolicy(RemovalPolicy.RETAIN);
```
sarayev, line 1600: [question] Retaining the `BucketDeployment` CRs pulls them off the teardown path, but the bucket's own `autoDeleteObjects` custom resource (`Custom::S3AutoDeleteObjects`) is still a delete-time handler on that path. Since the root cause was a delete-time handler wedging the stack, what makes us confident that one won't wedge too? I think it genuinely is safer (it only empties the bucket, no CloudFront invalidation) - could we say that in the comment so the next reader knows why only `BucketDeployment` is retained and not the auto-delete CR?

packages/core/src/scripts/sandbox.ts
osama-rizk: Done — single-quoted `backendPath` in both cdk `--app` strings (`cdk ls` and `cdk destroy`) so a path with a space can't split.

packages/core/src/scripts/sandbox.ts
osama-rizk: Good catch — closed it rather than just documenting. The S3 client is now `new S3Client({ followRegionRedirects: true })`, so a bucket outside `AWS_REGION` (e.g. a us-east-1 L@E stack's bucket) is emptied by transparently following the region redirect instead of `PermanentRedirect` → silent skip. It needs no `GetBucketLocation` permission (which the role doesn't grant), so no new IAM. I didn't try to *confirm* the standalone e2e never places a bucket cross-region — this makes it correct either way.

packages/core/src/scripts/sandbox.ts
osama-rizk: Agreed. Made the warning louder: it now logs the count + the first key/code and states the bucket "will block stack teardown until cleared manually" — so a skipped bucket is visible in CI output rather than a quiet one-liner. Kept it best-effort (returns) rather than throwing, since the retry/exhaust path already surfaces the final failure.

packages/core/src/scripts/sandbox.ts
osama-rizk: Added. Extracted the orchestration into `runDestroyWithRetries(deps)` with injectable `runDestroy` / `listStackNames` / `emptyBuckets` / `sleep`, and unit-tested the wiring: (1) on a destroy failure buckets are emptied **before** the retry (asserts the exact event order incl. the resolved stack names), (2) no empty/sleep when the first destroy succeeds, (3) throws after retries are exhausted. This locks in the sev2 behavior without spawning cdk.

packages/core/src/scripts/sandbox.ts
osama-rizk: Fair — left as-is for now. A destroy failure often involves *both* a non-empty bucket and ENIs, and reliably classifying "ENI-related vs not" means parsing CFN error text; the delay is bounded (1m/2m, max 2 retries), so the complexity didn't seem worth it. Easy follow-up if the wait becomes a pain in practice.

packages/hosting/src/constructs/hosting_construct.ts
osama-rizk: Great question — added a comment saying exactly this. We deliberately do NOT retain the bucket's `Custom::S3AutoDeleteObjects` CR: it only empties its own bucket (an S3-only op, no CloudFront invalidation / cross-service call), so it's far less wedge-prone, and it's the mechanism that actually removes the objects — retaining it would leave every bucket non-empty and blocking. Only the BucketDeployment CR (whose delete-time invalidation is the wedge) is retained.

sarayev (approved):

PR #471, opened 2026-09-01
fix(bb-distributed-table): deterministic GSI sort-key tie ordering in the mock

osama-rizk:
## Card

Bug Bash board (Starter Kit, DX P3): **"bb-distributed-table: position-tie ordering nondeterministic (mock vs prod)"** — *Map insertion order vs DDB sort-key order; tied positions render differently.*
🔗 https://github.com/orgs/aws-amplify/projects/141/views/15?pane=issue&itemId=195534546

## Problem

A **GSI** sort key need not be unique, so several items can share one sort-key value (e.g. a `position` field). In `query`, the mock collects items from a `Map` and sorts by the index sort key, returning `0` for equal values — so tied rows keep **Map insertion order**. That order depends on write order and on disk-reload order (`.bb-data`), so mock results are **nondeterministic and can diverge from deployed behavior**. DynamoDB orders index rows with equal sort keys by the **base-table primary key**.

## Fix

Tie-break on the base-table primary key (partition key, then sort key) via a small `compareByBaseKey` helper, instead of falling back to insertion order. The whole index order — ties included — still reverses under `order: 'desc'`. No change when the sort key is unique (the common base-table case).

## Testing

New `query (index sort-key ties)` block (`index.test.ts`): a `byPosition` GSI with several cards sharing `position`:
- ties order by the base key (`cardId`), and **the same result holds regardless of write order** (deterministic);
- `order: 'desc'` reverses ties too;
- primary ordering stays by the index sort key — the base key only breaks ties.

`@aws-blocks/bb-distributed-table` suite **144 pass / 0 fail**.

## Verification

`npm run lint` 0 errors · umbrella-changeset guard ✅. Internal change (private comparator) — no public API change. Documented in DESIGN.md D-DT-6.

## Changeset

`@aws-blocks/bb-distributed-table` **patch** + `@aws-blocks/blocks` **patch** (umbrella re-export).

## Why no real-DynamoDB e2e for the tie order

The fix makes the **mock deterministic** by tie-breaking on the base-table key. It matches DynamoDB's *observed* ordering, but AWS doesn't document a **contractual** guarantee for how index rows with equal sort keys are ordered. An e2e that asserts a specific tie order against real DynamoDB would therefore be asserting unspecified behavior — brittle, and liable to flake if the service's internal ordering ever shifts. So the e2e is intentionally out of scope; the value here is reproducible local ordering (covered by the write-order-convergence unit tests), not pinning DynamoDB's contract. Documented in DESIGN.md D-DT-6.

### DIFF
--- a/.changeset/dt-index-tie-ordering.md
+++ b/.changeset/dt-index-tie-ordering.md
@@ -5,4 +5,4 @@
 
 `DistributedTable` mock: order GSI sort-key ties deterministically to match DynamoDB.
 
-When a GSI sort key was shared by multiple items, the mock's `query` fell back to Map insertion order for the tied rows — so results depended on write order / disk-reload order and diverged from deployed behavior. The mock now tie-breaks on the base-table primary key (partition key, then sort key), matching how DynamoDB orders index rows with equal sort keys, and reverses the whole order (ties included) under `order: 'desc'`. No change for unique sort keys.
+When a GSI sort key was shared by multiple items, the mock's `query` fell back to Map insertion order for the tied rows — so results depended on write order / disk-reload order and diverged from deployed behavior. The mock now tie-breaks on the base-table primary key (partition key, then sort key), matching DynamoDB's observed ordering for index rows with equal sort keys, and reverses the whole order (ties included) under `order: 'desc'`. AWS doesn't document a contractual guarantee for the tie order, so the mock is intentionally deterministic even where DynamoDB's contract is unspecified. No change for unique sort keys.
--- a/.changeset/dt-index-tie-ordering.md
+++ b/.changeset/dt-index-tie-ordering.md
@@ -1,0 +1,8 @@
+---
+"@aws-blocks/bb-distributed-table": patch
+"@aws-blocks/blocks": patch
+---
+
+`DistributedTable` mock: order GSI sort-key ties deterministically to match DynamoDB.
+
+When a GSI sort key was shared by multiple items, the mock's `query` fell back to Map insertion order for the tied rows — so results depended on write order / disk-reload order and diverged from deployed behavior. The mock now tie-breaks on the base-table primary key (partition key, then sort key), matching how DynamoDB orders index rows with equal sort keys, and reverses the whole order (ties included) under `order: 'desc'`. No change for unique sort keys.
--- a/packages/bb-distributed-table/src/index.mock.ts
+++ b/packages/bb-distributed-table/src/index.mock.ts
@@ -266,7 +266,14 @@ export class DistributedTable<
 			const dir = options.order === 'desc' ? -1 : 1;
 			items.sort((a, b) => {
 				const av = (a as any)[skField], bv = (b as any)[skField];
-				return (av < bv ? -1 : av > bv ? 1 : 0) * dir;
+				const primary = av < bv ? -1 : av > bv ? 1 : 0;
+				// Tie-break on the base-table primary key. On a GSI the sort key need
+				// not be unique, so equal sort-key values must NOT fall back to Map
+				// insertion order (which varies by write order / disk reload) — that's
+				// the mock-vs-DynamoDB divergence. DynamoDB orders index ties by the
+				// base-table key, and the whole index (ties included) reverses under
+				// `order: 'desc'`.
+				return (primary !== 0 ? primary : this.compareByBaseKey(a, b)) * dir;
 			});
 		}
 
@@ -357,6 +364,24 @@ export class DistributedTable<
 				throw blocksError(DistributedTableErrors.ConditionalCheckFailed, 'The conditional request failed');
 			}
 		}
+	}
+
+	/**
+	 * Deterministic tie-break for `query` ordering: compare two items by the
+	 * base-table primary key (partition key, then sort key). Used when an index
+	 * sort-key value is shared by multiple items, so results don't depend on Map
+	 * insertion order. Returns a stable -1/0/1.
+	 */
+	private compareByBaseKey(a: T, b: T): number {
+		const pk = this.keyConfig.partitionKey;
+		const ap = (a as any)[pk], bp = (b as any)[pk];
+		if (ap !== bp) return ap < bp ? -1 : 1;
+		const sk = this.keyConfig.sortKey;
+		if (sk) {
+			const as = (a as any)[sk], bs = (b as any)[sk];
+			if (as !== bs) return as < bs ? -1 : 1;
+		}
+		return 0;
 	}
 
 	private serializeKey(key: TableKey<T, K>): string {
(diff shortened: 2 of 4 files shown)


.changeset/dt-index-tie-ordering.md
```diff
@@ -0,0 +1,8 @@
+---
+"@aws-blocks/bb-distributed-table": patch
+"@aws-blocks/blocks": patch
+---
+
+`DistributedTable` mock: order GSI sort-key ties deterministically to match DynamoDB.
+
+When a GSI sort key was shared by multiple items, the mock's `query` fell back to Map insertion order for the tied rows — so results depended on write order / disk-reload order and diverged from deployed behavior. The mock now tie-breaks on the base-table primary key (partition key, then sort key), matching how DynamoDB orders index rows with equal sort keys, and reverses the whole order (ties included) under `order: 'desc'`. No change for unique sort keys.
```
soberm: **Suggestion**

The changeset (and `DESIGN.md`) state definitively that "DynamoDB orders index rows with equal sort keys by the base-table primary key." This ordering is widely observed, but AWS's public docs are not unambiguous that it's a contractual guarantee for GSI sort-key ties.

If it isn't guaranteed, the mock is now *intentionally more deterministic* than DynamoDB's contract — which is a good thing for testing, and worth stating as such rather than as a strict match.

Per `A0-API-DESIGN.md` § "Living Document" (capture edge-case decisions) and Tenet T3 (documentation accuracy), consider softening to "matches DynamoDB's observed ordering" and noting the mock is deterministic even where DynamoDB's contract is unspecified. What's the source you're anchoring the guarantee on?

packages/bb-distributed-table/src/index.mock.ts
```diff
@@ -359,6 +366,24 @@ export class DistributedTable<
...
+	 * insertion order. Returns a stable -1/0/1.
+	 */
+	private compareByBaseKey(a: T, b: T): number {
+		const pk = this.keyConfig.partitionKey;
+		const ap = (a as any)[pk], bp = (b as any)[pk];
+		if (ap !== bp) return ap < bp ? -1 : 1;
+		const sk = this.keyConfig.sortKey;
+		if (sk) {
+			const as = (a as any)[sk], bs = (b as any)[sk];
+			if (as !== bs) return as < bs ? -1 : 1;
```
soberm, line 382: **Nit**

The `<`/`>` comparison here inherits the mock's existing string semantics (JS UTF-16 code-unit order), which can differ from DynamoDB's UTF-8 byte order for non-ASCII base-key values. This matches the pre-existing index-sort comparison at L268–L269, so it's not a regression introduced by this PR.

Consider a short note in `DESIGN.md` making this known boundary explicit, so the parity discussion documents where mock string ordering can diverge. Take it or leave it — purely a documentation nicety.

packages/bb-distributed-table/src/index.test.ts
```diff
@@ -859,4 +859,49 @@ describe('DistributedTable', () => {
...
+		test('ties on the index sort key order by the base-table key, regardless of write order', async () => {
+			const t1 = cardTable();
+			// All position=1; insert cardIds out of order.
+			for (const cardId of ['c3', 'c1', 'c2']) await t1.put({ boardId: 'b1', cardId, position: 1 });
+			assert.deepEqual((await q(t1)).map((c) => c.cardId), ['c1', 'c2', 'c3']);
+
+			// A different write order must yield the SAME result (deterministic).
+			const t2 = cardTable();
+			for (const cardId of ['c2', 'c3', 'c1']) await t2.put({ boardId: 'b1', cardId, position: 1 });
+			assert.deepEqual((await q(t2)).map((c) => c.cardId), ['c1', 'c2', 'c3']);
```
soberm, line 889: **Suggestion**

This is exactly the right test for the root cause — two write orders converging on one result. Since the whole fix depends on "the mock matches DynamoDB," have we considered a single e2e case in `test-apps/comprehensive/` that writes rows sharing a GSI sort-key value and asserts the same order against real DynamoDB?

That would confirm the premise the fix rests on and guard against a future service behavior change silently reopening the parity gap. Per `CONTRIBUTING.md` § Building Block Implementation Checklist (Comprehensive Test App) and Tenet T1. If e2e is intentionally out of scope for a mock-only fix, capturing that rationale in the PR description would close the loop.

packages/bb-distributed-table/src/index.mock.ts
```diff
@@ -359,6 +366,24 @@ export class DistributedTable<
...
+		}
+		return 0;
```
soberm, line 387: **Nit**

Every item in `this.data` is keyed by its serialized base primary key (`serializeKey`), so two distinct entries always differ in base PK or base SK — meaning `compareByBaseKey` can never actually fall through to this `return 0` for distinct rows. The line is correct defensive code; a one-line comment noting it's unreachable-for-distinct-rows would save the next reader from wondering why it isn't covered by a test.

.changeset/dt-index-tie-ordering.md
osama-rizk: Fair — softened both the changeset and DESIGN.md (D-DT-6) to "matches DynamoDB's **observed** ordering," and stated explicitly that AWS doesn't document a contractual guarantee for the tie order, so the mock is intentionally **deterministic even where DynamoDB's contract is unspecified** (a plus for reproducible tests, not something to rely on in prod logic). On the source: I don't have an AWS doc that guarantees it — it's the widely-observed behavior (GSI rows are physically ordered by sort key then the base-table key that makes them unique). So framing it as "observed + deterministic" rather than "guaranteed match" is the honest call. Thanks.

packages/bb-distributed-table/src/index.mock.ts
osama-rizk: Good call — added a "String-ordering boundary" note to DESIGN.md D-DT-6: the mock compares keys with JS `<`/`>` (UTF-16 code-unit order), which can differ from DynamoDB's UTF-8 byte order for non-ASCII keys, and this applies to both the index sort comparison (L268–269) and the new base-key tie-break. Flagged as a pre-existing mock property, not a regression.

packages/bb-distributed-table/src/index.test.ts
osama-rizk: Considered it, and I've captured the rationale in the PR description: because AWS doesn't guarantee the tie order (the point from your other comment), an e2e asserting a **specific** order against real DynamoDB would be asserting unspecified behavior — brittle and flake-prone if the service's internal ordering ever shifts. The value of this fix is reproducible **local** ordering (the write-order-convergence unit tests cover that), not pinning DynamoDB's contract. So the real-DDB e2e is intentionally out of scope; documented in DESIGN.md D-DT-6.

packages/bb-distributed-table/src/index.mock.ts
osama-rizk: Done — added the comment: `return 0` is unreachable for distinct rows because every item is keyed by its serialized base primary key, so two entries always differ in base PK or SK. (Also cross-referenced the UTF-16 ordering note there.)

soberm (approved):

sarayev (approved):

PR #487, opened 2026-09-03
fix(bb-file-bucket): secure defaults (enforceSSL, versioning on, access logs, CORS guard)

soberm:
## Problem

`bb-file-bucket` shipped insecure S3 defaults, surfaced by security scanner rules **R6/R8** (4 reports across 3 apps):

- **Versioning off by default** — no recovery from accidental overwrite or delete.
- **No server access logs** — no audit trail of object access.
- **Wildcard (`*`) CORS permitted for `PUT`** — the local dev file-server reflected the request `Origin` (falling back to `*`) while also sending `Access-Control-Allow-Credentials: true`, the classic credentialed-wildcard combination.
- **No in-transit encryption enforcement** — the bucket did not deny non-TLS requests.

**Issue #, if available:** Supersedes and extends #304.

## Changes

- **`enforceSSL: true`** always-on for the main bucket (denies non-TLS requests).
- **Versioning ON by default** (opt out with `versioned: false`). This also inverts the version-aware conditional option types so a default / no-options bucket gets `versionId`-aware typings that match the new runtime default — the type/runtime mismatch that led #304 to leave versioning opt-in is resolved here.
- **Opt-in `accessLogging`**: provisions a dedicated locked-down log bucket (`BLOCK_ALL`, `S3_MANAGED`, `enforceSSL`) with a lifecycle rule (`logRetentionDays`, default 90 days). Non-positive / non-integer `logRetentionDays` throws at synth.
- **Dev file-server** no longer sends `Access-Control-Allow-Credentials: true`, removing the credentialed reflected-origin/wildcard combination the scanner flagged.
- **Synth-time `Error`** when a `corsRule` allows origin `'*'` together with a mutating method (`PUT` / `POST` / `DELETE`); explicit origins are required instead. #304 documented against this rather than enforcing it — this PR enforces it.

### Breaking changes

- **Versioning is now on by default** — retained object versions accrue storage cost over time. Opt out with `versioned: false`.
- **Wildcard-origin CORS with a mutating method now fails synthesis** — a `corsRule` combining origin `'*'` with `PUT` / `POST` / `DELETE` throws; specify explicit origins instead.

## Validation

- **108 unit/synth tests pass** (was 92) via `node --test` on the compiled `dist`. Build is clean (`tsc --build`, no type errors).
- Developed **TDD red → green** (failing tests first, then implementation).
- New/updated tests cover each behavior: default `enforceSSL` deny policy; `accessLogging` provisioning a second locked-down bucket with its own `enforceSSL` policy, lifecycle expiry, and the main bucket's `LoggingConfiguration` referencing it; `logRetentionDays` synth-time validation (0 / negative / non-integer throw; ignored when logging is off); versioning-on-by-default at runtime and in the type layer (with explicit `versioned: false` opt-out); dev-server preflight no longer emitting `Access-Control-Allow-Credentials`; and the wildcard-CORS-with-mutating-method synth error (while wildcard + safe methods and explicit-origin + mutating methods remain allowed).

## Checklist

- [x] PR description included

---


packages/bb-file-bucket/src/index.cdk.ts
```diff
@@ -54,19 +60,78 @@ export class FileBucket<O extends FileBucketOptions = FileBucketOptions> extends
...
-			versioned: options?.versioned ?? false,
-			removalPolicy: destroy
-				? RemovalPolicy.DESTROY
-				: options?.removalPolicy === 'retain'
-					? RemovalPolicy.RETAIN
-					: undefined,
+			// All FileBucket traffic (SDK calls + presigned URLs) is HTTPS, so
+			// enforce TLS to close the in-transit exposure gap unconditionally.
+			enforceSSL: true,
+			versioned: options?.versioned ?? true,
```
osama-rizk: Versioning is now on by default, but the data bucket has no lifecycle rule to expire **noncurrent** versions (only the access-log bucket gets an expiry). Every overwrite/delete now retains prior versions indefinitely, so storage — and cost — grows monotonically with no bound. The changeset notes the cost accrual but doesn't cap it.

Secure-default patterns usually pair versioning with a `noncurrentVersionExpiration` (e.g. 30–90 days, opt-out) to keep recent-recovery without unbounded retention. Consider adding a default noncurrent-version expiration to the main bucket, or documenting the recommended lifecycle so consumers add one. Low severity — the security posture is right; this is the operational cost tail of the new default.

packages/bb-file-bucket/src/types.ts
```diff
@@ -94,23 +120,31 @@ export interface VersionedGetUrlOptions extends GetUrlOptions {
...
  * Resolves the get options type based on whether versioning is enabled.
  * Versioned buckets accept `{ versionId }`, non-versioned accept no options.
+ * Versioning is on by default, so only an explicit `versioned: false` selects
+ * the non-versioned (optionless) shape. A non-literal `boolean` (widened from
+ * a variable) or an absent `versioned` resolves to the versioned-aware shape,
+ * matching the default-on runtime behavior.
  */
 export type GetOptionsFor<O extends FileBucketOptions> =
-	O extends { versioned: true } ? VersionedGetOptions : undefined;
+	O extends { versioned: false } ? undefined : VersionedGetOptions;
```
osama-rizk, line 151: Confirming this is an accepted residual: keying on the literal `{ versioned: false }` means a dynamically-computed `versioned: someBoolean` (widened to `boolean`) resolves to the **versioned-aware** typings even when the runtime bucket ends up unversioned — so a caller could pass `versionId`-bearing options at the type level against a non-versioned bucket. It's documented, and the default-on direction is the safer one (matches the new runtime default), so this is likely fine — just flagging that the dynamic-`boolean` path reintroduces a smaller version of the type/runtime mismatch #304 hit, inverted. No change needed if that's understood/accepted.

packages/bb-file-bucket/src/index.cdk.ts
```diff
@@ -54,19 +60,78 @@ export class FileBucket<O extends FileBucketOptions = FileBucketOptions> extends
...
+			? RemovalPolicy.DESTROY
+			: options?.removalPolicy === 'retain'
+				? RemovalPolicy.RETAIN
+				: undefined;
+
+		// Opt-in server access logging: provision a dedicated, locked-down log
+		// bucket and expire its logs after the configured retention. Kept
+		// separate from the data bucket so log delivery can't loop back on it.
+		let serverAccessLogsBucket: s3.Bucket | undefined;
+		if (options?.accessLogging) {
```
osama-rizk: Framework-consistency: this bucket resolves its posture-relevant knobs locally instead of through the stack-wide `BlocksDefaults` model that #302 established and #385 (`logRetention` + `accessLogging` stack defaults, merged 2026-09-02) extended. #385's fields are already on this branch's `BlocksDefaults`, but bb-file-bucket doesn't read them — so sandbox vs production postures don't flow into the file bucket.

The framework contract is `options.field ?? scope.defaults.field` (per-block option wins; stack posture is the fallback), as adopted in bb-kv-store:
```ts
removalPolicy:      options?.removalPolicy      ?? this.defaults.removalPolicy,
deletionProtection: options?.deletionProtection ?? this.defaults.deletionProtection,
```
Four gaps here:
1. **accessLogging** — honors only `options.accessLogging`; should be `options.accessLogging ?? this.defaults.accessLogging` so a `production` posture that opts in gets logging on the file bucket (`BlocksDefaults.accessLogging` already exists via #385).
2. **logRetention** — introduces a parallel per-block `logRetentionDays: number` (default 90) instead of the framework `this.defaults.logRetention` (`RetentionDays`; sandbox `ONE_WEEK` / prod `ONE_YEAR`). Two competing log-retention concepts for the same package family, landed within a day of each other.
3. **removal / deletion** — still keyed off the legacy `cdk.Stack.of(this).node.tryGetContext('sandboxMode')` (line 57), which #302 explicitly replaced with `scope.defaults.removalPolicy` / `.deletionProtection`. bb-file-bucket hasn't migrated; bb-kv-store has.
4. **versioning** — hardcoded `versioned ?? true` regardless of posture, so throwaway sandbox buckets also accrue versions + cost (ties to the unbounded-noncurrent-versions comment above). If versioning should track posture, it wants a `BlocksDefaults` field (`options.versioned ?? this.defaults.versioned`) rather than a hardcoded default.

Net: this PR was developed in parallel to #385 rather than on top of it. Recommend rebasing onto #385 and routing accessLogging / logRetention / removal / deletion through `this.defaults` (the `?? scope.defaults.field` pattern), and deciding whether versioning becomes a posture default too. enforceSSL staying unconditional is fine — it's a security invariant, not a posture knob.

packages/bb-file-bucket/src/index.cdk.ts
soberm: Addressed in e929932. The data bucket now gets a noncurrent-version expiration lifecycle rule, default **90 days**, configurable via a new `noncurrentVersionExpirationDays` option, applied whenever versioning is enabled (the default). So versioning-on-by-default keeps recent-recovery without unbounded storage growth. The value is validated at synth (a non-positive or non-integer throws) regardless of the `versioned` setting; the rule itself is only emitted when versioning is on. README/DESIGN/changeset updated to document the 90-day cap.

packages/bb-file-bucket/src/types.ts
soberm: Acknowledged — understood and accepted, no code change. Keying the conditional option types on the literal `{ versioned: false }` means a widened `versioned: boolean` (or a dynamically-computed value) resolves to the versioned-aware typings even if the bucket ends up unversioned at runtime. We're accepting this residual deliberately: the default-on direction is the safer one (it matches the new runtime default), and narrowing a widened `boolean` back to a literal isn't expressible at the type level. The `versioned` JSDoc notes that the literal `false` is required to select the non-versioned typings. Leaving as-is.

packages/bb-file-bucket/src/index.cdk.ts
soberm: Addressed in e929932 — rebased onto latest main (picks up #385) and routed the posture knobs through `BlocksDefaults` using the `options?.field ?? this.defaults.field` contract, matching bb-kv-store:

- **accessLogging** → `options?.accessLogging ?? this.defaults.accessLogging`, so a stack posture that opts in enables logging on the file bucket without a per-block option.
- **logRetention** → removed the parallel per-block `logRetentionDays`; the access-log bucket's expiration now derives from `this.defaults.logRetention` (`RetentionDays`; sandbox `ONE_WEEK` / prod `ONE_YEAR`, with `INFINITE` keeping logs indefinitely).
- **removal** → dropped the legacy `tryGetContext('sandboxMode')` read; `removalPolicy`/`autoDeleteObjects` now fall back to `this.defaults.removalPolicy`. (S3 has no `deletionProtection` equivalent, so there's no deletion-protection knob to route — unlike DynamoDB.)
- **versioning** → kept `options?.versioned ?? true` (secure default-on) rather than a posture default: `BlocksDefaults` has no `versioned` field today, and adding one is a cross-package change to `@aws-blocks/core` that's out of scope here. The cost concern from making it default-on is instead bounded by the new `noncurrentVersionExpirationDays` (default 90). Happy to add a `BlocksDefaults.versioned` field in a follow-up if you'd like versioning to track posture.
- **enforceSSL** stays unconditional, as you noted.

osama-rizk (dismissed):

packages/bb-file-bucket/src/types.ts
```diff
@@ -10,12 +10,50 @@ import type { ChildLogger } from '@aws-blocks/bb-logger';
...
+	 *
+	 * Must be a positive integer; a non-positive or non-integer value is
+	 * rejected at synth. There is no "disable" sentinel — noncurrent-version
+	 * expiration is always on for a versioned bucket to cap version growth. To
+	 * opt out entirely, disable versioning (`versioned: false`), which drops
+	 * the rule. Ignored by the mock and browser runtimes (no AWS resource).
+	 */
+	noncurrentVersionExpirationDays?: number;
 	/** Wrap an existing S3 bucket instead of creating one. */
 	bucket?: ExternalBucketRef;
```
Simone319, line 68: When an external `bucket` is supplied, the constructor binds to it and returns early, so none of the secure defaults (`enforceSSL`, versioning, access logging, noncurrent-version expiry, CORS guard) apply — and nothing signals that to the caller. Since IDE autocomplete/hover surfaces this field's TSDoc, adding the caveat right there on `bucket` makes it visible to the customer at the moment they use the option (no deploy-time warning needed). Optionally echo it in the README.

packages/bb-file-bucket/src/index.mock.ts
```diff
@@ -96,7 +96,7 @@ export class FileBucket<O extends FileBucketOptions = FileBucketOptions> extends
 		if (!options?.bucket) validateBucketName(this.fullId);
 		this.log = options?.logger ?? new Logger(this, 'logger', { level: 'error' });
 		this.dataDir = getMockDataDir(this);
-		this.versioned = options?.versioned ?? false;
+		this.versioned = options?.versioned ?? true;
```
Simone319, line 146: The mock reads `versioned` but doesn't mirror the CDK's synth-time validation: `noncurrentVersionExpirationDays` isn't range-checked and there's no CORS wildcard + mutating-method guard here. That lets a unit test pass against the mock with config that fails at synth (a negative/zero expiry, or a wildcard `PUT` rule), which undercuts the mock's shift-left value. Consider mirroring both checks in the mock constructor.

packages/bb-file-bucket/src/types.ts
soberm: Done in 099ac9d. Added an `@remarks` caveat to the `bucket` option's TSDoc (so it surfaces on IDE hover) spelling out that when you wrap an existing bucket, FileBucket binds to it and returns early — none of the secure defaults (`enforceSSL`, versioning + noncurrent-version expiration, server access logging, `blockPublicAccess`, encryption, the wildcard-CORS guard) are applied; you own that bucket's security posture. Also echoed a short note in the README's "Wrapping an Existing Bucket" section.

packages/bb-file-bucket/src/index.mock.ts
soberm: Done in 099ac9d. The mock constructor now mirrors the CDK's two synth-time guards with identical messages and the same evaluation order (CORS wildcard + mutating-method guard, then the `noncurrentVersionExpirationDays` positive-integer check), gated the same way as the CDK path (skipped only when an external `bucket` is supplied; the noncurrent format check runs regardless of `versioned`). So a unit test against the mock now fails the same way `cdk synth` would for a negative/zero expiry or a wildcard `PUT`/`POST`/`DELETE` rule. Added mock tests covering the throw and allowed cases (including `versioned: false` still rejecting a malformed expiry) plus an order-parity test asserting the CORS error wins when both are violated.

osama-rizk (approved):

Simone319 (approved):

PR #488, opened 2026-09-03
fix(hosting): place CloudFront 5xx alarm in us-east-1 (#481)

### LINKED ISSUE #481: CloudFront 5xx alarm is created in the stack's region, where CloudFront metrics do not exist

@aws-blocks/core (the affected code is in `@aws-blocks/hosting`, which is not in the list)

### Describe the bug

`MonitoringConstruct` creates the `CloudFront5xxRate` alarm in the stack's region. `AWS/CloudFront` metrics are only published in us-east-1, and CloudWatch alarms cannot watch metrics in another region ([docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Cross-Account-Cross-Region.html)). For any stack outside us-east-1, the alarm never receives a datapoint.

The alarm uses `treatMissingData: NOT_BREACHING`, so it reports `OK` instead of `INSUFFICIENT_DATA`:

```
StateValue:            OK
StateUpdatedTimestamp: 2026-07-27T08:47:02Z   (= creation time, unchanged since)
StateReason:           no datapoints were received for 1 period and
                       1 missing datapoint was treated as [NonBreaching]
```

Related points:

- `monitoring.enabled` defaults to `true`, so every stack outside us-east-1 gets this alarm.
- Adding `region: 'us-east-1'` to the `Metric` is not a fix; aws-cdk-lib rejects it at synth (`Cannot create an Alarm in region X based on metric in region Y`). The alarm needs to live in a us-east-1 stack.
- `waf_construct.ts` handles the same constraint for CloudFront-scoped WAF by throwing `WafRegionError` on non-us-east-1 stacks. The alarm has no equivalent check.
- `monitoring_construct.test.ts` uses a us-west-2 stack and asserts the `Region: 'Global'` dimension. The dimension is correct, but the test does not cover which region the alarm is evaluated in, and a template assertion cannot detect that no metric will match.

### Expected behavior

The alarm is created in a region where its metric exists, or synth fails when that is not possible. An alarm that is never evaluated does not report `OK`.

### Reproduction steps

1. Deploy a hosting-enabled app to a region other than us-east-1 with default settings.
2. `aws cloudwatch describe-alarms --region <stack-region> --alarm-name-prefix <stack>-HostingMonitoringCloudFront5xxRate` → `StateValue: OK`, `StateUpdatedTimestamp` equals creation time.
3. `aws cloudwatch get-metric-statistics --region <stack-region> --namespace AWS/CloudFront --metric-name 5xxErrorRate --dimensions Name=DistributionId,Value=<id> Name=Region,Value=Global ...` → `Datapoints: []`
4. Same command with `--region us-east-1` → datapoints returned.

### Code snippet

```typescript
// packages/hosting/src/constructs/monitoring_construct.ts
const cf5xx = new Alarm(this, 'CloudFront5xxRate', {
  metric: new Metric({
    namespace: 'AWS/CloudFront',
    metricName: '5xxErrorRate',
    dimensionsMap: {
      DistributionId: props.distribution.distributionId,
      // CloudFront metrics live in us-east-1 regardless of stack.
      Region: 'Global',
    },
    ...
  }),
  treatMissingData: TreatMissingData.NOT_BREACHING,
  ...
});
```

### Environment

- `@aws-blocks/blocks` 0.2.1 (`core` 0.1.11, `hosting` 0.1.4). Verified by code inspection that `hosting` 0.2.0 (released 2026-09-01) has the same alarm definition.
- aws-cdk-lib 2.257.0, Node.js 26.7.0, stack region ap-northeast-1, static SPA + API (only this alarm is created)

### Additional context

**Possible fixes**

1. Create the CloudFront alarm in a us-east-1 stack (the repo already has one for Lambda@Edge, `edge-lambda-stack`).
2. Use `TreatMissingData.MISSING` for this alarm so it reports `INSUFFICIENT_DATA`, and throw at synth on non-us-east-1 stacks as `waf_construct.ts` does.
3. Document the limitation.

sarayev:
Fixes #481.

## Root cause
`AWS/CloudFront` metrics are published **only in us-east-1**, and a CloudWatch alarm can only evaluate a metric in its **own** region — confirmed by the [CloudWatch docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Cross-Account-Cross-Region.html) (*"Cross-Region functionality is not supported for alarms, so you cannot create an alarm in one Region that watches a metric in a different Region"*) and rejected by aws-cdk-lib at synth. `new Metric({ region: 'us-east-1' })` is honored for dashboards but throws for alarms.

`MonitoringConstruct` created the `CloudFront5xxRate` alarm in the hosting **stack's** region. For any stack outside us-east-1 the alarm never received a datapoint and, with `treatMissingData: NOT_BREACHING`, sat at `OK` forever instead of alarming — a monitoring alarm that silently never fires. `monitoring.enabled` defaults to `true`, so every off-region hosting stack was affected. (`waf_construct.ts` already handles the same CloudFront/us-east-1 constraint; the alarm had no equivalent.)

## Fix — two-topic design
Off-region, the CloudFront alarm is **deferred** to a dedicated, hosting-owned **us-east-1 support stack** (`UsEast1MonitoringStack`) that owns its **own** SNS topic. That topic's ARN is surfaced as `MonitoringTopicArnUsEast1` (an output **of the support stack**) for the operator to subscribe to, alongside the regional `MonitoringTopicArn`. No forwarder Lambda and no runtime message forwarding.

- Regional alarms (SSR error-rate, throttles, image, DLQ) are unchanged — their metrics are regional.
- **In-region (us-east-1) is unchanged**: single stack, alarm created locally.

## New prop
`monitoring.cloudFrontAlarm: 'usEast1Stack' (default) | 'skip'`
- `'usEast1Stack'`: create the support stack. **Requires `env: { account, region }`** off-region (a cross-region stack needs a concrete account). Env-agnostic off-region synth throws `HostingError('MonitoringEnvRequiredError')` with resolution guidance.
- `'skip'`: emit a synth warning, create no second stack (works env-agnostic).

## Breaking-ish notes for reviewers
- Default now **synthesizes a second CloudFormation stack** off-region and **requires `env: { account, region }`**.
- The us-east-1 alarm must reference the regional distribution id; CDK bridges that with its **standard cross-region export reader/writer custom resources** (added to both stacks automatically) — inherent to referencing a resource across regions, not our code.
- Fully env-agnostic stacks (region is an unresolved Token) keep today's behavior (alarm created locally); we can't decide region at synth. Documented limitation.

## Tests
- `monitoring_construct.test.ts`: deferral unit tests (`createCloudFrontAlarmLocally: false` → no CF alarm + `cloudFrontAlarmDeferred`, regional alarms still created).
- New `hosting_construct.cf_alarm_region.test.ts`: off-region two-stack synth (CF alarm in the us-east-1 stack + `MonitoringTopicArnUsEast1` output, one SNS topic, no forwarder subscription); env-agnostic off-region → `MonitoringEnvRequiredError`; `'skip'` mode warning + no second stack; in-region single-stack with local alarm.
- **All 853 hosting tests pass.**

Do not merge — opened for review.

### DIFF
--- a/.changeset/cf-alarm-us-east-1-off-region.md
+++ b/.changeset/cf-alarm-us-east-1-off-region.md
@@ -1,0 +1,23 @@
+---
+"@aws-blocks/hosting": minor
+"@aws-blocks/core": patch
+---
+
+Fix off-region CloudFront alarm placement (#481).
+
+`HostingConstruct` now correctly handles the AWS constraint that
+CloudFront metrics only publish in `us-east-1`: when the hosting stack
+is deployed to a non-`us-east-1` region the CloudFront 5xx alarm is
+placed in a synthesized `<stackName>-CfMonitoring` us-east-1 stack that
+owns its own encrypted SNS topic and exposes a `MonitoringTopicArnUsEast1`
+CloudFormation output for operator subscriptions.
+
+Off-region deployments now defer the CloudFront 5xx alarm to a synthesized
+`<stackName>-CfMonitoring` us-east-1 stack (own SNS topic +
+MonitoringTopicArnUsEast1 output); this requires env:{account,region} to
+be set off-region. Set monitoring.cloudFrontAlarm:'skip' to opt out (emits
+a warning).
+
+`@aws-blocks/core` gains the `cloudFrontAlarm` property on
+`HostingMonitoringOptions` to surface the new `'skip'` | `'usEast1Stack'`
+choice at the L3 config layer.
--- a/packages/core/src/hosting.ts
+++ b/packages/core/src/hosting.ts
@@ -702,10 +702,9 @@ export class Hosting extends Construct {
       monitoring: props.monitoring
         ? {
             ...props.monitoring,
-            // Forward cloudFrontAlarm through as-is: undefined inherits the
-            // L3 default ('usEast1Stack'). To change the Blocks-app default
-            // to 'skip' later, replace the fallback on the next line.
-            cloudFrontAlarm: props.monitoring.cloudFrontAlarm ?? undefined,
+            // cloudFrontAlarm flows through the spread as-is: undefined inherits
+            // the L3 default ('usEast1Stack'). To change the Blocks-app default
+            // to 'skip' later, set it explicitly here.
           }
         : undefined,
       skewProtection: props.skewProtection,
--- a/packages/core/src/hosting.ts
+++ b/packages/core/src/hosting.ts
@@ -382,6 +382,21 @@ export interface HostingProps {
     enabled?: boolean;
     /** ARN of an existing SNS topic to send alarm actions to. */
     snsTopicArn?: string;
+    /**
+     * How to handle the CloudFront 5xx alarm when the app is deployed
+     * outside us-east-1. AWS/CloudFront metrics only exist in us-east-1
+     * and a CloudWatch alarm can't watch a metric cross-region, so the
+     * alarm cannot live in an off-region stack.
+     *  - `'usEast1Stack'`: place the alarm in a dedicated us-east-1
+     *    support stack (`<stackName>-CfMonitoring`) with its own SNS
+     *    topic, surfaced as the `MonitoringTopicArnUsEast1` output.
+     *    Requires `env: { account, region }` on the stack.
+     *  - `'skip'`: omit the CloudFront alarm off-region (emit a warning).
+     * Ignored when the stack is already in us-east-1.
+     *
+     * When omitted, inherits the underlying default (`'usEast1Stack'`).
+     */
+    cloudFrontAlarm?: 'usEast1Stack' | 'skip';
   };
 
   /**
@@ -684,7 +699,15 @@ export class Hosting extends Construct {
       logging: props.logging,
       buildCache: props.buildCache,
       errorPages: skipPropsErrorPages ? undefined : props.errorPages,
-      monitoring: props.monitoring,
+      monitoring: props.monitoring
+        ? {
+            ...props.monitoring,
+            // Forward cloudFrontAlarm through as-is: undefined inherits the
+            // L3 default ('usEast1Stack'). To change the Blocks-app default
+            // to 'skip' later, replace the fallback on the next line.
+            cloudFrontAlarm: props.monitoring.cloudFrontAlarm ?? undefined,
+          }
+        : undefined,
       skewProtection: props.skewProtection,
     };
 
--- a/packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
+++ b/packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
@@ -9,13 +9,13 @@
  */
 import { afterEach, describe, it } from 'node:test';
 import assert from 'node:assert';
-import * as fs from 'fs';
-import * as path from 'path';
-import * as os from 'os';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+import * as os from 'node:os';
 import { App, Stack } from 'aws-cdk-lib';
 import { Annotations, Match, Template } from 'aws-cdk-lib/assertions';
 import { HostingConstruct } from './hosting_construct.js';
-import { DeployManifest } from '../manifest/types.js';
+import type { DeployManifest } from '../manifest/types.js';
 import { HostingError } from '../hosting_error.js';
 
 let tmpDir: string;
--- a/packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
+++ b/packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
@@ -1,0 +1,145 @@
+/**
+ * HostingConstruct — CloudFront alarm region wiring (issue #481).
+ *
+ * AWS/CloudFront metrics only publish in us-east-1 and a CloudWatch
+ * alarm can only evaluate a metric in its own region. These tests cover
+ * the two-topic fix: off-region, the CloudFront 5xx alarm is placed in a
+ * hosting-owned us-east-1 support stack with its own SNS topic; in
+ * us-east-1 the behavior is unchanged (single stack, alarm local).
+ */
+import { afterEach, describe, it } from 'node:test';
+import assert from 'node:assert';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { App, Stack } from 'aws-cdk-lib';
+import { Annotations, Match, Template } from 'aws-cdk-lib/assertions';
+import { HostingConstruct } from './hosting_construct.js';
+import { DeployManifest } from '../manifest/types.js';
+import { HostingError } from '../hosting_error.js';
+
+let tmpDir: string;
+
+const createStaticDir = (): string => {
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'hosting-cfregion-'));
+  fs.writeFileSync(path.join(tmpDir, 'index.html'), '<html></html>');
+  return tmpDir;
+};
+
+const spaManifest = (staticDir: string): DeployManifest => ({
+  version: 1,
+  compute: {},
+  staticAssets: { directory: staticDir },
+  routes: [{ pattern: '/*', target: 'static' }],
+  buildId: 'cfregion-test-1',
+});
+
+const CF_ALARM = Match.objectLike({ Namespace: 'AWS/CloudFront' });
+
+void describe('HostingConstruct — CloudFront alarm region (#481)', () => {
+  afterEach(() => {
+    if (tmpDir) fs.rmSync(tmpDir, { recursive: true, force: true });
+  });
+
+  // ---- (iv) In-region: unchanged single-stack behavior ----
+  void it('creates the CloudFront alarm locally in a single stack when region is us-east-1', () => {
+    const staticDir = createStaticDir();
+    const app = new App();
+    const stack = new Stack(app, 'TestStack', {
+      env: { account: '123456789012', region: 'us-east-1' },
+    });
+    new HostingConstruct(stack, 'Hosting', { manifest: spaManifest(staticDir) });
+
+    const template = Template.fromStack(stack);
+    // Alarm lives in this (us-east-1) stack.
+    template.resourcePropertiesCountIs('AWS::CloudWatch::Alarm', CF_ALARM, 1);
+    // No second stack was synthesized.
+    assert.strictEqual(
+      app.node.tryFindChild('TestStack-CfMonitoring'),
+      undefined,
+    );
+  });
+
+  // ---- (i) Off-region: two stacks, CF alarm in us-east-1 stack ----
+  void it('places the CloudFront alarm in a us-east-1 support stack when off-region', () => {
+    const staticDir = createStaticDir();
+    const app = new App();
+    const stack = new Stack(app, 'TestStack', {
+      env: { account: '123456789012', region: 'ap-northeast-1' },
+    });
+    new HostingConstruct(stack, 'Hosting', { manifest: spaManifest(staticDir) });
+
+    // The regional stack has NO CloudFront alarm...
+    const regional = Template.fromStack(stack);
+    regional.resourcePropertiesCountIs('AWS::CloudWatch::Alarm', CF_ALARM, 0);
+
+    // A sibling us-east-1 support stack exists and holds the alarm.
+    const support = app.node.tryFindChild('TestStack-CfMonitoring') as Stack;
+    assert.ok(support, 'expected a TestStack-CfMonitoring support stack');
+    assert.strictEqual(support.region, 'us-east-1');
+
+    const supportTemplate = Template.fromStack(support);
+    supportTemplate.resourcePropertiesCountIs(
+      'AWS::CloudWatch::Alarm',
+      Match.objectLike({
+        Namespace: 'AWS/CloudFront',
+        MetricName: '5xxErrorRate',
+        Threshold: 5,
+        TreatMissingData: 'notBreaching',
+      }),
+      1,
+    );
+    // Its own SNS topic (two-topic design). No forwarder subscription —
+    // consolidation is via two topics, not runtime message forwarding.
+    supportTemplate.resourceCountIs('AWS::SNS::Topic', 1);
+    supportTemplate.resourceCountIs('AWS::SNS::Subscription', 0);
+    // The us-east-1 topic ARN is surfaced as an output of the support stack.
+    supportTemplate.hasOutput(
+      '*',
+      Match.objectLike({ Description: Match.stringLikeRegexp('us-east-1') }),
+    );
+  });
+
+  // ---- (iii) 'skip' mode: warning, no second stack ----
+  void it("emits a warning and creates no support stack when cloudFrontAlarm is 'skip'", () => {
+    const staticDir = createStaticDir();
+    const app = new App();
+    const stack = new Stack(app, 'TestStack', {
+      env: { account: '123456789012', region: 'ap-northeast-1' },
+    });
+    new HostingConstruct(stack, 'Hosting', {
+      manifest: spaManifest(staticDir),
+      monitoring: { cloudFrontAlarm: 'skip' },
+    });
+
+    const template = Template.fromStack(stack);
+    template.resourcePropertiesCountIs('AWS::CloudWatch::Alarm', CF_ALARM, 0);
+    assert.strictEqual(
+      app.node.tryFindChild('TestStack-CfMonitoring'),
+      undefined,
+    );
+    Annotations.fromStack(stack).hasWarning(
+      '*',
+      Match.stringLikeRegexp('CloudFront 5xx alarm skipped'),
+    );
+  });
+
+  // ---- (ii) Off-region without a concrete account → throws ----
+  void it('throws MonitoringEnvRequiredError off-region when the account is unresolved', () => {
+    const staticDir = createStaticDir();
+    const app = new App();
+    // region resolved (off-region) but account left unresolved.
+    const stack = new Stack(app, 'TestStack', {
+      env: { region: 'ap-northeast-1' },
+    });
+    assert.throws(
+      () =>
+        new HostingConstruct(stack, 'Hosting', {
+          manifest: spaManifest(staticDir),
+        }),
+      (err: unknown) =>
+        err instanceof HostingError &&
+        err.code === 'MonitoringEnvRequiredError',
+    );
+  });
+});
--- a/packages/hosting/src/constructs/hosting_construct.ts
+++ b/packages/hosting/src/constructs/hosting_construct.ts
@@ -10,6 +10,8 @@ import {
   RemovalPolicy,
   Size,
   Stack,
+  Stage,
+  Token,
 } from 'aws-cdk-lib';
 import type { ICertificate } from 'aws-cdk-lib/aws-certificatemanager';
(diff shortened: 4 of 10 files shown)


packages/hosting/src/constructs/us_east_1_monitoring_stack.ts
```diff
@@ -0,0 +1,113 @@
...
+import { type ITopic, Topic } from 'aws-cdk-lib/aws-sns';
+import { createAlarmTopicKey } from './monitoring_construct.js';
+
+/**
+ * Props for {@link UsEast1MonitoringStack}.
+ */
+export type UsEast1MonitoringStackProps = StackProps & {
+  /**
+   * CloudFront distribution id to alarm on. Passed as a plain string so
+   * this us-east-1 stack takes no cross-region CDK reference on the
```
soberm, line 20: The doc says `distributionId` is 'passed as a plain string so this us-east-1 stack takes no cross-region CDK reference.' That's not what happens on the default path: the caller passes `this.distribution.distributionId`, an unresolved cross-stack token, and CDK *does* create cross-region export machinery (`CrossRegionExportWriter` in the regional stack, `ExportsReader` `Fn::GetAtt` here). Please reword to say this creates a cross-region reference that CDK bridges via export writer/reader custom resources. The 'plain string' escape hatch is only real when a caller supplies a concrete id.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -1167,6 +1191,63 @@ export class HostingConstruct extends Construct {
...
+              message:
+                `Cannot create the us-east-1 CloudFront monitoring stack: no ` +
+                `enclosing App/Stage was found for this construct.`,
+              resolution:
+                `Instantiate hosting within a CDK App (or Stage), or set ` +
+                `monitoring.cloudFrontAlarm: 'skip'.`,
+            });
+          }
+          // The us-east-1 CloudFront alarm must reference the regional
+          // distribution's id; CDK bridges that with its standard
```
soberm, line 1236: This comment ('CDK bridges that with its standard cross-region export reader/writer custom resources ... automatically') is correct, but it contradicts the `distributionId` prop doc in `us_east_1_monitoring_stack.ts` ('no cross-region reference'). Please reconcile the two so the same story is told in both places.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -1167,6 +1191,63 @@ export class HostingConstruct extends Construct {
...
+          // cross-region export reader/writer custom resources (added to
+          // both stacks automatically). The topic ARN is surfaced as an
+          // output OF the support stack (MonitoringTopicArnUsEast1) — not
+          // re-output here, to keep the wiring one-directional.
+          new UsEast1MonitoringStack(
+            stage,
+            `${hostingStack.stackName}-CfMonitoring`,
```
soberm, line 1243: `${hostingStack.stackName}-CfMonitoring` collides if two `HostingConstruct`s are instantiated in the same stack (duplicate construct id under the same Stage). Since `HostingConstruct` is an embeddable L3, consider folding the construct's node id/addr into the support-stack id to keep it unique.

packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    // A sibling us-east-1 support stack exists and holds the alarm.
+    const support = app.node.tryFindChild('TestStack-CfMonitoring') as Stack;
+    assert.ok(support, 'expected a TestStack-CfMonitoring support stack');
+    assert.strictEqual(support.region, 'us-east-1');
+
+    const supportTemplate = Template.fromStack(support);
+    supportTemplate.resourcePropertiesCountIs(
+      'AWS::CloudWatch::Alarm',
+      Match.objectLike({
+        Namespace: 'AWS/CloudFront',
```
soberm, line 85: This asserts only Namespace/MetricName/Threshold/TreatMissingData. Since the point is that the re-homed alarm is identical to the regional one, please also assert `dimensionsMap` (`Region: 'Global'` + `DistributionId`), `ComparisonOperator`, `Period`, `Statistic`, and `EvaluationPeriods` to lock parity against future drift.

packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
```diff
@@ -0,0 +1,145 @@
...
+      1,
+    );
+    // Its own SNS topic (two-topic design). No forwarder subscription —
+    // consolidation is via two topics, not runtime message forwarding.
+    supportTemplate.resourceCountIs('AWS::SNS::Topic', 1);
+    supportTemplate.resourceCountIs('AWS::SNS::Subscription', 0);
+    // The us-east-1 topic ARN is surfaced as an output of the support stack.
+    supportTemplate.hasOutput(
+      '*',
+      Match.objectLike({ Description: Match.stringLikeRegexp('us-east-1') }),
```
soberm, line 99: `hasOutput('*', { Description: /us-east-1/ })` matches on the description, but docs and operators depend on the logical id `MonitoringTopicArnUsEast1`. Assert `hasOutput('MonitoringTopicArnUsEast1', ...)` so a rename can't silently pass.

packages/core/src/hosting.ts
```diff
@@ -684,7 +699,14 @@ export class Hosting extends Construct {
       logging: props.logging,
       buildCache: props.buildCache,
       errorPages: skipPropsErrorPages ? undefined : props.errorPages,
-      monitoring: props.monitoring,
+      monitoring: props.monitoring
+        ? {
+            ...props.monitoring,
```
soberm, line 704: Minor: `{ ...props.monitoring }` is a functional no-op here — `cloudFrontAlarm` already flowed through the previous `monitoring: props.monitoring`. Fine to keep for the anchoring comment, but the `@aws-blocks/core: patch` changeset shouldn't imply core behavior changed; it's just surfacing the type.

.changeset/cf-alarm-us-east-1-off-region.md
```diff
@@ -0,0 +1,23 @@
...
+Fix off-region CloudFront alarm placement (#481).
+
+`HostingConstruct` now correctly handles the AWS constraint that
+CloudFront metrics only publish in `us-east-1`: when the hosting stack
+is deployed to a non-`us-east-1` region the CloudFront 5xx alarm is
+placed in a synthesized `<stackName>-CfMonitoring` us-east-1 stack that
+owns its own encrypted SNS topic and exposes a `MonitoringTopicArnUsEast1`
+CloudFormation output for operator subscriptions.
+
+Off-region deployments now defer the CloudFront 5xx alarm to a synthesized
```
soberm, line 15: Two things: (1) the off-region explanation is repeated in two adjacent paragraphs — collapse to one. (2) Please state explicitly that this is a **breaking change** (new default synthesizes a second stack off-region and throws `MonitoringEnvRequiredError` for env-agnostic off-region stacks) so it reads that way in the changelog. The `minor` bump is the right label for a breaking change at 0.x.

soberm (commented):
Nice catch on the root cause. Off-region, the old `CloudFront5xxRate` alarm sat at `OK` forever (metric only in us-east-1 + `treatMissingData: NOT_BREACHING`) — a monitoring alarm that silently never fired. The deferral to a hosting-owned us-east-1 stack with its own topic is a clean approach, region/account gating reads correctly, and alarm-config parity with the original is exact.

I built the branch (Node 22, aws-cdk-lib 2.257.0) and ran `hosting_construct.cf_alarm_region.test.ts` — build green, 4/4 pass. Worth noting for future readers: the us-east-1 alarm references `this.distribution.distributionId` (an unresolved cross-stack **token**), and CDK auto-bridges that with `CrossRegionExportWriter`/`ExportsReader` custom resources on this CDK version *without* `crossRegionReferences: true`. That works — but it means the code comments claiming "plain string / no cross-region reference" are inaccurate (see inline).

The one thing I'd like a maintainer call on before merge: **is breaking-by-default the behavior we want?** The new `'usEast1Stack'` default makes a previously-synthesizing env-agnostic off-region stack now throw `MonitoringEnvRequiredError`. The `minor` changeset is the correct label for that (0.x → minor = breaking), so semver is fine — this is purely a "do we want the default to break, or default to `'skip'`?" product decision. Please also make the changeset state plainly that this is a breaking change requiring `env` off-region.

Remaining items are all non-blocking suggestions/nits inline. Recommend an off-region E2E destroy→deploy pass to confirm the new cross-region custom resources don't hit the known "cannot delete export in use" / CREATE-vs-`--revert-drift` issues.

svidgen (changes requested):
Small design with API BR review please. Details in slack.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -1140,33 +1163,97 @@ export class HostingConstruct extends Construct {
...
+
+      // Off-region: place the CloudFront alarm in a us-east-1 support
+      // stack, always (no opt-out). The same subscriptions are applied to
+      // its topic so callers subscribe in one place and both topics are
+      // covered.
+      if (this.distribution && monitoring.cloudFrontAlarmDeferred) {
+        if (Token.isUnresolved(hostingStack.account)) {
+          // A cross-region stack needs a concrete account. Fail loud
+          // rather than silently drop CloudFront monitoring.
+          throw new HostingError('MonitoringEnvRequiredError', {
```
soberm: [major] The rework removed all opt-out, and this is likely the crux of the standing change request. An off-region stack with a resolved region but an ambient (unresolved) account — a legitimate single-synth, multi-account pipeline shape — now hard-throws `MonitoringEnvRequiredError` here, with `monitoring.enabled: false` the only recourse (which also drops the working regional SSR/DLQ alarms). The old warned `'skip'` mode preserved that path without being silent. Consider restoring a warned escape hatch (warn-and-skip the CF alarm) rather than removing opt-out entirely — a loud warning isn't the silent-alarm bug #481 is about.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -1140,33 +1163,97 @@ export class HostingConstruct extends Construct {
...
         enabled: true,
-        snsTopic: userTopic,
+        subscriptions,
         distribution: this.distribution,
         // Lambda@Edge functions don't accept the CW metric helpers we
         // use; only attach when the SSR compute is a regional Lambda.
         ssrFunction: ssrFn instanceof LambdaFunction ? ssrFn : undefined,
         imageFunction: imgFn instanceof LambdaFunction ? imgFn : undefined,
         revalidationDlq: this.revalidationDlq,
+        createCloudFrontAlarmLocally: !offRegion,
```
soberm, line 1215: [minor] Fully env-agnostic stacks (region is an unresolved Token) still fall through to `createCloudFrontAlarmLocally: true` and get a local CloudFront alarm that silently never fires off-region — the exact #481 bug. Given the PR's stated "always-on, no opt-out" guarantee, this gap deserves an explicit synth-time warning here (region unresolved → we can't place the alarm correctly), rather than a silent local alarm.

packages/core/src/hosting.ts
```diff
@@ -684,7 +701,12 @@ export class Hosting extends Construct {
...
```
soberm: [minor] `{ ...props.monitoring }` is still a functional no-op — a shallow copy that behaves identically to passing `props.monitoring` through (as the prior revision did). Only the comment changed. Either drop the spread (`monitoring: props.monitoring`) or note that it exists purely to anchor the comment; the `@aws-blocks/core` changeset shouldn't imply a behavior change in core here beyond the `monitoringTopic` → `monitoring` surface rename.

packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
```diff
@@ -0,0 +1,180 @@
...
+      env: { account: '123456789012', region: 'us-east-1' },
+    });
+    new HostingConstruct(stack, 'Hosting', { manifest: spaManifest(staticDir) });
+
+    const template = Template.fromStack(stack);
+    // Alarm lives in this (us-east-1) stack.
+    template.resourcePropertiesCountIs('AWS::CloudWatch::Alarm', CF_ALARM, 1);
+    // No second stack was synthesized.
+    assert.strictEqual(
+      app.node.tryFindChild('TestStack-CfMonitoring'),
```
soberm: [minor] This negative assertion can never match the now-suffixed support-stack id (`${stackName}-CfMonitoring-${addr}`), so it passes even if a support stack were wrongly synthesized in-region. Use the same prefix scan the off-region test uses and assert none is found, e.g. `assert.ok(!app.node.children.some(c => c instanceof Stack && c.node.id.startsWith('TestStack-CfMonitoring')))`.

packages/hosting/src/constructs/hosting_construct.ts
```diff
@@ -31,7 +33,12 @@ import type { IHostedZone } from 'aws-cdk-lib/aws-route53';
 import * as s3 from 'aws-cdk-lib/aws-s3';
 import { Bucket } from 'aws-cdk-lib/aws-s3';
 import { BucketDeployment, CacheControl, Source } from 'aws-cdk-lib/aws-s3-deployment';
-import { type ITopic, Topic } from 'aws-cdk-lib/aws-sns';
+import { type ITopic } from 'aws-cdk-lib/aws-sns';
+import type {
+  EmailSubscription,
+  UrlSubscription,
+} from 'aws-cdk-lib/aws-sns-subscriptions';
+import { Alarm } from 'aws-cdk-lib/aws-cloudwatch';
```
soberm: [nit] `Alarm` is imported as a value but used only in type positions (the `monitoring.alarms: Alarm[]` field type). Prefer `import type { Alarm }` to satisfy Biome's `useImportType` and avoid a runtime import.

.changeset/cf-alarm-us-east-1-off-region.md
```diff
@@ -0,0 +1,37 @@
...
+cannot watch a metric cross-region, so off-region the CloudFront 5xx alarm
+never fired (it sat at `OK` under `treatMissingData: NOT_BREACHING`).
+Off-region deployments now **always** place the CloudFront alarm in a
+synthesized `<stackName>-CfMonitoring-<addr>` us-east-1 stack with its own
+encrypted SNS topic; `monitoring.subscriptions` are applied to that topic
+too. This requires `env: { account, region }` on off-region stacks —
+env-agnostic off-region synth throws `MonitoringEnvRequiredError`. There is
+no opt-out.
+
+Migration: replace `monitoring: { snsTopicArn }` with
```
soberm, line 33: [minor] The migration note covers the `snsTopicArn` → `subscriptions` swap but omits the escape paths for the two capabilities this removes: (1) a central/BYO topic is now reached via `hosting.monitoring.alarms.forEach(a => a.addAlarmAction(new SnsAction(myTopic)))`; (2) resource-target (Lambda/SQS) subscriptions are still valid on the regional topic via `hosting.monitoring.alarmTopics[].addSubscription(...)` — only the cross-region us-east-1 auto-fan-out drops them. Adding a line each would make the migration self-serve.

packages/hosting/src/constructs/us_east_1_monitoring_stack.ts
```diff
@@ -0,0 +1,118 @@
...
+ * plumbing, no forwarder Lambda). The parent applies the same
+ * `subscriptions` list to this topic as to the regional one, so callers
+ * subscribe in one place and both topics are covered. The topic ARN is
+ * also surfaced as the `MonitoringTopicArnUsEast1` output for operators
+ * who want the raw ARN.
+ */
+export class UsEast1MonitoringStack extends Stack {
+  /** The us-east-1 topic the CloudFront alarm publishes to. */
+  readonly topic: ITopic;
+  /** KMS key encrypting the auto-created topic (undefined for a BYO topic). */
```
soberm: [nit] "(undefined for a BYO topic)" — there's no BYO-topic path in this stack (it always creates its own encrypted topic), so `encryptionKey` is never undefined here. Drop the parenthetical.

packages/hosting/src/constructs/hosting_construct.ts
sarayev: Good call, fixed in 1140ba05. I agree the hard throw was too blunt for the ambient-account pipeline shape, and it took down the working regional SSR/image/DLQ alarms as collateral. Off-region with an unresolved account now skips *only* the CloudFront alarm and emits a loud synth warning (`@aws-blocks/hosting:CloudFrontAlarmSkippedNoAccount`) instead of throwing; all other alarms are preserved. A visible warning is not the silent-alarm failure #481 is about: the operator is told exactly what is missing and how to get it (`env: { account, region }`). Kept "always-on / no opt-out" as the default, but this narrow warn-and-skip covers the case where the us-east-1 support stack genuinely cannot be synthesized (CDK cross-region synth needs a concrete account). Recorded the reasoning in docs/DECISIONS.md D-016.

packages/hosting/src/constructs/hosting_construct.ts
sarayev: Fixed in 1140ba05. When the region is an unresolved token we now emit a synth warning (`@aws-blocks/hosting:CloudFrontAlarmRegionUnresolved`) noting the CloudFront alarm is created locally and will never fire if the app deploys outside us-east-1, with the fix (set `env: { account, region }`). We still create it locally as best-effort since we cannot know the target region at synth, but it is no longer silent.

packages/core/src/hosting.ts
sarayev: Fixed in 1140ba05: dropped the spread, now `monitoring: props.monitoring` passes straight through. The core changeset is scoped to the `monitoringTopic` to `monitoring` surface rename on the output side, not a behavior change in the passthrough.

packages/hosting/src/constructs/hosting_construct.cf_alarm_region.test.ts
sarayev: Fixed in 1140ba05: the in-region test now scans by prefix (`!app.node.children.some(c => c instanceof Stack && c.node.id.startsWith("TestStack-CfMonitoring"))`) so it actually fails if a support stack were wrongly synthesized in-region, matching the off-region test.

.changeset/cf-alarm-us-east-1-off-region.md
sarayev: Fixed in 1140ba05: the changeset migration note now documents both escape paths. (1) Route to a central/BYO topic via `hosting.monitoring.alarms.forEach(a => a.addAlarmAction(new SnsAction(myTopic)))`. (2) Lambda/SQS subscriptions remain valid on the regional topic via `hosting.monitoring.alarmTopics[].addSubscription(...)`; only the cross-region us-east-1 auto-fan-out drops them. Also noted the warn-and-skip behavior there.

packages/hosting/src/constructs/hosting_construct.ts
sarayev: Fixed in 1140ba05: changed to `import type { Alarm }`.

packages/hosting/src/constructs/us_east_1_monitoring_stack.ts
sarayev: Fixed in 1140ba05: dropped the "(undefined for a BYO topic)" parenthetical. There is no BYO path here; the stack always creates its own encrypted topic, so the key is always set.

soberm (approved):

svidgen (approved):

PR #489, opened 2026-09-03
feat(core): revert drift on production cdk deploy

soberm:
## Problem

The `cdk watch`/hotswap dev loop mutates resources directly via service APIs, leaving the deployed stack drifted from its CloudFormation template. A subsequent full production deploy did not reconcile that drift.

**Issue #, if available:** Addresses bug-bash item `[Bug Bash] scaffold: cdk watch sandbox accumulates drift (no --revert-drift)`.

## Changes

Adds `--revert-drift` to the production `cdk deploy` argv built by `buildCdkDeployArgs` in `packages/core/src/scripts/deploy-stream.ts`, so a full production deploy reconciles configuration drift (e.g. from the `cdk watch`/hotswap dev loop) against the CloudFormation template. The flag is available in the pinned CDK CLI (`^2.1138.0`).

**Scope: production deploy path ONLY.** The sandbox/dev-loop path is intentionally left unchanged — it is throwaway and deliberately relies on hotswap drift. This flag applies only to `cdk deploy`; it cannot apply to the `cdk watch` loop itself, which is why the fix lives on the production deploy path rather than the sandbox.

Includes a `@aws-blocks/core` patch changeset.

## Validation

- Added a presence test asserting `--revert-drift` is present in the production deploy args.
- Added a negative guard test asserting the flag does NOT leak into the sandbox/dev-loop args.
- Build passes; the relevant tests pass.

Note: 4 pre-existing, unrelated `ERR_MODULE_NOT_FOUND` subprocess-wrapper test failures exist on baseline `main` and are untouched by this change.

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (and PR referenced) — N/A, no user-facing docs affected by this internal deploy-argv change

---


soberm:
Closing this PR. On review, --revert-drift on the production deploy path doesn't address the problem the bug bash item raised, and it introduced fragility that isn't worth carrying.

PR #490, opened 2026-09-03
feat(bb-kv-store): compose ifNotExists + ifValueEquals on put (OR)

osama-rizk:
## Card

Bug Bash board (Starter Kit, DX P3): **"bb-kv-store: conditional write can't compose `ifValueEquals` with `ifNotExists`."**
🔗 https://github.com/orgs/aws-amplify/projects/141/views/15?pane=issue&itemId=195533991

## Problem

`put`'s two conditional-write options were **mutually exclusive**, and the layers disagreed:
- **AWS** (`index.aws.ts`) used `if (ifNotExists) … else if (ifValueEquals) …` — so `ifValueEquals` was **silently ignored** when both were set.
- **Mock** (`index.mock.ts`) applied both as separate AND checks — so passing both **always failed** (a new key fails the value check; an existing key fails the not-exists check).

So the useful "create-or-update" conditional was unusable, and mock/deploy diverged.

## Fix

`ifNotExists` and `ifValueEquals` now **compose with OR** on both layers:

```
attribute_not_exists(pk) OR value = :expected
```

The write succeeds when the key is **absent OR its current value matches**, and fails only when the key **exists AND the value differs** — the optimistic *"create it, or update it only if unchanged"* pattern. (AND would be unsatisfiable: a key can't be both absent and hold a matching value.) Each option used **alone is unchanged**.

- `index.aws.ts` — build the `ConditionExpression` from the set options, joined with `OR`.
- `index.mock.ts` — `succeed if (absent) OR (value matches)`, matching the AWS expression exactly.

## Testing

New mock tests (`index.test.ts`): both-set on a **new** key writes (absent branch); both-set on an **existing matching** key updates (value branch); both-set on an **existing differing** key throws `ConditionalCheckFailed`. Existing single-option tests unchanged.

`@aws-blocks/bb-kv-store` suite **102 pass / 0 fail**.

## Verification

`npm run lint` 0 errors · umbrella-changeset guard ✅. No public API surface change (both options already existed); docs updated (README example + `ConditionalWriteOptions` JSDoc).

## Changeset

`@aws-blocks/bb-kv-store` **minor** (new capability) + `@aws-blocks/blocks` **patch** (umbrella re-export).


osama-rizk:
Thanks — all five addressed (pushed).

**[should-fix] `@throws` JSDoc (both `index.aws.ts` and `index.mock.ts`):** the two single-option lines now say "(alone)", and a third line documents the combined case — *"If BOTH `ifNotExists` and `ifValueEquals` are set (they compose with OR), only when the key exists AND its current value does not match."* So the hover doc matches actual behavior.

**[should-fix] AWS parity test:** added a `describe('conditional-write composition (AWS PutCommand shape)')` block in `parity.test.ts` using the existing `captureAws` intercept. It asserts the real `PutCommand` input for: both conditions → `ConditionExpression === 'attribute_not_exists(#pk) OR #value = :expected'` with `ExpressionAttributeNames === { '#pk':'pk', '#value':'value' }` and `ExpressionAttributeValues === { ':expected': JSON.stringify('prev') }`; plus the single-option shapes and the no-condition case. A typo in the `join(' OR ')` or a missing name/value now fails a test.

**[minor] DESIGN.md:** added a "Conditional-write composition" row to the Mock vs AWS table, noting both runtimes compose with OR and that this was previously a divergence.

**[nit] `assert.equal` → `assert.strictEqual`** on the two compose tests (lines 73/80).

`@aws-blocks/bb-kv-store` suite **106 pass / 0 fail** (was 102), lint ✅, rebased on `main`.

packages/bb-kv-store/src/index.aws.ts
```diff
@@ -103,13 +104,26 @@ export class KVStore<T = string> extends Scope {
...
-			command.ExpressionAttributeNames = { '#value': 'value' };
-			command.ExpressionAttributeValues = { ':expected': JSON.stringify(options.ifValueEquals) };
+			conditions.push('attribute_not_exists(#pk)');
+			names['#pk'] = 'pk';
+		}
+		if (options && 'ifValueEquals' in options) {
+			conditions.push('#value = :expected');
+			names['#value'] = 'value';
+			values[':expected'] = JSON.stringify(options.ifValueEquals);
+		}
```
soberm: **MEDIUM — AWS/mock predicate divergence on explicit `undefined`.**

The AWS layer detects `ifValueEquals` with `'ifValueEquals' in options`, while the mock (`index.mock.ts:159`) uses `options?.ifValueEquals !== undefined`. These diverge on `{ ifValueEquals: undefined }`: the mock treats it as a no-op (the documented invariant `parity.test.ts:83-124` tests), but here `'ifValueEquals' in options` is `true`, so `values[':expected'] = JSON.stringify(undefined)` (the JS value `undefined`) is set and the SDK errors while marshalling. With the new OR path, `{ ifNotExists: true, ifValueEquals: undefined }` behaves as plain `ifNotExists` on the mock but emits `attribute_not_exists(#pk) OR #value = :expected` with an undefined value on AWS — a layer-parity gap, and it's untested (the existing 'undefined is no-op' block only exercises the mock). The predicate pre-dates this PR, but this change touches the exact code and can align it.

Fix: gate the value branch on `!== undefined` to match the mock, so explicit-`undefined` is a no-op on both layers.

```suggestion
		if (options?.ifValueEquals !== undefined) {
			conditions.push('#value = :expected');
			names['#value'] = 'value';
			values[':expected'] = JSON.stringify(options.ifValueEquals);
		}
```

packages/bb-kv-store/src/parity.test.ts
```diff
@@ -247,3 +247,46 @@ describe('TTL parity between mock and AWS runtimes', () => {
 		}
 	});
 });
+
+// Conditional-write composition parity: the AWS runtime must emit an OR-composed
+// ConditionExpression when both ifNotExists and ifValueEquals are set, matching
+// the mock's behavior. A typo in the join or a missing name/value would slip
+// past the mock-only tests.
+describe('conditional-write composition (AWS PutCommand shape)', () => {
```
soberm, line 256: **LOW — missing AWS parity assertion for `{ ifValueEquals: undefined }`.**

The new AWS composition parity block covers the compose/single/none cases but not `{ ifValueEquals: undefined }` — exactly the case where mock and AWS can drift (see the AWS predicate comment). Add a case asserting `store.put('k','v',{ ifValueEquals: undefined })` produces `ConditionExpression === undefined` (no value attrs), pinning the AWS layer to the mock's documented no-op behavior.

soberm (commented):
The OR-composition logic itself is correct and cleanly symmetric across both layers, and the test coverage is a genuine strength — the new AWS-layer parity block asserts the exact `ConditionExpression` / attribute-name / value shape the SDK will send, closing the gap that mock-only tests can't catch. Extending `bb-kv-store`'s existing conditional-write options is the right place and scope for this create-or-update DX fix (not a new BB), and the changeset correctly scopes `bb-kv-store` **minor** + `blocks` **patch** per the repo release convention.

One lingering mock↔AWS parity gap remains on the `{ ifValueEquals: undefined }` input (see inline comments): the mock uses `options?.ifValueEquals !== undefined` (preserving the documented no-op invariant that `parity.test.ts` already tests and warns about) while the AWS layer keeps `'ifValueEquals' in options`, which treats explicit `undefined` as a set condition and would marshal an undefined SDK value. The predicate pre-dates this PR, but this change touches the exact code and could align it.

packages/bb-kv-store/src/index.aws.ts
osama-rizk: Good catch — aligned. The AWS layer now detects the value branch with `options?.ifValueEquals !== undefined` (matching the mock at `index.mock.ts`), so an explicit `{ ifValueEquals: undefined }` is a no-op on **both** layers instead of emitting `#value = :expected` with an undefined value that the SDK rejects. Applied your exact suggestion.

packages/bb-kv-store/src/parity.test.ts
osama-rizk: Added — new parity case asserts `store.put('k','v',{ ifValueEquals: undefined })` produces `ConditionExpression === undefined` and no `ExpressionAttributeValues`, pinning the AWS layer to the mock's documented no-op. bb-kv-store suite now 107 pass.

sarayev (commented):
Nice, clean fix. The core OR logic looks correct on both the AWS and mock layers, and the truth table matches in all four states (absent, present+match, present+differ, and the "absent key wins even with a non-matching `ifValueEquals`" case). One nice detail: `attribute_not_exists(#pk) OR #value = :expected` is safe on an absent key, since DynamoDB returns false (not an error) for a comparison on a missing attribute, so the OR still passes via the absent branch just like the mock. No blockers from me. A few suggestions below, mostly around parity and test coverage.

**1. `ifNotExists` truthiness differs between the two layers (parity nit)**
`index.aws.ts` (~line 108) vs `index.mock.ts` (~line 156)

The whole point of this PR is that the layers agree, but they check `ifNotExists` a little differently: AWS uses a truthy check, the mock uses strict `=== true`.

```ts
// index.aws.ts
if (options?.ifNotExists) {            // truthy

// index.mock.ts
const wantAbsent = options?.ifNotExists === true;   // strict
```

The types make a weird value hard to reach, but should we make AWS use `=== true` too so the layers can never drift on an odd input?

**2. Should we cover the AWS "conditions fail -> error maps correctly" path?**
`parity.test.ts` (~line 254 onward)

The new parity tests assert the PutCommand shape we send to DynamoDB, which is great for catching a typo in the OR string. But they never simulate DynamoDB actually rejecting the write, so the branch that maps a `ConditionalCheckFailedException` to `KVStoreErrors.ConditionalCheckFailed` is unverified for the compose case. The mock covers rejection; the AWS side only checks the request. Worth a test that stubs the client to throw and asserts the mapped error?

```ts
// stub PutItem to throw ConditionalCheckFailedException, then:
await assert.rejects(
  () => store.put('k', 'next', { ifNotExists: true, ifValueEquals: 'prev' }),
  (err: Error) => isBlocksError(err, KVStoreErrors.ConditionalCheckFailed),
);
```

**3. `ifValueEquals: null` is not tested on either layer**
`index.test.ts` / `parity.test.ts`

Since the guard is `!== undefined`, `null` counts as a real condition (`JSON.stringify(null) === 'null'`). That undefined-vs-null boundary is exactly the kind of thing that surprises callers later. Could we add one mock test (`ifValueEquals: null` matches a key stored as `null`) and one parity assertion that AWS still emits `:expected = 'null'`?

**4. Small doc wording in `types.ts` (~line 14)**
The JSDoc says `ifValueEquals` matches when the value "deep-equals," but both layers actually compare via `JSON.stringify`, which is order-sensitive (`{a,b}` vs `{b,a}` will not match). It is pre-existing, but this PR touches that exact block, so maybe reword to something like "serializes equal (`JSON.stringify`)" so we do not overpromise?

One last thing to double-check (not visible in the diff): the AWS layer compares the stored `#value` against `:expected = JSON.stringify(...)`, which only lines up with the mock if `put` writes `item.value` as the JSON string too. That construction lives earlier in `index.aws.ts` and is not in this diff. Almost certainly fine (the single-option path already relied on it), just worth a glance since the parity tests only check command shape, not a real round-trip.

soberm (dismissed):

osama-rizk:
Thanks @sarayev — all four addressed (pushed).

**1. `ifNotExists` truthiness parity.** Aligned the AWS layer to strict `options?.ifNotExists === true`, matching the mock — the two can't drift on an odd value now.

**2. AWS rejection → error mapping (compose path).** Added a parity test that stubs the DynamoDB client to throw a `ConditionalCheckFailedException` and asserts `isBlocksError(err, KVStoreErrors.ConditionalCheckFailed)` for `{ ifNotExists: true, ifValueEquals: 'prev' }`. (`KVStoreErrors.ConditionalCheckFailed === 'ConditionalCheckFailedException'`, and `put`'s catch rethrows non-`ValidationException` errors unchanged, so the SDK error already matches — the test pins it.)

**3. `ifValueEquals: null`.** Added a mock test (a key stored as `null` matches `{ ifValueEquals: null }`; a non-null value is rejected) and a parity assertion that AWS emits `:expected = 'null'`. Confirms the `!== undefined` guard treats `null` as a real CAS value.

**4. JSDoc wording.** Reworded both `ifValueEquals` docs (put + delete) from "deep-equals" to "serializes equal to this value (compared via `JSON.stringify`, so object key order matters)", so it no longer overpromises.

On your round-trip double-check: confirmed — `put` writes `item.value = JSON.stringify(value)` (`index.aws.ts:99`), so the stored `#value` lines up with `:expected = JSON.stringify(...)`, matching the mock. `@aws-blocks/bb-kv-store` suite now **110 pass**.

sarayev (approved):

PR #492, opened 2026-09-04
fix(hosting): DeleteOldBuilds no longer expires the live build (#480)

### LINKED ISSUE #480: DeleteOldBuilds lifecycle rule also expires the build that is currently being served

@aws-blocks/core (the affected code is in `@aws-blocks/hosting`, which is not in the list)

### Describe the bug

`StorageConstruct` adds this lifecycle rule to the hosting bucket:

```
DeleteOldBuilds: prefix `builds/`, expire after `buildRetentionDays` (default 30 days)
```

The rule does not exclude the build that KVS `meta.b` currently points to. If there is no deploy within `buildRetentionDays`, S3 expires the objects of the build in service. The router function keeps rewriting requests to that prefix, so every static path returns 403 (hosting <= 0.1.4) or 404 (>= 0.1.5). The stack, bucket policy and distribution are unchanged; only the objects are gone.

Related points:

- Recovery requires a redeploy. Removing the delete markers does not help, because `Expiration.Days` counts from the object's creation date and the objects expire again on the next lifecycle pass.
- `HostingProps` has no way to change the retention. `HostingConstruct` accepts `storage.buildRetentionDays`, but `packages/core/src/hosting.ts` only passes `retainOnDelete`. The only workaround is an L1 property override on the bucket.
- The mirror case is already guarded: `skewProtection.maxAge > buildRetentionDays` throws `InvalidSkewProtectionMaxAgeError`, since a cookie must not outlive the build it points to. The deploy interval has the same relation to `buildRetentionDays`, but no guard.
- `buildRetentionDays` and `DeleteOldBuilds` are not mentioned in the README, CHANGELOG or `docs/`.

### Expected behavior

The build referenced by KVS is not expired, regardless of when it was deployed. Superseded builds can still be cleaned up.

### Reproduction steps

1. Deploy a hosting-enabled app. Note `meta.b` in the KVS and confirm `builds/<id>/index.html` exists in the bucket.
2. Do not deploy again. Wait for the lifecycle rule to run (30 days by default; to shorten, override `ExpirationInDays` on the bucket to 1 and wait one day).
3. List object versions under `builds/<id>/`. Each object's latest version is a delete marker. KVS `meta.b` still points to `<id>`.
4. `curl -s -o /dev/null -w "%{http_code}" https://<distribution-domain>/` → `403` or `404`.

### Code snippet

```typescript
// packages/hosting/src/constructs/storage_construct.ts
lifecycleRules: [
  {
    id: 'DeleteOldBuilds',
    prefix: 'builds/',        // includes the current build
    expiration: Duration.days(buildRetentionDays),
    enabled: true,
  },
  ...
]
```

```typescript
// packages/core/src/hosting.ts — buildRetentionDays is not passed through
storage: props.retainOnDelete != null
  ? { retainOnDelete: props.retainOnDelete }
  : undefined,
```

### Log output

```
# KVS meta key
{"b":"ms2z0nnb-6eec9615", ...}

# Object versions under the current build prefix
builds/ms2z0nnb-6eec9615/index.html   DeleteMarker  IsLatest=true  2026-08-27T00:00:00Z
# (object created 2026-07-27; expired 31 days later)

# Static path
$ curl -s -D - -o /dev/null https://<distribution-domain>/
HTTP/2 403
server: AmazonS3
x-cache: Error from cloudfront
<Error><Code>AccessDenied</Code><Message>Access Denied</Message></Error>

# API path on the same distribution
$ curl -s -o /dev/null -w "%{http_code}\n" https://<distribution-domain>/aws-blocks/api
200
```

### Environment

- `@aws-blocks/blocks` 0.2.1 (`core` 0.1.11, `hosting` 0.1.4). Verified by code inspection that `hosting` 0.2.0 has the same rule and `core` 0.3.0 still does not pass `buildRetentionDays` (both released 2026-09-01).
- aws-cdk-lib 2.257.0, Node.js 26.7.0, ap-northeast-1, static SPA + API

### Additional context

**Possible fixes**

1. Exclude the current build from the rule: filter the rule by a tag, and tag a build's objects when it is superseded (at the `KvKeys` cutover).
2. Replace time-based expiration with keep-last-N cleanup at deploy time.
3. Pass `buildRetentionDays` through `HostingProps`, and document the constraint.

sarayev:
Fixes #480. The DeleteOldBuilds S3 lifecycle rule was prefix-only (builds/, 30-day expiration) with no tag filter, so it expired the live build that CloudFront's KVS meta.b points at, breaking static asset paths (403/404) while the API stayed 200. This scopes the lifecycle rule to superseded-tagged builds only and tags the outgoing build at KVS cutover (list+PutObjectTagging, no delete). Also plumbs buildRetentionDays through core HostingProps and adds a deployIntervalDays advisory warn. Hosting tests 853/0; adds a core buildRetentionDays pass-through test (passing).

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -313,6 +313,17 @@ new Hosting(stack, 'Web', {
 
 The `framework` option selects the frontend type: `'spa' | 'static' | 'nextjs'`. When omitted, the framework is auto-detected by reading your app's OWN `package.json` (not `node_modules`): a `next` dependency → `nextjs`; otherwise `spa`; and `static` when there is no `package.json`. Set `framework: 'spa'` explicitly to override auto-detection — e.g. when a stray `next` dependency would otherwise trigger an unwanted Next.js/OpenNext build. Full reference lives in the source JSDoc.
 
+Old builds are retained for rollback and expired by an S3 lifecycle rule after `buildRetentionDays` (default **30**). The build currently being served is never expired regardless of deploy cadence — only superseded builds are cleaned up. Raise the window with `buildRetentionDays` (it must be ≥ `skewProtection.maxAge` in days):
+
+```typescript
+new Hosting(stack, 'Web', {
+  root: join(__dirname, '..'),
+  api: blocksStack,
+  buildRetentionDays: 90,
+});
+```
+
+
 ## Building Blocks
 
 Import Building Blocks from their specific packages (or from the `@aws-blocks/blocks` umbrella):
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -313,7 +313,7 @@ new Hosting(stack, 'Web', {
 
 The `framework` option selects the frontend type: `'spa' | 'static' | 'nextjs'`. When omitted, the framework is auto-detected by reading your app's OWN `package.json` (not `node_modules`): a `next` dependency → `nextjs`; otherwise `spa`; and `static` when there is no `package.json`. Set `framework: 'spa'` explicitly to override auto-detection — e.g. when a stray `next` dependency would otherwise trigger an unwanted Next.js/OpenNext build. Full reference lives in the source JSDoc.
 
-Old builds are retained for rollback and expired by an S3 lifecycle rule after `buildRetentionDays` (default **30**). The build currently being served is never expired regardless of deploy cadence — only superseded builds are cleaned up. Raise the window with `buildRetentionDays` (it must be ≥ `skewProtection.maxAge` in days):
+Old builds are retained for rollback and expired by an S3 lifecycle rule after `buildRetentionDays` (default **30**). The build currently being served is never expired regardless of deploy cadence — only superseded builds are cleaned up. Only builds superseded by a normal `Update` cutover are auto-tagged (and thus auto-expired): pre-existing/orphaned builds and aborted or rolled-back deploys are never tagged and need manual cleanup (e.g. an S3 inventory / Batch Operations sweep). Raise the window with `buildRetentionDays` (it must be ≥ `skewProtection.maxAge` in days):
 
 ```typescript
 new Hosting(stack, 'Web', {
@@ -322,6 +322,8 @@ new Hosting(stack, 'Web', {
   buildRetentionDays: 90,
 });
 ```
+
+Core's top-level `buildRetentionDays` maps to the L3 construct's `storage.buildRetentionDays`.
 
 
 ## Building Blocks
--- a/packages/hosting/README.md
+++ b/packages/hosting/README.md
@@ -254,6 +254,36 @@ immediate — SSM Parameter Store has no recovery window.)
 └──────────────────────────────────────────────┘
 ```
 
+## Build retention & rollback
+
+Each deploy uploads its assets under an immutable `builds/<buildId>/` prefix and
+flips a CloudFront KeyValueStore pointer (`meta.b`) to the new build. Old builds
+are retained for `storage.buildRetentionDays` (default **30**) so you can roll
+back, then expired by an S3 lifecycle rule.
+
+- **The build currently being served is never expired**, no matter how long ago
+  it was deployed. Only *superseded* builds (ones a later deploy replaced) are
+  eligible for cleanup — they are tagged `aws-blocks:build-state=superseded` at
+  the cutover, and the lifecycle rule matches only that tag.
+- Raise `storage.buildRetentionDays` for a longer rollback window. It must be at
+  least `skewProtection.maxAge` (converted to days), or synth throws
+  `InvalidSkewProtectionMaxAgeError` — a skew cookie must not outlive the build
+  it pins to.
+- Optional `storage.deployIntervalDays` is an advisory hint: if you set it and
+  it is ≥ `buildRetentionDays`, synth prints a warning that superseded builds
+  may expire before your next deploy (shrinking the rollback window). It never
+  blocks a deploy and never affects the live build.
+
+```ts
+new Hosting(stack, 'Hosting', {
+  root: './',
+  storage: {
+    buildRetentionDays: 90,   // keep 90 days of rollback targets
+    deployIntervalDays: 30,   // advisory only
+  },
+});
+```
+
 ## Custom domains
 
 Configure a custom domain through the `domain` prop on `HostingConstruct`
--- a/packages/hosting/README.md
+++ b/packages/hosting/README.md
@@ -274,6 +274,14 @@ back, then expired by an S3 lifecycle rule.
   may expire before your next deploy (shrinking the rollback window). It never
   blocks a deploy and never affects the live build.
 
+> **Cleanup caveat:** a build is tagged superseded only during a normal deploy
+> cutover (an `Update` that flips `meta.b`). Builds that become stale outside
+> that path — every build already present before upgrading to this version, or
+> builds left by an aborted/rolled-back deploy — are never tagged, so the
+> lifecycle rule never expires them and they accumulate until you remove them
+> manually. This is safe (nothing deletes the live build), just not
+> self-cleaning for pre-existing artifacts.
+
 ```ts
 new Hosting(stack, 'Hosting', {
   root: './',
--- a/packages/hosting/README.md
+++ b/packages/hosting/README.md
@@ -283,8 +283,8 @@ back, then expired by an S3 lifecycle rule.
 > self-cleaning for pre-existing artifacts.
 
 ```ts
-new Hosting(stack, 'Hosting', {
-  root: './',
+new HostingConstruct(stack, 'Hosting', {
+  manifest,
   storage: {
     buildRetentionDays: 90,   // keep 90 days of rollback targets
     deployIntervalDays: 30,   // advisory only
--- a/packages/hosting/src/constructs/kv_keys.ts
+++ b/packages/hosting/src/constructs/kv_keys.ts
@@ -6,6 +6,7 @@ import { Code, Function as LambdaFunction } from 'aws-cdk-lib/aws-lambda';
 import * as iam from 'aws-cdk-lib/aws-iam';
 import { Provider } from 'aws-cdk-lib/custom-resources';
 import type { IKeyValueStore } from 'aws-cdk-lib/aws-cloudfront';
+import type { IBucket } from 'aws-cdk-lib/aws-s3';
 import { DEFAULT_NODE_RUNTIME } from './node_runtime.js';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
@@ -28,6 +29,13 @@ const HANDLER_BUNDLE = join(__dirname, 'kv_keys_handler_bundle.mjs');
 export type KvKeysProps = {
   /** The CloudFront KeyValueStore to write into. */
   store: IKeyValueStore;
+  /**
+   * The hosting bucket that holds `builds/<id>/...`. At the KVS cutover the
+   * handler tags the OUTGOING build's objects as superseded so the
+   * `DeleteOldBuilds` S3 lifecycle rule can expire them without ever touching
+   * the live build (#480). The handler is granted list + tag (never delete).
+   */
+  bucket: IBucket;
   /**
    * Desired key→value map. The custom resource diffs this against the
    * previously deployed entries (empty on Create) and applies the minimal set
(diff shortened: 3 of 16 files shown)


packages/hosting/src/constructs/kv_keys_handler.ts
```diff
@@ -193,5 +279,29 @@ export async function handler(event: Event): Promise<{ PhysicalResourceId: strin
...
+  // build's objects as superseded so the `DeleteOldBuilds` lifecycle rule can
+  // expire them. Order matters: we tag ONLY after applyUpdate() has committed
+  // the flip, so a failure here can never affect the build that is now live
+  // (the new build is never tagged). Best-effort — a failure leaves the old
+  // build untagged (not expired), which is safe; we log and continue so a
+  // transient S3 error never fails an otherwise-successful deploy.
+  if (event.RequestType === 'Update' && event.ResourceProperties.BucketName) {
+    const oldBuildId = activeBuildId(event.OldResourceProperties?.Entries);
+    const newBuildId = activeBuildId(entriesJson);
+    if (oldBuildId && oldBuildId !== newBuildId) {
```
osama-rizk, line 336: **Supersede-tagging is one-directional — nothing ever clears the tag off an incoming build.** The handler only ever calls `supersedeBuild(oldBuildId)`; there's no `DeleteObjectTagging` / tag-live path anywhere. So the safety property "the live build is never tagged" holds only for builds that have *never been superseded before*.

Failure scenario: deploy B1 (live) → deploy B2 (this cutover tags **B1 superseded**, B2 live) → operator rolls back to B1 by re-pointing KVS `meta.b` to the retained `builds/B1/` prefix **without re-uploading**. B1 is now live but still carries `aws-blocks:build-state=superseded` → `DeleteOldBuilds` expires the live build. That's exactly #480, in the path the retention window exists to serve.

**Precondition — this only bites if a rollback is a pointer flip.** If a rollback is always a full redeploy it's self-healing: `generateBuildId()` mints a fresh id (new untagged prefix), and even a deterministic same-id redeploy re-PUTs objects via `BucketDeployment`, which clears the tag.

So the question decides severity: **how is a rollback performed here?** If it's ever "flip the KVS pointer back to a retained build," the handler should also clear the tag on the *new* build at cutover (`DeleteObjectTagging` on `builds/<newBuildId>/`), symmetric to the tag-outgoing step. The README framing of `buildRetentionDays` as "rollback targets" is what led me here.

packages/hosting/src/constructs/kv_keys_handler.ts
```diff
@@ -59,6 +71,80 @@ const client = new CloudFrontKeyValueStoreClient({
...
+ * UNtagged, which is safe — an untagged build is simply never matched by the
+ * lifecycle rule, so it is not expired (it just isn't cleaned up yet). The NEW
+ * (live) build is never passed here, so it is never a lifecycle target.
+ *
+ * Bounded by the custom-resource Lambda timeout: on a very large build the run
+ * may not tag every object before timing out, leaving some objects untagged.
+ * That is safe by the same contract — untagged objects are never expired, and
+ * the live build is never tagged regardless.
+ */
+export async function supersedeBuild(
```
osama-rizk, line 112: **"Cleaned up after `buildRetentionDays`" oversells the guarantee.** Tagging is best-effort and bounded by the Lambda timeout, and (per your own honest caveat) pre-existing builds and aborted/rolled-back deploys are never tagged → never matched → never expired. Combined, self-cleaning only covers builds that were superseded by a clean cutover *and* fully tagged before timeout.

Not a correctness risk (nothing is wrongly deleted), but a real storage-cost drift with no visibility. Worth considering an inventory-driven or S3 Batch Operations sweep for untagged stale builds, or at least surfacing an orphan count via the existing opt-in `inventory` option.

packages/hosting/src/constructs/kv_keys_handler.test.ts
```diff
@@ -120,3 +120,42 @@ describe('kv_keys_handler — batches (50-key / 3 MB boundaries)', () => {
...
+  it('returns undefined when meta has no b', () => {
+    assert.equal(activeBuildId(JSON.stringify({ meta: JSON.stringify({ bp: '/' }) })), undefined);
+  });
+
+  it('returns undefined on malformed JSON (never throws)', () => {
+    assert.equal(activeBuildId('not json'), undefined);
+    assert.equal(activeBuildId(JSON.stringify({ meta: 'not json' })), undefined);
+  });
+
+  it('distinguishes old vs new build so only a real cutover triggers tagging', () => {
```
osama-rizk, line 162: **No test exercises the cutover tagging branch itself.** These tests cover `activeBuildId` extraction thoroughly, but the actual guard — `RequestType === 'Update' && old !== new → supersedeBuild(old)`, and *not* on Create/Delete or `old === new` — is untested end-to-end. `supersedeBuild` is exported yet never invoked in any test.

Add one with a mocked `S3Client` asserting `PutObjectTagging` is issued for `builds/<old>/` and **never** for `builds/<new>/`. That's the assertion that actually protects the live build (and would pin down F1).

packages/hosting/src/constructs/kv_keys.ts
```diff
@@ -80,6 +88,25 @@ export class KvKeys extends Construct {
...
     );
 
+    // #480: supersede-tagging of the OUTGOING build at cutover. Least
+    // privilege: the handler may LIST and TAG objects under `builds/*` only —
+    // it is deliberately granted NO delete permission. Actual expiry is done
+    // by the S3 `DeleteOldBuilds` lifecycle rule, so a bug in the handler can
+    // never delete the live build; worst case an old build lingers untagged.
+    handler.addToRolePolicy(
+      new iam.PolicyStatement({
+        actions: ['s3:ListBucket'],
```
osama-rizk, line 100: Minor / documenting blast radius: `s3:prefix: builds/*` here and `PutObjectTagging` on `builds/*` below let the handler list/tag *any* build, not just the one being superseded — unavoidable since `buildId` can't be IAM-templated. Correct call to withhold delete; just noting the tagging blast radius is all builds, not a single prefix. No action needed.

packages/core/README.md
```diff
@@ -313,6 +313,17 @@ new Hosting(stack, 'Web', {
 
 The `framework` option selects the frontend type: `'spa' | 'static' | 'nextjs'`. When omitted, the framework is auto-detected by reading your app's OWN `package.json` (not `node_modules`): a `next` dependency → `nextjs`; otherwise `spa`; and `static` when there is no `package.json`. Set `framework: 'spa'` explicitly to override auto-detection — e.g. when a stray `next` dependency would otherwise trigger an unwanted Next.js/OpenNext build. Full reference lives in the source JSDoc.
 
+Old builds are retained for rollback and expired by an S3 lifecycle rule after `buildRetentionDays` (default **30**). The build currently being served is never expired regardless of deploy cadence — only superseded builds are cleaned up. Raise the window with `buildRetentionDays` (it must be ≥ `skewProtection.maxAge` in days):
+
+```typescript
+new Hosting(stack, 'Web', {
+  root: join(__dirname, '..'),
+  api: blocksStack,
+  buildRetentionDays: 90,
```
osama-rizk, line 322: Doc nit: this example uses top-level `buildRetentionDays`, while the hosting package README uses `storage.buildRetentionDays`. Both are correct (core `Hosting` flattens it into the L3's `storage` prop), but a reader crossing between the two docs may stumble — a one-line note that core's `buildRetentionDays` maps to the L3 `storage.buildRetentionDays` would help.

osama-rizk (commented):
Reviewed the whole change — the core fix (tag-filter the `DeleteOldBuilds` rule + tag the outgoing build at cutover, never granting delete) is well-engineered, and the order-of-operations (tag only *after* `applyUpdate` commits the flip) and the dependency-free `build_tags.ts` module are exactly right. A few findings below, inline. The one I'd most like your read on is the rollback path (F1): supersede-tagging is one-directional, so whether the live build can ever end up tagged depends entirely on how a rollback is performed.

Nits/praise aside, nothing here blocks the intent — F1 is the only one that could reintroduce #480, and only under one precondition.

osama-rizk (dismissed):

packages/hosting/README.md
```diff
@@ -254,6 +254,44 @@ immediate — SSM Parameter Store has no recovery window.)
...
+> self-cleaning for pre-existing artifacts.
+
+```ts
+new Hosting(stack, 'Hosting', {
+  root: './',
+  storage: {
+    buildRetentionDays: 90,   // keep 90 days of rollback targets
+    deployIntervalDays: 30,   // advisory only
+  },
+});
```
soberm: [major] Non-runnable snippet — violates AGENTS.md rule 11 (runnable docs). This calls `new Hosting(...)` with `root` + `storage.{buildRetentionDays,deployIntervalDays}`, which compiles against neither API:

- Core `Hosting` (packages/core/src/hosting.ts) requires `root: string` and exposes retention as a flat top-level `buildRetentionDays?: number` — it has no `storage` field and no `deployIntervalDays` anywhere in packages/core, so `storage: {...}` is an excess-property error.
- The L3 `HostingConstruct` (packages/hosting/src/constructs/hosting_construct.ts) is the one that accepts `storage.buildRetentionDays`/`storage.deployIntervalDays`, but it requires `manifest` and takes no `root`.

Fix — pick one and match the surrounding prose:
- Core golden path: `new Hosting(stack, 'Web', { root: join(__dirname, '..'), api: blocksStack, buildRetentionDays: 90 })` and drop `deployIntervalDays` (core does not expose it; it is an L3-only advisory prop).
- L3 construct: `new HostingConstruct(stack, 'Hosting', { manifest, storage: { buildRetentionDays: 90, deployIntervalDays: 30 } })`.

osama-rizk (approved):

packages/hosting/src/constructs/kv_keys_handler.ts
sarayev: Addressed in ae4c395. Added a symmetric `clearBuildState(newBuildId)` at cutover: after `applyUpdate` commits the pointer flip and we `supersedeBuild(oldBuildId)`, we now issue `DeleteObjectTagging` on `builds/<newBuildId>/` (best-effort, warn-and-continue, same 15-in-flight chunking as the tagging path). This guarantees the incoming/live build is never left carrying `aws-blocks:build-state=superseded`, so the pointer-flip rollback scenario you described can no longer reintroduce #480. IAM adds `s3:DeleteObjectTagging` on `builds/*` (no delete-object). Covered by the new cutover-tagging tests (see F3).

packages/hosting/src/constructs/kv_keys_handler.ts
sarayev: Addressed in ae4c395 (docs). Added a caveat to both the hosting and core READMEs: self-cleaning only covers builds superseded by a clean cutover that were fully tagged before the Lambda timeout — pre-existing builds and aborted/rolled-back deploys are never tagged, so they are never matched and never expired. No wrongful-deletion risk; the residue is storage-cost drift. Noted the opt-in `inventory` option as the path to surface an orphan count / drive a sweep, rather than adding an unbounded delete here.

packages/hosting/src/constructs/kv_keys_handler.test.ts
sarayev: Addressed in ae4c395. Added a `cutover tagging (#480 F1)` suite (20 tests) that mocks `S3Client.prototype.send` / `CloudFrontKeyValueStoreClient.prototype.send` and asserts the actual guard: on `Update` with `old !== new`, `PutObjectTagging` is issued for `builds/<old>/` and **never** for `builds/<new>/`, `DeleteObjectTagging` clears the new build, and nothing fires on Create/Delete or when `old === new`. That pins the live-build safety property end-to-end.

packages/hosting/src/constructs/kv_keys.ts
sarayev: Acknowledged — no action, as you noted. The `builds/*` tagging blast radius is unavoidable since `buildId` can't be IAM-templated, and we deliberately withhold delete-object, so the worst case is a mis-tag (recoverable) rather than a deletion. Left the comment in place documenting this.

packages/core/README.md
sarayev: Addressed in ae4c395. Added a one-line note in core's README that its flat top-level `buildRetentionDays` maps to the L3's `storage.buildRetentionDays`, so a reader crossing between the core and hosting docs isn't tripped up.

packages/hosting/README.md
sarayev: Fixed in 12bf0bf. Agreed — this "Build retention & rollback" section documents the L3 `HostingConstruct` (it uses `storage.buildRetentionDays`/`deployIntervalDays` and the architecture diagram just above labels `HostingConstruct (CDK L3)`), so I kept `storage.*` and switched the constructor:

```ts
new HostingConstruct(stack, 'Hosting', {
  manifest,
  storage: { buildRetentionDays: 90, deployIntervalDays: 30 },
});
```

`manifest` is `HostingConstruct`'s only required prop (matches the other L3 examples in this README). Verified runnable: `npx tsc --build packages/hosting` → exit 0, and `npm run sync-docs:check` passes.

soberm (approved):

PR #493, opened 2026-09-04
feat(hosting,pipeline): managed-value JSON codec + public per-stage post-deploy hook

osama-rizk:
## What & why

Two additive, backward-compatible capabilities that let a higher-level construct compose `@aws-blocks/hosting` markers and `@aws-blocks/pipeline` stages across a build boundary without reaching into internals.

### 1. `@aws-blocks/hosting` — managed-value JSON codec

`secret()`/`config()` return markers branded with a `Symbol` (and optionally carrying a non-serializable `schema`). They therefore **do not survive `JSON.stringify`/`JSON.parse`** — the brand is dropped and `isManagedValue()` returns false on the far side.

Any consumer that carries a config object containing markers across a JSON boundary — e.g. an orchestrator that serializes per-stage config into a build environment variable and reads it back in a later phase — needs a lossless round-trip. This adds one:

- `encodeManagedValue(marker)` / `decodeManagedValue(encoded)`
- `managedValueReplacer` / `managedValueReviver` — drop-in `JSON.stringify`/`JSON.parse` replacer & reviver
- `isManagedValueJSON`, type `ManagedValueJSON`, and `MANAGED_VALUE_JSON_TAG`

Only the `kind` + `key` locator identity is transported. A marker's optional `schema` is not serializable and is intentionally dropped (re-declare it on the far side if the runtime JSON-parse behavior is needed there) — documented in the code.

### 2. `@aws-blocks/pipeline` — `postStage` hook

New optional `postStage` prop on `PipelineProps`:

```ts
postStage?: (ctx: PostStageContext<TConfig>) => Array<ShellStep | CodeBuildStep> | undefined
```

`PostStageContext` gives the hook the `stage`, its `stageConfig` (including `env` — the stage's target account/region), and the **resolved pipeline `source`** file set. The returned steps are attached as that stage's post-deploy steps.

This exists so a higher-level construct can run a **second per-stage deploy phase** (one that consumes the first phase's outputs) without matching the pipeline's internal construct naming to rediscover the stage's source — previously the only way to find the source was to walk the construct tree and string-match the `-Stage-` id convention, which silently breaks on any internal rename.

**Ordering fix:** when a stage has both `postStage` steps and a `bakeTime`, the bake step is now made to **depend on** the post-stage steps (`addStepDependency`). Previously post steps and the bake `sleep` were siblings that ran in parallel, so the bake timer could expire before the post-stage phase finished. The dependency serializes bake strictly after the post-stage steps.

## Tests

- **hosting** (`secret.test.ts`): raw round-trip loses the brand (motivation), `encode`→`decode` restores branded `secret`/`config` markers, encoded form is tagged & JSON-safe, full nested `JSON.stringify(replacer)`/`JSON.parse(reviver)` round-trip, reviver leaves non-markers untouched, `isManagedValueJSON` rejects malformed shapes.
- **pipeline** (`pipeline.test.ts`): hook invoked once per stage with a defined `source`; returned steps attach as post-deploy actions; **bake runs after post-stage** (RunOrder assertion); hook returning `undefined` adds nothing.

All hosting + pipeline unit tests pass (`node --test dist/*.test.js`); build (`tsc --build`), biome lint (changed files), and `check:exports-consistency` are clean. Changeset included (patch bump for both packages — additive, non-breaking under the 0.x policy where minor is the breaking channel).

## Compatibility

Purely additive: new exports on `@aws-blocks/hosting` and one new optional prop + exported context type on `@aws-blocks/pipeline`. No existing signatures change; existing pipelines synthesize identically when `postStage` is not set.

## Follow-ups (not in this PR)

Deferred so this stays a small, reviewable, well-tested surface:
- Optional customer-managed KMS key on imported secrets so the construct can grant `kms:Decrypt` (or fail synth clearly) for CMK-encrypted secrets.
- Explicit precedence/validation when a compute `environment` key collides with a managed marker key.
- Separating untrusted build/synth from deploy roles (permission boundary, exact target bootstrap roles) and preserving a repo-root-relative working directory for nested-workspace stage discovery.


packages/hosting/src/secret.ts
```diff
@@ -166,6 +166,79 @@ export function isManagedValue(v: unknown): v is ManagedValue {
...
+}
+
+/** Type guard: a value produced by {@link encodeManagedValue} (the JSON form). */
+export function isManagedValueJSON(v: unknown): v is ManagedValueJSON {
+	if (typeof v !== 'object' || v === null) return false;
+	const inner = (v as Record<string, unknown>)[MANAGED_VALUE_JSON_TAG];
+	return (
+		typeof inner === 'object' &&
+		inner !== null &&
+		((inner as ManagedValue).kind === 'secret' || (inner as ManagedValue).kind === 'config') &&
```
soberm: **Suggestion:** `inner` here is the JSON payload shape `{ kind; key }`, but it's cast to `ManagedValue` (the branded runtime marker) to read `.kind`/`.key`. It compiles and works structurally, but the cast is inaccurate. Prefer casting to the inner shape, e.g. `as { kind?: unknown; key?: unknown } | null`, and drop the per-access casts.

.changeset/managed-value-json-codec-and-poststage-hook.md
```diff
@@ -0,0 +1,29 @@
+---
+"@aws-blocks/hosting": patch
```
soberm, line 2: **Nit:** `patch` is the correct level here — under AWS Blocks' 0.x policy (minor = the breaking channel), a non-breaking additive change should be `patch`. Note the PR *description* says 'minor bump for both packages', which contradicts this changeset; recommend fixing the PR body wording to 'patch' to avoid confusing reviewers.

packages/hosting/src/secret.ts
```diff
@@ -166,6 +166,79 @@ export function isManagedValue(v: unknown): v is ManagedValue {
...
+}
+
+/** Encode a marker into a JSON-safe tagged object that survives `JSON.stringify`. */
+export function encodeManagedValue(v: ManagedValue): ManagedValueJSON {
+	return { [MANAGED_VALUE_JSON_TAG]: { kind: v.kind, key: v.key } };
+}
+
+/** Decode a tagged object (see {@link encodeManagedValue}) back into a branded marker. */
+export function decodeManagedValue(v: ManagedValueJSON): ManagedValue {
+	const { kind, key } = v[MANAGED_VALUE_JSON_TAG];
```
soberm: **Nit:** `schema` is intentionally (and clearly) dropped on decode. Flagging for awareness only — a consumer relying on a marker's `schema` after a JSON round-trip gets no runtime signal that it's gone. Docs are sufficient; no change required.

soberm (commented):
3 non-blocking notes inline (1 suggestion, 2 nits). Nothing blocking.

packages/hosting/src/secret.ts
osama-rizk: Applied in 67ffed36 — `inner` is now cast once to the wire shape `{ kind?: unknown; key?: unknown } | null | undefined` (not the branded `ManagedValue`), and the per-access `as ManagedValue` casts are gone. After the `typeof inner === 'object' && inner !== null` guard TS narrows it fine; runtime behavior is unchanged. Thanks — the cast was indeed inaccurate.

.changeset/managed-value-json-codec-and-poststage-hook.md
osama-rizk: Fixed the PR body — it now reads "patch bump for both packages" to match the changeset. Agreed `patch` is correct under the 0.x policy (minor = breaking channel; this is purely additive). The changeset itself is unchanged; only the description wording was contradictory.

packages/hosting/src/secret.ts
osama-rizk: Acknowledged — this is intentional and there's no runtime signal by design: `schema` (a Standard Schema instance) isn't JSON-serializable, and the codec only needs the `kind`+`key` locator identity to survive transport. The docstring on the codec calls this out ("only `kind`+`key` are transported … re-declare the schema on the far side"). Leaving as-is per your note.

packages/hosting/src/secret.ts
```diff
@@ -166,6 +166,84 @@ export function isManagedValue(v: unknown): v is ManagedValue {
...
+	return kind === 'secret' ? secret(key) : config(key);
```
Simone319: **[Blocking][B1] The codec is not lossless for schema-bearing markers and silently changes runtime behavior.**

The new wire representation contains only `kind` and `key`; decode reconstructs `secret(key)` or `config(key)` without the optional schema. Schema is operational rather than type-only: deployment uses its presence to emit a JSON flag (`packages/hosting/src/secret-resolve.ts:323-340`), and runtime uses that flag to parse the fetched string (`packages/hosting/src/secret-runtime.ts:60-75`). A schema-bearing marker transported through #3174 therefore comes back as a raw string rather than the schema-inferred JSON value.

This contradicts the changeset's “lossless” contract and makes the exact #3174 boundary semantically unsafe. Encode at least an explicit schema/JSON-parsing bit and retain equivalent metadata on decode, or narrow the public contract and reject schema-bearing markers rather than silently degrading them.

Simone319: **[Blocking][B2] The new public cross-build wire format has no protocol version, and malformed/future values do not fail predictably.**

The stable exported tag/type hard-code a first representation without a version. The direct decoder uses a binary ternary, so every unknown future `kind` becomes `config`; the reviver silently leaves unknown tagged shapes as ordinary data. Producers and consumers on different package versions can therefore silently reinterpret or ignore values instead of reporting incompatible wire data.

This is a public compatibility boundary that will be difficult to evolve after publication. Add a protocol version now, validate `unknown` input exhaustively, and produce an actionable unsupported-version/kind error instead of silently aliasing or ignoring future representations.

packages/hosting/src/secret.test.ts
```diff
@@ -100,3 +106,57 @@ void describe('paths, prefixes, env naming', () => {
...
+		assert.deepStrictEqual(restored, { a: 1, b: 'x', c: [1, 2] });
+	});
+
+	void it('isManagedValueJSON rejects malformed shapes', () => {
+		assert.strictEqual(isManagedValueJSON({ [MANAGED_VALUE_JSON_TAG]: { kind: 'nope', key: 'K' } }), false);
+		assert.strictEqual(isManagedValueJSON({ [MANAGED_VALUE_JSON_TAG]: { kind: 'secret' } }), false);
+		assert.strictEqual(isManagedValueJSON({ kind: 'secret', key: 'K' }), false);
+		assert.strictEqual(isManagedValueJSON(null), false);
+	});
+});
```
Simone319, line 178: **[Blocking][B3] No test exercises the real two-process stage-config/build-environment boundary that #3174 will consume.**

Codec tests stringify and parse in one process, while pipeline tests inspect synthesized attachment and run order; none serializes stage configuration into a build environment, revives it in a second process, and proves secret/config semantics survive (`packages/pipeline/src/pipeline.test.ts:2291-2399`).

`AGENTS.md:293-300` requires sandbox e2e coverage when serialization changes, and this missing path is where B1 manifests. Add an integration or sandbox test that transports nested secret/config markers—including a schema-bearing value—through the same environment boundary used by #3174 and verifies the downstream deployed/runtime behavior.

packages/hosting/src/secret.ts
```diff
@@ -166,6 +166,84 @@ export function isManagedValue(v: unknown): v is ManagedValue {
...
+	const inner = (v as Record<string, unknown>)[MANAGED_VALUE_JSON_TAG] as
+		| { kind?: unknown; key?: unknown }
+		| null
+		| undefined;
+	return (
+		typeof inner === 'object' &&
+		inner !== null &&
+		(inner.kind === 'secret' || inner.kind === 'config') &&
+		typeof inner.key === 'string'
+	);
```
Simone319: **[Minor][BN1] A tag collision silently replaces an ordinary object and drops sibling fields.**

`isManagedValueJSON` accepts any object containing the public tag plus a valid kind/key, even when the object has additional data; the reviver then replaces the whole object with a marker. The namespaced key makes collision rare, but the result is silent data loss.

Require the exact wire shape or explicitly reserve/document the tag and reject tagged objects with unexpected fields; add a collision regression test.

packages/hosting/src/secret.ts
Simone319: **[Minor][BN2] Malformed-input handling differs between the public decoder and reviver.**

Direct `decodeManagedValue` assumes a trusted shape, throws a raw `TypeError` when the tag is missing, and maps an unknown kind to config. The guard accepts any string key even though the constructors may reject its syntax, causing `JSON.parse` with the reviver to fail with a constructor error.

Accept `unknown` in the public decoder, validate exact shape/kind/key, and either return a typed codec error or consistently leave malformed values untouched. Document whether the reviver can throw and which malformed inputs trigger it.

Simone319 (commented):
Reviewing the public building blocks intended for amplify-backend#3174 at `67ffed36300264eee323ff0e68e4012146dd384d`. **B1–B4 are blocking for #3174 adoption; BN1 and BN2 are minor.**

Comment ④ could not be placed inline because `packages/hosting/README.md` has no changed hunk in this PR, so it is included here.

**[Blocking][B4] The shipped package documentation does not document either new public API or its contracts.** (`packages/hosting/README.md:33-62`; also `packages/pipeline/README.md:12-38`)

The hosting README export/usage sections omit the codec, wire compatibility, malformed-input behavior, collisions, and schema limitation; the pipeline README omits `postStage`, the source guarantee, and bake dependency ordering. These are exported, customer-facing APIs whose correct use depends on non-obvious ordering and serialization rules. Update the shipped READMEs with examples and explicit contracts before release; this is required by T3 (documentation lives with code) and the repository's public-API checklist (`AGENTS.md:27-28,293-300,315`).

packages/hosting/src/secret.ts
osama-rizk: Fixed in e00f2fe8. The wire form now carries the schema's **operational bit** (`json: true` when the marker declared a schema); `decodeManagedValue` rehydrates a marker with a passthrough schema so the far side re-emits the synth JSON-parse flag (`jsonFlagEnvVarName`) and the runtime `JSON.parse`s the value exactly as at the origin. The schema object itself still isn't serialized (it can't cross the boundary), and deep re-validation still needs it re-declared — but that's inherent (the runtime getter is parse-only regardless, per `finalizeValue`). So the round-trip is now lossless for the *operational* contract instead of silently degrading. Tests assert the bit is emitted, that a schema-bearing marker round-trips with a schema present, and (B3) that this survives a real process boundary.

osama-rizk: Fixed in e00f2fe8. Added `MANAGED_VALUE_JSON_VERSION` (currently `1`) to the wire payload and a single exhaustive validator. `decodeManagedValue(v: unknown)` now throws a typed `ManagedValueCodecError` on an unsupported version, unknown kind, invalid key, or malformed shape — the binary ternary that aliased any unknown `kind` to `config` is gone, and the reviver leaves unknown / future-versioned tagged shapes untouched rather than silently reinterpreting them. `MANAGED_VALUE_JSON_VERSION` and `ManagedValueCodecError` are exported. Tests cover unknown-version and unknown-kind rejection (both the guard and the throwing decoder) plus reviver pass-through of a future version.

packages/hosting/src/secret.test.ts
osama-rizk: Added in e00f2fe8 — `managed value JSON codec — real cross-process boundary (B3)`. It serializes a nested config (secret + config + a schema-bearing config + a plain value + a nested marker) into an **environment variable**, then spawns a separate `node` process that revives it via `managedValueReviver` and asserts secret/config identity, plain/nested values, and — critically — that the schema/json bit survives so the schema-bearing value keeps its runtime JSON-parse behavior. That is the exact stage-config → build-env → later-phase boundary #3174 consumes, in two real OS processes.

A full deploy-time sandbox e2e would need live infra; this is the self-contained integration proof that the serialization change is safe. I'm happy to add the deployed-runtime sandbox e2e in the #3174 consumer, where the end-to-end runtime is actually exercised.

packages/hosting/src/secret.ts
osama-rizk: Fixed in e00f2fe8. The validator now requires the **exact** wire shape — the codec tag must be the object's *only* own key — so a plain object that merely also carries the tag plus sibling fields is not treated as a wire value, and the reviver leaves it whole (no silent field loss). Added a collision regression test asserting the object passes through untouched and that direct `decodeManagedValue` rejects it.

packages/hosting/src/secret.ts
osama-rizk: Fixed in e00f2fe8. `decodeManagedValue` now accepts `unknown`, validates shape/version/kind/key exhaustively, and throws a typed `ManagedValueCodecError` (not a raw `TypeError`). The guard and the decoder share one validator, so the reviver only ever decodes already-validated values and therefore **never throws** — malformed or foreign data (including a syntactically invalid key) is left untouched. The reviver-never-throws vs decoder-throws contract is now documented in the codec docblock and the README.

osama-rizk:
Re **[B4]** (shipped README docs) — addressed in e00f2fe8:

- `packages/hosting/README.md` now has a **"Transport markers across a JSON / build boundary"** section documenting the codec exports, the versioned wire form, malformed-input behavior (decoder throws `ManagedValueCodecError`, reviver leaves non-wire values untouched), the collision rule (only a sole-tag object is revived), and the schema limitation (operational JSON-parse bit is preserved; deep re-validation needs the schema re-declared).
- `packages/pipeline/README.md` now documents **`postStage`** — the `PostStageContext` (including the resolved `source` guarantee so callers don't rediscover it by internal construct naming) and the **bake-after-post-stage** ordering — with a runnable example.

The other threads (B1–B3, BN1, BN2) are answered inline. Thanks for the thorough review — the versioning + schema-bit points in particular made this a much safer boundary.

Simone319 (approved):

soberm (approved):

PR #497, opened 2026-09-07
fix(bb-realtime): auto-reconnect + resubscribe production WebSocket on unexpected close

soberm:
## Problem

**Issue #, if available:** internal — useChat WebSocket timeout handling for long agent turns

The production Realtime middleware (`packages/bb-realtime/src/aws-middleware.ts`) did **not** reconnect after an unexpected WebSocket close — only the local mock did. API Gateway WebSockets enforce a **2h max connection duration** and a **10-min idle timeout**, so a long-running agent turn (now up to **8h on AgentCore**) could silently lose its stream when the socket dropped, with no client-side recovery.

This is **PR 1 of 2** (the transport-layer fix). A stacked PR on top wires `useChat` (bb-agent) to re-sync conversation state from the DB on reconnect and reset loading on send-path failures.

## Changes

Brings production to parity with the mock's recovery model:

- **Auto-reconnect on unexpected close** (not on a normal `1000`/`1005` close or a client-initiated `unsubscribe`) with exponential backoff `min(1000 * 2^(n-1), 30_000)`, capped at `MAX_RECONNECT = 5`.
- **Resubscribe** all stored channels with their replayed tokens on the new socket; **keep-alive ping timer re-established** on reconnect (covers the 10-min idle limit across reconnects).
- **New optional `onReconnect()` on `SubscribeOptions`**, fired once after a successful resubscribe (after the drop's `onDisconnect`). Purely additive — existing `subscribe(handler)` / `onDisconnect` callers are unaffected.

Potentially sensitive parts (all covered by tests + reviewed):

- `onDisconnect` fires on **every** drop (handlers cleared only on terminal close).
- A **stale replayed token** (2h TTL vs up to 8h sessions) is surfaced via `onDisconnect('error')` and drains `resubscribePending` so surviving channels still fire `onReconnect` — no silent drop.
- `onerror` no longer rejects `pendingEstablished`; `onclose` alone decides terminal-reject vs reconnect-preserve, so an in-flight `subscribe()` during a transient drop resolves on reconnect.
- Hitting `MAX_RECONNECT` (or a channel-less reconnect) **tears the connection down** so a later `subscribe()` rebuilds — no zombie pool entry.

## Validation

- New unit suite `packages/bb-realtime/src/reconnect.test.ts` (node:test, self-contained FakeWebSocket + mock timers) — **87/87 bb-realtime tests pass**. Covers: auto-reconnect after `1006` with retry cap; resubscribe with replayed token; `onReconnect` ordering; keep-alive re-arm; and the failure paths — every-drop notify, flapping-socket cap, stale-token surfacing, `onerror` preserves in-flight `established`, post-cap rebuild.
- `npm run build` ✅ · `npm run lint:deps` ✅ · `biome lint` ✅ (0 errors) · API report regenerated (`SubscribeOptions` gains `onReconnect?`). Changeset added (`@aws-blocks/bb-realtime` **minor**).
- Two review cycles completed (4 blocking findings resolved → APPROVE).
- **Sandbox e2e:** a true API Gateway 2h/idle close can't be cheaply simulated; the reconnect state machine is proven by the unit suite. A sandbox e2e confirming subscribe/publish end-to-end post-change is tracked for the stacked PR.

### Maintainer notes
- **Behavior change on a 0.x package** → changeset is `minor` per the repo's 0.x convention. Please eyeball `MAX_RECONNECT=5` / backoff adequacy for an 8h session crossing the 2h boundary (~4 forced reconnects) plus transient drops.

## Checklist

- [x] PR description included
- [x] Relevant documentation is changed or added (API.md regenerated; onReconnect JSDoc added)

---

### DIFF
--- a/.changeset/realtime-prod-reconnect.md
+++ b/.changeset/realtime-prod-reconnect.md
@@ -1,0 +1,22 @@
+---
+"@aws-blocks/bb-realtime": minor
+---
+
+feat(bb-realtime): production Realtime middleware now auto-reconnects and resubscribes after an unexpected WebSocket drop
+
+The production (`aws-middleware`) Realtime transport previously gave up when the WebSocket
+closed unexpectedly (e.g. an API Gateway idle timeout or an abnormal `1006` closure) — only the
+local mock middleware recovered. It now transparently reconnects with exponential backoff (capped
+retries), rebuilds the socket URL from the retained connection token, and resubscribes every active
+channel by replaying its stored per-channel token so the server can re-authorize. The keep-alive
+ping timer is re-armed on the fresh socket. A normal client-initiated close (`1000`/`1005`, e.g.
+`unsubscribe()`) is still treated as expected and does not trigger a reconnect.
+
+`SubscribeOptions` gains an optional `onReconnect` callback, fired once after a successful
+reconnect when every channel on the connection has been re-confirmed by the server (after the
+corresponding `onDisconnect` for the drop that triggered it; never on the initial subscribe).
+
+This change is behavior-additive: existing `subscribe(handler)` / `SubscribeOptions` callers are
+unaffected and need no changes. Note for maintainers: on a `0.x` package this ships as a `minor`
+per this repo's convention (minor is the breaking/behavior-change channel pre-1.0), since it alters
+the runtime reconnect behavior of the production transport.
--- a/packages/bb-realtime/src/aws-middleware.ts
+++ b/packages/bb-realtime/src/aws-middleware.ts
@@ -410,7 +410,11 @@ function subscribeTo(
 			}
 		},
 		established,
-		connection: conn.ws!,
+		// Live getter, not a snapshot: openSocket assigns a fresh conn.ws on every
+		// reconnect, so reading conn.ws here means `.connection` always reflects the
+		// current socket rather than the stale (closed) one captured at subscribe
+		// time. Coalesce null → undefined to match the optional `connection?: WebSocket` type.
+		get connection() { return conn.ws ?? undefined; },
 	};
 }
 
--- a/packages/bb-realtime/src/aws-middleware.ts
+++ b/packages/bb-realtime/src/aws-middleware.ts
@@ -29,6 +29,13 @@ export interface RealtimeChannelClient<T = unknown> {
 
 const KEEP_ALIVE_MS = 9 * 60 * 1000;
 
+// ── Auto-reconnect caps (mirrors mock-middleware.ts) ────────────────────────
+
+/** Maximum reconnect attempts after an unexpected drop before giving up. */
+const MAX_RECONNECT = 5;
+/** Ceiling for exponential backoff between reconnect attempts. */
+const MAX_DELAY_MS = 30_000;
+
 // ── Shared connection pool — keyed by wsUrl ─────────────────────────────────
 
 interface PendingSubscribe {
@@ -39,8 +46,14 @@ interface PendingSubscribe {
 interface Connection {
 	ws: WebSocket | undefined;
 	connected: boolean;
+	/** API Gateway endpoint, retained so a reconnect can rebuild the socket URL. */
+	wsUrl: string;
+	/** Connection-level auth token, replayed in the URL on every (re)connect. */
+	connectToken: string;
 	/** Per-channel message handlers. */
 	subscriptions: Map<string, Set<MessageHandler>>;
+	/** Per-channel auth token, replayed on resubscribe after a reconnect. */
+	channelTokens: Map<string, string>;
 	/** Per-channel established promise callbacks. */
 	pendingEstablished: Map<string, PendingSubscribe[]>;
 	/** Subscriptions queued before WebSocket is open. */
@@ -49,6 +62,14 @@ interface Connection {
 	keepAliveTimer: ReturnType<typeof setInterval> | null;
 	/** Registered onDisconnect callbacks (called on unexpected close). */
 	disconnectHandlers: Set<(reason: DisconnectReason) => void>;
+	/** Registered onReconnect callbacks (called after a successful resubscribe). */
+	reconnectHandlers: Set<() => void>;
+	/** Consecutive reconnect attempts since the last successful open. */
+	reconnectAttempts: number;
+	/** Pending reconnect timer, tracked so it can be cleared on teardown. */
+	reconnectTimer: ReturnType<typeof setTimeout> | null;
+	/** Channels awaiting resubscribe confirmation after a reconnect; when it drains, onReconnect fires. `null` outside a reconnect. */
+	resubscribePending: Set<string> | null;
 }
 
 const connections = new Map<string, Connection>();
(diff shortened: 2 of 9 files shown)


packages/bb-realtime/src/aws-middleware.ts
```diff
@@ -63,27 +108,107 @@ function getOrCreateConnection(wsUrl: string, connectToken: string): Connection
...
+		for (const channel of channels) {
+			ws.send(JSON.stringify({ action: 'subscribe', channel, token: conn.channelTokens.get(channel) }));
+		}
+		// Flush subscribes queued while the socket was down (skip any already
+		// resent above).
+		for (const sub of conn.pendingSubs) {
+			if (!conn.subscriptions.has(sub.channel)) {
+				ws.send(JSON.stringify({ action: 'subscribe', channel: sub.channel, token: sub.token }));
+			}
+		}
```
sharonyajain: **onReconnect fires connection-wide, so a channel whose own resubscribe was rejected gets a false-recovery signal**

reconnectHandlers and disconnectHandlers are stored per-CONNECTION, not per-channel. On a reconnect where >=2 channels share one pooled socket and one channel's replayed token is stale, the failed channel is removed and gets onDisconnect('error'), then settleResubscribe fires onReconnect for ALL handlers (gated only on resubscribeAnySucceeded>=1) — including the dead channel's. The healthy-sibling half (a spurious onDisconnect('error') on channel A caused by channel B) is documented and tested as intended, but the FAILED channel's onReconnect firing is neither: the test suite only ever attaches onReconnect to the channel that succeeds (reconnect.test.ts, 'stale token on resubscribe...'). types.ts/DESIGN.md document onReconnect as connection-level ('at least one channel re-confirmed').

```typescript
if (conn.resubscribeAnySucceeded) {
	conn.reconnectHandlers.forEach(h => { try { h(); } catch {} });
}
// ...error branch:
conn.subscriptions.delete(msg.channel);
conn.channelTokens.delete(msg.channel);
conn.disconnectHandlers.forEach(h => { try { h('error'); } catch {} });
```

**Why it matters:** The dead channel's owner sees onReconnect() AFTER its onDisconnect('error') -> treats the channel as recovered and backfills once from the store (exactly as DESIGN.md instructs), then receives no further live messages because the channel was removed from subscriptions -> silent stuck stream, the precise failure class this PR exists to prevent.

**Suggestion:**We can route reconnect/disconnect notifications per-channel — key reconnectHandlers (and the resubscribe-error onDisconnect) by channel so only the affected channel's callbacks fire. If the connection-level contract is deliberate, at minimum suppress onReconnect for any channel that received onDisconnect('error') in the same reconnect, and add a test where the FAILED channel carries the onReconnect handler.

packages/bb-realtime/src/aws-middleware.ts
```diff
@@ -63,27 +108,107 @@ function getOrCreateConnection(wsUrl: string, connectToken: string): Connection
...
+		// can fire only after the server re-confirms each resubscribe.
+		const channels = [...conn.subscriptions.keys()];
+		if (isReconnect) {
+			conn.resubscribePending = new Set(channels);
+			conn.resubscribeAnySucceeded = false;
 		}
-		conn!.pendingSubs.length = 0;
-		// Start keep-alive
-		conn!.keepAliveTimer = setInterval(() => {
+		for (const channel of channels) {
```
sharonyajain, line 200: **unsubscribe of a not-yet-open channel leaves a stale pendingSubs entry that onopen resubscribes**

subscribeTo always adds a channel to `subscriptions` (and `channelTokens`) before pushing to `pendingSubs`, so onopen's subscriptions-driven loop already resubscribes every channel and this guarded pendingSubs flush is dead on the normal path. The only case the guard (`!subscriptions.has`) is true is when a channel is unsubscribed before the socket's first open while other channels remain (subscriptions.size>0, so no teardown fires): the stale pendingSubs entry then sends a subscribe frame for the just-unsubscribed channel. unsubscribe() never purges pendingSubs.

```typescript
for (const sub of conn.pendingSubs) {
	if (!conn.subscriptions.has(sub.channel)) {
		ws.send(JSON.stringify({ action: 'subscribe', channel: sub.channel, token: sub.token }));
	}
}
conn.pendingSubs.length = 0;
```

**Why it matters:** The server registers a subscription that has no local handler (its Set was deleted on unsubscribe), so its messages are silently discarded client-side and no unsubscribe is ever sent -> a phantom server-side subscription (wasted API Gateway/Lambda/DynamoDB fan-out) that persists until the socket terminally closes. Pre-existing in the old flush-all onopen; this PR narrows but does not eliminate it while rewriting the same code.

**Suggestion:** We can delete the channel from pendingSubs in unsubscribe(), or drop pendingSubs entirely now that `subscriptions` + `channelTokens` drive resubscribe on every open (initial and reconnect) — the queue is redundant and its only live path is this bug.

.changeset/realtime-prod-reconnect.md
```diff
@@ -0,0 +1,22 @@
...
+
+The production (`aws-middleware`) Realtime transport previously gave up when the WebSocket
+closed unexpectedly (e.g. an API Gateway idle timeout or an abnormal `1006` closure) — only the
+local mock middleware recovered. It now transparently reconnects with exponential backoff (capped
+retries), rebuilds the socket URL from the retained connection token, and resubscribes every active
+channel by replaying its stored per-channel token so the server can re-authorize. The keep-alive
+ping timer is re-armed on the fresh socket. Only a client-initiated teardown — unsubscribing the last
+channel (or the internal reset/give-up paths) — is treated as terminal; an unexpected drop reconnects on ANY close code, including a clean `1000`/`1005`.
+
+`SubscribeOptions` gains an optional `onReconnect` callback, fired once after a successful
```
sharonyajain, line 15: **Changeset says onReconnect needs every channel re-confirmed, but it fires when at least one does**

The changeset release note states onReconnect fires 'when every channel on the connection has been re-confirmed', but the implementation gates on resubscribeAnySucceeded (>=1 channel) and the types.ts JSDoc, DESIGN.md and README all say 'at least one channel'. The partial-stale test explicitly asserts onReconnect fires with one channel still failed.

```markdown
fired once after a successful
reconnect when every channel on the connection has been re-confirmed by the server
```

**Why it matters:** The changeset is the release note downstream consumers read; an author trusting it would assume onReconnect guarantees ALL channels are live and could skip handling partial-stale failure -> mishandled recovery on exactly the 8h-session/stale-token path.

**Suggestion:** Changing the changeset wording to 'when at least one channel on the connection has been re-confirmed', matching the code and the other three docs.

packages/bb-realtime/src/aws-middleware.ts
soberm: Fixed in c6906d5. `disconnectHandlers`/`reconnectHandlers` are now keyed per-channel (`Map<channel, Set<...>>`), so a channel whose resubscribe is rejected receives `onDisconnect('error')` and **not** `onReconnect` — no more false-recovery signal on a removed channel. `onReconnect` fires per-channel only for the channels that actually re-confirmed, and a sibling's rejection no longer reaches healthy channels. Added a test where the FAILED channel carries the `onReconnect` handler and asserts it gets `onDisconnect('error')` only, while the succeeding sibling still recovers.

packages/bb-realtime/src/aws-middleware.ts
soberm: Fixed in c6906d5 by dropping `pendingSubs` entirely. `subscriptions` + `channelTokens` already drive resubscribe on every open (initial connect and reconnect), so the queue was redundant and its only live path was this bug. With it gone, unsubscribing a not-yet-open channel no longer leaves a stale entry that `onopen` would resubscribe, so the phantom server-side subscription can't occur.

.changeset/realtime-prod-reconnect.md
soberm: Fixed in c6906d5. Reworded to the per-channel contract: `onReconnect` fires for a channel once THAT channel's resubscribe is re-confirmed, and a channel whose resubscribe is rejected gets `onDisconnect('error')` and not `onReconnect`. The JSDoc, DESIGN.md and README were updated to match.

PR #498, opened 2026-09-07
fix(core): key generate-spec method schemas by namespace (fixes #445)

### LINKED ISSUE #445: blocks-generate-spec silently assigns schemas across namespaces that share a method name

@aws-blocks/core

### Describe the bug

`blocks-generate-spec` can silently assign one namespace's parameter and result schemas to another namespace when both namespaces expose an operation with the same bare method name.

In the reproduction, both namespaces are direct top-level `ApiNamespace` exports and both qualified OpenRPC methods are discovered:

- `widgets.create`
- `subscriptions.create`

Their TypeScript contracts are deliberately incompatible. Nevertheless, the generated OpenRPC entry for `widgets.create` receives the `email` input and numeric `subscriptionId` result belonging to `subscriptions.create`, rather than its own `label` input and string `widgetId` result.

Unlike an `unknown` fallback, this produces a valid-looking but incorrect OpenRPC document. Downstream documentation, validators, and generated native clients can therefore trust the wrong contract without an obvious generation failure.

### Expected behavior

Extracted method schemas must be associated with the fully qualified RPC method name, including the namespace.

---

For the reproduction:

- `widgets.create` must retain `CreateWidgetInput` and `CreateWidgetResult`.
- `subscriptions.create` must retain `CreateSubscriptionInput` and `CreateSubscriptionResult`.

The schema for one namespace must never overwrite or be reused by another namespace merely because both operation names are `create`.

### Reproduction steps

Minimal reproduction: https://github.com/costleya/aws-blocks-method-schema-collision-repro

1. Clone the reproduction repository.
2. Run `pnpm install`.
3. Run `pnpm repro`.
4. The control assertions confirm that both direct, qualified methods exist in the generated OpenRPC document.
5. The final assertion fails because `widgets.create` contains `subscriptions.create`'s incompatible parameter and result schemas.

### Code snippet

```typescript
import { ApiNamespace, Scope } from '@aws-blocks/blocks'

const scope = new Scope('method-schema-collision-repro')

interface CreateWidgetInput {
  label: string
}

interface CreateWidgetResult {
  widgetId: string
}

interface CreateSubscriptionInput {
  email: string
}

interface CreateSubscriptionResult {
  subscriptionId: number
}

export const widgets = new ApiNamespace(scope, 'widgets', () => ({
  async create(input: CreateWidgetInput): Promise<CreateWidgetResult> {
    return { widgetId: input.label }
  },
}))

export const subscriptions = new ApiNamespace(scope, 'subscriptions', () => ({
  async create(
    input: CreateSubscriptionInput,
  ): Promise<CreateSubscriptionResult> {
    return { subscriptionId: input.email.length }
  },
}))
```

### Log output

```plain text
AssertionError [ERR_ASSERTION]: BUG: widgets.create should retain its own schemas rather than subscriptions.create schemas

widgets.create actual input:
  { email: string }
widgets.create expected input:
  { label: string }

widgets.create actual result:
  { subscriptionId: number }
widgets.create expected result:
  { widgetId: string }
```

### Environment

- AWS Blocks version: `@aws-blocks/blocks@0.2.7`, `@aws-blocks/core@0.1.18`
- Node.js version: `24.18.0`
- Package manager (npm/yarn/pnpm): `pnpm@11.24.0`
- OS: macOS
- CDK version (if applicable): N/A

### Additional context

The current implementation extracts schemas into a map keyed by the bare method name and later looks them up using that same bare name, even though `generateSpec` has already computed the qualified method name. Relevant source:

- [`extract-ts-types.ts`](https://github.com/aws-devtools-labs/aws-blocks/blob/main/packages/core/src/scripts/extract-ts-types.ts)
- [`generate-spec.ts`](https://github.com/aws-devtools-labs/aws-blocks/blob/main/packages/core/src/scripts/generate-spec.ts)

The minimal safe association is the fully qualified method name, such as `widgets.create` and `subscriptions.create`. Falling back to an unqualified `create` key would remain ambiguous and could preserve the silent corruption.

### Suggested fix

Key extracted method information by the fully qualified RPC method name, for example `widgets.create` and `subscriptions.create`, and have `generateSpec` look up only the qualified name it already computes.

There should be no fallback to a bare method name. Every generated RPC method belongs to a namespace, and a fallback such as `create` is inherently ambiguous even when a particular application currently has only one matching operation.

Suggested regression coverage:

- two direct top-level namespaces with the same `create` method name;
- intentionally incompatible parameter and result types;
- assertions that both qualified methods retain their own schemas regardless of declaration/traversal order;
- an assertion that the extractor does not emit or consume a bare `create` schema key.

### Validated patch

A working pnpm patch against `@aws-blocks/core@0.1.18` is available on the [`codex/proposed-fix` branch](https://github.com/costleya/aws-blocks-method-schema-collision-repro/tree/codex/proposed-fix). Its [`pnpm-workspace.yaml`](https://github.com/costleya/aws-blocks-method-schema-collision-repro/blob/codex/proposed-fix/pnpm-workspace.yaml) registers the issue-specific [`@aws-blocks/core@0.1.18` patch](https://github.com/costleya/aws-blocks-method-schema-collision-repro/blob/codex/proposed-fix/patches/%40aws-blocks__core%400.1.18.patch).

It patches both the TypeScript source and the shipped `dist` implementation. The essential change is to qualify schemas at extraction time and use the already-computed qualified name at generation time:

```diff
- result.set(methodName, info)
+ result.set(`${namespaceName}.${methodName}`, info)

  const qualifiedName = `${routingName}.${methodName}`
- const tsInfo = tsTypes.get(methodName)
+ const tsInfo = tsTypes.get(qualifiedName)
```

osama-rizk:
Fixes #445.

## Problem

`blocks-generate-spec`'s TypeScript extractor (`extract-ts-types.ts`) keyed each method's param/result schemas by the **bare method name** (`result.set(methodName, info)`). When two `ApiNamespace`s each expose an operation with the same name — `widgets.create` and `subscriptions.create` — the second overwrote the first in the map, so `generate-spec` looked up `tsTypes.get('create')` and **silently assigned one namespace's schemas to the other** in the OpenRPC document. Their TS contracts are deliberately incompatible, so the generated client types were wrong with no error.

## Fix

Key method schemas by the **namespace-qualified name** (`namespace.method`), derived from the enclosing `export const <name> = new ApiNamespace(...)` binding — the same name `generate-spec` routes by (`routingName = exportName`). `generate-spec` now resolves `tsTypes.get(qualifiedName) ?? tsTypes.get(methodName)`: qualified wins (no collision), with a **bare fallback** so any method the extractor couldn't attribute to a namespace (e.g. created behind indirection) still resolves exactly as before — no previously-working case regresses.

Covers both extractor paths (the direct `new ApiNamespace(...)` AST walk and the type-checker-resolved indirect path).

## Testing

- New regression test (`extract-ts-types.test.ts`): two namespaces both exposing `create` with incompatible params → distinct `widgets.create` / `subscriptions.create` entries, each keeping its own param type, and **no** bare `create` entry.
- Updated the existing extractor tests to assert the qualified keys.
- Extractor + `generate-spec` + `generate-client` suites green.

## Verification

`npm run lint` 0 errors. No public API change (internal codegen). Changeset added.

## Changeset

`@aws-blocks/core` **patch** + `@aws-blocks/blocks` **patch** (umbrella re-export).

### DIFF
--- a/packages/core/src/scripts/extract-ts-types.test.ts
+++ b/packages/core/src/scripts/extract-ts-types.test.ts
@@ -143,9 +143,34 @@ describe('extractSkipCodegenMethods — pure-AST scan', () => {
 		try {
 			const skips = extractSkipCodegenMethods(join(dir, 'index.ts'));
 			assert.strictEqual(skips.size, 1);
-			assert.ok(skips.has('devOnly'));
-			assert.ok(!skips.has('normalMethod'));
-			assert.ok(!skips.has('alsoNormal'));
+			assert.ok(skips.has('api.devOnly'), 'skip set is namespace-qualified');
+			assert.ok(!skips.has('api.normalMethod'));
+			assert.ok(!skips.has('api.alsoNormal'));
+		} finally {
+			rmSync(dir, { recursive: true, force: true });
+		}
+	});
+
+	it('qualifies skip names by namespace so a tag on one namespace does not drop another (regression: #445)', () => {
+		const dir = createTempProject({
+			'index.ts': `
+				class ApiNamespace { constructor(scope: any, name: any, handler: any) {} }
+
+				export const widgets = new ApiNamespace(null, 'widgets', () => ({
+					/** @blocksSkipCodegen */
+					async create() { return null; },
+				}));
+				// subscriptions.create is NOT tagged and must survive.
+				export const subscriptions = new ApiNamespace(null, 'subscriptions', () => ({
+					async create() { return null; },
+				}));
+			`,
+		});
+		try {
+			const skips = extractSkipCodegenMethods(join(dir, 'index.ts'));
+			assert.ok(skips.has('widgets.create'), 'tagged method is skipped, qualified');
+			assert.ok(!skips.has('subscriptions.create'), 'untagged same-named method must NOT be skipped');
+			assert.ok(!skips.has('create'), 'no bare key that would drop both');
 		} finally {
 			rmSync(dir, { recursive: true, force: true });
 		}
--- a/packages/core/src/scripts/extract-ts-types.test.ts
+++ b/packages/core/src/scripts/extract-ts-types.test.ts
@@ -29,6 +29,47 @@ const API_NS_MOCK = `
 	const ApiNamespace: ApiNamespaceConstructor = class {} as any;
 `;
 
+describe('extractMethodTypes — namespaces sharing a method name (regression: #445)', () => {
+	it('keys methods by namespace, so a shared method name does not cross-assign schemas', () => {
+		const dir = createTempProject({
+			'tsconfig.json': JSON.stringify({
+				compilerOptions: { target: 'ESNext', module: 'ESNext', moduleResolution: 'bundler', strict: true },
+			}),
+			'index.ts': `
+				${API_NS_MOCK}
+
+				// Two namespaces both expose \`create\`, with deliberately incompatible contracts.
+				export const widgets = new ApiNamespace(null, 'widgets', (context) => ({
+					async create(label: string): Promise<{ widgetId: string }> {
+						return { widgetId: 'w' };
+					},
+				}));
+				export const subscriptions = new ApiNamespace(null, 'subscriptions', (context) => ({
+					async create(topicCount: number): Promise<{ subId: number }> {
+						return { subId: 1 };
+					},
+				}));
+			`,
+		});
+
+		try {
+			const types = extractMethodTypes(join(dir, 'index.ts'));
+			// Distinct qualified keys — no collision.
+			assert.ok(types.has('widgets.create'), 'widgets.create should be keyed by namespace');
+			assert.ok(types.has('subscriptions.create'), 'subscriptions.create should be keyed by namespace');
+			assert.ok(!types.has('create'), 'no bare-name entry that could cross-assign');
+
+			// Each keeps its OWN parameter type.
+			assert.strictEqual(types.get('widgets.create')!.params[0].name, 'label');
+			assert.strictEqual(types.get('widgets.create')!.params[0].schema.type, 'string');
+			assert.strictEqual(types.get('subscriptions.create')!.params[0].name, 'topicCount');
+			assert.strictEqual(types.get('subscriptions.create')!.params[0].schema.type, 'number');
+		} finally {
+			rmSync(dir, { recursive: true, force: true });
+		}
+	});
+});
+
 describe('extractMethodTypes — direct ApiNamespace calls', () => {
 	it('extracts param and return types from inline ApiNamespace', () => {
 		const dir = createTempProject({
@@ -58,15 +99,15 @@ describe('extractMethodTypes — direct ApiNamespace calls', () => {
 
 		try {
 			const types = extractMethodTypes(join(dir, 'index.ts'));
-			assert.ok(types.has('getUser'), 'should find getUser');
-			assert.ok(types.has('createUser'), 'should find createUser');
+			assert.ok(types.has('api.getUser'), 'should find getUser');
+			assert.ok(types.has('api.createUser'), 'should find createUser');
 
-			const getUser = types.get('getUser')!;
+			const getUser = types.get('api.getUser')!;
 			assert.strictEqual(getUser.params.length, 1);
 			assert.strictEqual(getUser.params[0].name, 'id');
 			assert.strictEqual(getUser.params[0].schema.type, 'string');
 
-			const createUser = types.get('createUser')!;
+			const createUser = types.get('api.createUser')!;
 			assert.strictEqual(createUser.params.length, 2);
 			assert.strictEqual(createUser.params[0].schema.type, 'string');
 			assert.strictEqual(createUser.params[1].schema.type, 'number');
@@ -155,8 +196,8 @@ describe('extractMethodTypes — @blocksSkipCodegen JSDoc tag', () => {
 
 		try {
 			const types = extractMethodTypes(join(dir, 'index.ts'));
-			assert.strictEqual(types.get('normalMethod')?.skipCodegen, undefined);
-			assert.strictEqual(types.get('getLastCode')?.skipCodegen, true);
+			assert.strictEqual(types.get('api.normalMethod')?.skipCodegen, undefined);
+			assert.strictEqual(types.get('api.getLastCode')?.skipCodegen, true);
 		} finally {
 			rmSync(dir, { recursive: true, force: true });
 		}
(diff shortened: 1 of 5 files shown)


packages/core/src/scripts/generate-spec.ts
```diff
@@ -334,9 +341,8 @@ export async function generateSpec(
 			// type-resolution errors and silently empty its Map — we always
 			// want the JSDoc tag to be honored, so the AST-only scan acts as
 			// a safety net.
-			if (tsTypes.get(methodName)?.skipCodegen || skipCodegenNames.has(methodName)) continue;
+			if (tsInfo?.skipCodegen || skipCodegenNames.has(methodName)) continue;
```
sarayev: `skipCodegenNames` holds bare method names (from `extractSkipCodegenMethods`), so this `has(methodName)` check isn't namespace-aware. If two namespaces both expose `create` and only one is tagged `@blocksSkipCodegen`, doesn't the other one get dropped from the spec too? That looks like the same collision #445 is about, just hiding in this safety-net branch. Could we qualify the skip set the same way and check `has(qualifiedName) || has(methodName)`?

packages/core/src/scripts/generate-spec.ts
```diff
@@ -324,6 +324,13 @@ export async function generateSpec(
 		for (const [methodName, methodFn] of Object.entries(methodMap)) {
 			if (typeof methodFn !== 'function') continue;
 
+			const qualifiedName = `${routingName}.${methodName}`;
+			// Prefer the namespace-qualified TS types; fall back to the bare method
+			// name for entries the extractor couldn't attribute to a namespace.
+			// Qualified keys stop two namespaces that share a method name (e.g.
+			// widgets.create / subscriptions.create) from cross-assigning schemas.
+			const tsInfo = tsTypes.get(qualifiedName) ?? tsTypes.get(methodName);
```
sarayev, line 335: Nice fallback. Worth a quick note that two namespaces the extractor *can't* attribute (default export, factory-wrapped) still land on bare keys and could overwrite each other here, same as #445. Fine to leave as-is, but a one-line comment would save the next person some head-scratching.

packages/core/src/scripts/extract-ts-types.ts
```diff
@@ -280,11 +281,35 @@ function extractMethodsFromHandler(
...
+ * (`export const widgets = new ApiNamespace(...)` → `"widgets"`). This is the
+ * name the dev server / Lambda handler route by, and the one `generate-spec`
+ * uses to qualify method names. Returns `undefined` when the namespace is
+ * created behind indirection (factory return, etc.), in which case methods key
+ * by their bare name.
+ */
+function enclosingBindingName(node: ts.Node): string | undefined {
+	let current: ts.Node | undefined = node.parent;
+	while (current) {
+		if (ts.isVariableDeclaration(current) && ts.isIdentifier(current.name)) return current.name.text;
```
sarayev, line 314: Small edge case: this keys off the `const` binding name, but generate-spec looks methods up by the export routing name. For `const widgets = ...; export { widgets as gadgets }` the keys won't line up and the types quietly fall back to `unknown`. Rare, and the old bare-keying happened to dodge it. Maybe just a comment flagging the limitation?

packages/core/src/scripts/extract-ts-types.test.ts
```diff
@@ -29,6 +29,47 @@ const API_NS_MOCK = `
...
+						return { subId: 1 };
+					},
+				}));
+			`,
+		});
+
+		try {
+			const types = extractMethodTypes(join(dir, 'index.ts'));
+			// Distinct qualified keys — no collision.
+			assert.ok(types.has('widgets.create'), 'widgets.create should be keyed by namespace');
```
sarayev, line 58: Good repro — this fails on main. Since the fix also changed how `generate-spec` looks these keys up, would it be worth a matching collision test on the generate-spec side? Two namespaces with a same-named method taking different params, asserting each emitted method keeps its own schema. That's the part customers actually hit in #445.

packages/core/src/scripts/generate-spec.ts
soberm: [minor] `skipCodegenNames.has(methodName)` here is still keyed by the bare method name. `tsInfo` above is now namespace-resolved, but this safety-net check is not — so if two namespaces share a method name and only one carries `@blocksSkipCodegen`, the other can be wrongly dropped from the spec. That's the same #445 collision class, just in the fallback branch. Consider qualifying the skip set and checking `has(qualifiedName) || has(methodName)`.

packages/core/src/scripts/extract-ts-types.ts
soberm, line 314: [minor] `enclosingBindingName` keys off the `const` binding name, but `generate-spec` routes by the export name. An export rename — `const widgets = new ApiNamespace(...); export { widgets as gadgets }` — keys `widgets.*` while routing looks up `gadgets.*`, quietly falling back to the bare/`unknown` path. Worth a one-line comment flagging the limitation (or resolving the export alias if it's cheap).

packages/core/src/scripts/extract-ts-types.test.ts
soberm, line 58: [minor] This proves the *extractor* produces distinct qualified keys — nice repro. Consider also adding a collision test on the `generate-spec` side: two namespaces, same-named method, different params, asserting each operation keeps its own schema in the generated OpenRPC spec. That locks in the actual user-visible #445 (which manifested in the spec, not the map).

packages/core/src/scripts/generate-spec.ts
soberm, line 335: [nit] Good fallback comment. One added sentence noting that namespaces the extractor can't attribute (default export, factory-wrapped) still land on bare keys and could overwrite each other would flag the residual edge for the next reader.

packages/core/src/scripts/generate-spec.ts
osama-rizk: Great catch — real bug, fixed. `extractSkipCodegenMethods` now keys namespace-qualified too (bare only when the namespace is unattributable), and generate-spec checks `skipCodegenNames.has(qualifiedName) || has(methodName)`. So a `@blocksSkipCodegen` tag on `widgets.create` no longer drops `subscriptions.create`. Added a regression test asserting exactly that (tagged `widgets.create` skipped, untagged `subscriptions.create` survives, no bare `create` key).

osama-rizk: Great catch — real bug, fixed. `extractSkipCodegenMethods` now keys namespace-qualified too (bare only when the namespace is unattributable), and generate-spec checks `skipCodegenNames.has(qualifiedName) || has(methodName)`. So a `@blocksSkipCodegen` tag on `widgets.create` no longer drops `subscriptions.create`. Added a regression test asserting exactly that (tagged `widgets.create` skipped, untagged `subscriptions.create` survives, no bare `create` key).

packages/core/src/scripts/extract-ts-types.test.ts
osama-rizk: Added it. New `generateSpec` regression test: two namespaces each with a `create` taking different params (string vs number), asserting the emitted `widgets.create` keeps `label: string` and `subscriptions.create` keeps `topicCount: number` — the user-visible #445 in the OpenRPC output, not just the map. (It feeds a typed `.ts` foundation to the extractor and a custom `loader` for runtime discovery, since `node --test` can't import a raw `.ts`.)

osama-rizk: Added it. New `generateSpec` regression test: two namespaces each with a `create` taking different params (string vs number), asserting the emitted `widgets.create` keeps `label: string` and `subscriptions.create` keeps `topicCount: number` — the user-visible #445 in the OpenRPC output, not just the map. (It feeds a typed `.ts` foundation to the extractor and a custom `loader` for runtime discovery, since `node --test` can't import a raw `.ts`.)

packages/core/src/scripts/extract-ts-types.ts
osama-rizk: Added a comment on `enclosingBindingName` flagging it: it keys by the local `const` binding, so an `export { widgets as gadgets }` rename keys `widgets.*` while routing looks up `gadgets.*` → qualified lookup misses and falls back to the bare key (resolves if no collision, else `unknown`). Left resolving the alias as a noted follow-up since it's rare and the fallback is safe.

osama-rizk: Added a comment on `enclosingBindingName` flagging it: it keys by the local `const` binding, so an `export { widgets as gadgets }` rename keys `widgets.*` while routing looks up `gadgets.*` → qualified lookup misses and falls back to the bare key (resolves if no collision, else `unknown`). Left resolving the alias as a noted follow-up since it's rare and the fallback is safe.

packages/core/src/scripts/generate-spec.ts
osama-rizk: Done — added a sentence to the fallback comment noting that namespaces the extractor can't attribute (default export, factory-wrapped, export-rename) still land on bare keys and could overwrite each other there, same class as #445.

osama-rizk: Done — added a sentence to the fallback comment noting that namespaces the extractor can't attribute (default export, factory-wrapped, export-rename) still land on bare keys and could overwrite each other there, same class as #445.

soberm (approved):

sarayev (approved):

PR #500, opened 2026-09-07
fix(data-bbs): map OCC conflicts to JSON-RPC 409 instead of 500

soberm:
## Problem

Optimistic-concurrency (OCC) / conditional-write conflicts in the four data Building Blocks were returned to clients as JSON-RPC **`code: 500`** (Internal Server Error) instead of **`409`** (Conflict).

**Root cause:** the JSON-RPC serializer (`packages/core/src/rpc.ts`) computes `code = error instanceof ApiError ? error.status : 500`. Only `ApiError` carries an HTTP status; every OCC throw site emitted a plain / `blocksError` `Error` (name set, **no status**), so it fell through to the generic 500. The auth BBs were already fixed to throw `ApiError(…, 409/401, { name })`; the data BBs never got the same treatment.

**Issue #, if available:** Bug Bash P3 — *Optimistic-concurrency error returned as JSON-RPC code:500 (should be 409)* (1 report, R7).

## Changes

Wrap each OCC throw site in `ApiError(message, 409, { name: <existing error name>, cause, retriable: true })`, following the established auth-BB pattern. `error.name` is preserved so `isBlocksError()` and the `{Class}Errors` constants keep matching; core is intentionally **not** modified (`ApiError` already exists).

| BB | Change |
|---|---|
| `bb-kv-store` | mock `conditionalConflict()` helper (put/delete); aws catches DynamoDB `ConditionalCheckFailedException` on put + delete → 409 |
| `bb-distributed-table` | `conditionalCheckFailed()` helper in `errors.ts`, used in mock + aws (put/delete) |
| `bb-distributed-data` | `translateDsqlError` maps SQLSTATE `40001` (SerializationFailure) → 409; mock engine routed through it for parity |
| `bb-data` | `serializationConflict()` helper wired into all three engines (`pg-error-translator`, `pglite`, `data-api`) for `40001` |

**Potentially problematic / call-outs:**
- **Client-observable behavior change:** these OCC paths now serialize to **409** instead of **500**. Released as a `minor` bump on the four `bb-*` packages (0.x: `minor` is the breaking-change channel). Flagging for maintainer review.
- `cause` stays server-side — non-enumerable via `ApiError`, and the `rpc.ts` serializer never emits `cause` (no `$metadata`/ARN leak).
- Mock ↔ aws parity: identical `error.name` + `409` + `retriable` in both paths for every BB.

## Validation

- **TDD:** failing tests captured first (status ≠ 409 / not an `ApiError`), then made green.
- **Added tests:** unit tests per BB asserting `status === 409`, correct `name`, and `retriable`; **new e2e** tests in `test-apps/comprehensive` (`kv-store.test.ts`, `distributed-table.test.ts`) asserting `409` over the real JSON-RPC wire — no type casts in customer-facing test code.
- **Local runs (all pass):** `npm run build`, `npm run lint:deps`, Biome `lint`, `npm test` (fail 0), `npm run update:api` (no API report churn).
- `npm run test:e2e:local`: **501 pass / 0 fail**, including the 3 new 409 tests. (The suite process exits 1 only via a pre-existing Realtime WebSocket open-handle backstop at teardown — unrelated to this change.)
- Changeset added: `.changeset/occ-conflicts-409.md` (`minor` × 4 bb-* packages). The `@aws-blocks/blocks` umbrella bump is handled automatically by changesets' dependency resolution.

## Checklist

- [x] PR description included

---


test-apps/comprehensive/aws-blocks/index.ts
```diff
@@ -1730,6 +1730,19 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
     return { success: true };
   },
 
+  // Test-only: forces the mock DSQL engine to raise a serialization conflict on
+  // the next commit, so the e2e suite can prove the OCC→409 mapping over the
+  // JSON-RPC wire. simulateConflict() is a mock-only hook (absent from the
+  // deployed aws surface); the comprehensive app type-checks against mock types,
+  // so this compiles without a cast.
+  async dsqlForceConflict() {
+    dsql.simulateConflict();
```
sarayev: `dsqlForceConflict()` leans on `dsql.simulateConflict()`, which only exists on the mock engine. On a real sandbox deploy the Lambda runs the `index.aws.ts` runtime where that hook is not defined, so I think this call throws a TypeError and comes back as a 500, which would flip the matching `assert e.status === 409` red under `test:e2e:sandbox`. Did this run against a real sandbox, or is the DSQL 409 check local-only right now? If it is local-only, could we skip it when the sandbox target is active (and soften the 'proven over the JSON-RPC wire' wording in DESIGN.md), or force a genuine 40001 with two concurrent transactions instead of the mock hook? Also a little wary of exposing a mock-only method on the deployed API surface with no guard.

test-apps/comprehensive/test/dsql.test.ts
```diff
@@ -147,5 +151,22 @@ export function dsqlTests(getApi: () => typeof apiType) {
...
+
+    // OCC serialization conflicts (SQLSTATE 40001) must serialize to JSON-RPC
+    // code 409 (Conflict) over the wire — reconstructed client-side as an
+    // ApiError with status 409 — not a generic 500. The error name is preserved
+    // so isBlocksError still matches. dsqlForceConflict() drives the mock's
+    // simulateConflict() hook to raise the conflict deterministically.
+    test('DSQL - serialization conflict returns status 409 over the wire', async () => {
+      const api = getApi();
+      try {
+        await api.dsqlForceConflict();
```
sarayev: Same concern I left on `dsqlForceConflict` in `test-apps/comprehensive/aws-blocks/index.ts`: since it drives the mock-only `simulateConflict()` hook, does this test actually pass under `test:e2e:sandbox` against the real DSQL runtime, or is it local-only? If sandbox cannot produce a deterministic 40001, worth guarding this test to the local/mock target so the sandbox suite does not go red here.

packages/bb-kv-store/src/index.aws.ts
```diff
@@ -120,6 +120,26 @@ export class KVStore<T = string> extends Scope {
...
+			// name (== KVStoreErrors.ConditionalCheckFailed) so isBlocksError()
+			// keeps matching, and keep the driver error as `cause` (server-side).
+			// DynamoDB collapses every conditional failure under one exception with
+			// no sub-reason, so retriability is decided from the condition THIS
+			// call set: an `ifValueEquals` optimistic-lock conflict is retriable
+			// (re-read and retry can succeed); an `ifNotExists` uniqueness
+			// assertion is not (a blind identical retry fails identically). Matches
+			// the mock path's per-branch `conditionalConflict()`.
+			if (err instanceof Error && err.name === KVStoreErrors.ConditionalCheckFailed) {
+				const retriable = !options?.ifNotExists && !!(options && 'ifValueEquals' in options);
```
sarayev: On the aws side `retriable` is recomputed from the options, but the mock in `index.mock.ts` sets it from whichever branch actually threw. They agree for single-condition calls, but disagree when conditions are combined: e.g. `put` with `ifNotExists: true` plus `ifValueEquals` on a missing key fails the value check, so the mock reports `retriable: true` while this computes `!true && ... = false`. The changeset promises 'matching the mock branch-for-branch', so want to align them? Deriving `retriable` with the same precedence the mock uses (existence assertion wins, so not retriable) would close it. Minor, since name and status stay correct and only the retriable hint is off. Small aside: this uses `'ifValueEquals' in options` (key presence) while the mock uses `!== undefined` (value), so an explicit `ifValueEquals: undefined` gets classified differently.

packages/bb-kv-store/src/index.aws.ts
```diff
@@ -147,7 +167,26 @@ export class KVStore<T = string> extends Scope {
...
+		} catch (err: unknown) {
+			// A failed conditional delete is a Conflict, not an
+			// InternalServerError: map DynamoDB's raw
+			// ConditionalCheckFailedException to an ApiError with status 409 (see
+			// the put path above). Preserves the name for isBlocksError() and keeps
+			// the driver error as `cause`. Retriability is decided from the
+			// condition THIS call set: an `ifValueEquals` optimistic-lock conflict
+			// is retriable; an `ifExists` existence assertion is not.
+			if (err instanceof Error && err.name === KVStoreErrors.ConditionalCheckFailed) {
+				const retriable = !conditions?.ifExists && !!(conditions && 'ifValueEquals' in conditions);
```
sarayev: Same as the put path above: `retriable` here can diverge from the mock when `ifExists` plus `ifValueEquals` are combined, and `'ifValueEquals' in conditions` vs the mock's `!== undefined` classifies an explicit `undefined` differently. Worth aligning both the same way.

packages/bb-distributed-table/src/index.aws.ts
```diff
@@ -129,6 +129,19 @@ export class DistributedTable<
...
+			// ConditionalCheckFailedException to an ApiError with status 409 so the
+			// JSON-RPC serializer emits code 409 instead of 500. Name is preserved
+			// for isBlocksError(), the driver error is kept as `cause`. DynamoDB
+			// collapses every conditional failure under one exception with no
+			// sub-reason, so retriability is decided from the condition THIS call
+			// set: an `ifFieldEquals` optimistic-lock conflict is retriable, an
+			// `ifNotExists` uniqueness assertion is not. Matches the mock path. Other
+			// errors (e.g. oversized items) still flow through remapItemTooLarge.
+			if (err instanceof Error && err.name === DistributedTableErrors.ConditionalCheckFailed) {
+				throw conditionalCheckFailed(!!options?.ifFieldEquals, err);
```
sarayev: Same `retriable` parity concern as in `packages/bb-kv-store/src/index.aws.ts`: this recomputes retriability from the options rather than the failing branch. The wrinkle here is it resolves the combined case the opposite way from KV: `ifNotExists` plus `ifFieldEquals` on an existing key fails the uniqueness assertion, but `!!options?.ifFieldEquals` still reports retriable=true, whereas the mock would say false. So the two Blocks do not land on the same rule. Could we settle on one precedence (existence assertion wins, so not retriable) across both? Minor.

packages/bb-distributed-table/src/index.aws.ts
```diff
@@ -143,7 +156,20 @@ export class DistributedTable<
...
+			await this.docClient.send(new DeleteCommand(command));
+		} catch (err: unknown) {
+			// A failed conditional delete is a Conflict, not an
+			// InternalServerError: map DynamoDB's raw
+			// ConditionalCheckFailedException to an ApiError with status 409 (see
+			// the put path above). Retriability is decided from the condition THIS
+			// call set: an `ifFieldEquals` optimistic-lock conflict is retriable, an
+			// `ifExists` existence assertion is not. Matches the mock path.
+			if (err instanceof Error && err.name === DistributedTableErrors.ConditionalCheckFailed) {
+				throw conditionalCheckFailed(!!options?.ifFieldEquals, err);
```
sarayev: Same as the put path just above (and as noted in `packages/bb-kv-store/src/index.aws.ts`): `ifExists` plus `ifFieldEquals` combined would report retriable=true from `!!options?.ifFieldEquals` while the mock keys off the branch that threw. Aligning the precedence would keep the two paths and the two Blocks consistent.

packages/bb-data/src/errors.ts
```diff
@@ -15,6 +17,24 @@ export const DatabaseErrors = {
...
+ * conflict. Maps to HTTP 409 (Conflict) so the JSON-RPC serializer emits code
+ * 409 instead of a generic 500, preserves the `SerializationFailure` name so
+ * `isBlocksError(e, DatabaseErrors.SerializationFailure)` keeps matching on both
+ * server and client, keeps the original engine error as `cause` (server-side),
+ * and flags the conflict retriable (the caller can retry the transaction).
+ * Shared by every engine translator (PGlite, pg-client, Data API) so all paths
+ * produce an identically shaped 409.
+ */
+export function serializationConflict(cause: Error): ApiError {
+  return new ApiError(cause.message, 409, {
```
sarayev: `cause.message` is passed straight through as the client-visible ApiError message. Keeping `cause` server-side is the right call and there is no ARN or $metadata leak, so this is mostly benign. Only flag: Data API 40001 messages can be fairly verbose compared to a fixed string (the DynamoDB paths pass a fixed 'The conditional request failed', so those are fine). Fine to leave, or normalize to a fixed message if you would rather not surface raw driver text to clients. Optional.

packages/bb-distributed-data/src/errors.ts
```diff
@@ -35,7 +37,17 @@ export const TRANSACTION_ROW_LIMIT = 3000;
...
   const code = (e as any).code as string | undefined;
   if (code === PG_SERIALIZATION_FAILURE) {
-    e.name = DistributedDatabaseErrors.SerializationFailure;
+    // An OCC / serialization-failure conflict (SQLSTATE 40001) is a Conflict,
+    // not an InternalServerError: throw an ApiError with status 409 so the
+    // JSON-RPC serializer emits code 409 instead of a generic 500. Preserve the
+    // SerializationFailure name so isBlocksError() keeps matching on both server
+    // and client, keep the original error as `cause` (server-side), and flag it
+    // retriable — the caller can retry the transaction (see `retryOnConflict`).
+    throw new ApiError(e.message, 409, {
```
sarayev: Same minor note as in `packages/bb-data/src/errors.ts`: `e.message` is the raw driver message going to the client. Benign for a fixed 40001 string; only worth normalizing if you would rather not expose driver text. Optional.

packages/bb-distributed-data/src/engines/dsql-mock-engine.ts
```diff
@@ -146,7 +146,10 @@ export class DsqlMockEngine implements DatabaseEngine {
         new Error('SerializationFailureException: OCC conflict — transaction not committed.'),
         { code: PG_SERIALIZATION_FAILURE, name: DistributedDatabaseErrors.SerializationFailure }
       );
-      throw err;
+      // Route through translateDsqlError so the mock surfaces the SAME 409
+      // ApiError (name preserved, retriable) the real DSQL engine produces for
+      // SQLSTATE 40001 — keeping mock and aws paths behaviorally identical.
+      translateDsqlError(err);
```
sarayev, line 152: Tiny nit: `translateDsqlError(err);` works because its return type is `never`, but `throw translateDsqlError(err);` reads clearer and matches how the engine translators are invoked elsewhere.

test-apps/comprehensive/aws-blocks/index.ts
soberm: Good catch — you are right that simulateConflict() is mock-only and would TypeError→500 on the aws runtime under test:e2e:sandbox, plus it exposed a mock-only method on the deployed API. Removed dsqlForceConflict() and its e2e test entirely in 1c5b7d4; bb-distributed-data’s 40001→409 mapping is now documented as unit/translator-covered (mirroring bb-data), which is the ceiling the mock can prove deterministically.

test-apps/comprehensive/test/dsql.test.ts
soberm: Removed this test (and dsqlForceConflict) in 1c5b7d4 — it was local/mock-only and would have gone red under test:e2e:sandbox. Documented the mapping as unit-covered instead.

packages/bb-kv-store/src/index.aws.ts
soberm: Fixed in 1c5b7d4. Both mock and aws now use one precedence — existence assertion wins: retriable = (ifValueEquals !== undefined) && !ifNotExists. So combined ifNotExists+ifValueEquals ⇒ retriable:false in both paths. Also switched from `"ifValueEquals" in options` to value-presence (!== undefined) so an explicit undefined is classified identically. Added a combined-condition unit test.

packages/bb-kv-store/src/index.aws.ts
soberm: Fixed in 1c5b7d4: delete path now retriable = (ifValueEquals !== undefined) && !ifExists, value-presence based, identical to the mock — combined ifExists+ifValueEquals ⇒ false.

packages/bb-distributed-table/src/index.aws.ts
soberm: Fixed in 1c5b7d4. Aligned bb-distributed-table to the SAME existence-wins precedence as bb-kv-store: retriable = (ifFieldEquals !== undefined) && !ifNotExists, so combined ifNotExists+ifFieldEquals ⇒ false in both mock and aws. The two Blocks now land on one rule.

packages/bb-distributed-table/src/index.aws.ts
soberm: Fixed in 1c5b7d4: delete path retriable = (ifFieldEquals !== undefined) && !ifExists, matching the mock branch-for-branch.

packages/bb-data/src/errors.ts
soberm: Normalized in 1c5b7d4: the 40001 ApiError now carries a fixed client-visible message ("The transaction failed due to a serialization conflict"); the raw driver message is retained only as the server-side cause.

packages/bb-distributed-data/src/errors.ts
soberm: Same fix in 1c5b7d4 — fixed client-visible message, raw driver text kept as server-side cause only.

packages/bb-distributed-data/src/engines/dsql-mock-engine.ts
soberm: Done in 1c5b7d4 — changed to `throw translateDsqlError(err);` in the commit-conflict path for clarity.

packages/bb-data/DESIGN.md
```diff
@@ -61,6 +62,8 @@ Data API errors that carry no SQLState are classified by their SDK exception nam
 
 The `DatabaseBase` subclass only adds `TransactionFailed` naming for errors that escape the engine's transaction methods without a recognized name.
 
+An OCC serialization failure (SQLSTATE `40001`) is translated to an `ApiError` with status **409 (Conflict)**, retriable, preserving the `SerializationFailure` name so the JSON-RPC serializer emits 409 instead of a generic 500. This mapping is verified by translator/engine **unit tests** (`pg-error-translator.test.ts`, `data-api-engine.test.ts`), not an over-the-wire e2e test: the single-connection PGlite mock has no conflict-injection hook and cannot deterministically produce a `40001` serialization conflict over the wire, so unit tests are the verification ceiling for this Block. (By contrast, DistributedDatabase's DSQL mock exposes a `simulateConflict()` hook, so its 409 mapping is proven over the JSON-RPC wire in the comprehensive e2e suite.)
```
sarayev: Small follow-up from the last round: this parenthetical says DistributedDatabase's 409 mapping is 'proven over the JSON-RPC wire in the comprehensive e2e suite', but the DSQL over-the-wire test (`dsqlForceConflict`) was removed in this round, and `bb-distributed-data/DESIGN.md` now states the mapping is covered by unit tests only (no over-the-wire e2e). So this line now contradicts its sibling doc. Could we drop the 'By contrast ... proven over the JSON-RPC wire' sentence, or reword it to note DistributedDatabase is likewise unit-tested?

packages/bb-kv-store/src/index.aws.ts
```diff
@@ -120,6 +120,26 @@ export class KVStore<T = string> extends Scope {
...
+			// name (== KVStoreErrors.ConditionalCheckFailed) so isBlocksError()
+			// keeps matching, and keep the driver error as `cause` (server-side).
+			// DynamoDB collapses every conditional failure under one exception with
+			// no sub-reason, so retriability is derived from the conditions THIS
+			// call set, matching the mock branch-for-branch: existence assertion
+			// wins — a conflict is retriable only for a pure `ifValueEquals`
+			// optimistic-lock check (value presence, so an explicit `undefined` is
+			// treated as absent) and NOT when `ifNotExists` is also set.
+			if (err instanceof Error && err.name === KVStoreErrors.ConditionalCheckFailed) {
+				const retriable = options?.ifValueEquals !== undefined && !options?.ifNotExists;
```
sarayev: minor: this path still passes the raw DynamoDB `err.message` to the client, whereas the DB blocks (`bb-data`, `bb-distributed-data`) and the KV mock were normalized to a fixed `'The conditional request failed'` string in this round. It matches today only because DynamoDB's ConditionalCheckFailedException message happens to be exactly that, but if the SDK message changes (or someone sets ReturnValuesOnConditionCheckFailure) it would leak item/driver text and drift from the mock. Mind using the fixed string here and keeping `cause: err`, to match the other paths?

packages/bb-kv-store/src/index.aws.ts
```diff
@@ -147,7 +167,26 @@ export class KVStore<T = string> extends Scope {
...
+		} catch (err: unknown) {
+			// A failed conditional delete is a Conflict, not an
+			// InternalServerError: map DynamoDB's raw
+			// ConditionalCheckFailedException to an ApiError with status 409 (see
+			// Retriability is derived from the conditions THIS call set (see the
+			// put path): existence assertion wins — retriable only for a pure
+			// `ifValueEquals` optimistic-lock check (value presence) and NOT when
+			// `ifExists` is also set. Matches the mock path.
+			if (err instanceof Error && err.name === KVStoreErrors.ConditionalCheckFailed) {
+				const retriable = conditions?.ifValueEquals !== undefined && !conditions?.ifExists;
```
sarayev, line 203: minor: same as the put path above (raw `err.message` to the client instead of the fixed string every other path uses). Worth normalizing this one the same way for parity.

packages/bb-kv-store/README.md
```diff
@@ -78,13 +78,13 @@ await store.delete('temp', { ifExists: true });
 await store.delete('lock', { ifValueEquals: expectedVal });
 ```
 
-All condition failures throw with `error.name === KVStoreErrors.ConditionalCheckFailed`.
+All condition failures throw with `error.name === KVStoreErrors.ConditionalCheckFailed`. They serialize to JSON-RPC **409 (Conflict)** over the wire (not 500) and are flagged retriable, so `error.status === 409` on the client and `isBlocksError` still matches by name.
```
sarayev: minor: this reads as if all conditional failures are retriable, but after this round `ifNotExists`/`ifExists` are deliberately NOT retriable (only `ifValueEquals` is), and the error-table row just below says the same. Could we qualify it to something like 'retriable only for `ifValueEquals` optimistic-lock conflicts; existence assertions are not retriable', so the doc matches the code?

packages/bb-distributed-table/README.md
```diff
@@ -129,7 +129,7 @@ await table.delete(key, { ifExists: true });
 await table.delete(key, { ifFieldEquals: { status: 'archived' } });
 ```
 
-All condition failures throw with `error.name === DistributedTableErrors.ConditionalCheckFailed`.
+All condition failures throw with `error.name === DistributedTableErrors.ConditionalCheckFailed`. They serialize to JSON-RPC **409 (Conflict)** over the wire (not 500) and are flagged retriable, in both the mock and AWS runtime.
```
sarayev: minor: same retriable overclaim as in `bb-kv-store/README.md` (and in the error-table row below): this reads as if every conditional failure is retriable, but `ifNotExists`/`ifExists` are not, only `ifFieldEquals` is. Worth the same qualification for consistency with D-DT-9 and the code.

packages/bb-data/DESIGN.md
soberm: Fixed in 4534e95 — reworded so bb-data/DESIGN.md no longer claims DistributedDatabase is proven over the JSON-RPC wire; it now states the 40001→409 mapping is unit/translator-covered, consistent with bb-distributed-data/DESIGN.md (the dsqlForceConflict e2e was removed last round).

packages/bb-kv-store/src/index.aws.ts
soberm: Fixed in 4534e95 — the put path now throws the fixed "The conditional request failed" string (the same one the mock uses), keeping cause: err. So it no longer relies on the DynamoDB SDK message happening to match, and stays aligned with the mock even if the SDK message changes or ReturnValuesOnConditionCheckFailure is set.

packages/bb-kv-store/src/index.aws.ts
soberm: Fixed in 4534e95 — delete path normalized to the same fixed string, cause: err preserved.

packages/bb-kv-store/README.md
soberm: Fixed in 4534e95 — the sentence (and the error-table row) now scope retriable to ifValueEquals optimistic-lock conflicts only; ifNotExists/ifExists existence assertions are documented as not retriable, matching the code.

packages/bb-distributed-table/README.md
soberm: Fixed in 4534e95 — same qualification applied here for ifFieldEquals (existence assertions not retriable), consistent with the code and D-DT.

sarayev (dismissed):
Thanks for working through both rounds of feedback so quickly. All my comments are addressed: the OCC conflicts map cleanly to 409, the retriable flag is scoped correctly (optimistic-lock only), the mock and AWS paths are now byte-for-byte consistent, and the docs line up with the code. Nice work. Approving.

test-apps/comprehensive/test/kv-store.test.ts
```diff
@@ -168,6 +168,38 @@ export function kvStoreTests(getApi: () => typeof apiType) {
       });
     });
 
+    // OCC conflicts must serialize to JSON-RPC code 409 (Conflict) over the
+    // wire — reconstructed client-side as an ApiError with status 409 — not a
+    // generic 500. The error name is preserved so isBlocksError still matches.
+    describe('OCC conflicts map to HTTP 409', () => {
```
osama-rizk, line 174: `retriable` is never asserted over the wire, and the retriable path has no e2e coverage at all. Both over-the-wire tests (here and `distributed-table.test.ts`) assert `status === 409` + name only, and both exercise **existence** conditions (`ifNotExists`/`ifExists`) — the *non*-retriable paths. So the whole subject of the retriable-scoping follow-up, and specifically the `retriable: true` value-equals path, is proven only in unit tests.

Since `rpc.ts` conditionally serializes `retriable` and `decodeRpcResponse` reconstructs it, one added wire assertion on an `ifValueEquals` conflict (`e.retriable === true`) would lock both the mechanism and the single field most likely to regress silently. Cheap, high-value.

osama-rizk (commented):
Reviewed the whole change end-to-end, including that `retriable` actually round-trips the JSON-RPC wire (`rpc.ts` serializes it at the encode side and `decodeRpcResponse` reconstructs it on the client) — so the feature genuinely reaches callers. The mock↔AWS parity derived from a single `ifX !== undefined && !existenceAssertion` expression, the existence-wins handling of the type-forbidden combined case, keeping `cause` server-side, and the honest documentation of the bb-data / bb-distributed-data e2e ceiling are all excellent. Nothing here blocks. Findings inline are LOW / nits plus one scope question (F3). The two I'd act on: add one wire assertion for `retriable` (F1) and confirm the 23505→500 boundary is deliberate (F3).

packages/bb-distributed-table/src/index.mock.ts
```diff
@@ -342,19 +350,22 @@ export class DistributedTable<
...
 
-	private checkFieldEquals(keyStr: string, fields: Partial<T>): void {
+	private checkFieldEquals(keyStr: string, fields: Partial<T>, retriable: boolean): void {
 		const entries = Object.entries(fields).filter(([, v]) => v !== undefined);
 		if (entries.length === 0) {
 			throw blocksError(DistributedTableErrors.InvalidQuery, DistributedTableMessages.emptyIfFieldEquals);
 		}
 
 		const existing = this.data.get(keyStr);
 		if (!existing) {
```
osama-rizk, line 367: `retriable: true` is flagged even when the row is **missing**. Here `checkFieldEquals` throws `conditionalConflict(retriable)` with `retriable` already computed `true` for a pure `ifFieldEquals`, so a field-equals conflict against a non-existent item is marked retriable — but a blind retry fails identically (the row is still absent). Same shape on the KVStore `ifValueEquals`-vs-missing-key path.

Not a bug: it's forced on AWS (DynamoDB collapses every conditional failure into one exception, so missing-vs-stale is indistinguishable) and the mock deliberately matches for parity. But it means the `retriable` contract is slightly overstated. Worth one caveat sentence in the docs: `retriable` marks the *conflict kind*, not a guarantee a retry will succeed.

packages/bb-data/src/engines/pg-error-translator.ts
```diff
@@ -15,12 +18,17 @@ const PG_CONNECTION_EXCEPTION_CLASS = '08';
...
  * // All other errors → QueryFailed
  */
 export function translatePgError(e: unknown, engineName: string): never {
   if (e instanceof Error) {
     const code = (e as any).code as string | undefined;
+    if (code === PG_SERIALIZATION_FAILURE) {
+      // OCC conflict: surface as a retriable 409 (Conflict), not a generic 500.
+      throw serializationConflict(e);
+    }
     if (code === PG_UNIQUE_VIOLATION) {
```
osama-rizk, line 32: Scope question (not a defect): sibling conflict-class errors still return 500. `23505` (UniqueConstraintViolation) here — and in `data-api-engine.ts` — still does `e.name = …; throw e` → plain `Error` → **500** over the wire. A duplicate-key is also a client-side conflict (arguably 409/400), i.e. the same 5xx-for-a-4xx-cause smell this PR fixes for OCC.

Scoping this PR to OCC/40001 is a defensible boundary, so I'm not asking to expand it — just: is `23505 → 500` deliberate, or is a follow-up planned to align the 4xx/5xx classification across the data BBs? (ConnectionFailed staying 500 is correct — that one genuinely is server/infra.)

packages/bb-distributed-data/src/engines/dsql-mock-engine.ts
```diff
@@ -146,7 +146,10 @@ export class DsqlMockEngine implements DatabaseEngine {
         new Error('SerializationFailureException: OCC conflict — transaction not committed.'),
         { code: PG_SERIALIZATION_FAILURE, name: DistributedDatabaseErrors.SerializationFailure }
       );
-      throw err;
+      // Route through translateDsqlError so the mock surfaces the SAME 409
+      // ApiError (name preserved, retriable) the real DSQL engine produces for
+      // SQLSTATE 40001 — keeping mock and aws paths behaviorally identical.
+      throw translateDsqlError(err);
```
osama-rizk: Nit: `throw translateDsqlError(err)` — `translateDsqlError` is typed `(e) => never` and always throws, so the outer `throw` is unreachable. The other call sites in this same file (lines ~124/130/138) call it bare. Harmless, but inconsistent — drop the `throw` here for uniformity.

packages/bb-distributed-data/src/errors.ts
```diff
@@ -35,7 +37,19 @@ export const TRANSACTION_ROW_LIMIT = 3000;
...
-    e.name = DistributedDatabaseErrors.SerializationFailure;
+    // An OCC / serialization-failure conflict (SQLSTATE 40001) is a Conflict,
+    // not an InternalServerError: throw an ApiError with status 409 so the
+    // JSON-RPC serializer emits code 409 instead of a generic 500. Preserve the
+    // SerializationFailure name so isBlocksError() keeps matching on both server
+    // and client, keep the original error as `cause` (server-side), and flag it
+    // retriable — the caller can retry the transaction (see `retryOnConflict`).
+    // The client-visible message is a fixed, stable string (the raw driver text
+    // is verbose); the original error is retained as `cause` for diagnostics.
+    throw new ApiError('The transaction failed due to a serialization conflict', 409, {
```
osama-rizk, line 48: Nit / consistency: the 40001→409 construction is inlined here as `new ApiError('The transaction failed due to a serialization conflict', 409, …)`, while `bb-data` factors the identical shape into a `serializationConflict()` helper. The fixed message string now lives in 3 places and will drift independently. Single call site here makes a local helper low-value, and these are separate packages with distinct error-name constants, so I wouldn't do a cross-package refactor — just noting it so the message stays in sync if it ever changes.

test-apps/comprehensive/test/kv-store.test.ts
soberm: Great call — added in cb6d2b6. kv-store.test.ts now has an over-the-wire test that forces an ifValueEquals value-mismatch conflict and asserts e.status === 409 AND e.retriable === true, so the retriable field (serialized in rpc.ts, reconstructed by decodeRpcResponse) is now locked over the wire, not just in unit tests. Reused the existing kvPut method — no new API surface, no casts.

packages/bb-distributed-table/src/index.mock.ts
soberm: Agreed it is not a bug (forced by DynamoDB collapsing conditional failures into one indistinguishable exception, and the mock matches for parity), but the contract was overstated. Added a caveat in cb6d2b6 to bb-kv-store/DESIGN.md, bb-distributed-table/DESIGN.md, and the ApiError.retriable JSDoc: retriable marks the conflict KIND, not a guarantee a retry succeeds — a value/field-equals conflict against a missing row is still flagged retriable though a blind retry fails identically.

packages/bb-data/src/engines/pg-error-translator.ts
soberm: Deliberate — this PR is scoped to OCC/40001. You are right that 23505 UniqueConstraintViolation is the same 5xx-for-a-4xx-cause smell (and ConnectionFailed correctly stays 500). Filed a follow-up to align the duplicate-key classification (→ 409/400) across the data BBs: #508.

packages/bb-distributed-data/src/engines/dsql-mock-engine.ts
soberm: Done in cb6d2b6 — dropped the throw, so it is now a bare translateDsqlError(err) matching the sibling `: never` call sites in this file. (Note: this reverses an earlier nit that asked to add the throw; going with uniformity across the sibling sites here.)

packages/bb-distributed-data/src/errors.ts
soberm: Fair point on the duplicated fixed string across the three packages. Leaving as-is per your own note — single call site here makes a local helper low-value, and these are separate packages with distinct error-name constants, so a cross-package refactor is not worth it. Flagging so we keep the message in sync if it ever changes.

osama-rizk (approved):

sarayev (approved):
Re-reviewed after the main merge. The #490 OR-composition on put is integrated correctly, the put retriable derivation is right under the new OR semantics, mock and aws stay in parity, and the new over-the-wire test covers 409 + retriable. Looks good, re-approving.

PR #507, opened 2026-09-07
feat(observability): Compute-driven logging, tracing, and dashboard

Simone319:
## The model
**Logging — always on; retention is a compute prop.**
Every compute captures stdout to its own log group unconditionally — there is no "enable logging" step. Retention moves to the compute: `LambdaCompute` gains a `logRetention` prop that overrides the stack-wide `defaults.logRetention`. The CDK `Logger` becomes a no-op placeholder, and its `retention` option is removed. Any number of Loggers coexist freely.
**Log level — per-instance only.**
Log level is set solely per `Logger` via the `level` option (default `'info'`). There is no app-wide `BlocksDefaults.logLevel` and no `LOG_LEVEL` env var — Blocks stamps no log-level config, and the runtime `Logger` no longer reads `LOG_LEVEL`. Multiple Loggers with different levels coexist without any shared config.
**Tracing — presence-gated, fleet-wide.**
Creating **any** `Tracer` in the app enables X-Ray on **every** compute. X-Ray provisions real, costed infrastructure, so it stays off until the app opts in by constructing a Tracer. Implemented with a core tracer registry: `registerTracer()` records presence, `finalizeTracing()` enables tracing on all computes at finalize (before the dashboard finalize). `Compute.enableTracing()` is idempotent, so multiple Tracers are safe. The X-Ray publish permission is granted once on the shared execution role in `finalizeTracing` (not per compute), since every compute assumes the same role.
**Dashboard — compute-state-driven, finalize-deferred.**
`DashboardOptions` gains `logs?: boolean` / `traces?: boolean` (default `true`) — app-wide display toggles applied uniformly to every compute section. The dashboard covers **every** compute in the app (resolved at finalize, so construction order never matters); no `computes` selector is exposed yet, since it would leak the internal `Compute` type before customers can construct one to pass. The `CwDashboard` resource is created eagerly in the constructor (only its widget *body* is deferred to finalize), so the dashboard `url`, redirect route, and config registration never point at a resource that does not exist.
Each compute renders a health section always, a logs section (unless `logs: false`), and a traces section only when tracing is enabled on it (unless `traces: false`). Metrics remain app-scoped and are passed explicitly. The deprecated `LoggerBBRef` / `TracerBBRef` types are removed — the dashboard reads compute state directly.
## ⚠️ Behavior / API changes
- `Logger` no longer reconfigures log retention; the `retention` option was removed from `LoggingOptions`. Set `logRetention` on the compute instead.
- A `Tracer` now enables X-Ray on **all** computes, not one.
- `Logger` no longer reads the `LOG_LEVEL` env var; log level is set solely via the per-`Logger` `level` option (default `'info'`). Blocks stamps no app-wide log-level config, and `BlocksDefaults` has no `logLevel` field.
- Removed the deprecated `LoggerBBRef` / `TracerBBRef` dashboard types.
## Testing
`npm run build`, `npm run lint:deps`, `npm run check:api`, and unit tests for all touched packages pass on Node 22 (core, bb-logger, blocks, bb-dashboard, bb-tracer, bb-lambda-compute, bb-async-job, bb-cron-job). A CDK-synth test asserts presence-gated fleet-wide tracing — a real `Tracer` flips `TracingConfig: Active` on every compute at finalize (verified with two computes), and an app with no Tracer leaves them untouched.
Deployed to AWS accounts to verify the dashboard appearance.


packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -61,11 +68,11 @@ export class LambdaCompute extends Compute {
...
 		// LogGroup passed as the function's `logGroup`) makes its retention follow
-		// the stack-wide default instead of AWS's infinite default, and gives
-		// bb-logger one group to reconfigure rather than a second, colliding one.
-		// Torn down with the stack (logs are not durable state).
+		// this compute's setting instead of AWS's infinite default. Retention is a
+		// compute-level prop (per-compute override) falling back to the stack-wide
+		// default. Torn down with the stack (logs are not durable state).
 		this.logGroup = new LogGroup(this, 'HandlerLogGroup', {
-			retention: this.defaults.logRetention,
+			retention: options?.logRetention ?? this.defaults.logRetention,
```
osama-rizk, line 75: Consistency: this handler log group now honors the per-compute override (`options?.logRetention ?? this.defaults.logRetention`) — good — but the API Gateway **access-log** group (line 126, inside `if (this.defaults.accessLogging)`) still uses `this.defaults.logRetention` **directly**, with no `options?.logRetention` fallback. So a user who sets `logRetention: TWO_YEARS` on the compute retunes the handler logs but silently **not** the access logs, which keep the stack default.

Defensible — access logs are an API-Gateway concern and arguably should track stack posture rather than the compute prop — but it's a surprising asymmetry given both are "this compute's logs." Either apply the same `options?.logRetention ?? …` fallback at line 126 for symmetry, or add a one-line comment stating access-log retention intentionally follows the stack default only.

packages/core/src/cdk/tracer-registry.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Mark that the app contains a `Tracer`. Tracing is **presence-gated**: a Tracer
+ * anywhere in the app means every compute should be traced (X-Ray provisions
+ * real, costed infra, so it's off unless the app opts in by creating a Tracer).
+ * Multiple Tracers are fine — this just records the boolean. Stored per stack
+ * (keyed by a Symbol), like the config/compute registries.
+ *
+ * @param scope - Any construct in the stack (used to locate the stack).
+ */
+export function registerTracer(scope: Construct): void {
+	(cdk.Stack.of(scope) as unknown as Record<symbol, boolean>)[REGISTRY_KEY] = true;
```
osama-rizk, line 20: The tracer presence (and the dashboard registry) key on `cdk.Stack.of(scope)` — the same per-CDK-stack scoping as the config/compute registries. That inherits the multi-backend-in-one-stack limitation I flagged on #391: `defaults` are anchored per `BlocksBackend`, but presence + `getComputes` + `finalizeTracing` are stack-scoped. So with two `BlocksBackend`s in one CDK stack, a `Tracer` (or `Dashboard`) in backend A enables X-Ray on / renders sections for backend B's computes too — and each backend's `create()` runs `finalizeTracing`/`finalizeDashboards` over the whole stack.

This is consistent with the existing model, so it's not a regression — but the observability rework widens the blast radius of that scope mismatch (tracing now fans out fleet-wide per stack). Worth confirming multi-backend-per-stack is unsupported for observability the same way it is for config, or scoping these registries per-backend to match `defaults`. Untested either way (the isolation tests cover multi-*stack*, not multi-backend-in-one-stack).

packages/bb-dashboard/src/index.cdk.ts
```diff
@@ -65,20 +71,57 @@ export class Dashboard extends Scope {
...
-			start: config.defaultTimeRange,
-			widgets: widgetRows,
+		// Build the widget body in a finalizer, not here. The body depends on
+		// which computes are traced (`dashboardSection` gates its traces section on
+		// each compute's tracing flag), and that flag is flipped when the app's
+		// `Tracer` is finalized during the backend-module import. This finalizer
+		// runs after that import completes, so the dashboard observes tracing
+		// regardless of the order the customer constructed things in — and its
+		// compute list captures every compute in the app.
+		registerDashboardFinalizer(this, () => {
```
osama-rizk, line 98: The Dashboard emits its URL `CfnOutput`, the redirect route, and the `registerConfig` URL **eagerly** in the constructor, but defers the actual `CfnDashboard` resource to this finalizer. If `finalizeDashboards` is ever not run for this stack (a construction path outside `create()`, or a test that forgets to drive it — which the docstring notes is required), the deploy still emits a live dashboard URL + redirect route pointing at a dashboard resource that was **never created** — a silent 404-style dangling link rather than a synth error.

Real `create()` always finalizes, so this is low, but the eager/deferred split makes the failure silent. Consider building at least the `CwDashboard` resource (even empty) eagerly and only deferring the widget *body*, or asserting at synth that registered finalizers ran.

packages/bb-tracer/src/index.cdk.ts
```diff
@@ -1,26 +1,21 @@
...
+import type { TracerOptions } from './types.js';
 
-export type { TracerOptions, Segment, AnnotationValue } from './types.js';
+export type { AnnotationValue, Segment, TracerOptions } from './types.js';
 
 export class Tracer extends Scope {
 	constructor(scope: ScopeParent, id: string, options?: TracerOptions) {
 		super(id, { parent: scope });
 
 		if (options?.enabled !== false) {
```
osama-rizk, line 24: Minor footgun under presence-gating: `new Tracer(scope, id, { enabled: false })` only skips **this** Tracer's registration — because tracing is a fleet-wide OR across all Tracers, any other Tracer (or one constructed without the flag) still enables X-Ray on every compute. So `enabled: false` doesn't reliably disable tracing for a scope; it just removes one vote. Given the model is documented as presence-gated/fleet-wide this is defensible, but the `enabled` option reads like a per-Tracer on/off switch and could surprise. Worth a doc note that `enabled: false` = "this Tracer contributes nothing" rather than "tracing off."

osama-rizk (commented):
Reviewed the whole rework — read the core registries and both finalize paths closely. The design is sound: the finalize ordering is correct and consistent in both entry points (config → tracing → dashboards, with tracing before the dashboard precisely because `dashboardSection` reads `tracerEnabled`), `enableTracing()` is idempotent, `finalizeDashboards` drains its list so a double-finalize can't rebuild, and `tracerEnabled` is private with a protected read-only getter so tracing can't be turned on without marking the compute traced. It also correctly adopts the #302/#385 defaults model — `options?.logRetention ?? this.defaults.logRetention` is the canonical two-tier precedence, with posture-aware presets (ONE_WEEK sandbox / ONE_YEAR prod). Nice — this is the good-citizen counterpart to the bb-file-bucket case that ignored that model.

Findings below are all LOW / consistency, plus one non-blocking history note:

**Stale commit message (L0 intent, no code impact).** Commit 1's body says "App-wide default log level is a new `BlocksDefaults.logLevel`, stamped once as the `LOG_LEVEL` runtime config." But HEAD has no `logLevel` field and no `LOG_LEVEL` env var (later commits removed that approach, and the PR body's ⚠️ section correctly says so). On a squash-merge that commit body becomes the permanent history and will mislead a future reader/bisector — please reword it to match the final model before merge.

packages/core/src/cdk/tracer-registry.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * If the app contains a `Tracer`, enable tracing on **every** compute in the
+ * stack. Runs once at the end of `create()` (after the backend module has
+ * imported, so all computes are registered). `Compute.enableTracing()` is
+ * idempotent, so this is safe regardless of how many Tracers exist.
+ *
+ * @param scope - Any construct in the stack (used to locate the stack).
+ */
+export function finalizeTracing(scope: Construct): void {
```
sarayev: Quick question on scoping: registerTracer sets the flag on cdk.Stack.of(scope) and finalizeTracing enables X-Ray on every compute in that stack. Since two BlocksBackends can share one cdk.Stack (there is a TwoBackendsStack test, and the fullId docstring points at Amplify Gen2's backend.createStack('blocks')), wouldn't a Tracer in backend A switch on costed X-Ray for backend B's computes? Could we key this on the owning BlocksStack/BlocksBackend (the same anchor defaults/executionRole use) and add a two-backends-in-one-stack test? Same concern applies to the dashboard registry.

packages/core/src/cdk/dashboard-registry.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/** A Dashboard's deferred widget-body build, run once after the app is constructed. */
+type DashboardFinalizer = () => void;
+
+/**
+ * Get or create the deferred-dashboard list for a given stack. Stored on the
+ * stack object (keyed by a Symbol), so each stack in a multi-stack synth gets
+ * its own — mirrors the config + compute registries.
+ */
+function getRegistry(stack: cdk.Stack): DashboardFinalizer[] {
```
sarayev, line 16: Cross-ref to the tracer-registry note: this registry is also keyed on cdk.Stack.of, so with two backends in one stack a Dashboard in one backend would render the other backend's computes too. Same fix: scope it to the owning backend.

packages/bb-tracer/src/index.cdk.ts
```diff
@@ -1,26 +1,21 @@
...
-			const cfnFunction = this.handler.node.defaultChild as CfnFunction;
-			cfnFunction.tracingConfig = { mode: 'Active' };
-
-			this.handler.addToRolePolicy(new PolicyStatement({
-				actions: ['xray:PutTraceSegments', 'xray:PutTelemetryRecords'],
-				resources: ['*'],
-			}));
+			// Tracing is presence-gated: creating a Tracer signals that the app
```
sarayev, line 25: This registerTracer(this) call is really the heart of the PR, but I could not find a CDK test covering it: registerTracer/finalizeTracing are not referenced in any test, and the dashboard cdk test calls compute.enableTracing() directly to avoid loading the real Tracer. Could we add a synth test that a constructed Tracer flips TracingConfig: Active on the compute(s) at finalize, is fleet-wide with more than one compute, and that enabled: false alone enables nothing?

packages/bb-logger/README.md
```diff
@@ -118,31 +118,32 @@ The logger handles edge cases gracefully:
...
 
-The shared handler log group already carries the stack-wide default retention
-(`defaults.logRetention` — one week in sandbox, one year in production). Set
-`retention` only to override it for this handler:
+**Logging is always on, and retention is a compute-level setting — not a Logger
+option.** Every compute captures its handler's stdout to its own CloudWatch log
+group, which carries the stack-wide default retention (`defaults.logRetention` —
+one week in sandbox, one year in production).
+
+To override retention, set it on the **compute**, not the Logger:
```
sarayev, line 128: I think this retention snippet is not runnable as written: it never imports LambdaCompute, LambdaCompute is @internal (not re-exported from @aws-blocks/blocks), and the umbrella builds the default compute with no options, so a customer cannot pass logRetention today. Since Logger({ retention }) was removed, does that mean customers lose retention control entirely? Could we either expose a customer-reachable retention seam, or reword the docs to say retention is not configurable yet and drop the snippet?

packages/bb-dashboard/DESIGN.md
```diff
@@ -76,9 +76,114 @@ Design document for Dashboard. For usage, see [README.md](./README.md).
 **Rationale:**
 - **When it fits** — Teams that want operational visibility into a deployed application without hand-building CloudWatch dashboards.
 - **When it does not** — Fully custom widget layouts are better served by the CloudWatch console directly; data-inspection admin UIs belong in `AdminSite`, not here. (complements D-DB-3, which covers why we lean on CloudWatch's native dashboard over a custom UI)
-- **Composition guidance** — Connect all three observability BBs (Metrics, Logger, Tracer) for full visibility; use `title` to distinguish dashboards across multi-stage deployments; keep the default widget set for standard apps and use `widgets` only for custom additions.
+- **Composition guidance** — Health + logs render for every compute automatically; add a `Tracer` anywhere in the app for the traces sections, and pass Metrics source(s) to the dashboard. Use the `logs` / `traces` toggles to hide sections and `computes` to restrict which computes appear; use `title` to distinguish dashboards across multi-stage deployments.
```
sarayev: This describes a public computes selector, but types.ts intentionally omits it (there is an explicit NOTE, and D-DB-10 agrees) to avoid leaking the internal Compute type. The same computes wording appears again around L320-321 and in the options table (~L362), and D-DB-11 says options.computes ?? getComputes(this) while the code just uses getComputes(this). Since DESIGN.md ships with the package, could we reword these to logs/traces only, or mark computes clearly as future work?

packages/core/src/cdk/dashboard-registry.ts
```diff
@@ -0,0 +1,66 @@
...
+ * imported. Clearing the list makes a repeated call a no-op, so a dashboard's
+ * body is never built twice.
+ *
+ * A Dashboard constructed outside `create()` (e.g. directly in a unit test) must
+ * call this explicitly before synth — the same way `config-registry.test.ts`
+ * drives `finalizeConfigRegistry`.
+ *
+ * @param scope - Any construct in the stack (used to locate the stack).
+ */
+export function finalizeDashboards(scope: Construct): void {
```
sarayev, line 63: Edge case: if a Dashboard (or Tracer) is constructed after create() has resolved, the finalizer list is already drained, so it silently produces no resource. Worth tracking a per-stack 'finalized' flag and either running the finalizer immediately or emitting a synth warning?

packages/core/src/cdk/tracer-registry.ts
```diff
@@ -0,0 +1,39 @@
...
+	const stack = cdk.Stack.of(scope);
+	if (!hasTracer(stack)) return;
+	for (const compute of getComputes(scope)) compute.enableTracing();
```
sarayev, line 46: Minor multi-compute nit: since all computes share one execution role, enabling tracing per compute adds N identical xray:PutTraceSegments statements to the same policy. Harmless at one compute, but this is the multi-compute path, so maybe dedupe the grant (guard on the role, or grant once at finalize)?

sarayev (commented):
Reviewed as a senior engineer against AGENTS.md and the docs/ guidelines, focusing on correctness, edge cases, security, performance, maintainability, and tests. Overall this is a clean, well-structured redesign: the finalize ordering (config -> tracing -> dashboards) is correct, enableTracing() is idempotent, and the single-compute path looks solid. I have left inline notes below: a couple I would like to sort out before merge (registry scoping across multiple backends, and a test for the fleet-wide tracing path), a doc/capability question on log retention, plus some smaller suggestions and nits. I framed several as questions, so please push back where I have misread the intent!

### Additional notes
- packages/bb-tracer/README.md:1 - Nit: DESIGN.md notes tracing is presence-gated and fleet-wide, but the README does not mention that constructing any Tracer enables X-Ray on every compute (a real cost implication). One line here would help users.

.changeset/observability-compute-driven.md
```diff
@@ -0,0 +1,25 @@
+---
+"@aws-blocks/core": minor
+"@aws-blocks/blocks": minor
```
sarayev, line 3: These are real public-API breaks (removed LoggerBBRef/TracerBBRef/RetentionDays exports, removed Logger.retention / BlocksDefaults.logLevel / LOG_LEVEL, narrowed DashboardOptions) shipped as minor. The changeset body flags them clearly, which is great. Is minor the intended channel for breaking changes on 0.x here, or should this be a major? Might be worth a maintainer sign-off, since AGENTS.md asks to surface removed exports.

packages/core/src/cdk/index.ts
```diff
@@ -52,288 +56,315 @@ export { ApiError, isBlocksError, hasAuthError, DEFAULT_API_ERROR_NAME } from '.
...
+	}
+	/** The default compute's API Gateway REST API. To be removed once consumers move to the multi-compute model. */
+	get gateway(): cdk.aws_apigateway.RestApi {
+		return this.requireDefaultCompute().apiGateway;
+	}
+	/** The default compute's RPC endpoint URL. To be removed once consumers move to the multi-compute model. */
+	get apiUrl(): string {
+		return this.requireDefaultCompute().apiUrl;
+	}
+	/** The default compute's handler CloudWatch log group. `bb-logger` reconfigures its retention. */
```
sarayev: Small doc drift: this still says bb-logger 'reconfigures its retention', but after this PR the CDK Logger is a no-op and retention comes from the compute's logRetention / defaults.logRetention. The same stale wording is also at index.ts L304, blocks-backend.ts L226, and bb-logger/src/index.aws.ts L36 (that last one still mentions the removed retention option). Mind updating all four?

packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -61,11 +68,11 @@ export class LambdaCompute extends Compute {
...
 			removalPolicy: cdk.RemovalPolicy.DESTROY,
 		});
```
sarayev, line 78: (re: line 122) logRetention bounds the handler log group, but the API Gateway access-log group still uses defaults.logRetention. Latent today since presets default accessLogging: false, but if someone turns on access logging and sets logRetention the two groups diverge. Apply the same options?.logRetention ?? defaults.logRetention here too, or note that logRetention only scopes the handler group?

.changeset/observability-compute-test-stubs.md
```diff
@@ -0,0 +1,13 @@
+---
```
sarayev, line 1: bb-async-job and bb-cron-job only changed their *.test.ts here (FakeCompute stubs), with no runtime/published code change. Is the patch bump intentional, or can this changeset be dropped so we do not republish them with no consumer-facing change?

packages/bb-dashboard/DESIGN.md
```diff
@@ -76,9 +76,114 @@ Design document for Dashboard. For usage, see [README.md](./README.md).
...
+
+One dashboard, grouped by compute, with metrics as a trailing app-wide section:
+
+```
+# <app> dashboard
+## Compute — api (Lambda)
+   health   (always)
+   logs     (only if a Logger targets this compute)
+   traces   (only if a Tracer targets this compute)
+## Compute — worker (Container)
```
sarayev, line 118: Nit: this labels logs as 'only if a Logger targets this compute', but the PR makes logs always-on per compute. Also the Target layout diagram (L117-119) orders health -> logs -> traces, while widgets.ts renders health -> traces -> logs (the README and the 'as implemented' section agree on traces-before-logs). Mind aligning the diagram?

packages/bb-dashboard/src/widgets.ts
```diff
@@ -328,39 +126,58 @@ function sectionHeader(text: string): IWidget[] {
 // ── Main builder ────────────────────────────────────────────────────────
 
 /**
- * Build the complete set of dashboard widget rows from resolved configuration.
+ * Build the complete set of dashboard widget rows.
```
sarayev, line 129: (re: line 100) Nit: a metrics source with no metricConfigs renders a placeholder graph with an empty metric name (a blank widget). Could we skip the metrics section when there are no configs?

packages/bb-lambda-compute/src/index.cdk.ts
Simone319: Fixed in 31217434 — the API Gateway access-log group now uses the same `options?.logRetention ?? this.defaults.logRetention` fallback as the handler log group, so setting `logRetention` on the compute retunes both groups rather than only the handler one.

packages/core/src/cdk/tracer-registry.ts
Simone319: You're right, and it's broader than tracer + dashboard: the **compute** and **config** registries key on `cdk.Stack.of` too, so a one-registry fix would only half-solve it. (Config is arguably the worst — its stack-level `finalized` flag makes a second backend's `finalizeConfigRegistry` a no-op, so that backend's computes never get `BLOCKS_CONFIG_BUCKET`/`KEY` stamped.)

The normal topology is one `aws-blocks/` module = one backend = one `cdk.Stack`, so the blast radius today is effectively nil. Rather than patch one registry here, we're deferring a uniform fix that re-anchors all four registries (compute/config/tracer/dashboard) on the owning `BlocksStack`/`BlocksBackend`. Tracking issue: https://github.com/aws-amplify/private-amplify-backend-staging/issues/1111. Thanks for the cross-ref to #391.

packages/bb-dashboard/src/index.cdk.ts
Simone319: Fixed in 31217434 — the `CwDashboard` resource is now created **eagerly** in the constructor; only the widget *body* is deferred to the finalizer. So the URL `CfnOutput`, redirect route, and `registerConfig` URL always point at a real resource even when `finalizeDashboards` isn't driven (a Dashboard constructed outside `create()`, or a test that forgets to). In that case only the widget body is left empty — nothing dangles.

packages/bb-tracer/src/index.cdk.ts
Simone319: Agreed it's a footgun — keeping the presence-gated model but documenting it clearly. As of 31217434 the class docstring and the `enabled` JSDoc spell out that `enabled: false` only removes *this* Tracer's vote; it is **not** a global off switch, and the only way to keep X-Ray off is to construct no `Tracer` at all.

packages/core/src/cdk/tracer-registry.ts
Simone319: Good catch — yes, today a `Tracer` in backend A would switch on costed X-Ray for backend B's computes when both share one `cdk.Stack`, because presence + `getComputes` + `finalizeTracing` all key on `cdk.Stack.of`. In the normal topology (one backend module = one backend = one `cdk.Stack`) they coincide, so this only bites the exotic two-backends-in-one-stack case (the `TwoBackendsStack` test / the Amplify `createStack('blocks')` path in the `fullId` docstring).

It's the same root cause as the compute and config registries, so we're deferring a uniform fix rather than patching just the tracer. Tracking issue: https://github.com/aws-amplify/private-amplify-backend-staging/issues/1111.

packages/core/src/cdk/dashboard-registry.ts
Simone319: Same root cause as the tracer-registry note — both key on `cdk.Stack.of`. We're deferring a uniform re-anchoring of all four registries (compute/config/tracer/dashboard) onto the owning backend rather than fixing one in isolation. Tracking issue: https://github.com/aws-amplify/private-amplify-backend-staging/issues/1111.

packages/bb-tracer/src/index.cdk.ts
Simone319: Added in 31217434 — `packages/bb-lambda-compute/src/tracing.cdk.test.ts`. It drives the core seams (`registerTracer` + `finalizeTracing`) and asserts against the synth template that: (1) tracing flips `TracingConfig: Active` on the compute at finalize, (2) it's fleet-wide across two computes, and (3) an app with no Tracer leaves `TracingConfig` untouched. It calls the core seams directly rather than constructing a real `Tracer` for the same reason the dashboard cdk test calls `enableTracing()` directly — bb-tracer's `cdk` export condition can't be activated at ESM import time in the shared mock-conditioned test process, so importing it would resolve the local-mock variant and never touch the compute.

packages/bb-logger/README.md
Simone319: Reworded in 31217434 — the snippet no longer imports the internal `LambdaCompute`. It now shows the only retention knob customers have today: overriding `logRetention` on the stack-wide `defaults` preset, which applies to every compute's handler log group. And no — customers don't lose retention control. The README now states explicitly that a per-compute `logRetention` prop arrives with the public compute-configuration surface; until then `defaults.logRetention` is the knob.

packages/bb-dashboard/DESIGN.md
Simone319: Fixed in 31217434 — swept DESIGN.md to match `types.ts`. Removed the public `computes` selector wording (the options table and the ~L320 mention), and corrected D-DB-11: the code just uses `getComputes()` (there is no `options.computes ?? getComputes(this)`), and the finalizer covers every compute in the app.

packages/core/src/cdk/dashboard-registry.ts
Simone319: Addressed the dangling-resource half directly in 31217434: the `CwDashboard` is now created eagerly in the constructor, so a Dashboard (or the dashboard reading a late Tracer's state) constructed after `create()` has drained still produces a real resource — its URL/redirect never dangle, only its widget body is left empty.

We deliberately did **not** add a per-stack `finalized` flag + run-immediately/warn: it broke the supported "construct after `create()`, then re-run `finalizeDashboards`" pattern the unit tests rely on, and the eager resource removes the actual footgun you flagged.

packages/core/src/cdk/tracer-registry.ts
Simone319: Done in 31217434 — moved the grant out of the per-compute `applyTracing` and into `finalizeTracing`, which adds `xray:PutTraceSegments`/`PutTelemetryRecords` **once** on the shared execution role. `applyTracing` now only flips the function's `TracingConfig`. So N traced computes no longer add N identical statements to the same policy.

.changeset/observability-compute-driven.md
Simone319: Flagging for maintainer sign-off on the channel. Per the repo's pre-1.0 convention, breaking changes have been shipping as **minor** on 0.x, and the changeset body calls out each break explicitly. Happy to bump this to **major** if that's the preferred policy — just want a maintainer decision since it's a judgment call.

packages/core/src/cdk/index.ts
Simone319: Fixed all four in 31217434 — `core/src/cdk/index.ts` (both spots), `blocks-backend.ts`, and `bb-logger/src/index.aws.ts:36` (which also dropped the mention of the removed `retention` option). They now describe the compute-owned handler log group with retention coming from `logRetention`/`defaults.logRetention`.

packages/bb-lambda-compute/src/index.cdk.ts
Simone319: Fixed in 31217434 — the access-log group now applies `options?.logRetention ?? this.defaults.logRetention`, same as the handler group, so the two no longer diverge when someone enables access logging and sets `logRetention`.

.changeset/observability-compute-test-stubs.md
Simone319: Can't drop it — the repo's changeset-coverage guard fails CI if a changed package (even test-only) has no changeset. The bump is a patch, and the changeset body notes it's a test-only `FakeCompute`-stub adaptation with no consumer-facing change. Open to a maintainer override if we'd rather exclude test-only diffs from the guard.

packages/bb-dashboard/DESIGN.md
Simone319: Fixed in 31217434 — updated the label to reflect that logs are always-on per compute, and re-ordered the Target-layout diagram to health → traces → logs to match `widgets.ts` (and the README / "as implemented" section).

packages/bb-dashboard/src/widgets.ts
Simone319: Done in 31217434 — `widgets.ts` now skips the metrics section entirely when a source has no `metricConfigs`, instead of emitting a placeholder graph with an empty metric name.

osama-rizk (approved):

sarayev (approved):
Thanks for the quick turnaround, Simone! Went through the latest revision and the fixes all look good: the fleet-wide tracing test, the single X-Ray grant on the shared role, the eager dashboard resource, the matching logRetention fallback on the access-log group, and the doc cleanups. Happy to defer the cross-backend registry scoping to the uniform follow-up since it's tracked. Approving.

PR #510, opened 2026-09-08
feat(hosting): swappable, removable & composable front door — CloudFront.

osama-rizk:
## Why

The **front door** — the public entry point that terminates TLS, caches, applies WAF, serves a custom domain, and routes requests to the app's static assets, SSR/image compute, and backend API — was hard-coded to a single global CloudFront distribution. That made it a load-bearing wall rather than a choice: you couldn't drop it for a cheap static site, put the app behind an ALB you already run, use a serverless API Gateway, or stack CloudFront in front of a regional router.

This PR makes the front door **removable, swappable, and nestable** through one optional prop, so an app chooses the door it wants — or none — from the same code, in its own AWS account.

## What changed

- A single optional **`frontDoor`** prop on `Hosting` selects the door. **Omit it and nothing changes** — you get today's global CloudFront door (CloudFormation byte-identical; no migration).
- The app's own resources (static assets, SSR/image compute, backend API) are **identical across every door** — only the front door in front of them changes.
- The chosen door is checked **at synth against what the app actually needs** (inferred from the build output + `Hosting` props). If the door can't serve a needed capability, the deploy **fails at synth** with a message naming the capability and a door that supports it — never a silent failure in production. A door that merely lacks a capability the app never asked for is a clean fit; a reduced-but-usable capability must be accepted explicitly (`degrade`).
- Fully backward compatible: the default is CloudFront, every addition is optional, and no existing exports or signatures changed.

---

### The one knob: `frontDoor`

`HostingProps` gains a single optional `frontDoor` prop:

```ts
interface HostingProps {
  // …existing props…
  /** The front door. Omit → 'cloudfront' (the default global CDN). */
  frontDoor?: FrontDoor;
}

type FrontDoor =
  | 'cloudfront'                                  // default — global CloudFront CDN
  | 'none'                                        // no front door — served directly from S3 website hosting
  | { kind: 'alb'; vpc?; internal?; certificate?; degrade? }         // regional ALB, no CDN
  | { kind: 'apiGateway'; api?: 'rest' | 'http'; certificate?; degrade? } // API Gateway, no CDN
  | { edge: 'cloudfront'; router: 'alb'; vpc?; internal? };          // CloudFront edge over an ALB router
```

| Value | Selects | Why it exists |
|---|---|---|
| _(omitted)_ / `'cloudfront'` | Global CloudFront CDN over the app (today's behavior) | Correct, safe default — TLS, SSR, auth, static, WAF, custom domain all work. |
| `'none'` | **No front door** — served directly from S3 website hosting (public bucket, HTTP) | Cheapest static/SPA (internal tools, previews). An S3 bucket is an *origin*, not a front door; `s3-website` is the internal mechanism. |
| `{ kind: 'alb' }` | Regional **ALB**, no CloudFront | Behind an existing ALB, regional-only, or private deploys — no global edge needed. |
| `{ kind: 'apiGateway' }` | Amazon **API Gateway**, no CloudFront | Serverless sibling of the ALB door — no VPC/NAT, scale-to-zero, pay-per-request. `api: 'http'` (default) is rootless; `api: 'rest'` is for use behind a custom domain / CloudFront. |
| `{ edge: 'cloudfront', router: 'alb' }` | CloudFront edge **in front of** a regional ALB router | Keep the global edge **and** a regional router — the AWS one-click "ALB + CloudFront" shape, as IaC. |

`degrade?: CapabilityId[]` (on the `alb` / `apiGateway` doors) is a conscious opt-in to a capability a door provides only in reduced form — without it, the synth-time check fails rather than silently dropping the capability.

### Examples

```ts
// Default — global CloudFront CDN (unchanged; omit frontDoor)
new Hosting(stack, 'Web', { root, api: blocksStack });

// Cheapest static/SPA — no front door (served directly from S3, HTTP)
new Hosting(stack, 'Web', { root, api: blocksStack, frontDoor: 'none' });

// Regional ALB (no CloudFront) — behind existing infra / private / regional-only
new Hosting(stack, 'Web', { root, api: blocksStack, frontDoor: { kind: 'alb' } });

// Serverless API Gateway (no CloudFront) — no VPC, scale-to-zero
new Hosting(stack, 'Web', { root, api: blocksStack, frontDoor: { kind: 'apiGateway' } });

// CloudFront edge over a regional ALB router
new Hosting(stack, 'Web', { root, api: blocksStack, frontDoor: { edge: 'cloudfront', router: 'alb' } });
```

### Use cases → value

| Use case | `frontDoor` |
|---|---|
| Simplest static site / SPA, cheapest possible (internal tools, previews) | `'none'` |
| Public static / SPA **or** SSR on a global CDN (custom domain, WAF) — today's default | `'cloudfront'` |
| Behind an existing ALB / regional / private, no global edge | `{ kind: 'alb' }` |
| Serverless, no CDN and no always-on VPC/NAT baseline | `{ kind: 'apiGateway' }` |
| Keep the global edge **and** a regional API router | `{ edge: 'cloudfront', router: 'alb' }` |

### Capability trade-offs when you switch the front door

Every door renders the same app; what differs is which capabilities it can serve. The synth-time check makes any loss **explicit** — a capability the app *demands* that the door can't provide fails the build unless accepted via `degrade`. Baseline is **`'cloudfront'`**, which supports everything the standard way.

Legend: ✅ full · ◑ different mechanism, full · ⚠️ reduced (opt-in via `degrade`, else synth fails) · ❌ unsupported (synth fails if demanded)

| Capability (vs CloudFront) | `'none'` (S3) | `{ kind: 'alb' }` | `{ kind: 'apiGateway' }` | `{ edge:'cloudfront', router:'alb' }` |
|---|:--:|:--:|:--:|:--:|
| Global edge cache | ❌ | ⚠️ | ⚠️ | ✅ |
| SSR (server render) | ❌ | ✅ | ✅ | ✅ |
| Response streaming | ❌ | ✅ | ❌ | ✅ |
| Same-origin API (no CORS) | ❌ | ✅ | ✅ | ✅ |
| Custom domain + TLS | ❌ | ✅ | ✅ | ✅ |
| Security / per-route response headers | ❌ | ⚠️ | ⚠️ | ✅ |
| WAF | ❌ | ✅ (regional) | ⚠️ | ✅ (edge) |
| Geo restriction | ❌ | ❌ | ❌ | ✅ |
| Atomic release / skew-pin | ⚠️ / ❌ | ◑ / ⚠️ | ◑ / ⚠️ | ◑ / ⚠️ |
| Image optimization | ❌ | ✅ | ✅ | ✅ |
| Access logging | ❌ | ✅ | ❌ | ✅ |
| Long request (>29 s) | ❌ | ✅ | ❌ | ✅ |
| Large payload (>10 MB) | ❌ | ⚠️ (1 MB) | ⚠️ (10 MB) | ⚠️ (1 MB) |

One-line summary of each switch:
- **`'none'`** — static hosting only: no TLS, edge cache, SSR/image, same-origin API, WAF, or atomic deploy. Cross-origin cookie auth can't work over HTTP.
- **`{ kind: 'alb' }`** — lose the global edge cache, per-route/security header injection, skew-pinning, and geo (CloudFront-edge features); HTTP-only without a cert. Keep SSR, streaming, same-origin API, regional WAF, image-opt.
- **`{ kind: 'apiGateway' }`** — like ALB, plus no response streaming, ~29 s / ~10 MB request caps. Gains: serverless, no VPC/NAT, scale-to-zero.
- **`{ edge:'cloudfront', router:'alb' }`** — no net loss at the edge (CloudFront still provides cache, TLS, WAF, geo, custom domain, security headers). Cost: the ALB's always-on baseline + one extra hop.

### Deliberately NOT added
- **No public Lambda Function URL door** — a Function URL can't front SSR (OAC + non-empty POST/PUT breaks SigV4), so it's not a public option.
- **No cost/latency tuning knobs** — the front door is a small, honest, discoverable choice, not a tuning surface.
- **No new required fields / no default change** — omitting `frontDoor` is byte-identical to today.

---

## Verification (deployed sample apps)

Each door was deployed to a sandbox as its own sample app and driven with Playwright — the full self-check probe suite (routing / static / caching / headers / API) plus the complete AuthBasic sign-up → session → protected call → sign-out lifecycle:

| Door | Probe suite | Auth lifecycle |
|---|:--:|:--:|
| `'cloudfront'` | 11 / 11 | 4 / 4 |
| `{ kind: 'alb' }` | 10 / 1¹ | 4 / 4 |
| `{ kind: 'apiGateway' }` (HTTP API) | 10 / 1¹ | 4 / 4 |
| `{ edge:'cloudfront', router:'alb' }` | 11 / 11 | 4 / 4 |
| `'none'` (S3 website) | —² | —² |

¹ The single "miss" is the security-headers probe — the expected, explicit trade-off (an ALB / API Gateway can't inject per-response security headers; the app doesn't demand them, so the check treats it as a clean fit, not a failure).
² Not verifiable in the test account: an org S3 Block-Public-Access guardrail auto-remediates public buckets (denies public bucket policies), so the public-bucket `'none'` door can't be served there — an environment constraint, not code. `'none'` remains a dev/permissive-account door (and its cross-origin cookie auth is unsupported over HTTP by design).

Unit + synth tests cover every door and the synth-time capability check; the default CloudFront path is guarded byte-identical.

---

## Backward compatibility

`frontDoor` defaults to `'cloudfront'`; the default path is CloudFormation byte-identical; every new export is additive; no signatures changed. No breaking changes.

## Follow-ups (out of scope here)
- Non-CloudFront doors building their own logging / WAF / alarms / error pages (today those tiers are honestly `unsupported`, so a demanding app is rejected, not silently served without them).
- Private / VPC origins (CloudFront → internal ALB → EKS) — needs CloudFront VPC Origins.
- Container (ECS/EKS) origins behind the door; a shared multi-tenant API Gateway door.


packages/hosting/src/constructs/alb_construct.ts
```diff
@@ -0,0 +1,206 @@
...
+    // the API is same-origin with the frontend (session cookies flow, no CORS). A
+    // lone `'*'` origin is the single-compute case (`/aws-blocks/*` + auth subtree
+    // → one backend); a named namespace routes `/aws-blocks/api/{ns}/*`. These
+    // rules win over the route/catch-all rules (lowest priority numbers).
+    let apiPriority = API_PRIORITY_BASE;
+    for (const origin of plan.backend?.origins ?? []) {
+      const ns = origin.namespace;
+      const paths = ns === '*' ? ['/aws-blocks/*', '/aws-blocks-auth/*'] : [`/aws-blocks/api/${ns}/*`];
+      const idSuffix = ns === '*' ? 'Default' : ns;
+      const apiProxy = new LambdaFunction(this, `ApiProxy${idSuffix}`, {
```
Simone319, line 141: So we have `ALB -> Lambda -> APIGW -> Lambda`? 
It will increase the latency as both Lambda can encounter cold start. Can we skip the middle layer and have ALB -> Lambda directly?

When we introduce container to the compute, there will be an  `ALB -> Lambda -> ALB -> Container` if we have container? 
For customers who use container for lower API latency, the lambda in the front will re-introduce cold start. Can we reuse the ALB if the backend has one already? So that `ALB -> Container`  remains.

In general I'm trying to reduce the layers in the sandwich, as each layer introduce more surface for tuning, and things that can go wrong.

packages/hosting/src/constructs/alb_construct.ts
```diff
@@ -0,0 +1,206 @@
...
+    super(scope, id);
+    const { plan, bucket } = props;
+    const buildId = plan.release.buildId;
+    const compute = props.computeFunctions ?? new Map<string, IFunction>();
+
+    const serverFn = props.serverComputeName ? compute.get(props.serverComputeName) : undefined;
+    const imageFn = props.imageComputeName ? compute.get(props.imageComputeName) : undefined;
+
+    // ── VPC (BYO or a default 2-AZ VPC) ──
+    this.vpc = props.vpc ?? new ec2.Vpc(this, 'Vpc', { maxAzs: 2, natGateways: 1 });
```
Simone319, line 70: that auto-created VPC always includes the private-with-egress subnets and therefore **an idle NAT Gateway** that nothing in the stack uses — the ALB is public and the Lambdas aren't VPC-attached. So omitting vpc silently costs you a NAT Gateway (~hourly + per-GB) for no functional benefit.

packages/hosting/src/constructs/function_url_construct.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Lambda Function URL front-door construct — the cheapest HTTPS, no-CloudFront
+ * door for a STATIC site / SPA.
+ *
+ * A Function URL exposes ONE Lambda over a built-in HTTPS endpoint
+ * (`https://<id>.lambda-url.<region>.on.aws/`) with no ALB, no NAT, no API
+ * Gateway — pay-per-request, scale-to-zero. Because it is a single origin with
+ * no path routing, this door serves a static/SPA site through one asset-proxy
```
Simone319, line 8: `single origin with no path routing` Does it mean in this mode the lambda will not serve as the frontdoor to route to backend API?

packages/hosting/src/constructs/apigw_construct.ts
```diff
@@ -0,0 +1,130 @@
...
+};
+
+/** Convert a route-table glob pattern to an API Gateway HTTP API route path (or null for the catch-all). */
+const toApiGwPath = (pattern: string): string | null => {
+  if (pattern === '/*' || pattern === '*') return null; // → $default
+  if (pattern.endsWith('/*')) return `${pattern.slice(0, -2)}/{proxy+}`;
+  return pattern; // exact
+};
+
+export class ApiGatewayConstruct extends Construct {
```
Simone319, line 46: In general, here's some cons on using APIGW as a frontdoor for routing:
~29 s hard request timeout — long-running work (agent loops, big batch jobs) can't pass through it.

   - 10 MB payload cap — large uploads/downloads can't pass through it.

   - Bills per request at the router, on top of each backend API compute's own ingress cost.

   - Regional — no global edge.

   - A private ALB origin needs an NLB (REST API VPC Link) or a switch to HTTP API.

   - Non-text payloads need explicit binary-media-type / base64 handling.

   - Out of the box, metrics are only API/stage aggregates; a per-namespace (per-resource) latency/error breakdown requires enabling Detailed CloudWatch Metrics (opt-in, billed extra), whereas a per-compute ingress emits that at the API level for free.

   - Stage-wide config (WAF, TLS/domain, access logs, default throttle) applies to every origin at once — it can't differ per compute.

   - Shared blast radius — a bad router deploy, a WAF change, or hitting the account-level API Gateway throttle degrades every namespace together.

   - Its throttle/log/WAF config is a cross-cutting resource that must encode every origin's needs.

packages/hosting/src/constructs/alb_construct.ts
osama-rizk: Great push — we dug into this. You're right that today it's `ALB → forwarder Lambda → backend API GW → backend Lambda`, and the forwarder is the layer to attack. Findings:

**Why the forwarder exists today:** an ALB target group can only target Lambda / IP / instance — not an arbitrary HTTPS URL. The backend is exposed only as a URL token (`apiUrl`) to the hosting layer, so the forwarder is the only thing that can relay `/aws-blocks/*` to it while keeping the API same-origin (so `SameSite=Lax` cookies flow, no CORS).

**Can we skip it? Yes — direct-target the backend Lambda** (`targets.LambdaTarget(backendFn)`), which also lets us drop the **backend API Gateway entirely** on this door — not just the forwarder. Two hops removed, not one. Crucially, ALB→Lambda has no SigV4/OAC request-body problem (the reason we *can't* do the equivalent on the pure-CloudFront door, where CF→Lambda-URL OAC fails to sign POST bodies and 403s the whole JSON-RPC API). What it takes:
- The backend Lambda's **request** contract is already ALB-compatible (it reads `event.path`/`httpMethod`/`multiValueHeaders`/`multiValueQueryStringParameters`/`body`/`isBase64Encoded`).
- The one gap is the **response**: the handler emits single-value `headers` for most responses and `multiValueHeaders` only for `Set-Cookie`, so we'd normalize it to always emit `multiValueHeaders` (else auth cookies drop under a multi-value target group).
- Wiring: widen `BlocksStackApi` to carry the `IFunction` (or accept the stack and read `.handler`) + add the already-stubbed `BackendIngress kind:'lambda'` and thread it through.

**Container case:** same seam. `ALB → forwarder → ALB → Container` collapses to `ALB → Container` by reusing the backend's existing ALB/target — a `BackendIngress kind:'ip'|'alb'` variant on the exact same widening, so no forwarder and no cold-start re-introduction for container backends.

### Can we drop the backend API Gateway? Per front door
The deciding factor is how the door reaches the backend Lambda, and whether that path forces SigV4/OAC request-body signing (which breaks our POST-only JSON-RPC API).

| Front door | Path to backend | SigV4/OAC body problem? | Can drop backend API GW? |
|---|---|---|---|
| **`none`** | browser → API GW **directly** | n/a — API GW *is* the public endpoint | ❌ No |
| **CloudFront** (default) | CF → HTTP origin | **Yes** — CF→Lambda-URL OAC won't sign POST bodies → 403 | ❌ No |
| **ALB** | ALB → **Lambda target group (direct)** | No — ALB invokes the Lambda; no SigV4 | ✅ Yes |
| **API Gateway** (door) | front API GW → **Lambda proxy integration** | No — AWS_PROXY invokes Lambda natively | ✅ Yes |
| **CF → ALB** (composed) | CF → ALB → **Lambda target (direct)** | No — CF talks to the ALB (plain custom origin); ALB→Lambda direct | ✅ Yes |

**Why the ❌ two can't drop it:** `none` has no other public endpoint — the API Gateway *is* the reachable HTTPS surface. Pure CloudFront can't target a Lambda directly; its only private-Lambda option is a Lambda Function URL, and CF's OAC signs the origin request with SigV4 whose payload hash covers the body — OAC doesn't sign the body, so every POST 403s. The escape (`AuthType: NONE`) makes the Function URL a public, CF-bypassing endpoint. Either way you need a real HTTP origin, which is what the backend API Gateway is.

**Why the ✅ three can:** each reaches the backend through a layer that invokes the Lambda directly (ALB target, or API-Gateway AWS_PROXY integration) — the body rides natively, no SigV4 mismatch. For CF→ALB, CF only ever talks to the ALB (a normal custom origin), so the OAC-body problem never arises.

### Pros of eliminating the forwarder + backend API Gateway
- **Fewer hops → lower latency**: removes a full Lambda invoke (and its cold-start surface) plus the API Gateway proxy from every request.
- **Lower cost**: the backend is REST API Gateway v1 (~$3.50/M requests, the priciest tier) *plus* the forwarder's own Lambda invocations — both gone.
- **No 29s API-Gateway timeout / 10 MB cap** on the API path.
- **Simpler mental model**: one fewer resource to tune (throttle/stage/logging) and one fewer thing that can fail.
- **Multi-compute-friendly**: the door becomes the single router that direct-targets N Lambdas/containers — zero per-compute API Gateways.

### Cons / what we accept
- **Cross-stack coupling**: backend and hosting are frequently separate stacks; a direct `LambdaTarget` hard-links them (a backend replacement now ripples into the hosting stack).
- **Public `core` contract change**: `BlocksStackApi` must carry the `IFunction` + a shared-runtime response-normalization tweak → maintainer sign-off.
- **Lose per-stage throttle**: API Gateway's stage throttle goes away; ALB WAF rate-based rules cover abuse but at a different granularity.
- **Response-contract correctness risk**: if the handler isn't normalized to `multiValueHeaders`, multiple `Set-Cookie` on the auth path collapse — must be done carefully with an e2e cookie test.

**Net:** the "reduce the sandwich" direction is right and the seam is already in place, but the real unlock is a **compute-layer change** — make the backend's ingress *optional* so a front door that adopts the Lambda/container directly suppresses the redundant API Gateway (droppable on ALB / API-Gateway / CF→ALB; kept on `none` / pure-CloudFront). That's beyond this PR's scope (touches `core` public API + the shared handler + multi-compute), so I'd like to land it as a fast-follow rather than expand this one. I'll file it with this write-up. Sound good?

packages/hosting/src/constructs/alb_construct.ts
osama-rizk: Good catch — fixed in `fed0a476`. You're exactly right: the public ALB lives in the public subnets and none of the front-door Lambdas are VPC-attached, so the `PRIVATE_WITH_EGRESS` tier and its NAT Gateway were billed-but-idle.

The default VPC is now **public-only** — no private subnets, `natGateways: 0`:

```ts
this.vpc =
  props.vpc ??
  new ec2.Vpc(this, 'Vpc', {
    maxAzs: 2,
    natGateways: 0,
    subnetConfiguration: [{ name: 'public', subnetType: ec2.SubnetType.PUBLIC, cidrMask: 24 }],
  });
```

A future VPC-attached backend (container / private Lambda) that needs egress can bring its own VPC with a NAT deliberately. Added a regression test asserting `AWS::EC2::NatGateway` count is 0 for the auto-created VPC, and a changeset note.

packages/hosting/src/constructs/function_url_construct.ts
osama-rizk: Two things here:

**Direct answer to your question:** yes — a Lambda Function URL is a *single* origin with no path-based routing, so in that mode the door could only serve the static/SPA site through one asset-proxy Lambda. It could **not** route `/aws-blocks/*` to the backend API — there's no way to fan a Function URL out to a second origin. To reach a private backend you'd also hit the same CF/OAC POST-body signing problem we discussed on the ALB thread.

**But this comment is now on deleted code** — we removed the Function URL door entirely in `5e265d4` (`feat(hosting)!: remove the Function URL front door`). The deciding reasons:
- A standalone browser-facing Function URL must be `authType: NONE` (a browser can't SigV4-sign; `AWS_IAM`/OAC only works with CloudFront in front) — so it's a public, unauthenticated endpoint.
- Provisioning a public Function URL is **auto-remediated as a security event (Sev-2 risk) in guardrailed accounts**, so it's not a viable door in practice.

For the cheap-static use case we point people at **s3-website** (dev/niche) or **CloudFront → S3 + OAC** (prod) instead. Resolving this thread as obsolete.

packages/hosting/src/constructs/apigw_construct.ts
osama-rizk: All accurate — and this is exactly why API Gateway is a **non-default, gated door**, not a recommendation over CloudFront. A couple of points:

**Cost isn't the pitch.** To be clear, we're *not* positioning this as cheaper than CloudFront — it isn't (HTTP API ≈ $1/M, REST ≈ $3.50/M vs CF ≈ $0.75–1/M + transfer), and as you note it bills at the router on top of each backend's ingress. I'll soften the "cheap" wording in the docstring.

**What it's actually for:** a **regional, CDN-free** entry point — teams/regions that can't or won't run CloudFront (data-residency / in-region-only compliance, GovCloud, CDN-averse security postures), or apps already standardized on API Gateway. "No global edge" is a latency con but a data-residency pro.

Issue #508, opened 2026-09-08
Align sibling conflict-class errors (23505 UniqueConstraintViolation) to 409 across data BBs


soberm:
### Summary

Follow-up from PR #500 (which maps optimistic-concurrency / conditional-write conflicts to JSON-RPC **409** instead of 500). A sibling conflict-class error is still returned as **500**: `23505` **UniqueConstraintViolation** (duplicate key).

In `packages/bb-data/src/engines/pg-error-translator.ts` and `packages/bb-data/src/engines/data-api-engine.ts`, the `23505` path does `e.name = DatabaseErrors.UniqueConstraintViolation; throw e;` → a plain `Error` → the JSON-RPC serializer emits **code 500** (only `ApiError` carries an HTTP status; see `packages/core/src/rpc.ts`).

A duplicate-key violation is a **client-side** conflict, not an internal server error — arguably **409 (Conflict)** (or 400). This is the same "5xx for a 4xx cause" smell PR #500 fixed for OCC/40001, just for a different SQLSTATE.

### Scope

PR #500 was deliberately scoped to OCC / `40001` serialization conflicts. This issue tracks aligning the remaining conflict-class errors across the data BBs.

### Proposed fix

Wrap the `23505` throw sites in `ApiError(<fixed message>, 409, { name: DatabaseErrors.UniqueConstraintViolation, cause: e })` (mirroring the OCC→409 pattern from PR #500), preserving `error.name` for `isBlocksError()`. Audit for any other SQLSTATE classes that are client-caused conflicts vs genuine infra failures.

Note: `ConnectionFailed` and similar infra errors correctly stay 500 — this is only about client-caused conflict classes.

### References
- PR #500 (OCC → 409)
- Raised by @osama-rizk in review of PR #500

PR #511, opened 2026-09-08
fix(data-bbs): map 23505 UniqueConstraintViolation to JSON-RPC 409 (#508)

soberm:
## Problem

Follow-up to #500. Duplicate-key / unique-constraint violations (SQLSTATE **`23505`**, `UniqueConstraintViolation`) were thrown as plain named `Error`s, so the JSON-RPC serializer (`packages/core/src/rpc.ts`) mapped them to code **500** — the same "5xx for a 4xx cause" smell #500 fixed for OCC/`40001`. A duplicate key is a **client-caused conflict**, so it should be **409 (Conflict)**.

**Issue #, if available:** #508

## Changes

Mirrors the OCC→409 pattern from #500.

- **`@aws-blocks/bb-data`** — new `uniqueConstraintConflict(cause)` helper (parallel to `serializationConflict`) returning `ApiError('The item violates a unique constraint', 409, { name: DatabaseErrors.UniqueConstraintViolation, cause })`. Routed through it at every `23505` site: `pg-error-translator.ts`, `pglite-engine.ts`, and `data-api-engine.ts` (both the SQLState-parsed `code === '23505'` path and the message-matched duplicate-key path).
- **`@aws-blocks/bb-distributed-data`** — `translateDsqlError`'s `23505` branch now throws the equivalent 409 `ApiError` (mirrors its existing `40001` branch).

`error.name` is preserved end-to-end so `isBlocksError(e, DatabaseErrors.UniqueConstraintViolation)` keeps matching on both server and client. The raw driver error is kept only as the server-side `cause` (non-enumerable via `ApiError`; the serializer never emits it), and the client-visible message is a fixed string (no column/driver text leak). `retriable` is `false` — a blind retry of a duplicate insert fails identically.

**Out of scope / unchanged:** `ConnectionFailed`, `QueryFailed` and other genuine infra errors correctly stay **500**; the `40001` serialization/OCC path (already 409 from #500) is untouched.

## Validation

- **TDD**: unit tests updated from the old plain-`Error`/name-only assertions to assert `ApiError` + `status === 409` + preserved name (`pg-error-translator.test.ts`, `pglite-engine.test.ts`, `data-api-engine.test.ts` ×3, `bb-distributed-data/errors.test.ts`) — confirmed red before the fix, green after.
- **Over-the-wire e2e** (per AGENTS.md §11 — serialization/behavior change): `test-apps/comprehensive/test/database.test.ts` and `dsql.test.ts` provoke a real duplicate-key insert and assert `e instanceof ApiError && e.status === 409 && isBlocksError(e, …UniqueConstraintViolation)`. No type casts. The PGlite/DSQL mocks enforce the primary key, so the path is genuinely exercised locally.
- Local: `npm run build`, `lint`, `lint:deps`, changed-package unit tests (bb-data 381/0, bb-distributed-data 164/0), and both `changeset-guard` checks (`verify-coverage`, `verify-umbrella`) pass.
- Changeset: `.changeset/unique-violation-409.md` — `@aws-blocks/bb-data` + `@aws-blocks/bb-distributed-data` `minor` (pre-1.0 breaking = minor); `@aws-blocks/blocks` `patch` (umbrella).

## Checklist

- [x] PR description included

---

### DIFF
--- a/packages/bb-data/src/engines/pg-error-translator.test.ts
+++ b/packages/bb-data/src/engines/pg-error-translator.test.ts
@@ -3,10 +3,11 @@
 
 /**
  * Unit tests for the shared PostgreSQL error translator (used by PgClientEngine
- * and PGliteEngine). Focus: an OCC / serialization-failure conflict (SQLSTATE
- * 40001) must surface as an ApiError with HTTP status 409 (Conflict), not a
- * generic 500, while preserving the SerializationFailure name and flagging the
- * conflict retriable.
+ * and PGliteEngine). Focus: conflict codes must surface as an ApiError with an
+ * HTTP status (409 Conflict), not a generic 500 — an OCC / serialization-failure
+ * conflict (SQLSTATE 40001, retriable) and a duplicate-key / unique-constraint
+ * violation (SQLSTATE 23505, not retriable) — while preserving the standardized
+ * error name so `isBlocksError` keeps matching.
  */
 import { test } from 'node:test';
 import assert from 'node:assert';
@@ -28,14 +29,18 @@ test('translatePgError: serialization failure (40001) → ApiError status 409, r
   );
 });
 
-test('translatePgError: unique violation (23505) → UniqueConstraintViolation (unchanged, non-ApiError)', () => {
-  const err = Object.assign(new Error('duplicate key'), { code: '23505' });
+test('translatePgError: unique violation (23505) → ApiError status 409, name preserved, not retriable', () => {
+  const err = Object.assign(new Error('duplicate key value violates unique constraint "t_pkey"'), { code: '23505' });
   assert.throws(
     () => translatePgError(err, 'PgClientEngine'),
     (e: unknown) => {
-      assert.ok(e instanceof Error);
-      assert.strictEqual((e as Error).name, DatabaseErrors.UniqueConstraintViolation);
-      assert.ok(!(e instanceof ApiError), 'unique violation should not be remapped to an ApiError');
+      assert.ok(e instanceof ApiError, 'expected an ApiError');
+      assert.strictEqual(e.status, 409);
+      assert.strictEqual(e.name, DatabaseErrors.UniqueConstraintViolation);
+      assert.notStrictEqual(e.retriable, true, 'a duplicate-key retry fails identically → not retriable');
+      // Raw driver error is retained server-side as `cause`, not leaked into the message.
+      assert.strictEqual(e.cause, err);
+      assert.notStrictEqual(e.message, err.message);
       return true;
     },
   );
--- a/packages/bb-distributed-data/src/errors.ts
+++ b/packages/bb-distributed-data/src/errors.ts
@@ -51,7 +51,20 @@ export function translateDsqlError(e: Error): never {
       retriable: true,
     });
   } else if (code === PG_UNIQUE_VIOLATION) {
-    e.name = DistributedDatabaseErrors.UniqueConstraintViolation;
+    // A duplicate-key / unique-constraint violation (SQLSTATE 23505) is a
+    // Conflict, not an InternalServerError: throw an ApiError with status 409
+    // so the JSON-RPC serializer emits code 409 instead of a generic 500.
+    // Preserve the UniqueConstraintViolation name so isBlocksError() keeps
+    // matching on both server and client, and keep the original error as
+    // `cause` (server-side). Unlike the 40001 branch above, this is NOT
+    // retriable — a duplicate key is deterministic, so a blind retry of the
+    // same insert fails identically (ApiError defaults retriable=false). The
+    // client-visible message is a fixed, stable string; the raw driver text
+    // (which can name columns / constraints) is retained only as `cause`.
+    throw new ApiError('The item violates a unique constraint', 409, {
+      name: DistributedDatabaseErrors.UniqueConstraintViolation,
+      cause: e,
+    });
   } else if (code && code.startsWith(PG_CONNECTION_EXCEPTION_CLASS)) {
     e.name = DistributedDatabaseErrors.ConnectionFailed;
   } else {
--- a/test-apps/comprehensive/test/database.test.ts
+++ b/test-apps/comprehensive/test/database.test.ts
@@ -3,6 +3,8 @@
 
 import { test, describe } from 'node:test';
 import assert from 'node:assert';
+import { isBlocksError, ApiError } from '@aws-blocks/core';
+import { DatabaseErrors } from '@aws-blocks/bb-data';
 import type { api as apiType } from 'aws-blocks';
 
 // Compile-time type assertion helpers. `Equal` is the standard invariant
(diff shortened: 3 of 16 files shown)


sharonyajain (dismissed):

packages/bb-distributed-data/src/errors.ts
```diff
@@ -51,7 +51,20 @@ export function translateDsqlError(e: Error): never {
...
+    // Conflict, not an InternalServerError: throw an ApiError with status 409
+    // so the JSON-RPC serializer emits code 409 instead of a generic 500.
+    // Preserve the UniqueConstraintViolation name so isBlocksError() keeps
+    // matching on both server and client, and keep the original error as
+    // `cause` (server-side). Unlike the 40001 branch above, this is NOT
+    // retriable — a duplicate key is deterministic, so a blind retry of the
+    // same insert fails identically (ApiError defaults retriable=false). The
+    // client-visible message is a fixed, stable string; the raw driver text
+    // (which can name columns / constraints) is retained only as `cause`.
+    throw new ApiError('The item violates a unique constraint', 409, {
```
osama-rizk: bb-data factors this into a `uniqueConstraintConflict()` helper, but here it's inlined — the same helper-vs-inline split I flagged on #500 for `serializationConflict`. Now two fixed conflict messages (`'The item violates a unique constraint'` and `'The transaction failed due to a serialization conflict'`) each live both as a bb-data helper and as a bb-distributed-data inline — four spots to keep in sync by hand. These are separate packages so a *shared* helper isn't warranted, but bb-distributed-data would benefit from its own local `uniqueConstraintConflict` / `serializationConflict` helpers in this `errors.ts` (mirroring bb-data) so each package has a single source for its conflict shapes. Consistency/maintainability, not correctness.

packages/bb-data/src/engines/pg-error-translator.test.ts
```diff
@@ -28,14 +29,18 @@ test('translatePgError: serialization failure (40001) → ApiError status 409, r
...
   assert.throws(
     () => translatePgError(err, 'PgClientEngine'),
     (e: unknown) => {
-      assert.ok(e instanceof Error);
-      assert.strictEqual((e as Error).name, DatabaseErrors.UniqueConstraintViolation);
-      assert.ok(!(e instanceof ApiError), 'unique violation should not be remapped to an ApiError');
+      assert.ok(e instanceof ApiError, 'expected an ApiError');
+      assert.strictEqual(e.status, 409);
+      assert.strictEqual(e.name, DatabaseErrors.UniqueConstraintViolation);
+      assert.notStrictEqual(e.retriable, true, 'a duplicate-key retry fails identically → not retriable');
```
osama-rizk: This asserts `retriable !== true`, which passes for `false`, `undefined`, `0` — anything but `true`. But the contract is *definitively* `false` (that's the whole point of distinguishing 23505 from the retriable 40001), and `ApiError` sets `retriable = options?.retriable ?? false`. Note the 40001 test two lines up correctly asserts `strictEqual(e.retriable, true)` — so pin this one the same way: `assert.strictEqual(e.retriable, false)`. As written, a future refactor that dropped the `ApiError` default (making it `undefined`) would silently change the client-observable contract while this test stayed green. Same tightening applies to the sibling assertions in pglite-engine / data-api-engine / bb-distributed-data tests and the e2e specs.

osama-rizk (commented):
Model follow-up. I verified all five `23505` throw sites are routed through the 409 mapping (pg-error-translator, pglite, data-api's SQLState *and* message-matched paths, and bb-distributed-data), `retriable` is correctly left `false` (deterministic duplicate → a blind retry fails identically, unlike the retriable `40001`), and the fixed-message/cause-server-side behavior is now actually asserted (`cause === err`, `message !== err.message`). It also closes both loose ends from #500: it answers the 23505→409 scope question I raised, and — unlike #500 — ships over-the-wire e2e coverage for the 409 path (the PGlite/DSQL mocks enforce the PK, so it's genuinely exercised). Two LOW notes only.

packages/bb-distributed-data/src/errors.ts
soberm: Done in 57896cf — added local serializationConflict() and uniqueConstraintConflict() helpers to bb-distributed-data/src/errors.ts (mirroring bb-data), and translateDsqlError now throws via them instead of the two inline ApiError constructions. Kept them package-local (distinct DistributedDatabaseErrors constants, no cross-package shared helper), so each package now has a single source for its conflict shapes. Messages/names/409/retriable unchanged.

packages/bb-data/src/engines/pg-error-translator.test.ts
soberm: Done in 57896cf — tightened the 23505/unique-violation assertions to assert.strictEqual(e.retriable, false) across pg-error-translator, pglite-engine, data-api-engine (both sites), bb-distributed-data errors.test, and the comprehensive e2e specs (database.test.ts + dsql.test.ts). The 40001 assertions stay strictEqual(true). Now a refactor that dropped ApiErrors retriable default would fail these instead of silently passing.

osama-rizk (dismissed):

test-apps/comprehensive/test/database.test.ts
```diff
@@ -105,6 +107,27 @@ export function databaseTests(getApi: () => typeof apiType) {
...
+    // ApiError with status 409 — not a generic 500. The error name is preserved
+    // so isBlocksError still matches. The PGlite mock enforces the PK constraint,
+    // so this exercises the real translate path locally. (issue #508)
+    test('Database - duplicate insert returns status 409 over the wire', async () => {
+      const api = getApi();
+      const id = `t-409-${Date.now().toString(36)}`;
+      await api.dbInsert(id, 'first', 1);
+      try {
+        await api.dbInsert(id, 'dup', 2);
+        assert.fail('Expected a conflict error');
```
sharonyajain: **assert.fail inside the try is swallowed by the same catch, so a no-throw regression surfaces as a misleading error**

The over-the-wire e2e test drives the conflict with try/catch and puts assert.fail in the try. If a future regression makes the duplicate insert stop throwing, assert.fail's AssertionError is caught by the same catch, so the failure surfaces as the later `assert.ok(e instanceof ApiError, ...)` assertion ('expected ApiError, got AssertionError') rather than the intended 'Expected a conflict error'. Cleanup (api.dbDelete) is also outside a finally, so it is skipped whenever an in-catch assertion fails. The package's own unit tests already use the idiomatic assert.rejects(fn, validator) pattern for exactly this shape. test-apps/comprehensive/test/dsql.test.ts contains the identical structure.

```typescript
try {
  await api.dbInsert(id, 'dup', 2);
  assert.fail('Expected a conflict error');
} catch (e) {
  assert.ok(e instanceof ApiError, `Expected ApiError, got ${e}`);
  ...
}
```

**Why it matters:** assert.fail placed inside the guarded try -> a no-throw regression is caught and re-reported as the wrong assertion, and the row is not cleaned up -> slower diagnosis of a real 409-regression and possible leftover state, plus divergence from the repo's established assert.rejects idiom.

**Suggestion:** We can use assert.rejects(() => api.dbInsert(id, 'dup', 2), (e) => { assert.ok(e instanceof ApiError); assert.strictEqual(e.status, 409); assert.ok(isBlocksError(e, DatabaseErrors.UniqueConstraintViolation)); assert.strictEqual(e.retriable, false); return true; }) as the unit tests do (or move assert.fail out of the try and put dbDelete in a finally). The same can be done to dsql.test.ts.

soberm: Good catch — fixed in 89f3d4f. Both e2e conflict tests (database.test.ts and dsql.test.ts) now use the repo's idiomatic assert.rejects(fn, validator) shape (matching the package unit tests) and move the dbDelete/dsqlDelete cleanup into a finally. So a no-throw regression now fails as a real rejection-expected error rather than being swallowed and mis-reported, and cleanup always runs. No behavior change, still cast-free.

sharonyajain (approved):

osama-rizk (approved):

PR #512, opened 2026-09-08
feat(core): provision a managed CloudFront API front door

Simone319:
> **⚠️ Merge together with #513 — do not merge this alone.** This is the first of a stacked pair. On its own, this PR provisions the API front-door distribution but does **not** switch the client to it and does **not** reuse an existing Hosting distribution. So for an app that uses `Hosting`, merging this without #513 creates a **second** CloudFront distribution alongside Hosting's own — two distributions fronting the same app. #513 adds both the client switch and the Hosting-reuse branch (`registerHostingDistribution` / the aspect's Hosting-present short-circuit) that suppresses the duplicate. Merge both together, or hold this until #513 is ready.

## What this does

Provisions a managed CloudFront **API front door**: a single CloudFront distribution that becomes the stable public origin for the backend's HTTP surface. Today a browser talks directly to the API Gateway, whose hostname changes whenever the backend's compute topology changes; routing all backend traffic through one long-lived distribution keeps the public hostname fixed, which is what lets `Secure` auth cookies survive adding, scaling, or swapping a compute.

In this PR the distribution is **provisioned and its URL published, but nothing routes through it yet** — the client still calls the API Gateway directly. That makes this an additive, no-traffic-change step that can be deployed and smoke-tested before the client is switched over in the follow-up.

- **`BlocksDefaults.provisionApiFrontDoor` (boolean)** gates the feature: **`true` in the `production` preset, `false` in `sandbox`**. A sandbox's dev server is same-origin `localhost` and a deployed sandbox is disposable, so a stable origin buys nothing there and isn't worth the CloudFront propagation time. Apps that spread a `BlocksPresets` preset need no change; a hand-written `BlocksDefaults` literal must add the field.
- **`scheduleApiFrontDoor`** (in `@aws-blocks/core/cdk`) registers a one-shot **CDK aspect** from `create()`, so the decision is deferred to synth (when the whole app exists). When provisioning is on, the aspect builds one `cloudfront.Distribution` whose single default behavior proxies the stack's API origin — forwarding cookies + `Authorization` and disabling caching (the backend is dynamic and auth-bearing) — and emits a `ApiFrontDoorUrl` output. When off (sandbox / opt-out) it does nothing.
- The origin is built from the stack's API URL by **`httpOriginFromApiUrl`**, the single place the framework turns a Blocks API URL into a CloudFront origin (a follow-up reuses it for Hosting).
- The distribution and its output are **scoped to the owning `BlocksStack` / `BlocksBackend`**, so several API-front-door-enabled backends can share one stack without colliding on a logical id.

## Testing

Synth unit tests (`core/src/cdk/api-front-door.cdk.test.ts` — in **core**, alongside the implementation; it uses the same inline `StubLambdaCompute` harness as `blocks-stack.test.ts`, so no concrete compute package is needed):
- production → exactly one distribution + a `ApiFrontDoorUrl` output, and the client-facing `ApiUrl` output still points at the API Gateway (nothing routes through the distribution yet); sandbox → none; an explicit `provisionApiFrontDoor: false` opts a production app out.
- two API-front-door-enabled backends in one stack synthesize to two distinct distributions/outputs (guards the per-owner scoping).
- `httpOriginFromApiUrl` builds a CloudFront origin from a Blocks API URL.

Build, lint, and affected unit tests pass on Node 22. There is no end-to-end coverage here because nothing routes through the distribution yet — the client switch and its e2e land in the follow-up PR.

**Non-breaking:** purely additive; single-compute apps keep reaching the API Gateway directly.


svidgen (approved):

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -110,6 +110,27 @@ export interface BlocksDefaults {
 	 * to derive one on demand even when this is unset.
 	 */
 	vpc?: BlocksVpcOptions;
+
+	/**
+	 * Whether to provision a managed CloudFront **API front door** — one
+	 * distribution that becomes the stable public origin for the backend HTTP
+	 * surface, path-routing `/aws-blocks/api/{namespace}/*` to the owning
```
soberm, line 117: [minor] This JSDoc says the front door path-routes `/aws-blocks/api/{namespace}/*` to each compute's ingress, but this PR creates one default behavior for one API origin; its synth test explicitly says per-namespace fan-out is deferred. Please describe the current single-origin behavior and mark namespace routing as future work so the public docs are correct at HEAD.

packages/core/src/cdk/blocks-defaults.ts
```diff
@@ -110,6 +110,27 @@ export interface BlocksDefaults {
...
+	 * `Secure` session cookie is never dropped.
+	 *
+	 * **On in `production`, off in `sandbox`** — a sandbox's dev server is
+	 * same-origin `localhost` and a deployed sandbox is disposable, so origin
+	 * stability is moot and the API front door's minutes-long CloudFront propagation
+	 * isn't worth paying on every throwaway deploy. Single-compute production
+	 * apps still get the API front door so a later single→multi transition never
+	 * changes the origin.
+	 *
+	 * When a frontend is present, Blocks reuses Hosting's distribution instead of
```
soberm, line 129: [minor] This JSDoc claims that a frontend reuses Hosting's distribution. That safeguard lands in stacked PR #513; #512 alone creates a second distribution for Hosting apps, as this PR's description warns. Please make this future tense or move the statement to #513 so the public API docs match this PR's behavior.

packages/core/src/cdk/api-front-door.cdk.test.ts
```diff
@@ -0,0 +1,188 @@
...
+/** The `ApiUrl` output value as a searchable string (references are tokens). */
+function apiUrlOutput(template: Template): string {
+	return JSON.stringify(template.findOutputs('ApiUrl'));
+}
+
+describe('API front door (scheduled aspect, gated on defaults.provisionApiFrontDoor)', () => {
+	test('production provisions one distribution + an ApiFrontDoorUrl output; the client is not switched', async () => {
+		const stack = await makeStack('ApiFrontDoorProd', BlocksPresets.production);
+		const template = Template.fromStack(stack);
+		template.resourceCountIs('AWS::CloudFront::Distribution', 1);
```
soberm, line 128: [minor] The production synth test proves that a distribution exists, but not that its security-sensitive behavior configuration remains correct. A regression enabling caching or replacing `ALL_VIEWER_EXCEPT_HOST_HEADER` would still pass despite this API carrying cookies and `Authorization`. Please assert the generated distribution's disabled cache policy, origin-request policy, allowed methods, and HTTPS redirect behavior.

packages/core/src/cdk/api-front-door.cdk.test.ts
```diff
@@ -0,0 +1,188 @@
...
+		template.resourceCountIs('AWS::CloudFront::Distribution', 2);
+		const apiFrontDoorOutputs = Object.values(template.findOutputs('*')).filter(
+			(o) => o.Description === 'Blocks API front door URL (CloudFront)',
+		);
+		assert.strictEqual(apiFrontDoorOutputs.length, 2, 'one ApiFrontDoorUrl output per backend');
+	});
+});
+
+describe('httpOriginFromApiUrl', () => {
+	test('builds an HTTP origin from a Blocks API URL', () => {
```
soberm, line 183: [minor] `httpOriginFromApiUrl()` is only tested for object existence and `bind()`, which does not validate its host/stage parsing logic. Please assert the resulting synthesized origin domain and origin path so a future `Fn.select`/split-index regression cannot silently route CloudFront to the wrong API origin.

PR #513, opened 2026-09-08
feat(core): switch the client to the API front door + reuse Hosting's

Simone319:
> **Draft.** Builds on #512 (its base is that branch), so this PR's diff is only the changes layered on top. Review/merge #512 first.

## What this does

Switches the client onto the managed CloudFront API front door and lets an app that also uses **Hosting reuse Hosting's existing CloudFront distribution** instead of provisioning a second one. #512 provisions the distribution but leaves the client on the API Gateway; this PR makes the front door the client's actual origin and adds the Hosting-reuse path.

- New **`apiFrontDoor?: 'cloudfront' | 'none'`** on `BlocksStack.create` / `BlocksBackend.create` — the customer-facing way to select the API front door. `'cloudfront'` provisions the managed distribution as the stable public origin; `'none'` leaves the client calling the API Gateway directly. An explicit value wins; omitting it follows the preset's `defaults.provisionApiFrontDoor` (on in `production`, off in `sandbox`), so existing apps are unaffected. A string union rather than a boolean so future API front-door modes don't need another flag.
- `BlocksStack`'s `ApiUrl` output is now **lazy** → the API front-door origin when one is provisioned, otherwise the API Gateway URL. The deploy step writes that value into the client's `config.json`, so the switch needs **no client-code change**.
- The API front-door aspect gains a **Hosting-reuse branch**: when a `Hosting` construct is present **in the same stack** it already fronts `/aws-blocks/api/*` on its own distribution, so the managed API front door reuses it and creates no second distribution. Hosting publishes its distribution via the new `registerHostingDistribution`, and its `addApiBehaviors` reuses `httpOriginFromApiUrl` (introduced in #512) so both API front-door paths behave identically.
- **Hosting in a different stack:** reuse detection is per-stack (the handle is published on the stack), so a `Hosting` construct in another stack can't be seen — leaving the API front door on there would provision a second, redundant distribution in front of the same API. Documented in core's README (and the relevant JSDoc): set `apiFrontDoor: 'none'` on the backend and let Hosting front the API.
- Deprecates `BlocksStack` / `BlocksBackend` `gateway` + `apiUrl` (to be removed in a later change).

### What does *not* change

Only the **origin** moves. The request path and the RPC envelope are untouched:

| | Before | After this PR |
|---|---|---|
| Host | `{apigw-id}.execute-api.{region}.amazonaws.com/{stage}` | `{dist}.cloudfront.net` |
| Path | `/aws-blocks/api` | `/aws-blocks/api` — **unchanged** |
| Namespace | in the JSON-RPC body | in the body — **unchanged** |

The resolved URL keeps the reserved RPC prefix (`` `${origin}${BLOCKS_RPC_PREFIX}` ``), and this PR touches **no client code** — the client already resolves `apiUrl` from `config.json`, so repointing it at CloudFront is a data change, not a code change.

### The client still posts to the base URL — the namespace path comes later

Concretely, every namespace shares one URL today and is distinguished only by the RPC body, so after this PR the calls look like:

```
POST https://{dist}.cloudfront.net/aws-blocks/api
{"jsonrpc":"2.0","method":"api.getUser","params":["123"],"id":1}

POST https://{dist}.cloudfront.net/aws-blocks/api      ← same URL
{"jsonrpc":"2.0","method":"authApi.signIn","params":[…],"id":2}
```

When multiple computes are enabled, the client will append its namespace to that same base URL, so requests become distinguishable by **path** (which is what a CloudFront behavior matches on — it can't route on a request body):

```
POST https://{dist}.cloudfront.net/aws-blocks/api/api       → compute A
POST https://{dist}.cloudfront.net/aws-blocks/api/authApi   → compute B
```

That later change is small and contained:

- `config.json` still carries **one** `apiUrl` — the client derives the per-namespace path rather than reading a namespace→URL map. One origin is what keeps a `Secure` session cookie valid across namespaces.
- The generated `client.js` doesn't change: `ApiNamespaceClient('authApi')` already receives its namespace, it just doesn't use it in the URL yet.
- **Customer code doesn't change at all** — `api.getUser('123')` is identical before and after.
- The namespace stays in the body too: the path becomes the *routing* key, the body remains the *dispatch* key.
- No server change is needed either — the runtime already treats every path under `/aws-blocks/api` as RPC, so the per-namespace paths work against already-deployed backends.

Deferred with that same work: per-namespace CloudFront behaviors and per-compute origin URLs. A per-namespace path only becomes meaningful once there are multiple computes to route between — with one compute every namespace resolves to the same origin, so switching the path here would be churn with no functional gain.

## Testing (this is where the e2e lands)

- **Unit/synth** (`core/src/cdk/api-front-door.cdk.test.ts`): production, no Hosting → one distribution and the `ApiUrl` output resolves to the API front-door origin; sandbox → none; **Hosting present → no Blocks-owned distribution**; two API-front-door-enabled backends in one stack → two distinct distributions; plus the new option both ways — `apiFrontDoor: 'none'` opts a production app out, and `apiFrontDoor: 'cloudfront'` forces one on in sandbox. Core, bb-lambda-compute, and blocks suites all green.
- **e2e (new):** the comprehensive test app now uses the **production preset in production mode** (overriding only `removalPolicy`→`DESTROY` and `deletionProtection`→`false` so CI can tear the stack down, while keeping the sandbox `allowedOrigins`). So the production e2e provisions the managed API front door, the client resolves the CloudFront origin, and the whole suite — **including the auth-cookie round-trip** — runs through it. Added a production-only assertion that `config.json`'s `apiUrl` is a `*.cloudfront.net` host, not an `execute-api` one.
- **Coverage by tier:** sandbox e2e → no-front-door path; production e2e → managed API front door; the `hosting-*` Playwright suites → the Hosting-reuse path (cookie through CloudFront).

**CI note:** this PR is stacked on top of #512 as a native GitHub stack, and the full `pr-checks` suite — including **Build, Unit Tests, E2E Local** — runs against it, so the API front-door e2e runs here.

**Non-breaking** coordinated redeploy (no production traffic changes on merge).


Simone319:
Pushed `d8c2ca25` with two cleanups that belong in this PR rather than downstream:

- **`ApiApiFrontDoorAspect` → `ApiFrontDoorAspect`** — the class name had `Api` twice. Introduced in `d4aa2c0c` on this branch. The class isn't exported, so this is a no-op for consumers.
- **Dropped internal tracker identifiers from published JSDoc** — the `@deprecated` notices on `BlocksStack`/`BlocksBackend` `gateway`/`apiUrl` and the front-door test header carried internal work-item labels that mean nothing to someone reading the package. Introduced in `6deccb58` on this branch. The surrounding wording is unchanged.

No behavior change. Verified on this branch: build, lint, lint:deps, check:api, changeset guard, and the core CDK + hosting tests (151/151). The existing changeset already covers `@aws-blocks/core`.

Both were originally caught while building the multi-compute routing fan-out on top of this branch; moving them here keeps that PR scoped to its own change.

svidgen (dismissed):

svidgen (commented):
Whoops. Reverting approval. Tabs rearranged on me. This was mean to approve prior PR in the stack. 😰

PR #515, opened 2026-09-08
fix(bb-agent): widen UseChatOptions.api sendMessage/resume return types to Promise<unknown>

sharonyajain:
## Problem

`useChat`'s `UseChatOptions.api` types `sendMessage` and `resume` as `Promise<void>`, but the natural backend methods return objects — `agent.stream()` resolves to `{ channelId }` and the `resume` wrappers to `{ ok: true }`. `Promise<{ channelId }>` is not assignable to `Promise<void>` (TS2322: `'{ channelId: string; }' is not assignable to 'void'`), so a customer with a natural backend is forced into an await-and-discard wrapper:

```ts
// forced today:
sendMessage: async (convId, msg, chId) => { await api.sendMessage(convId, msg, chId); },
// wanted:
sendMessage: (convId, msg, chId) => api.sendMessage(convId, msg, chId),
```

`useChat` awaits both calls purely for completion and **discards the resolved value** (`index.hooks.ts` — `await options.api.sendMessage(...)` and `await options.api.resume(...)`), so the strict `Promise<void>` is unnecessarily narrow and buys nothing.

The README backend example currently dodges this by having `sendMessage` return `void` (it awaits `agent.stream()` without returning `{ channelId }`) — the very workaround this change removes. The comprehensive test-app's `agentStream` returns `{ channelId }` and `agentResume` returns `{ ok: true }`, the natural shapes that hit the friction.

**Issue #, if available:**
no linked issue: reported via bug bash, no issue was filed.

## Changes

- `packages/bb-agent/src/index.hooks.ts`: widen `UseChatOptions.api.sendMessage` and `resume?` return types from `Promise<void>` to `Promise<unknown>`. `unknown` is assignable-from both `{ ... }` results and `void`, and since the hook discards the value this only widens what callers may return — no runtime behavior change, and existing `void`-returning backends keep compiling (backward compatible).
- `packages/bb-agent/README.md`: simplify the backend `sendMessage` example to `return { channelId }`, showing the natural, more useful shape now that the clean frontend wiring compiles.

The public `ChatInstance.sendMessage` / `respondToInterrupt` stay `Promise<void>` — those genuinely surface no value to the UI consumer. The change is scoped to the boundary where customer backends plug in.

## Validation

- Verified the original friction reproduces the exact `TS2322` (`Promise<{ channelId: string; }>` not assignable to `Promise<void>`) and that `Promise<unknown>` accepts both a natural `{ channelId }` / `{ ok }` backend and an existing `void` backend, confirming the fix works and is backward compatible.
- Typechecked the edited `index.hooks.ts` in isolation (it only imports `./types.js`) — clean.
- No API report drift: `API.md` is generated by API Extractor from `dist/index.aws.d.ts` (the CDK/server entry point); `UseChatOptions` lives in the client entry point `index.hooks.ts` and does not appear in `API.md`, so no regeneration is needed.

## Manual verification

N/A — type-only change plus a doc example; unit-level type assertions cover the behavior.

## Checklist

- [x] PR description included
- [ ] Tests are changed or added — N/A, this is a type-signature widening; no runtime code path changes and no new test hook exists for the client entry-point type surface
- [x] Relevant documentation is changed or added (README backend example updated)

### DIFF
--- a/packages/bb-agent/README.md
+++ b/packages/bb-agent/README.md
@@ -763,7 +763,8 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
     return { conversationId: await agent.createConversationId(userId) };
   },
   async sendMessage(conversationId: string, message: string, channelId: string, userId: string) {
-    await agent.stream(message, { conversationId, channelId, userId });
+    const result = await agent.stream(message, { conversationId, channelId, userId });
+    return { channelId: result.channelId };
   },
   async getConversation(conversationId: string) {
     const messages = await agent.getConversation(conversationId);
--- a/packages/bb-agent/src/index.hooks.ts
+++ b/packages/bb-agent/src/index.hooks.ts
@@ -29,10 +29,10 @@ export interface ChatMessage {
 /** Options for creating a chat instance. */
 export interface UseChatOptions {
 	api: {
-		sendMessage(conversationId: string, message: string, channelId: string): Promise<void>;
+		sendMessage(conversationId: string, message: string, channelId: string): Promise<unknown>;
 		createConversation(): Promise<{ conversationId: string }>;
 		getConversation(id: string): Promise<{ messages: { role: string; content: string; metadata?: Record<string, any> }[] }>;
-		resume?(channelId: string, responses: Array<{ interruptId: string; approved: boolean; trust?: boolean; toolName?: string; input?: any }>, conversationId?: string): Promise<void>;
+		resume?(channelId: string, responses: Array<{ interruptId: string; approved: boolean; trust?: boolean; toolName?: string; input?: any }>, conversationId?: string): Promise<unknown>;
 		getPendingInterrupts?(conversationId: string): Promise<{ interrupts: Array<{ id: string; name: string; reason?: any }> }>;
 	};
 	/**
--- a/.changeset/usechat-api-return-types.md
+++ b/.changeset/usechat-api-return-types.md
@@ -1,0 +1,8 @@
+---
+"@aws-blocks/bb-agent": patch
+"@aws-blocks/blocks": patch
+---
+
+`useChat`: widen `UseChatOptions.api.sendMessage` and `resume` return types from `Promise<void>` to `Promise<unknown>`.
+
+The natural backend methods return objects (`agent.stream()` → `{ channelId }`, `resume` wrappers → `{ ok: true }`), but `Promise<{ channelId }>` is not assignable to `Promise<void>` (TS2322), which forced customers into an await-and-discard wrapper. `useChat` awaits both calls only for completion and discards the resolved value, so `Promise<unknown>` — assignable-from both object results and `void` — lets natural-shape backends wire up directly while existing `void`-returning backends keep compiling. Type-only change; no runtime behavior change.
--- a/packages/bb-agent/src/index.test.ts
+++ b/packages/bb-agent/src/index.test.ts
@@ -1185,8 +1185,32 @@ describe('model-factory', () => {
 // ── useChat ──────────────────────────────────────────────────────────────────
 
 import { useChat } from './index.hooks.js';
+import type { UseChatOptions } from './index.hooks.js';
 
 describe('useChat', () => {
+	// Type-only regression guard for the api return-type contract (PR that widened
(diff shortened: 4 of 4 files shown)


packages/bb-agent/src/index.hooks.ts
```diff
@@ -29,10 +29,10 @@ export interface ChatMessage {
 /** Options for creating a chat instance. */
 export interface UseChatOptions {
 	api: {
-		sendMessage(conversationId: string, message: string, channelId: string): Promise<void>;
+		sendMessage(conversationId: string, message: string, channelId: string): Promise<unknown>;
```
osama-rizk, line 32: The whole change is a type contract, validated manually — but nothing committed locks it. A future refactor could narrow this back to `Promise<void>` (or someone tidying types), silently reintroducing the exact TS2322 for natural-backend customers, and no build/test would fail because the proof lives only in the PR's local check.

Consider a `.test-d.ts` / `expectAssignable`-style assertion proving both a `{ channelId }`-returning backend and a `void`-returning backend satisfy `UseChatOptions['api']` — it's the only durable regression guard for a type-only fix. If the repo has no type-test infrastructure, that's a fair reason to skip, but worth saying so explicitly, since "N/A, type-only" undersells that the type contract is precisely what can regress here.

packages/bb-agent/README.md
```diff
@@ -763,7 +763,8 @@ export const api = new ApiNamespace(scope, 'api', (context) => ({
     return { conversationId: await agent.createConversationId(userId) };
   },
   async sendMessage(conversationId: string, message: string, channelId: string, userId: string) {
-    await agent.stream(message, { conversationId, channelId, userId });
+    const { channelId: id } = await agent.stream(message, { conversationId, channelId, userId });
```
osama-rizk: Nit: this destructures `channelId` and renames it to `id`, then rebuilds `{ channelId: id }` — a rename-and-rewrap that adds noise and slightly obscures the point of the example (return the natural shape). Since the doc change is meant to showcase how clean the natural wiring now is, simplify: `return agent.stream(message, { conversationId, channelId, userId });` if `stream` resolves to exactly `{ channelId }`, or `const { channelId } = await agent.stream(...); return { channelId };`.

osama-rizk (commented):
Clean, correct fix. `unknown` (not `any`) is the right principled widening — it's the top type, so it accepts every backend shape while still keeping any future reader honest if they ever consume the value; and I confirmed the hook genuinely discards both results (`await options.api.sendMessage(...)` and `await options.api.resume(...)`), so this is behavior-neutral. Good call keeping the public `ChatInstance.sendMessage`/`respondToInterrupt` at `Promise<void>` — relaxing only the customer-backend boundary is exactly the right scope, and it's backward-compatible by construction. Two low notes below.

packages/bb-agent/src/index.hooks.ts
sharonyajain: Fixed in 3ad8b59d5b607aa626f85ab863052af12e5062a6. Added a compile-time regression guard in `packages/bb-agent/src/index.test.ts` (the existing `describe('useChat')` block) — it's compiled by `tsc --build` ahead of the runtime tests, so no new type-test infra was needed. It asserts both a `{ channelId }`-returning backend and a `void`-returning backend satisfy `UseChatOptions['api']`. I verified it bites: narrowing either member back to `Promise<void>` reproduces TS2322 on the object backend and fails the build. You're right that "N/A, type-only" undersold the regression surface — the type contract is exactly what can regress, so it now has committed proof.

packages/bb-agent/README.md
sharonyajain: Fixed in 3ad8b59d5b607aa626f85ab863052af12e5062a6. Dropped the rename-and-rewrap. `agent.stream()` resolves to `AgentStreamResult` (`{ channelId, channel, complete, ... }`), not exactly `{ channelId }`, so it can't be returned whole; used `const result = await agent.stream(...); return { channelId: result.channelId };` — which also avoids shadowing the `channelId` parameter that a literal `const { channelId } = ...` destructure would introduce. The example now reads as the natural shape without the noise.

osama-rizk (approved):

pranavosu (approved):

PR #516, opened 2026-09-08
fix(auth-oidc): make cognitoFederated() deployable via a deploy-time custom resource (#447)

### LINKED ISSUE #447: auth-oidc: cognitoFederated() is undeployable — CloudFormation rejects {{resolve:ssm-secure}} in UserPoolIdentityProvider.ProviderDetails

Building Block: auth-oidc

### Describe the bug

`cognitoFederated()` emits a CloudFormation template that CloudFormation refuses to
deploy. The CDK layer writes the IdP credentials into
`AWS::Cognito::UserPoolIdentityProvider.ProviderDetails` as `{{resolve:ssm-secure:...}}`
dynamic references, but CloudFormation only supports `ssm-secure` dynamic references in
a small allowlist of resource properties, and this property is not one of them.

`cdk synth` succeeds, so the problem is invisible until deploy time, where it fails
during change set creation — before any resource is created.

This appears to affect every `cognitoFederated()` user: there is no option to pass the
credentials any other way. `CognitoFederatedOpts.clientId` / `clientSecret` are typed as
`AppSettingLike`, and the doc comment states "CDK reads the parameter name for
CloudFormation dynamic references", so the unsupported reference is the only code path.

### Expected behavior

`npm run deploy` provisions the Cognito User Pool, app client, domain and the Google
identity provider, resolving the IdP credentials from the SSM SecureString parameters.

A few possible directions (not a preference, just what seems available):

- Register the identity provider through a custom resource (e.g. `AwsCustomResource`
  calling `CreateIdentityProvider` / `UpdateIdentityProvider`) that reads the
  SecureString parameter at deploy time, instead of a CloudFormation dynamic reference.
- Or, if that is not desirable, surface the limitation at synth time and document that
  `cognitoFederated()` cannot source credentials from `secret: true` `AppSetting`s.

### Reproduction steps

1. `npm create @aws-blocks/blocks-app@latest my-app` (or add the backend to an existing
   project with `npx @aws-blocks/create-blocks-app . --template backend`).
2. Add the `AuthOIDC` block with a `cognitoFederated()` provider, as in the code snippet
   below (this is the example from the `bb-auth-oidc` README, "Cognito-mediated
   federation").
3. Create the two SSM SecureString parameters that the secret `AppSetting`s map to, so
   the dynamic references have something to resolve:
   ```
   aws ssm put-parameter --type SecureString --name /<stack>-<scope>-google-client-id     --value '<google client id>'
   aws ssm put-parameter --type SecureString --name /<stack>-<scope>-google-client-secret --value '<google client secret>'
   ```
4. `npx cdk synth` — succeeds. The template contains
   `"client_id": "{{resolve:ssm-secure:/...}}"` and
   `"client_secret": "{{resolve:ssm-secure:/...}}"` under
   `AWS::Cognito::UserPoolIdentityProvider` → `ProviderDetails`.
5. `npm run deploy` — fails while creating the change set. No resources are created; the
   stack is left in `REVIEW_IN_PROGRESS`.

### Code snippet

```typescript
import {
  ApiNamespace,
  AppSetting,
  AuthOIDC,
  cognitoFederated,
  Scope,
} from "@aws-blocks/blocks";

const scope = new Scope("my-app");

const googleClientId = new AppSetting(scope, "google-client-id", { secret: true });
const googleClientSecret = new AppSetting(scope, "google-client-secret", { secret: true });

export const auth = new AuthOIDC(scope, "auth", {
  providers: [
    cognitoFederated({
      name: "google",
      identityProvider: "Google",
      cognitoDomain: "my-app-abc123",
      region: "ap-northeast-1",
      clientId: googleClientId,
      clientSecret: googleClientSecret,
    }),
  ],
});

export const authApi = auth.createApi();

export const api = new ApiNamespace(scope, "api", (context) => ({
  async whoami() {
    const user = await auth.requireAuth(context);
    return { userId: user.userId, email: user.email };
  },
}));
```

### Log output

```plain text
text
my-app-abc123-prod: creating CloudFormation changeset...
Failed to create change set cdk-deploy-change-set:
my-app-abc123-prod  (AWS::CloudFormation::Stack)
  SSM Secure reference is not supported in:
  [AWS::Cognito::UserPoolIdentityProvider/Properties/ProviderDetails/client_secret,AWS::Cognito::UserPoolIdentityProvider/Properties/ProviderDetails/client_id]

❌ Deployment failed.
DeployProcessError: npx cdk deploy --require-approval never --ci --progress events --outputs-file .blocks-sandbox/outputs.json --context projectRoot=<project> exited with code 1
    at runStreaming (.../@aws-blocks/core/dist/scripts/deploy-stream.js:324:19)
  exitCode: 1


Relevant synth output:

 text
"ProviderDetails": {
  "client_id": "{{resolve:ssm-secure:/my-app-abc123-prod-my-app-google-client-id}}",
  "client_secret": "{{resolve:ssm-secure:/my-app-abc123-prod-my-app-google-client-secret}}",
  ...
}
```

### Environment

- AWS Blocks version: `@aws-blocks/blocks` 0.3.1 (`@aws-blocks/bb-auth-oidc` 0.1.8, `@aws-blocks/core` 0.2.0)
- Node.js version: 24.20.0
- Package manager (npm/yarn/pnpm): npm 11.19.0
- OS: macOS 26.6.1 (arm64)
- CDK version (if applicable): CLI 2.1139.0, `aws-cdk-lib` 2.257.0

### Additional context

Workaround used in the meantime: switching the provider from `cognitoFederated()` to
`google()`. That path resolves the credentials at runtime through `AppSetting.get()`
rather than through CloudFormation, so it is unaffected — the synthesized template then
contains no `ssm-secure` references at all and deploys cleanly.

osama-rizk:
## Problem (#447)

`cognitoFederated()` produced a CloudFormation template that **always failed to deploy**. The CDK layer registered the federated IdP with a native `AWS::Cognito::UserPoolIdentityProvider`, writing the IdP `client_id` / `client_secret` into `ProviderDetails` as `{{resolve:ssm-secure:...}}` dynamic references. CloudFormation only permits `ssm-secure` references on a small allowlist of properties that **excludes `ProviderDetails`**, so `cdk synth` succeeded but every deploy failed at change-set creation — before any resource was created — leaving the stack in `REVIEW_IN_PROGRESS`:

```
SSM Secure reference is not supported in:
[AWS::Cognito::UserPoolIdentityProvider/Properties/ProviderDetails/client_secret, .../client_id]
```

This affected **every** `cognitoFederated()` user — credentials are always sourced from `secret` `AppSetting`s, so `ssm-secure` was the only path.

## Fix — register the IdP with a deploy-time custom resource

The IdP is now registered by a **Lambda-backed custom resource** instead of a native resource. The handler reads + decrypts the IdP credential SecureString parameters via the SDK **at deploy time** and calls Cognito's `CreateIdentityProvider` / `UpdateIdentityProvider` / `DeleteIdentityProvider`. **Only the parameter *names* cross into CloudFormation — the secret values never appear in the template.**

- **Lifecycle:** Create (idempotent — falls back to Update on `DuplicateProviderException`); Update (Create fallback on `ResourceNotFoundException`; a `ProviderName`/`ProviderType` change forces replacement via a new `PhysicalResourceId`); Delete (ignores `ResourceNotFoundException`). Reads params with a short retry to tolerate the bulk secret-init resource landing slightly later.
- **Least-privilege IAM:** `cognito-idp:*IdentityProvider` on the pool ARN, `ssm:GetParameter` on the specific parameter ARNs, `kms:Decrypt` conditioned on `kms:ViaService = ssm.<region>` (covers the default `aws/ssm` key and CMKs).
- **Ordering:** the app client depends on each custom resource, so the IdP exists before the client lists it in `SupportedIdentityProviders`.
- **Nothing reused from `@aws-blocks/hosting`.** The pattern mirrors the in-package `bb-app-setting` secret-init resource: inline handler, `@aws-sdk/*` provided by the Node.js Lambda runtime → no new dependencies, no esbuild, no `package.json` change.

### 1. Reproduced the original bug against live CloudFormation
Synthesized the **pre-fix** template and drove `create-change-set` with the real framework-emitted `ProviderDetails`. It failed with the exact issue error (`SSM Secure reference is not supported in [...ProviderDetails...]`). Confirmed the "not supported" validation only fires *after* the `ssm-secure` refs resolve (with the params absent you instead get `Parameters cannot be found`), which is why the reporter's repro creates the params first.

### 2. Unit / synth tests (`packages/bb-auth-oidc/src/index.cdk.test.ts`)
- The synthesized template contains **no** `AWS::Cognito::UserPoolIdentityProvider` and **no** `{{resolve:ssm-secure}}` reference.
- The IdP-registration custom resource carries the parameter **names** (`/…-client-id`), never the values.
- The handler role has the three expected grants (`cognito-idp:CreateIdentityProvider`, `ssm:GetParameter`, scoped `kms:Decrypt`).
- Control: a self-hosted `google()` provider provisions no Cognito resources.
- bb-auth-oidc suite: **126/126**. `npm run build`, biome lint, `lint:deps`, `check:api` all clean (no API drift).

### 3. End-to-end deploy on a real account (us-west-2)

**Deploy A — mechanism check** (standalone harness, SecureString params pre-created): `cdk deploy` succeeded with no change-set rejection; the Google IdP was registered on the pool with `client_id` equal to the value stored in the SecureString.

**Deploy B — real `AppSetting(secret:true)` integration** (the faithful test): a harness with genuine `new AppSetting(scope, 'google-client-id', { secret: true })` instances, **no pre-created params**, so the framework's own `BlocksSecretsBulk` custom resource creates the SecureString parameters *during this deploy* and the IdP-registration custom resource reads them (real ordering). Result:

```
✅  BbAuthOidcE2eReal447          # deploy succeeded, no ssm-secure rejection
… BlocksSecretsBulk           CREATE_COMPLETE   # framework creates the SecureString params
… app/auth/idp-google/Default CREATE_COMPLETE   # my CR reads them + registers the IdP

# IdP client_id in Cognito == the value BlocksSecretsBulk generated in SSM:
IdP client_id   = LxkkjYbABaA94xYJ7nygbC0TFq-03eGfFGruXb5l9_0
SSM param value = LxkkjYbABaA94xYJ7nygbC0TFq-03eGfFGruXb5l9_0   ← exact match
```

This exercises the full real chain: `AppSetting(secret:true)` → `BlocksSecretsBulk` provisions the SecureString param → the IdP-registration custom resource reads + decrypts it at deploy time (ordering handled) → `CreateIdentityProvider` registers the IdP → nothing sensitive in the template. Not covered: a browser sign-in through Cognito Hosted UI (needs real Google OAuth credentials).

## Changeset
`@aws-blocks/bb-auth-oidc` **minor** (template shape change — the IdP is now a custom resource, not a native `UserPoolIdentityProvider`; snapshot assertions on that resource will diff) + `@aws-blocks/blocks` patch. Public API unchanged.

Fixes #447


osama-rizk:
### Reproduction evidence (offline)

Synthesizing the **pre-fix** template (guard temporarily removed) emits exactly the references the issue reports as rejected, under `AWS::Cognito::UserPoolIdentityProvider`:

```json
"ProviderDetails": {
  "client_id": "{{resolve:ssm-secure:/app-google-client-id}}",
  "client_secret": "{{resolve:ssm-secure:/app-google-client-secret}}",
  "authorize_scopes": "openid email profile"
}
```

This is the root cause, confirmed at the synth layer. CloudFormation only permits `ssm-secure` dynamic references on a fixed allowlist of properties that excludes `UserPoolIdentityProvider.ProviderDetails`, so this template fails at change-set creation (as the reporter observed).

**Scope of verification:** the bug's mechanism (above) and the guard both live at the synth layer, and the regression test verifies the guard fires (true red→green). The live `cdk deploy` change-set rejection itself was **not** re-run here — it needs a sandbox + real IdP credentials — but it is documented CFN behavior and matches the issue's log output.

osama-rizk:
### End-to-end verification against live CloudFormation ✅

Reproduced the exact failure the issue reports, using the framework-synthed `ProviderDetails` in a minimal template (`UserPool` + `UserPoolIdentityProvider`) via `create-change-set` (no resources created — the rejection fires at change-set creation):

```
Status=FAILED
StatusReason: SSM Secure reference is not supported in:
[AWS::Cognito::UserPoolIdentityProvider/Properties/ProviderDetails/client_secret,
 AWS::Cognito::UserPoolIdentityProvider/Properties/ProviderDetails/client_id]
```

Byte-for-byte the issue's log output.

One clarifying detail worth capturing: the "not supported" validation only fires **after** the `ssm-secure` references resolve. With the SecureString params absent, the change-set instead fails with `Parameters: [ssm-secure:/...] cannot be found`; creating the params (as the reporter's repro step 3 does) then surfaces the real "not supported in ProviderDetails" rejection. Either way the deploy is impossible.

With this PR's guard, `cdk synth` errors out before emitting the template, so neither failure can occur — verified by the regression test (true red→green). All sandbox artifacts (stack + SSM params) were torn down.

osama-rizk:
### Updated: implemented the real fix (deploy-time custom resource)

This PR now ships the **proper fix** rather than the synth-time fail-fast guard. `cognitoFederated()` is deployable again.

**What changed:** the IdP is registered by a Lambda-backed custom resource instead of a native `AWS::Cognito::UserPoolIdentityProvider`. The handler reads + decrypts the IdP credential SecureString params via the SDK at deploy time and calls `CreateIdentityProvider` / `UpdateIdentityProvider` / `DeleteIdentityProvider`. **Only the parameter names cross into CloudFormation — the secret values never appear in the template.**

- **Handler lifecycle:** Create (idempotent — falls back to Update on `DuplicateProviderException`), Update (Create fallback on `ResourceNotFoundException`; a `ProviderName`/`ProviderType` change forces replacement), Delete (ignores `ResourceNotFoundException`). Reads params with a short retry to tolerate the bulk secret-init resource landing slightly later.
- **Least-privilege IAM:** `cognito-idp:*IdentityProvider` on the pool ARN, `ssm:GetParameter` on the specific parameter ARNs, `kms:Decrypt` conditioned on `kms:ViaService = ssm.<region>` (covers the default `aws/ssm` key and CMKs).
- **Ordering:** the app client depends on each custom resource, so the IdP exists before the client lists it in `SupportedIdentityProviders`.
- **Nothing reused from `@aws-blocks/hosting`** — the pattern mirrors the in-package `bb-app-setting` secret-init resource (inline handler, `@aws-sdk/*` provided by the Lambda runtime, no new deps / no esbuild).

### End-to-end verification (real account)

Deployed the construct to a sandbox:

```
✅  BbAuthOidcE2e447    (no change-set rejection)
BbAuthOidcE2e447.PoolId = us-west-2_ufp8hMJFZ
```

Confirmed the IdP was actually registered, with the client id resolved from the SSM SecureString at deploy time:

```
$ aws cognito-idp describe-identity-provider --user-pool-id us-west-2_ufp8hMJFZ --provider-name Google
{ "Name": "Google", "Type": "Google",
  "ClientId": "dummy-google-client-id",          # ← the exact value stored in the SecureString param
  "Scopes": "openid email profile",
  "Attrs": { "email": "email", "name": "name", "username": "sub" } }
```

Then `cdk destroy` — the Delete path deregistered the IdP and tore the stack down cleanly; the dummy SSM params were removed.

### Tests / checks
- New CDK regression tests: no `ssm-secure` reference in the synthesized template, the custom-resource shape (names not values), and the three IAM grants. bb-auth-oidc suite **126/126**.
- `npm run build`, biome lint, `lint:deps`, `check:api` all clean (no API drift).
- Changeset updated: `@aws-blocks/bb-auth-oidc` **minor** (template shape changes — IdP is now a custom resource) + `@aws-blocks/blocks` patch.

Fixes #447

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -184,70 +342,66 @@ export class AuthOIDC<
...
+
+		paramNames.push(clientIdParam, clientSecretParam);
+
+		return new CustomResource(this, `idp-${provider.name}`, {
+			serviceToken,
+			properties: {
+				UserPoolId: pool.userPoolId,
+				ProviderName: provider.identityProvider,
+				ProviderType: providerType,
+				ClientIdParam: clientIdParam,
```
sarayev, line 315: The custom resource only takes the SSM parameter **names** as properties, never a value or version. That's great for keeping secrets out of the template, but I think it creates a rotation gap: when someone runs `blocks secret` to change a credential, only the SSM *value* changes, so the CR's properties are byte-for-byte identical and CloudFormation issues no Update. The handler never re-runs, and Cognito keeps the old secret, even on an explicit `cdk deploy`.

It gets worse on the *first* deploy: `BlocksSecretsBulk` seeds every managed secret with a random 32-byte placeholder (`crypto.randomBytes(32)`, `Overwrite: false` in bb-app-setting `index.cdk.ts`). So the first deploy registers the IdP with that random placeholder, and because of the no-Update gap above, a later real `blocks secret` + deploy never overwrites it. The README sells "set them with `blocks secret` before deploying," but only the first deploy honors it, and it honors the *placeholder*.

I think this is also why Deploy B passes while browser sign-in stays "not covered": the IdP `client_id` matched the auto-generated SSM value (the placeholder), not a credential a user would actually log in with.

Could we pass the SSM parameter **Version** (non-secret) as a CR property so a rotation bumps it and forces the (already idempotent) Update to re-run? And add an explicit `addDependency` from the IdP CR onto `BlocksSecretsBulk` so ordering isn't left to the retry loop?

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -61,6 +66,108 @@ export type {
...
+			await idp.send(new DeleteIdentityProviderCommand({ UserPoolId: p.UserPoolId, ProviderName: p.ProviderName }));
+		} catch (e) {
+			if (e.name !== 'ResourceNotFoundException') throw e;
+		}
+		return { PhysicalResourceId: physicalId };
+	}
+
+	const clientId = await readSecret(p.ClientIdParam);
+	const clientSecret = await readSecret(p.ClientSecretParam);
+	if (!clientId || !clientSecret) {
```
sarayev: The friendly "set them with `blocks secret`" error only triggers when the param exists but is empty. If it's genuinely missing, `readSecret` retries ~6x/10s then throws a raw `ParameterNotFound`. And since bulk-init always seeds a non-empty placeholder, this empty check almost never runs at all. Worth throwing the same actionable message on a terminal `ParameterNotFound` so the advertised UX actually happens?

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -144,9 +251,60 @@ export class AuthOIDC<
...
+			}),
+			code: Code.fromInline(IDP_REGISTRATION_HANDLER),
+		});
+		// Register / update / deregister the IdP on the pool.
+		idpFn.addToRolePolicy(new PolicyStatement({
+			actions: [
+				'cognito-idp:CreateIdentityProvider',
+				'cognito-idp:UpdateIdentityProvider',
+				'cognito-idp:DeleteIdentityProvider',
+				'cognito-idp:DescribeIdentityProvider',
```
sarayev: The handler never calls Describe, but it's in the role. Since the changeset leads with "least-privilege," this is the kind of thing an auth-block reviewer will flag. Drop it, or use it to drive the Create-vs-Update decision instead of the exception-catch fallbacks?

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -184,70 +342,66 @@ export class AuthOIDC<
...
+				break;
 		}
+
+		paramNames.push(clientIdParam, clientSecretParam);
+
+		return new CustomResource(this, `idp-${provider.name}`, {
+			serviceToken,
+			properties: {
+				UserPoolId: pool.userPoolId,
+				ProviderName: provider.identityProvider,
```
sarayev, line 313: The construct id uses `provider.name` (unique), but `ProviderName` uses `provider.identityProvider`. Two configs both set to `'Google'` would synth two CRs that write the same Cognito provider, and the second quietly wins via the `DuplicateProviderException` -> Update fallback. A synth-time uniqueness check (or deriving `ProviderName` from `provider.name`) would turn a silent overwrite into a clear error. Intentional, or worth guarding?

packages/bb-auth-oidc/src/index.cdk.test.ts
```diff
@@ -0,0 +1,137 @@
...
+				region: 'us-east-1',
+				clientId: appSettingStub('app-google-client-id'),
+				clientSecret: appSettingStub('app-google-client-secret'),
+			}),
+		],
+	});
+	return Template.fromStack(stack);
+}
+
+test('CDK: cognitoFederated() emits NO native IdP resource and NO ssm-secure reference (#447)', () => {
```
sarayev, line 78: These tests only assert the synthesized template shape. The riskier code, the `readSecret` retry/terminal behavior, the Create<->Update idempotency fallbacks, and the Delete `ResourceNotFoundException` swallow, has zero coverage, and those are the paths most likely to regress. If the handler moved to a real `.ts` file (see the other comment on the inline handler), it could be unit-tested with mocked SSM/Cognito clients across all three request types.

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -61,6 +66,108 @@ export type {
...
 } from './types.js';
 
+/**
+ * Inline handler for the IdP-registration custom resource. Runs at deploy time:
+ * reads the IdP credential SecureString parameters (by name, with decryption)
+ * and calls Cognito's Create/Update/DeleteIdentityProvider. The secret values
+ * are read here via the SDK and never transit CloudFormation. `@aws-sdk/*` is
+ * provided by the Node.js Lambda runtime, so nothing is bundled.
+ */
+const IDP_REGISTRATION_HANDLER = `
```
sarayev: Inline handler as a template string is a bit fragile: no type-checking or lint on it, the nested backtick escaping is easy to break, and `Code.fromInline` caps at 4096 bytes (this is ~2.6KB and growing). A bundled `NodejsFunction` or sibling `.handler.ts` would make it both testable and lint-covered. It does match the bb-app-setting pattern though, so fine to defer if you'd rather stay consistent for now.

packages/bb-auth-oidc/src/index.cdk.test.ts
```diff
@@ -0,0 +1,137 @@
...
+	return { stack, parent };
+}
+
+// An AppSetting-shaped stub: the CDK layer only reads `fullId`; `get()` is the
+// runtime path and is never called at synth.
+function appSettingStub(fullId: string): AppSettingLike {
+	return { fullId, get: async () => 'unused-at-synth' };
+}
+
+function synthFederated(): Template {
```
sarayev, line 61: The synth tests only cover Google. I confirmed Facebook/Amazon/OIDC produce the same values the old L2 constructs did, so nothing's broken today, but nothing pins Facebook's `public_profile email`, Amazon's `profile`, or OIDC's `oidc_issuer` / `attributes_request_method` against a future edit. A couple more per-provider assertions would lock that in.

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -61,6 +66,108 @@ export type {
...
+			return (r.Parameter && r.Parameter.Value) || '';
+		} catch (e) {
+			if (e.name === 'ParameterNotFound' && i < 5) {
+				await new Promise((res) => setTimeout(res, 2000));
+				continue;
+			}
+			throw e;
+		}
+	}
+	return '';
```
sarayev: A few small things in `readSecret`: the trailing `return ''` after the loop is unreachable (the loop always returns or throws on the last attempt); it only retries on `ParameterNotFound`, so a transient throttle/5xx on `GetParameter` throws immediately (probably fine given SDK default retries, just flagging); and `PhysicalResourceId` omits `ProviderType`, so a type change with an unchanged name routes to Update (which can't change type) rather than forcing replacement (edge-only here since `ProviderName === identityProvider`).

packages/bb-auth-oidc/src/index.cdk.ts
osama-rizk: Good catch — this was the important one. Two changes:

1. **Rotation/first-deploy propagation:** the custom resource now carries a per-synth `Trigger` property, so every `cdk deploy` re-reads the SecureString and re-registers the IdP (idempotent `UpdateIdentityProvider`). Previously an out-of-band value change altered no property, so CFN never re-ran the handler and Cognito stayed on the seeded placeholder — the IdP could never receive real credentials. Trade-off (documented): the resource shows as updated every deploy.
2. **Docs corrected:** a `secret: true` `AppSetting` is an SSM SecureString at `/<fullId>`, **not** a Secrets Manager entry, so `blocks secret` does not apply. Set it via `aws ssm put-parameter … --type SecureString --overwrite` or the `AppSetting` runtime `put()`.

On the explicit `addDependency` onto `BlocksSecretsBulk`: I could not wire it cleanly — bb-app-setting keys that shared resource under a module-private `Symbol` with no exposed handle, so I kept the deploy-time retry as the ordering guard. It is now exercised for real: I re-ran the e2e with a genuine `AppSetting(secret: true)` (no pre-created params), so `BlocksSecretsBulk` creates the SecureString during the same deploy and the handler reads it — and yes, you were right that the earlier Deploy B matched the *placeholder*; that behavior is now documented.

packages/bb-auth-oidc/src/index.cdk.ts
osama-rizk: Fixed. A terminal `ParameterNotFound` now throws the actionable message (including the exact `aws ssm put-parameter … --overwrite` command), instead of a raw SDK error. I also corrected the store guidance (SSM SecureString, not `blocks secret`). Note the empty-string check remains as a secondary guard, but per your other comment the docs no longer over-promise a fast-fail for the placeholder case.

packages/bb-auth-oidc/src/index.cdk.ts
osama-rizk: Dropped `cognito-idp:DescribeIdentityProvider` from the role — the handler uses the Create/Update exception-catch fallbacks, so Describe was dead permission.

packages/bb-auth-oidc/src/index.cdk.ts
osama-rizk: Added a synth-time guard: two `cognitoFederated()` providers sharing the same `identityProvider` now throw (Cognito allows one provider per name per pool). Test added.

packages/bb-auth-oidc/src/index.cdk.test.ts
osama-rizk: Done — the handler moved to a real `.ts` file (see the inline-handler comment) and is now unit-tested with fake SSM/Cognito clients across Create (+ `DuplicateProviderException` → Update), Update (+ `ResourceNotFoundException` → Create), Delete (+ not-found swallow), and `readSecret` retry/terminal behavior.

packages/bb-auth-oidc/src/index.cdk.ts
osama-rizk: Extracted to `src/idp-registration-lambda.ts`, bundled via esbuild (`build:lambda`, `--external:@aws-sdk/*`) to `dist/idp-registration-lambda/` and referenced with `Code.fromAsset`. Now type-checked, lint-covered, unit-tested, and off the 4KB `fromInline` cap.

packages/bb-auth-oidc/src/index.cdk.test.ts
osama-rizk: Added per-provider assertions pinning Facebook (`public_profile email`), LoginWithAmazon (`profile`), and OIDC (`oidc_issuer` + `attributes_request_method=GET`) so a future edit cannot drift them.

packages/bb-auth-oidc/src/index.cdk.ts
osama-rizk: All three addressed: removed the unreachable trailing `return `; `readSecret` now retries transient errors too (not only `ParameterNotFound`); and `PhysicalResourceId` now includes `ProviderType` (`<pool>|<name>|<type>`) so a type change forces a clean replacement.

packages/bb-auth-oidc/src/index.cdk.ts
osama-rizk: Follow-up: I've now wired the explicit `addDependency` I'd punted on. Each IdP custom resource depends on bb-app-setting's shared `BlocksSecretsBulk` resource, located by construct id (`Stack.of(this).node.tryFindChild('BlocksSecretsBulk')` — it's a direct child of the stack), so the SecureString parameters are guaranteed written before the handler reads them, rather than leaning on the read-retry (which remains as a fallback when no secret AppSetting exists). Unit test asserts the `DependsOn`; DESIGN updated.

Note: the `Trigger` nonce still handles the separate rotation concern (a value change alters no property, so it forces the re-read each deploy) — the dependency only fixes first-deploy ordering.

packages/bb-auth-oidc/src/index.cdk.ts
```diff
@@ -184,70 +258,85 @@ export class AuthOIDC<
...
+			},
+		});
+
+		// A `secret: true` AppSetting's SecureString value is written by the shared
+		// bb-app-setting bulk-init custom resource (`BlocksSecretsBulk`, a direct
+		// child of the stack). Depend on it so the parameter exists before this
+		// handler reads it — a hard ordering guarantee rather than leaving it to the
+		// handler's read-retry. (Skipped only if the customer somehow configured no
+		// secret AppSetting, in which case the retry remains the fallback.)
+		const bulkSecrets = cdk.Stack.of(this).node.tryFindChild('BlocksSecretsBulk');
```
Simone319: The IdP custom resource's ordering dependency is wired by locating BlocksSecretsBulk through a hard-coded construct-id string, and the if (bulkSecrets) guard silently skips the dependency when it isn't found. The comment covers the no-secret-AppSetting case, but if bb-app-setting ever renames or relocates that construct id, this returns undefined and the hard ordering guarantee silently degrades to the deploy-time read-retry with no error. Consider a shared exported constant for the id, or failing loud when secret-sourced providers are present but the sibling can't be found, so this cross-package coupling can't break silently.

osama-rizk: Good call — removed the hard-coded string. `bb-app-setting` now exports `SECRETS_BULK_CONSTRUCT_ID` as the single source of truth: it's used at the construct's creation site (`new cdk.CustomResource(stack, SECRETS_BULK_CONSTRUCT_ID, …)`) and re-exported from every entry, and `bb-auth-oidc` imports it for the `tryFindChild` lookup. That makes the coupling compile-time, so a rename in bb-app-setting moves the producer and every consumer together — it can't silently drift.

I kept the graceful skip rather than failing loud, because the lookup legitimately misses in two benign cases: a non-BlocksStack test harness, and an AuthOIDC with no secret AppSetting at all. In a real BlocksStack the cookie-signing secret AppSetting always creates the bulk resource before this runs, and the deploy-time read-retry remains the documented fallback. Now that the id is a shared constant, the remaining risk (a relocation that changes the parent, not just the id) would be a much more visible refactor — happy to add a fail-loud guard for the "secret-sourced providers present but sibling missing" case too if you'd prefer belt-and-suspenders. Changeset + API report updated (bb-app-setting minor for the new export).

Simone319 (dismissed):
Approving — the coupling concern is resolved: BlocksSecretsBulk is now depended on via the exported SECRETS_BULK_CONSTRUCT_ID constant (compile-time, no silent drift). Keeping the graceful skip is fine given the benign miss cases. Thanks for the thorough iteration.

sarayev:
Re-review — thanks for the thorough iteration. All 8 of my earlier inline comments are addressed (rotation Trigger, terminal `ParameterNotFound` → actionable message, dropped `DescribeIdentityProvider`, dup-provider synth guard, extracted+unit-tested handler, per-provider assertions, `readSecret` cleanup + `ProviderType` in `PhysicalResourceId`, explicit `BlocksSecretsBulk` dependency via the exported `SECRETS_BULK_CONSTRUCT_ID` constant). The design is solid and the credential-never-in-template approach is the right call.

**One blocker remains, and it's the same must-fix from my first pass — it is currently red on head `a1483576`.** The two required checks (`Build and Test`, `Build, Unit Tests, E2E Local`) are failing because the CDK synth tests can't find the Lambda asset:

```
«CannotFindAsset» Cannot find asset at .../packages/bb-auth-oidc/dist/idp-registration-lambda
npm error command sh -c node --test --test-concurrency=1 dist/**/*.test.js
```

Root cause: `index.cdk.ts:190` does `Code.fromAsset(join(__dirname, 'idp-registration-lambda'))`, and that directory is produced only by `build:lambda` (esbuild). `build` chains it (`tsc --build && npm run build:lambda`), but the `test` script runs `node --test dist/**/*.test.js` with no guarantee `build:lambda` ran, so in the CI test job `dist/idp-registration-lambda/` is absent and 7 synth tests throw `CannotFindAsset` → the job exits 1.

Fix is a one-liner in `packages/bb-auth-oidc/package.json` — make the asset a precondition of the test run:

```json
"pretest": "npm run build:lambda",
```

(or `"test": "npm run build:lambda && node --test --test-concurrency=1 dist/**/*.test.js"` if you'd rather keep it explicit). Push that, confirm both required checks go green, and this is good to merge — Simone319 has already approved on design.

osama-rizk:
Fixed — thanks for the precise diagnosis (and the design approval). Added `"pretest": "npm run build:lambda"` to `packages/bb-auth-oidc/package.json`, so the esbuild asset (`dist/idp-registration-lambda/`) is always produced before `node --test` runs, independent of CI build orchestration.

Verified locally by reproducing the CI condition: `rm -rf dist/idp-registration-lambda && npm test` → `pretest` regenerates the asset → 138/138 pass (previously this threw `CannotFindAsset` on the 7 synth tests). Pushed on `b7d95b5c`; the two required checks should go green now.

osama-rizk:
@sarayev addressed — the `pretest` hook is pushed (`b7d95b5c`) and builds the `idp-registration-lambda` asset before `node --test`, so the `CannotFindAsset` failure is resolved. Package Integrity is already green and the unit/E2E-local check is re-running past the point it previously failed. Thanks again for the review!

sarayev:
Re-reviewed at head `b7d95b5` (read the code, not just the threads) — **LGTM, my blocker is cleared.**

- The `CannotFindAsset` CI failure is fixed by the `pretest: npm run build:lambda` hook, so the esbuild asset exists before the CDK synth tests run. All required checks are green (including `Build, Unit Tests, E2E Local`, E2E Sandbox, E2E Production).
- All 8 of my original comments remain addressed, and the `SECRETS_BULK_CONSTRUCT_ID` coupling is real: defined once in `secrets-bulk.ts`, re-exported from every bb-app-setting entry, used at the creation site and imported by bb-auth-oidc — a rename moves producer and consumer together.
- IAM is genuinely least-privilege: `cognito-idp:{Create,Update,Delete}IdentityProvider` on the pool ARN, `ssm:GetParameter` on the lazily-resolved specific param ARNs, `kms:Decrypt` gated on `kms:ViaService = ssm.<region>`.
- Handler: Create/Update/Delete with idempotent cross-fallbacks, retries transient errors, terminal `ParameterNotFound` → actionable message. Changeset bumps (auth-oidc minor, app-setting minor, blocks patch) are correct with an accurate pre-1.0 rationale.

One non-blocking note for the record (not a change request): the per-synth `Trigger: Date.now()` nonce means the IdP custom resource reports as updated on **every** `cdk deploy` and issues an `UpdateIdentityProvider` each time. That's the deliberate, documented trade-off for rotation propagation and it's idempotent — fine as-is; just flagging that a future "only re-run when the SSM version changes" optimization exists if the per-deploy call ever becomes noise.

Nothing else from me. Ready to merge.

sarayev (approved):

PR #522, opened 2026-09-09
fix(core): follow a factory-returned, destructured ApiNamespace in spec extraction (#444)

### LINKED ISSUE #444: blocks-generate-spec emits unknown schemas for a destructured ApiNamespace returned by a factory

@aws-blocks/core

### Describe the bug

`blocks-generate-spec` loses the TypeScript parameter and result schemas for an exported ApiNamespace when the namespace is constructed inside a class factory, returned as a property, destructured, and then exported.

The namespace exists normally at runtime and its qualified method appears in the generated OpenRPC document. Only its schema information is lost: both the parameter and result are emitted as { "type": "unknown" }.

An equivalent directly constructed and exported ApiNamespace in the same foundation file produces the correct schemas. The two operations deliberately have different method names, so this is not caused by a same-named-method collision.

This makes ordinary TypeScript composition change the generated API contract even though both values are valid ApiNamespace instances.

### Expected behavior

Schema generation should follow the exported ApiNamespace values and their TypeScript method signatures regardless of whether a namespace is constructed directly or returned through a factory abstraction.

For the reproduction:

- directNamespace.getDirectGreeting(name: string): Promise<number> should emit a string parameter and number result.
- indirectNamespace.getIndirectGreeting(times: number): Promise<string> should emit a number parameter and string result.

Neither schema should be unknown.

### Reproduction steps

Minimal reproduction: https://github.com/costleya/aws-blocks-indirect-namespace-schema-repro

Clone the reproduction repository.

1. Run pnpm install.
2. Run pnpm repro.
3. The direct namespace control passes.
4. The assertion for indirectNamespace.getIndirectGreeting fails because the generated parameter schema is unknown instead of number. Its result schema is also unknown instead of string.

### Code snippet

```typescript
import { ApiNamespace, Scope } from '@aws-blocks/core'

class NamespaceFactory {
  constructor(private readonly scope: Scope) {}

  createNamespaces() {
    return {
      indirectNamespace: new ApiNamespace(
        this.scope,
        'indirectNamespace',
        () => ({
          async getIndirectGreeting(times: number): Promise<string> {
            return `hello ${times}`
          },
        }),
      ),
    }
  }
}

const scope = new Scope('indirect-namespace-schema-repro')

export const directNamespace = new ApiNamespace(
  scope,
  'directNamespace',
  () => ({
    async getDirectGreeting(name: string): Promise<number> {
      return name.length
    },
  }),
)

const { indirectNamespace } = new NamespaceFactory(scope).createNamespaces()
export { indirectNamespace }
```

### Log output

```plain text
PASS control: directNamespace emits string parameter and number result schemas

AssertionError [ERR_ASSERTION]: BUG: destructured factory namespace parameter schema is unknown
+ actual - expected

+ 'unknown'
- 'number'


The generated OpenRPC result schema for indirectNamespace.getIndirectGreeting is also { "type": "unknown" }; the expected result type is string.
```

### Environment

- AWS Blocks version: @aws-blocks/blocks@0.2.7, @aws-blocks/core@0.1.18
- Node.js version: 24.18.0
- Package manager (npm/yarn/pnpm): pnpm@11.24.0
- OS: macOS
- CDK version (if applicable): N/A

### Additional context

The generated document contains both qualified runtime methods, so namespace discovery/runtime construction succeeds. The failure is specifically TypeScript schema extraction for the indirect export shape.

The current extractor's indirect pass is limited to particular top-level declaration shapes and does not follow this destructured factory result to the namespace methods: [`packages/core/src/scripts/extract-ts-types.ts`](https://github.com/aws-devtools-labs/aws-blocks/blob/main/packages/core/src/scripts/extract-ts-types.ts).

This affects consumers of the generated OpenRPC document, including native client generation: the declared TypeScript types degrade to `Any`/unknown even though the type information is present in the program.

### Suggested fix

Extend TypeScript extraction to follow exported binding values through callable/factory return types until it reaches the `ApiNamespace` handler method object. The traversal should:

- support object binding patterns and separately exported destructured bindings;
- carry the exported namespace name while descending through factory/container types;
- distinguish namespace container properties from actual API operations;
- use call-stack-scoped type/declaration cycle guards so recursive types terminate without suppressing valid sibling namespaces.

This can be implemented with the TypeScript compiler APIs already used by the extractor; changing the application into a particular source-code shape should not be required.

Suggested regression coverage:

- a direct namespace control;
- a namespace returned by a function factory;
- a class `createNamespaces()` result destructured and exported from another module;
- two sibling namespaces created through the same factory;
- a recursive non-namespace container type proving traversal terminates.

### Validated patch

A working pnpm patch against `@aws-blocks/core@0.1.18` is available on the [`codex/proposed-fix` branch](https://github.com/costleya/aws-blocks-indirect-namespace-schema-repro/tree/codex/proposed-fix). Its [`pnpm-workspace.yaml`](https://github.com/costleya/aws-blocks-indirect-namespace-schema-repro/blob/codex/proposed-fix/pnpm-workspace.yaml) registers the issue-specific [`@aws-blocks/core@0.1.18` patch](https://github.com/costleya/aws-blocks-indirect-namespace-schema-repro/blob/codex/proposed-fix/patches/%40aws-blocks__core%400.1.18.patch).

It patches both the TypeScript source and the shipped `dist` implementation. The relevant changes:

- thread the exported namespace name through direct, factory, destructured-binding, and resolved-type extraction;
- recursively inspect callable/factory return types and namespace container properties;
- store extracted operations under qualified keys;
- make declaration and type recursion guards call-stack scoped with `finally` cleanup, allowing sibling namespaces to reuse a factory safely;
- extend `@blocksSkipCodegen` discovery through factory and destructured exports while retaining exact qualified names.

The core shape is:

```typescript
function extractMethodsFromResolvedType(
  type: ts.Type,
  checker: ts.TypeChecker,
  result: Map<string, MethodTypeInfo>,
  namespaceName: string,
  activeTypes = new Set<ts.Type>(),
) {
  if (activeTypes.has(type)) return
  activeTypes.add(type)

  try {
    // Follow callable returns and namespace container properties,
    // preserving namespaceName while descending.
  } finally {
    activeTypes.delete(type)
  }
}
```

osama-rizk:
## Problem (#444)

`blocks-generate-spec` loses the TypeScript parameter/result schemas for an `ApiNamespace` that is constructed inside a factory, returned as a property, **destructured**, and then exported:

```ts
const { indirectNamespace } = new NamespaceFactory(scope).createNamespaces();
export { indirectNamespace };
```

The namespace's method appears in the generated OpenRPC document, but its param and result schemas come out as `{ "type": "unknown" }`. A directly-constructed, directly-exported namespace in the same file is fine.

### Root cause

`extractMethodTypes` has two passes. The first (AST walk) finds `new ApiNamespace(...)` syntactically and attributes methods to the binding name via `enclosingBindingName` — but for a namespace built inside a factory method, that walk hits a function/block boundary and returns `undefined`, so the method is keyed by its **bare** name. The second (type-checker) pass was meant to attribute indirect exports, but it only handled a **simple identifier** binding:

```ts
if (!decl.initializer || !ts.isIdentifier(decl.name)) continue;  // ← destructuring falls through
```

So `const { indirectNamespace } = …` was skipped, leaving the method on a bare key. My earlier #498 added a bare-key fallback in `generate-spec`, which masks the *minimal* repro, but the bare key is collision-prone: a factory namespace sharing a method name with another namespace cross-assigns or degrades to `unknown` — the exact #445 class of bug, via the factory path (the residual edge the generate-spec comment already calls out for "factory-wrapped" exports).

## Fix

Extend the indirect pass to handle **object binding patterns**. For each destructured binding, resolve the property's type off the initializer and extract the methods keyed by the **local (exported) namespace name** — so a factory-returned namespace keys qualified (`indirectNamespace.getIndirectGreeting`) exactly like a directly-constructed one, and same-named methods across namespaces no longer collide. Renamed bindings (`const { foo: bar } = …`) read property `foo` but key under `bar`.

## Validation

Repro'd first, then fixed (via `extractMethodTypes` on the issue's exact shape):

```
# before: KEYS: [ 'getIndirectGreeting' (bare), 'directNamespace.getDirectGreeting' ]
# after:  KEYS: [ …, 'indirectNamespace.getIndirectGreeting' ]  → params [number], ret string
```

Tests added:
- `extract-ts-types.test.ts` — the #444 factory+destructure shape keys qualified with real schemas; plus a **collision-safety** test (a factory `subscriptions.create(number)` alongside a direct `widgets.create(string)` keep distinct qualified keys, no cross-assign).
- `generate-spec.test.ts` — end-to-end: the factory/destructured namespace emits real (non-`unknown`) schemas, with a same-named-method collision to prove the fix.

Core suite **789/789**; `check:api` clean (no public API change); lint introduces no new findings.

## Changeset
`@aws-blocks/core` patch + `@aws-blocks/blocks` patch.

Fixes #444


packages/core/src/scripts/extract-ts-types.ts
```diff
@@ -214,18 +214,46 @@ export function extractMethodTypes(sourcePath: string): Map<string, MethodTypeIn
...
-			const initType = checker.getTypeAtLocation(decl.initializer);
-			extractMethodsFromResolvedType(initType, checker, result, decl.name.text);
+			// (b) Object binding pattern: `const { indirectNamespace } = factory()`.
+			// Each destructured binding is a separate namespace whose name is the
+			// LOCAL binding name (the name it's exported under). Resolve the property
+			// type off the initializer and extract methods keyed by that name, so a
+			// factory-returned namespace keys qualified (`indirectNamespace.method`)
+			// like a directly-constructed one — instead of landing on a bare key that
+			// collides with same-named methods in other namespaces (#444, #445).
+			if (ts.isObjectBindingPattern(decl.name)) {
```
Simone319, line 261: The AST-walk pass still emits the factory namespace's method under its bare, unqualified key, and this new pass adds the qualified key alongside it rather than replacing it — so both keys live in the result map. Since #498's bare-key fallback in generate-spec is also still in place, it would help to confirm the qualified key always wins for these namespaces (otherwise the #445-class collision is still reachable via the lingering bare key), and to note whether the #498 fallback is now redundant and removable. A short comment or a test asserting the bare key no longer influences the output would lock that down.

packages/core/src/scripts/extract-ts-types.ts
```diff
@@ -214,18 +214,46 @@ export function extractMethodTypes(sourcePath: string): Map<string, MethodTypeIn
...
+				const initType = checker.getTypeAtLocation(decl.initializer);
+				for (const element of decl.name.elements) {
+					if (!ts.isIdentifier(element.name)) continue;
```
Simone319, line 264: This handles object binding patterns but not array patterns (const [ns] = factory()) or nested object patterns (const { a: { b } } = ...), which still fall through to the bare-key path. That is probably rare for namespace exports and fails soft, but a brief comment noting the boundary (and a test) would document the limit and prevent a silent gap later.

packages/core/src/scripts/extract-ts-types.ts
osama-rizk: Confirmed and documented. The qualified key always wins for an attributable namespace: `generate-spec` looks up `tsTypes.get(qualified) ?? tsTypes.get(bare)`, and this pass now supplies the qualified key, so the lingering bare entry never influences the emitted schema. The existing generate-spec collision test proves this end-to-end (a factory `subscriptions.create(number)` alongside a direct `widgets.create(string)` — each keeps its own schema in the output), and I strengthened the extractor collision test to assert the direct namespace's qualified schema is unaffected by the factory's bare key. So the #445-class collision isn't reachable via the bare key here.

On removing #498's fallback: I'd keep it. It's the last resort for namespaces this pass still can't attribute (default exports, deeper indirection, the array/nested patterns below) — dropping it would regress those back to `unknown`. Added a code comment stating this so it's not mistaken for dead code.

packages/core/src/scripts/extract-ts-types.ts
osama-rizk: Good boundary call — documented and tested. Added a comment noting only a top-level identifier and a shallow object binding pattern are attributed; array patterns (`const [ns] = factory()`) and nested patterns (`const { a: { b } } = …`) fall through to the bare-key path, where #498's fallback still resolves the schema when there's no collision (fails soft, no crash). Added a regression test with an array pattern asserting exactly that: no qualified `tupleNs.ping` key, but the schema is still recovered under the bare `ping` key. Widening the traversal can follow if a real nested/array namespace-export case shows up.

Simone319 (approved):
Approving — both review points are addressed: qualified-key precedence is documented and the generate-spec `?? bare` lookup guarantees the qualified key wins (with a strengthened collision test), and the array/nested destructuring boundary is now documented and covered by a test. Nicely done.

packages/core/src/scripts/extract-ts-types.test.ts
```diff
@@ -580,3 +580,139 @@ describe('extractMethodTypes — open-shape intersections', () => {
...
+			// Each namespace keeps its OWN param type under a distinct qualified key.
+			assert.strictEqual((types.get('widgets.create')?.params[0].schema as any)?.type, 'string');
+			assert.strictEqual((types.get('subscriptions.create')?.params[0].schema as any)?.type, 'number');
+			// The first-pass AST walk still emits a lingering bare `create` for the
+			// factory namespace, but it never overrides a qualified key: the
+			// direct namespace's qualified `widgets.create` is unaffected above, and
+			// `generate-spec` prefers the qualified key (see the collision test in
+			// generate-spec.test.ts, which asserts this end-to-end in the output).
+			// So the bare entry can't reintroduce the #445 cross-assignment here.
+			assert.strictEqual((types.get('widgets.create')?.params[0].schema as any)?.type, 'string');
```
soberm, line 670: [nit] This asserts the exact same thing as line 662 (`widgets.create` param type is `string`). The comment above frames it as "the bare entry can't reintroduce the #445 cross-assignment here," but a literal re-assert of an earlier line doesn't add coverage — the value hasn't changed between the two asserts. Either drop this line, or make it assert something the earlier one didn't (e.g. that no bare `create` qualified key leaked, or `types.has('create')` reflecting the documented lingering-bare-key behavior).

packages/core/src/scripts/extract-ts-types.test.ts
```diff
@@ -580,3 +580,139 @@ describe('extractMethodTypes — open-shape intersections', () => {
...
+		} finally {
+			rmSync(dir, { recursive: true, force: true });
+		}
+	});
+
+	it('array/nested destructuring falls through to the bare key (documented boundary)', () => {
```
soberm, line 676: [nit] Nicely documented fail-soft boundary. Just flagging for visibility: array/tuple and nested destructuring stay unattributed (bare-key fallback only). Fine to defer, but if a factory ever returns a tuple of namespaces (`const [a, b] = factory()`) with a method-name collision, that path would hit the #445 class again. Worth a follow-up issue link in the comment if one exists.

soberm (approved):

packages/core/src/scripts/extract-ts-types.test.ts
osama-rizk: sure, will address in a follow up PR.

packages/core/src/scripts/extract-ts-types.test.ts
osama-rizk: sure, will address in a follow up PR.

packages/core/src/scripts/extract-ts-types.test.ts
osama-rizk: Addressed in follow-up #553 — replaced the redundant re-assert with a direct check of the documented lingering-bare-key behavior (types.has('create')), which the earlier assert didn't cover.

packages/core/src/scripts/extract-ts-types.test.ts
osama-rizk: Addressed in follow-up #553 — filed #552 for attributing array/tuple & nested destructuring, and linked it from the boundary comment.

PR #526, opened 2026-09-10
chore(test-apps/pipeline): remove redundant @aws-blocks/core dependency

Simone319:
## What

Removes the direct `"@aws-blocks/core": "*"` entry from the `dependencies` block of `test-apps/pipeline/package.json`.

## Why

`@aws-blocks/core` is already pulled in transitively: this test-app depends on `@aws-blocks/blocks`, and `packages/blocks/package.json` declares `"@aws-blocks/core": "^0.4.0"`. The direct entry is therefore redundant.

## Diff

```diff
   "dependencies": {
     "@aws-blocks/blocks": "*",
-    "@aws-blocks/core": "*",
     "@aws-blocks/pipeline": "*",
     "aws-cdk-lib": "2.257.0",
     "constructs": "^10.6.0"
   },
```

## Note for reviewers

The generated files under `test-apps/pipeline/aws-blocks/` (`index.ts`, `pipeline.cdk.ts`, `pipeline-default.cdk.ts`) import from `@aws-blocks/core/cdk` directly, so resolution now relies on workspace hoisting of the transitive dependency. Please confirm the pipeline synth test job stays green before merging.

### DIFF
--- a/test-apps/pipeline/package.json
+++ b/test-apps/pipeline/package.json
@@ -10,7 +10,6 @@
   },
   "dependencies": {
     "@aws-blocks/blocks": "*",
-    "@aws-blocks/core": "*",
     "@aws-blocks/pipeline": "*",
     "aws-cdk-lib": "2.257.0",
     "constructs": "^10.6.0"


osama-rizk (approved):

Simone319:
Closing this — the change turns out to have no benefit and a small accuracy cost.

**Zero install/download impact.** `@aws-blocks/core` is a local workspace package, not a registry download. `package-lock.json` has `"node_modules/@aws-blocks/core": {"resolved": "packages/core", "link": true}` — no `resolved` URL, no `integrity` hash, so there is nothing to fetch on any machine, cold cache or not. And the test-apps are never installed standalone: they run in place as workspace members via the single root `npm ci` (`pr-checks.yml`), so `core` is satisfied by one hoisted symlink whether it is listed directly or pulled transitively through `@aws-blocks/blocks`.

**No effect on the public npm numbers either.** The `core` > `blocks` download gap is structural dependency fan-in — `core` is declared by 22 packages (`blocks`, `auth-common`, and all 20 `bb-*`) — not CI traffic. Every CI path that touches `@aws-blocks/*` resolves via workspace symlinks, `file:` links, or the scoped `@aws-blocks:registry=http://localhost:4873/...` override, so none of it hits public npm.

**Arguably incorrect.** The generated files `test-apps/pipeline/aws-blocks/{index.ts,pipeline.cdk.ts,pipeline-default.cdk.ts}` import `Pipeline`, `Scope`, and `RawRoute` from `@aws-blocks/core/cdk` **directly**. Dropping the direct dependency makes the manifest less accurate about what the app actually imports.

Net: no benefit, minor accuracy loss. If declaration hygiene is wanted later, the right fix is switching those imports to `@aws-blocks/blocks/cdk` (which re-exports all three via `export * from '@aws-blocks/core/cdk'` in `packages/blocks/src/index.cdk.ts`) — and it belongs in the generator, since those files are generated.

PR #536, opened 2026-09-11
fix(create-blocks-app): pin aws-cdk CLI as a template devDependency

### LINKED ISSUE #228: Pin aws-cdk CLI as a template devDependency (avoid npx auto-install + version skew)

sharonyajain:
## Problem

None of the `create-blocks-app` templates declare the `aws-cdk` CLI in `devDependencies` — they only depend on `aws-cdk-lib` (the library). Because the CLI isn't declared, the `cdk` invocations in the generated deploy/sandbox
scripts get fetched on demand by `npx` on first run. For a freshly scaffolded app this means:

- **First-run install latency + a warning** — the first deploy prints `npm warn exec The following package was not found and will be installed: cdk@<version>` and pays a one-time install cost.
- **Version skew** — the on-demand `cdk` CLI floats independently of the pinned `aws-cdk-lib`, so the CLI and the library can drift apart across machines and over time.

**Issue #, if available:** Fixes #228

## Changes

Add `aws-cdk` to `devDependencies` in every template that ships a deploy path, pinned to the **same** version as that template's existing `aws-cdk-lib` so the CLI and library stay in lockstep:

- `auth-cognito`, `bare`, `default`, `demo`, `nextjs`, `react` → `aws-cdk` at `^2.1141.0`
- `backend` → `aws-cdk` at exact `2.1141.0` (mirrors its existing exact `aws-cdk-lib` pin)

The `amplify` template is untouched — it declares no CDK dependencies. The change is purely additive; no source, script, or lockfile changes. A`@aws-blocks/create-blocks-app` patch changeset is included.

## Validation

- Confirmed all seven affected templates declared `aws-cdk-lib` but no `aws-cdk` before this change; each now pins `aws-cdk` to its template's `aws-cdk-lib` version.
- Re-validated all seven `package.json` files as well-formed JSON after the edit.
- No automated tests added: this is a template manifest dependency change with no runtime code path to exercise in unit/integration tests. The intended behavioral outcome — a fresh scaffold's first deploy no longer prints the
  `npm warn exec ... will be installed: cdk` line — is verified by scaffolding a template and running `npm run deploy`.

## Checklist

- [x] PR description included

---


.changeset/pin-aws-cdk-cli-templates.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aws-blocks/create-blocks-app": patch
+---
+
+Pin the `aws-cdk` CLI as a template `devDependency` so scaffolded apps no longer auto-install the CLI via `npx` on first deploy (avoiding first-run install latency and an uncontrolled floating CLI version). The `aws-cdk` CLI is versioned independently of `aws-cdk-lib`, so it is pinned to its own release line (`^2.1141.0`) rather than to the library's version.
```
sarayev: The changeset gets the versioning right here (CLI on its own release line, independent of `aws-cdk-lib`), but the PR description says the opposite — it claims the CLI is pinned "to the same version as that template's existing `aws-cdk-lib`". They're on separate lines (`aws-cdk` is `2.114x`, `aws-cdk-lib` is `2.257`), so they can't match. Could you fix the "Changes" section of the description to line up with this changeset? Otherwise the next reviewer will assume the two move in lockstep.

packages/create-blocks-app/templates/backend/package.json
```diff
@@ -36,6 +36,7 @@
     "typescript": "^5.3.0",
     "tsx": "^4.7.0",
     "esbuild": "^0.27.1",
+    "aws-cdk": "^2.1141.0",
```
sarayev: The templates aren't consistent on version ranges: `backend` pins `aws-cdk-lib` exactly (`2.257.0`) while the other six use a caret, and this PR adds `aws-cdk` with a caret everywhere. Since these are scaffolding templates copied into a fresh app with no committed lockfile, a caret means two people scaffolding the same template at different times can land on different CLI versions — the version skew this PR is meant to reduce.

Worth picking one convention: exact pins across the board (most reproducible for a fresh scaffold, matches what `backend` already does), or carets across the board (auto-patch updates, matches the other six). Either is defensible — the one combination that doesn't make sense is `backend`'s exact library sitting next to a floating CLI. Your call on the direction.

PR #537, opened 2026-09-11
fix(astro): allow an empty dist/client for pure-SSR builds

osama-rizk:
A server/hybrid Astro app with **no static assets** — no `public/` files and no prerendered pages — produces an **empty `dist/client`**. The astro adapter treated that as a missing build output and threw:

```
[AstroBuildOutputMissingError] Astro server build output is missing or empty at <projectDir>/dist/client.
```

which blocks synth/deploy for an otherwise-valid SSR app.

## Why it is wrong

For `output: 'server'` (or `'hybrid'`), the required artifact is `dist/server/entry.mjs`. `dist/client` only holds static assets (files under `public/`, prerendered pages, hashed client JS/CSS). A pure-SSR page with none of those legitimately yields an empty `dist/client`, and that is fine: CloudFront routes every request to the SSR Lambda, and the downstream manifest handling (`staticAssets.directory`, prerender globbing, error-page detection) all tolerate an empty/absent client dir.

## Fix

Extract the server/hybrid output validation into `ensureSsrBuildOutput`:

- require `dist/server/entry.mjs` (the real SSR artifact) — throw if missing;
- **ensure `dist/client` exists** (create it when absent) instead of throwing when it is empty.

The `static` output path is unchanged (an empty `dist/` there is still a real failure).

## Tests

Added unit tests for `ensureSsrBuildOutput` covering empty / absent / populated `dist/client` (no throw) and a missing server entry (still throws). Adapter tests: 14/14 pass.

## Repro

An Astro `output: 'server'` project whose only page is fully server-rendered (no `public/` dir) builds `dist/server/entry.mjs` + an empty `dist/client`, and previously failed synth with `AstroBuildOutputMissingError`; adding any file under `public/` was the only workaround. With this change it builds and deploys as-is.

### DIFF
--- a/packages/hosting/src/adapters/astro.test.ts
+++ b/packages/hosting/src/adapters/astro.test.ts
@@ -15,6 +15,7 @@ import {
   installSharpForAstroSsr,
   patchAstroRemoteImageRedirects,
   ASTRO_REDIRECT_PATCH_MARKER,
+  ensureSsrBuildOutput,
 } from './astro.js';
 
 void describe('astroUsesSharpService — decide whether to ship sharp (issue #3)', () => {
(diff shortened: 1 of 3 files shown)


packages/hosting/src/adapters/astro.test.ts
```diff
@@ -208,3 +209,56 @@ export { loadRemoteImage };
...
+  });
+
+  it('throws AstroBuildOutputMissingError when the server entry is missing', () => {
+    const { clientDir, serverDir, serverEntry } = scaffold({ entry: false, client: 'empty' });
+    assert.throws(
+      () => ensureSsrBuildOutput(clientDir, serverDir, serverEntry, 'server'),
+      /AstroBuildOutputMissingError|build output is missing or empty/,
+    );
+  });
+});
```
soberm, line 265: **Suggestion**

These four tests are a solid, well-scoped check of `ensureSsrBuildOutput` in isolation, including the missing-entry failure mode. What do you think about also adding one case a level up that drives `astroAdapter` itself for a pure-SSR build — `skipBuild: true` with a scaffolded `dist/server/entry.mjs` and an empty/absent `dist/client` — asserting it does not throw and that `manifest.staticAssets.directory` points at the (now-created) client dir?

The reported bug was `astroAdapter` throwing `AstroBuildOutputMissingError`, not the helper; an adapter-level case would lock the customer-observable symptom and prove the manifest is still valid (SSR catch-all route + `staticAssets.directory`) when the client dir is empty.

Per `AGENTS.md` § "Fixing a Bug": *"Prefer e2e over unit for repros… Use unit tests only for purely internal logic."*

```ts
it('astroAdapter: pure-SSR build with empty dist/client synthesizes without throwing', () => {
  // scaffold dist/server/entry.mjs and an empty dist/client under tmp
  const manifest = astroAdapter({ projectDir: tmp, skipBuild: true });
  assert.equal(manifest.staticAssets.directory, path.join(tmp, 'dist', 'client'));
  assert.ok(fs.existsSync(manifest.staticAssets.directory));
});
```

packages/hosting/src/adapters/astro.ts
```diff
@@ -697,6 +692,33 @@ const directoryHasFiles = (dir: string): boolean => {
...
+  serverDir: string,
+  serverEntry: string,
+  output: AstroOutput,
+): void => {
+  if (!fs.existsSync(serverEntry)) {
+    throw buildOutputMissingError(serverDir, output);
+  }
+  if (!fs.existsSync(clientDir)) {
+    fs.mkdirSync(clientDir, { recursive: true });
+  }
```
soberm: **Suggestion**

Creating the client dir here is the right call for a pure-SSR build, but it's the one spot where an *unintended* empty client (a broken server build that dropped its `_astro/*`/`public/` assets) becomes indistinguishable from a legitimate one — the old check would have thrown. Since this file already surfaces notable synth conditions to the operator (e.g. the `process.stderr.write` notices around L153 and L258), consider emitting a one-line breadcrumb when we materialize an absent client dir, so a silently-empty client stays visible in build logs without reintroducing the false-positive failure:

```ts
if (!fs.existsSync(clientDir)) {
  fs.mkdirSync(clientDir, { recursive: true });
  process.stderr.write(
    'ℹ️  No dist/client emitted; treating as a pure-SSR build (CloudFront routes all requests to the SSR Lambda).\n',
  );
}
```

Minor, take-it-or-leave-it: unlike the rest of this module (which raises `HostingError` with an actionable `resolution`), a `mkdirSync` failure here would surface as a raw Node error. Probably fine given the build just wrote to `dist/`, but worth a deliberate choice.

packages/hosting/src/adapters/astro.ts
```diff
@@ -697,6 +692,33 @@ const directoryHasFiles = (dir: string): boolean => {
...
+export const ensureSsrBuildOutput = (
+  clientDir: string,
+  serverDir: string,
+  serverEntry: string,
+  output: AstroOutput,
+): void => {
+  if (!fs.existsSync(serverEntry)) {
+    throw buildOutputMissingError(serverDir, output);
+  }
+  if (!fs.existsSync(clientDir)) {
```
Simone319: Nit: this existsSync guard is redundant — fs.mkdirSync(clientDir, { recursive: true }) is already idempotent and does not throw when the directory exists, so this can just be an unconditional fs.mkdirSync(clientDir, { recursive: true }). Non-blocking. (Also +1 to the two open points already raised on this PR — an adapter-level pure-SSR repro test, and a breadcrumb / HostingError wrap so a silently-created empty client dir can't mask a genuinely broken build.)

packages/hosting/src/adapters/astro.test.ts
osama-rizk: Added the adapter-level test. It drives `astroAdapter({ projectDir, skipBuild: true })` with a scaffolded `dist/server/entry.mjs` and **no** `dist/client` (plus a minimal `node_modules/astro` for the version gate and an `output: 'server'` config), and asserts it does **not** throw, that `manifest.staticAssets.directory` points at the now-created client dir, and that the SSR compute target is present — so the customer-observable symptom (`AstroBuildOutputMissingError`) and a still-valid manifest are both locked, not just the `ensureSsrBuildOutput` helper. Good call on the AGENTS.md guidance.

packages/hosting/src/adapters/astro.ts
osama-rizk: Done — added the breadcrumb. When the client dir is genuinely absent we now emit `ℹ️  No dist/client emitted; treating as a pure-SSR build (CloudFront routes all requests to the SSR Lambda).` to stderr, so a silently-empty client from a broken build stays visible in build logs (matching the L153/L258 notices) without reintroducing the false-positive failure. It fires only on a materialized-absent dir, not when an empty client dir already exists.

On the HostingError wrap for `mkdirSync`: I left it as a raw call — it writes into `dist/` immediately after a successful build, so a failure there is essentially unreachable, and wrapping every fs call would add noise. Flagging it as the deliberate choice you invited.

packages/hosting/src/adapters/astro.ts
osama-rizk: Fixed — dropped the redundant guard; it's now an unconditional `fs.mkdirSync(clientDir, { recursive: true })` (idempotent). I kept a cheap `const clientAbsent = !fs.existsSync(clientDir)` *before* the mkdir purely to gate the new breadcrumb (so it fires only when we actually materialize an absent dir), not to guard the mkdir itself. And thanks for the +1s — both are now addressed (adapter-level repro test + the breadcrumb).

soberm (approved):

Simone319 (approved):

PR #540, opened 2026-09-14
docs(bb-agent): add a concrete React useChat example

sharonyajain:
## Problem

The `useChat` docs (`packages/bb-agent/README.md`) only showed `useChat` in its
framework-agnostic callback form and warned that it is "a factory function, not a
React hook — call it **once**, not on every render." The docs never demonstrated
the fix, so developers got stuck on two things the abstract example doesn't show:

- How `onMessagesChange` / `onLoadingChange` / `onInterrupt` bridge to a framework's reactivity.
- How to hold the single `useChat` instance without recreating it every render — the exact footgun the note warns about but doesn't demonstrate.

**Issue #, if available:** no linked issue (raised as a draft-issue request; documented directly as a docs improvement)

## Changes

- Added a **primary React example** to the "Full Examples" section: it holds the instance in a lazily-initialized `useRef` (so it survives re-renders), bridges `onMessagesChange` / `onLoadingChange` into `useState`, tears the subscription down with `chat.destroy()` in a `useEffect` cleanup, and renders the message list plus a send handler. This directly resolves the "call once, not a hook" footgun.
- Added a **one-line Next.js note**: same component, keep the `'use client'` directive because `useChat` opens a browser WebSocket and holds client state — no separate snippet.
- **Kept** the existing framework-agnostic (callback-form) example as the baseline; `useChat` is deliberately framework-neutral.
- Deliberately out of scope: lit-html / Nuxt / SvelteKit / Astro examples.
- Because `packages/bb-agent/README.md` is synced into `@aws-blocks/blocks` packaged content by `scripts/gen-block-docs.mjs`, the changeset bumps both `@aws-blocks/bb-agent` and `@aws-blocks/blocks` (patch) — matching the `pr-checks.yml` block-docs sync gate.

## Validation

- Docs-only change; no automated tests apply. Verified the example against the actual source (`packages/bb-agent/src/index.hooks.ts`): every option (`onMessagesChange` / `onLoadingChange` / `onInterrupt`), every called method (`sendMessage`, `respondToInterrupt`, `loadConversation`, `destroy`), the `ChatMessage` shape, and the `@aws-blocks/bb-agent/client` export path all match the real API.
- Ran `node scripts/gen-block-docs.mjs` and confirmed the synced `packages/blocks/docs/bb-agent/README.md` (gitignored, built at publish time) picks up the new section, which is why the `@aws-blocks/blocks` bump is required.

## Checklist

- [x] PR description included
- [ ] Tests are changed or added — N/A, documentation-only change


packages/bb-agent/README.md
```diff
@@ -804,7 +804,87 @@ await chat.sendMessage('Hello!');
...
+export function Chat() {
+  const [messages, setMessages] = useState<ChatMessage[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
+  const [input, setInput] = useState('');
+
+  // Create the instance exactly once. The ref survives every re-render,
+  // so the subscription and conversation state are never torn down.
+  // Initialize with `undefined` explicitly — @types/react 19 dropped the
+  // zero-argument useRef overload, so useRef<T>() alone fails to type-check.
+  const chatRef = useRef<ReturnType<typeof useChat>>(undefined);
```
osama-rizk: Please run this through `tsc` under strict @types/react 19 — the line whose comment claims R19-correctness likely doesn't compile as written. `useRef<ReturnType<typeof useChat>>(undefined)` passes `undefined` to a type parameter (`ChatInstance`) that excludes it. Under R19's tightened `useRef` overloads this either:

- errors outright (`Argument of type 'undefined' is not assignable to 'ChatInstance'`), or
- resolves to `RefObject<ChatInstance | undefined>` — in which case the later **unguarded** `const chat = chatRef.current;` makes `chat` possibly-`undefined`, so `chat.destroy()` / `chat.sendMessage(...)` need a non-null assertion.

The robust form is `useRef<ReturnType<typeof useChat> | undefined>(undefined)` **and** `const chat = chatRef.current!` after the lazy-init block. (Marked as "verify" because the exact 19.x patch decides which of the two failures fires — but one of them almost certainly does, and a docs example that explicitly calls out R19 should compile under R19.)

packages/bb-agent/README.md
```diff
@@ -804,7 +804,87 @@ await chat.sendMessage('Hello!');
...
+- **One instance, held in a ref.** `useRef` + the lazy `if (!chatRef.current)` guard is the React idiom for "construct once." Because the identifier is `use`-prefixed, `eslint-plugin-react-hooks` (bundled in the default Next.js and CRA configs) flags the guarded call as a conditional hook (`react-hooks/rules-of-hooks`). `useChat` is a factory, not a hook, so this is a false positive — the inline `eslint-disable-next-line` above the call silences it. What you must **not** do is call `useChat(...)` unguarded on every render: that recreates the instance each time and is the footgun the factory note warns about.
+- **Callbacks are your reactivity bridge.** `useChat` mutates its own message list in place; `onMessagesChange` / `onLoadingChange` hand you the new value so you can `setState` and trigger a render. Passing `setMessages` / `setIsLoading` directly is enough.
+- **Clean up on unmount** with `chat.destroy()` in a `useEffect` cleanup, so the Realtime subscription is closed.
+- **Approvals:** add `onInterrupt: setInterrupts` (with a `const [interrupts, setInterrupts] = useState([])`) to render an approval UI, then call `chat.respondToInterrupt([{ interruptId, approved: true }])`.
```
osama-rizk: `useState([])` here infers `never[]`, so `setInterrupts` is typed `(v: never[]) => void` and won't accept what `onInterrupt` passes (`Array<{ id: string; name: string; reason?: any }>`) — a TS error for anyone following this literally. Give it a typed initial value: `useState<Array<{ id: string; name: string; reason?: unknown }>>([])`. Lower stakes than the `useRef` line since it's prose guidance rather than the main block, but still copy-paste-facing.

osama-rizk (commented):
Really good example — I cross-checked every API against `index.hooks.ts` and it's accurate: `ChatMessage` fields (`id`/`role`/`content`), all four `ChatInstance` methods including `destroy()`, the `api`/`subscribe`/`onMessagesChange`/`onLoadingChange`/`onInterrupt` option shape, and the `./client` export subpath all match. The lazy `useRef` guard is the correct fix for the footgun (not the eager `useRef(useChat(...))` mistake), the `useEffect` cleanup + deps are right, the rules-of-hooks false positive is honestly explained, and the Next.js `'use client'` note is accurate.

Two issues are in the example's *TypeScript* — worth fixing because the whole value here is a copy-paste-correct snippet. The flagship `useRef` line (F1) is the one to verify before merge.

packages/bb-agent/README.md
soberm: [minor] `respondToInterrupt` throws at runtime unless `api.resume` is configured — but the example's `api` object above only wires `sendMessage` / `createConversation` / `getConversation`. A reader who follows this section and adds the Approvals bullet verbatim will hit:

```
respondToInterrupt requires api.resume to be configured
```

(`index.hooks.ts`: `if (!options.api.resume) throw new Error('respondToInterrupt requires api.resume to be configured')`.)

Suggest adding one clause so the approvals path is self-contained, e.g.: "…and add `resume: (chId, responses, convId) => api.resume(...)` to the `api` object above (mirroring your backend's resume method)."

Non-blocking — approvals are explicitly a follow-on to the primary example.

soberm (dismissed):

osama-rizk (dismissed):

packages/bb-agent/README.md
sharonyajain: Fixed in c8932d2e. Good catch. You were right that the previous form just traded one R19 error for another. The ref is now typed `ReturnType<typeof useChat> | undefined` with an `undefined` initializer, and the read after the lazy-init block is `const chat = chatRef.current!` so `chat.destroy()` / `chat.sendMessage()` stay non-null.

packages/bb-agent/README.md
sharonyajain: Fixed in c8932d2e. The approvals bullet now types the state explicitly: `useState<Array<{ id: string; name: string; reason?: unknown }>>([])`, so `setInterrupts` accepts the `onInterrupt` payload instead of inferring `never[]`.

sharonyajain: Fixed in c8932d2e. The approvals bullet now tells the reader to wire `resume: (chId, responses, convId) => api.resume(chId, responses, convId)` into the `api` object, noting `respondToInterrupt` throws if it is absent so following the section verbatim no longer hits the runtime error.

osama-rizk (approved):

soberm (approved):

PR #544, opened 2026-09-14
feat(compute): mount a per-namespace /aws-blocks/api/{namespace} ingress

Simone319:
## Problem

A CloudFront behavior routes on **path**, not on request body. Today every RPC call goes to a single endpoint — `POST /aws-blocks/api` — with the target namespace inside the JSON-RPC body, so from the front door's perspective all namespaces look identical. There is no way to say "send `orders` to compute A and `auth` to compute B", because there is no URL that distinguishes them.

Before per-namespace routing to multiple computes can exist, each namespace needs a distinct, predictable address on the gateway. This PR creates those addresses.

## Changes

**`Compute.mountNamespaceRoutes()`** — a no-op on the base class, since a compute with no HTTP ingress has nothing to mount. Called at synth finalize from both `BlocksStack.create()` and `BlocksBackend.create()`, *after* every `ApiNamespace` has recorded itself on its compute (so the namespace list is complete).

**`LambdaCompute` overrides it** to add `/aws-blocks/api/{namespace}` and `/aws-blocks/api/{namespace}/*` on its API Gateway, both proxying to the same Lambda. To mount under it, the `/aws-blocks/api` resource is now captured as a private field instead of a local.

**Why nest under the `api` segment** rather than a new root like `/aws/{namespace}`: it scopes the tree to API traffic, so a namespace name can never collide with a sibling framework endpoint such as `/aws-blocks/dashboard`. It also means the runtime needs no change — the handler already dispatches every path under `/aws-blocks/api` as an RPC request — so these paths work even against an already-deployed backend.

Mounting is **idempotent**, so a repeated finalize pass (or a duplicate namespace name) won't create duplicate resources.

### This is additive — and deliberately has no client change

Worth stating explicitly, because it's the obvious question:

- `/aws-blocks/api` keeps serving exactly as before.
- The **client is untouched** — it still posts to `/aws-blocks/api` with the namespace in the body (`core/src/client/index.ts`).
- **Dispatch is untouched** — it still reads the namespace from the body.

So after this PR there are two functionally identical ways in, and all existing traffic continues to use the original one. The client's move to a per-namespace path lands later, together with the CloudFront behaviors that make the distinction meaningful. Keeping the two apart means this change deploys with zero request-path movement and can be verified on its own.

## Validation

**New unit tests** (`bb-lambda-compute/src/index.cdk.test.ts`, 3 added) covering the contract:
- a `/aws-blocks/api/{namespace}` resource is created per recorded namespace
- no namespaces recorded → nothing is created (compared against an identically-built baseline stack's gateway-resource count, so it can't pass vacuously)
- repeated `mountNamespaceRoutes()` calls don't duplicate the namespace resource

**Real synth check.** Synthesized `test-apps/extending-blocks-guide-blocksbackend`, whose backend declares a namespace named `api`, and confirmed the template now contains the mounted resource and its proxy alongside the original endpoint:

```
PathPart="aws-blocks"   .../aws-blocks/Resource
PathPart="api"          .../aws-blocks/api/Resource        ← original endpoint, still there
PathPart="api"          .../aws-blocks/api/api/Resource    ← mounted namespace ingress
PathPart="{proxy+}"     .../api/api/{proxy+}/Resource      ← its sub-path proxy
```

Verified on Node 22: `npm run build`, `npm run lint`, `npm run lint:deps`, `npm run check:api`, and the changeset guard all pass. Tests: bb-lambda-compute **41/41** (38 + 3 new), core cdk **53/53**, blocks umbrella **42/42 + 2/2**.

## Checklist

- [x] PR description included
- [ ] Relevant documentation is changed or added (and PR referenced) — no customer-facing surface changes yet (the new paths are infrastructure for later per-namespace routing; the documented endpoint is unchanged)

---


svidgen (commented):
Would it make any sense to just leverage `RawRoute` under the hood here? I suspect it could bring this down to about a 1 line change inside `ApiNamespace`, and then we don't have *N* path routing utils and mechanisms to maintain, etc... also forces us to ensure `RawRoute` actually routes to the correct compute handler as well, perhaps? 😅

svidgen (approved):

Simone319:
@svidgen 
> Would it make any sense to just leverage `RawRoute` under the hood here?

CloudFront routes on path, not body — so each namespace needs a distinct gateway address for the front door to send it to the owning compute - and CloudFront forwards the path unchanged, so the owning compute's gateway must have a resource matching `/aws-blocks/api/{ns}` or it 403s . Not sure how RawRoute could be used instead, as it registers no gateway resource.

packages/bb-lambda-compute/src/index.cdk.ts
```diff
@@ -238,4 +243,28 @@ export class LambdaCompute extends Compute {
...
+	 * compute that owns a namespace. Called at synth finalize once namespaces are
+	 * recorded; idempotent, so a repeated finalize won't duplicate resources.
+	 */
+	override mountNamespaceRoutes(): void {
+		if (this.namespaces.length === 0) return;
+		const integration = new apigateway.LambdaIntegration(this.fn);
+		for (const ns of this.namespaces) {
+			// Idempotent across repeated finalize passes / duplicate namespaces.
+			if (this.rpcResource.getResource(ns)) continue;
+			const nsResource = this.rpcResource.addResource(ns);
```
soberm, line 265: [minor] `ApiNamespace` names are not currently validated, but this newly uses them as API Gateway path parts. A name such as `café` (a valid JavaScript identifier), `orders/v1`, or `My Orders` now makes `addResource(ns)` throw during synth because API Gateway accepts only `[a-zA-Z0-9:._-$]+`; one such namespace aborts the entire stack and the error does not identify the namespace. Please validate the name before mounting the route (with an error that names it) or document and test the newly required path-safe charset.

PR #551, opened 2026-09-15
fix(kotlin): generate serializers for nested transferables

### LINKED ISSUE #548: kotlin codegen: nested transferable references a serializer that is never generated
### Summary

When a transferable appears as a **field of an operation-scoped result record** (rather than as the method's return type), the Kotlin generator emits a `@Serializable(with = ...)` annotation pointing at a serializer object that it never generates. `Serializers.kt` is not emitted at all, and nothing else declares the referenced name, so the generated client does not compile.

This is not specific to third-party or unrecognized transferables — it reproduces with `realtime/channel`.

### Reproduction

Add a fixture at `native/codegen-fixtures/99-repro/spec.json`:

```json
{
  "openrpc": "1.3.2",
  "info": { "title": "test", "version": "1.0.0" },
  "methods": [
    {
      "name": "api.getSession",
      "params": [],
      "result": {
        "name": "GetSessionResult",
        "schema": {
          "type": "object",
          "properties": {
            "sessionId": { "type": "string" },
            "channel": {
              "x-blocks-transferable": "realtime/channel",
              "x-blocks-type-args": [
                { "type": "object", "properties": { "text": { "type": "string" } }, "required": ["text"] }
              ]
            }
          },
          "required": ["sessionId", "channel"]
        }
      }
    }
  ]
}
```

Then:

```bash
cd native/kotlin && ./gradlew :codegen:regenerateFixtures --rerun-tasks
```

### Actual result

Two files are generated — `Api.kt` and `Servers.kt`. No `Serializers.kt`.

`Api.kt`:

```kotlin
package com.example.app

import com.aws.blocks.kotlin.BlocksClient
import com.aws.blocks.kotlin.BlocksRequest
import com.aws.blocks.kotlin.BlocksServer
import com.aws.blocks.kotlin.json.BlocksJson
import com.aws.blocks.kotlin.realtime.RealtimeChannel
import kotlin.String
import kotlinx.serialization.Serializable
import kotlinx.serialization.json.decodeFromJsonElement

public class Api(
  private val server: BlocksServer = Servers.local,
) {
  private val client: BlocksClient = BlocksClient(server)

  public suspend fun getSession(): GetSession.Result {
    val request = BlocksRequest(method = "api.getSession", params = emptyList(), id = BlocksRequest.nextId())
    val result = client.execute(request)
    return BlocksJson.decodeFromJsonElement(result)
  }

  public object GetSession {
    @Serializable
    public data class Result(
      public val sessionId: String,
      @Serializable(with = RealtimeChannelChannelSerializer::class)
      public val channel: RealtimeChannel<Result.Channel>,
    ) {
      @Serializable
      public data class Channel(
        public val text: String,
      )
    }
  }
}
```

`RealtimeChannelChannelSerializer` is referenced unqualified on line 27, is not imported, and is not declared by any generated file.

### Expected result

Either a `Serializers.kt` containing the referenced serializer object, or no annotation referring to one.

### Cause

Two code paths disagree about which transferables exist.

The annotation is emitted while walking a record's fields, so it fires for any transferable-typed field (`KotlinCodeGenerator.kt`, in the record-emission path around the `isTransferableType(field.type)` check).

The serializer objects come from `collectTransferableEntries`, which walks only `model.typeDefinitions`:

```kotlin
for (typeDef in model.typeDefinitions) {
    when (val t = typeDef.type) {
        is ResolvedType.Record -> t.fields.forEach { visit(it.type) }
        else -> {}
    }
}
```

`GetSession.Result` is scoped to the operation and is emitted through the operation's nested-type path, not registered as a shared type definition. Confirmation from the same run: no `Types.kt` is generated for this spec, which means the type index — built from `model.typeDefinitions` — held no records at all. So `collectTransferableEntries` returns empty, and `generate` skips `Serializers.kt` because `transferableEntries.isNotEmpty()` is false.

### Scope

Any transferable nested in an operation-scoped record. The same shape with an unrecognized tag produces the same dangling reference under the name `UnknownLinkSerializer` (the generated serializer name for an unrecognized tag is `Unknown` plus a suffix derived from the type arguments, with the tag itself omitted — so two unrecognized tags with matching type arguments would also collapse to one name once that path emits real code).

### Suggested direction

Collect the serializer entries from the same traversal that emits the annotation, so the two cannot diverge — or extend the collection to include operation-scoped nested records. Keying entries by tag rather than by a name derived only from type arguments would also remove the collision noted above.

### Environment

- `main` @ `6496713695aa75706ea174172efe6eddad4920f4`
- Verified via `:codegen:regenerateFixtures`

ikenyal:
## Problem

An operation-scoped result record can contain a transferable field such as `realtime/channel`. The Kotlin generator emits `@Serializable(with = ...)` for that field, but it previously collected serializer declarations in a separate traversal. Those paths could diverge for nested types, union variants, or repeated short names, leaving a missing serializer or the wrong payload type.

**Issue #, if available:** Fixes #548

## Changes

- Register serializer entries at the same annotation site that references them.
- Build relative and fully qualified operation type paths in one traversal, and key direct type resolution by the resolved type rather than a short name.
- Key serializer entries by transferable tag and resolved types; name allocation now distinguishes parameterized payloads.
- Generate transferable serializers for union variant fields.
- Expand fixture 25 with same-named nested payloads, union variants, and `List<String>` / `List<Int>` payloads, with Kotlin, Swift, and Dart goldens.
- Improve fixture assertion failures with fixture and serializer context, including wrapped annotations.
- Add a patch changeset for `aws-blocks-kotlin`.

## Validation

- `./gradlew :codegen:regenerateFixtures`
- `./gradlew :codegen:build`
- `dart test test/golden_file_test.dart` (Dart 3.13.3)
- Generated and compared the Swift fixture output with the repository's Swift generator.

## Checklist

- [x] PR description included
- [x] Documentation impact assessed; no documentation change is needed because this fixes generated source for an existing supported schema.

---


native/codegen-fixtures/25-nested-transferable-operation-result/spec.json
```diff
@@ -0,0 +1,53 @@
+{
```
mattcreaser, line 1: HI @ikenyal, you need to add the golden files for dart/swift for this new fixture. Please run `native/codegen-fixtures/regenerate-all.sh` and add the dart/swift directories to this PR.

Alternatively, you could modify the existing `17-transferables` fixture to include the nested case, and again run `native/codegen-fixtures/regenerate-all.sh`. Either works.

Thanks!

mattcreaser (changes requested):

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
```diff
@@ -1278,6 +1372,22 @@ class KotlinCodeGenerator(
             }
         }
 
+        fun visitNestedType(node: NestedTypeNode, opContext: OperationTypeContext?) {
+            when (val type = node.type) {
+                is ResolvedType.Record -> type.fields.forEach { visit(it.type, opContext) }
+                else -> {}
+            }
+            node.children.forEach { visitNestedType(it, opContext) }
+        }
```
mattcreaser: New collection walk skips union variants, so variant transferables still name a missing serializer

visitNestedType matches only ResolvedType.Record, so it never descends into a node whose type is ResolvedType.Union, nor into variant.nestedTypes. The annotation side does emit for those shapes, so the two halves still disagree for any transferable reached through a union variant.

Repro, verified against this branch: an operation result declared as oneOf, where one variant holds a nested object with a realtime/channel field, generates

    @Serializable(with = RealtimeChannelPayloadChannelSerializer::class)
    public val channel: RealtimeChannel<Payload.Channel>,

and no Serializers.kt at all, so the client does not compile. A transferable declared directly on a variant is worse: generateSealedVariant never applies the @Serializable(with = ...) annotation, and RealtimeChannel is not @Serializable, so that fails to compile too.

A second contributing cause is generateSealedVariant calling generateNestedTypeSpec(nested, index) with neither parentPath nor serializerPath, which leaves the annotation resolving against the package root.

Both shapes are also broken on main, so this is not a regression. Raising it because the PR is scoped to exactly this defect class and adds variant handling elsewhere, so the case currently looks covered when it is not. Worth fixing here, or adding a fixture that pins the known-bad output so it is tracked.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
```diff
@@ -854,44 +856,93 @@ class KotlinCodeGenerator(
...
+    ): OperationTypeContext? {
+        if (operation.nestedTypes.isEmpty()) return null
+        val context = OperationTypeContext(apiClassName)
+
+        fun walkNodes(nodes: List<NestedTypeNode>, parentPath: List<String>) {
+            for (node in nodes) {
+                val currentPath = parentPath + node.name
+                val className = ClassName(packageName, apiClassName, *currentPath.toTypedArray())
+                context.typeMap[node.name] = className
+                context.serializerTypeMap[node.name] = className
```
mattcreaser: serializerTypeMap is keyed by short name, so same-named payloads in one operation still collapse

walkNodes writes every node into one flat map keyed by node.name, so two nested types sharing a short name inside a single operation overwrite each other. The description says operation-local types with the same short name no longer collide, but that holds only across operations, not within one.

Repro, verified against this branch, for Result { channel: RealtimeChannel<{text}>, inner: { channel: RealtimeChannel<{count}> } }, where both anonymous payloads are named Channel:

    public data class Result(
      @Serializable(with = RealtimeChannelApiGetSessionResultInnerChannelSerializer::class)
      public val channel: RealtimeChannel<Result.Inner.Channel>,
      public val inner: Result.Inner,
    )

Result.channel should carry Result.Channel. One serializer is emitted for two distinct payload types, and the generated Result.Channel class is never referenced.

The wrong field type comes from the pre-existing typeMap, which has the same short-name keying. What changes here is the failure mode: on main this did not compile, because no serializer object existed at all; now it compiles and decodes the wrong payload at runtime. The new assertion in CodegenFixturesTest cannot catch that, since it only checks that some object with the referenced name exists somewhere in the output.

Keying on the resolved fully qualified path rather than the short name would fix both this and the field type.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
```diff
@@ -854,44 +856,93 @@ class KotlinCodeGenerator(
...
+                if (node.children.isNotEmpty()) {
+                    walkNodes(node.children, currentPath)
+                }
+                if (node.type is ResolvedType.Union) {
+                    for (variant in node.type.variants) {
+                        if (variant.nestedTypes.isNotEmpty()) {
+                            walkNodes(variant.nestedTypes, currentPath + variant.name)
                         }
                     }
                 }
```
mattcreaser: Suggestion: Union-variant branch in the serializer context builder is unreachable

This branch fills serializerTypeMap for variant-scoped nested types, but the only reader of a context built by buildSerializerOperationTypeContext is visitNestedType, which never descends into unions. The same holds for the variant branches added to walkNodes in buildOperationTypeContext and to registerChildren in generateNestedTypeSpec: nothing consults serializerTypeMap on a variant path, because generateSealedVariant does not emit transferable annotations at all.

So this reads as though union variants are handled when they are not, which is easy to trust on a later change. Either wire the collection side through unions, per the comment on visitNestedType, and keep these branches, or drop them.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
```diff
@@ -1225,48 +1316,51 @@ class KotlinCodeGenerator(
...
-                    else -> "JsonElement"
+        val suffix = type.typeArgs.joinToString("") { arg ->
+            when (val resolved = resolveSerializerNameType(arg, index, opContext)) {
+                is ClassName -> {
+                    resolved.simpleNames.joinToString("")
                 }
+                else -> "JsonElement"
             }
-        } else ""
+        }
```
mattcreaser: Suggestion: Serializer name still collapses to JsonElement while the decoded type is now precise

The name falls back to JsonElement for any payload that does not resolve to a ClassName, but typeArgument in collectTransferableEntries is now the precise resolveResolvedType result. Two payloads that both collapse to the same name are deduplicated by the seen set, so only the first survives, and the surviving object is typed for that first payload only.

Repro, verified against this branch, for Holder { a: RealtimeChannel<List<String>>, b: RealtimeChannel<List<Int>> }: both properties are annotated with RealtimeChannelJsonElementSerializer, and the single generated object is KSerializer<RealtimeChannel<List<String>>>, so applying it to b does not compile.

This collision predates the PR and does not compile on main either, so it is not a regression. Flagging it because this is the expression the PR rewrites, and because issue 548 called out the same keying problem: entries keyed by a synthesized name rather than by tag plus resolved payload type. Deriving the suffix from the full rendered TypeName, or keying entries on (transferableName, returnType), removes the whole class.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/TransferableSerializerGenerator.kt
```diff
@@ -14,9 +12,9 @@ class TransferableSerializerGenerator(
 ) {
```
mattcreaser: Suggestion: packageName is now unused in TransferableSerializerGenerator

Removing resolveTypeArgClassName took away the last read of packageName, so this constructor property is now dead. Worth dropping it here and at the construction site in KotlinCodeGenerator.

mattcreaser:
One design concern before this lands. Issue 548 suggested collecting the serializer entries from the same traversal that emits the annotation, so that the two cannot diverge. This change goes the other way: correctness now depends on three separately written path builders agreeing, buildOperationTypeContext, buildSerializerOperationTypeContext, and the childContext inside generateNestedTypeSpec. Two of the inline comments below are that disagreement showing up already, once for union variants and once for same-named nested payloads within one operation. Suggestion: have the annotation site record the entry it references, or have both sides read one map produced by a single walk, and key entries on the transferable tag plus the resolved payload type instead of a synthesized name string.

Two smaller notes on the new assertion in CodegenFixturesTest. On failure it prints only "expected true but was false", with no fixture or serializer name, so withClue or shouldContain would read better. And the regex misses annotations that KotlinPoet wraps across lines, which gets more likely as these names grow, as they do here.

native/codegen-fixtures/25-nested-transferable-operation-result/spec.json
ikenyal: Good catch. Seeing the fixture in the cross-language suite made the gap clear. The update adds Dart and Swift goldens alongside the Kotlin output for fixture 25.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
ikenyal: The variant repro was especially helpful because it showed that extending the collection walk would still leave two paths that could diverge. Serializer entries are now registered at the serialization annotation site, including variant fields. Fixture 25 now covers both nested and direct variant transferables.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
ikenyal: The root/inner example exposed the more serious failure mode: a short-name collision could compile while decoding with the wrong payload type. Operation contexts now resolve direct nested types by their resolved type rather than their short name, and the serializer registry is keyed by the transferable tag and resolved types. The fixture includes both root and nested Channel payloads.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
ikenyal: Agreed. I removed the separate serializer collection path rather than leaving variant branches with no reader. Variant properties now emit their serializer annotation and register the matching entry through the same path as record properties.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/TransferableSerializerGenerator.kt
ikenyal: Removed the unused constructor argument and updated the construction site.

native/kotlin/codegen/src/main/java/com/aws/blocks/kotlin/generator/KotlinCodeGenerator.kt
ikenyal: This was a useful boundary case for the registry key. The fixture now includes realtime channels whose payloads are lists of strings and integers; they produce distinct serializer entries and names based on the resolved parameterized payload types.

ikenyal:
The design concern was well founded. The update removes the separate collector entirely: serializers are registered when the property annotation is emitted, and the registry uses the transferable tag plus resolved types. That also lets the variant path use the same code. Fixture 25 now covers same-named nested payloads, variant fields, and parameterized list payloads. The fixture assertion now includes fixture and serializer context and accepts wrapped annotations; Dart and Swift goldens are included as well.

mattcreaser (approved):
Looks good!

VarshithaPamisetty (approved):

Issue #552, opened 2026-09-16
spec extraction: attribute array/tuple & nested destructuring of factory-returned ApiNamespaces


osama-rizk:
Follow-up from PR #522 (#444).

`extractMethodTypes` (packages/core/src/scripts/extract-ts-types.ts) attributes a factory-returned `ApiNamespace` to its namespace only for a top-level identifier or a **shallow object** binding pattern. Array/tuple patterns and nested destructuring fall through to the bare-key path:

```ts
const [ns] = factory();        // array/tuple pattern → bare key only
const { a: { b } } = factory(); // nested → bare key only
```

Today this is fail-soft: the schema is still recovered under the bare method name, and generate-spec's #498 fallback resolves it **when there is no method-name collision**. But if a factory returns a tuple of namespaces with a colliding method name (e.g. `const [a, b] = factory()` where both expose `create`), that path would hit the #445 class of cross-assignment again.

Documented as an accepted boundary in `extract-ts-types.test.ts` ('array/nested destructuring falls through to the bare key'). Filing so the comment can link a tracked follow-up.

PR #553, opened 2026-09-16
docs(create-block): add README + DESIGN; address extract-ts-types test nits (follow-up to #422/#522)

osama-rizk:
Follow-up to #422 (create-block) and #522 (#444 spec extraction), addressing three deferred review items.

## 1. create-block was missing a DESIGN.md (and README.md)
`@aws-blocks/create-block` shipped `files: ["dist","templates","LICENSE"]` — no `README.md` or `DESIGN.md`, unlike first-party packages (`bb-kv-store` ships both). This adds:
- **`README.md`** — user-facing: modes (contributor/customer/external), options, and what gets generated; defers to `--help` as the authoritative reference.
- **`DESIGN.md`** — internals: mode detection, templating/tokens, the six contributor-mode touchpoints, per-mode dependency pinning, and testing.
- Both added to `package.json` `files` so they ship to npm.

## 2 & 3. extract-ts-types test nits (from #522)
Test/comment only — no runtime change:
- **[r4003768090]** The last assert in the `#445` regression test was a literal re-assert of an earlier line. Replaced it with a direct check of the *documented* lingering-bare-key behavior (`types.has('create')`), which the earlier assert didn't cover.
- **[r4003768097]** Added a tracking-issue link (#552) to the array/tuple & nested-destructuring fail-soft boundary comment, per the reviewer's suggestion. Filed #552 for the fuller fix.

## Validation
- `extract-ts-types.test.ts`: **18/18** pass (incl. the strengthened assert).
- `create-block`: build + **28/28** tests, biome clean.
- Changeset covers both changed packages; `validate-structure` + `verify-coverage` guards pass locally; README/DESIGN links resolve.


packages/create-block/README.md
```diff
@@ -0,0 +1,46 @@
+# @aws-blocks/create-block
```
soberm, line 1: [blocking] Adding this README makes the required `Catalog in sync` check fail (`node scripts/sync-catalog.mjs --check`, run 35079779473): its inclusion rule is any `packages/*` with a `README.md` not in `EXCLUDED`, and `create-block` is not excluded, so the generated catalog gains a `create-block` row that `packages/blocks/README.md` does not have. `create-block` is a scaffolding CLI, not a `bb-*` Building Block; its sibling `create-blocks-app` is already in `EXCLUDED`. Please add `'create-block'` to the intentionally matching `EXCLUDED` sets in both `scripts/sync-catalog.mjs` and `scripts/gen-block-docs.mjs`. Do not resolve this by running `npm run sync-docs`, which would list a scaffolding CLI as a Building Block.

osama-rizk: Fixed — added 'create-block' to the EXCLUDED set in both scripts/sync-catalog.mjs and scripts/gen-block-docs.mjs (kept in sync), matching create-blocks-app. `node scripts/sync-catalog.mjs --check` now passes and the catalog gains no create-block row. Did not run sync-docs.

scripts/gen-block-docs.mjs
```diff
@@ -48,7 +48,7 @@ const packagesDir = join(__dirname, '..', 'packages');
 const blocksDir = join(packagesDir, 'blocks');
 const outDir = join(blocksDir, 'docs');
 
-const EXCLUDED = new Set(['blocks', 'data-common', 'foundations', 'create-blocks-app', 'bb-lambda-compute']);
+const EXCLUDED = new Set(['blocks', 'data-common', 'foundations', 'create-blocks-app', 'create-block', 'bb-lambda-compute']);
```
Simone319: Correct to exclude create-block here — adding a README.md to it makes it eligible for the getPackages() has-README heuristic, so it would otherwise leak into the catalog/docs. Two follow-ups: (1) this is now the second lockstep edit of the duplicated EXCLUDED set, and this file's own note says to extract a shared module if the pair grows — might be the moment to do that so the two can't drift; (2) worth confirming no other README-presence enumerator (e.g. validate-structure / verify-coverage) now newly picks up create-block. Both non-blocking.

osama-rizk: Both addressed in 488ee17a:

1. Extracted the shared block set into scripts/block-packages.mjs (EXCLUDED + getBlockPackages()); sync-catalog.mjs and gen-block-docs.mjs now both import it, so they can't drift. Kept it dependency-free (Node stdlib, no build step).
2. Checked the other README/package enumerators — only these two use the has-README heuristic. changeset-guard.ts enumerates via git-diff + the .changeset dir (not README presence), so it doesn't pick up create-block. Verified: `node scripts/sync-catalog.mjs --check` passes and `gen-block-docs.mjs` emits 23 block docs with no create-block entry.

soberm (approved):

Simone319 (approved):

PR #554, opened 2026-09-16
fix(bb-kv-store): align delete() conditional detection with the mock and put() (parity)

osama-rizk:
## Problem

The AWS `delete()` path detected the value-equality condition with `'ifValueEquals' in conditions` (key **presence**), while the mock and AWS `put()` use `!== undefined` (with a comment on `put` explicitly saying "to match the mock"). Two mock↔AWS parity breaks:

1. **`delete(key, { ifValueEquals: undefined })`** — a silent unconditional delete on the mock, but on AWS it emitted `#value = :expected` with `:expected = JSON.stringify(undefined)` (which is `undefined`) → a DynamoDB DocumentClient **marshalling error** rather than a delete.
2. **Both `ifExists` and `ifValueEquals` set** — the `if/else if` applied only `attribute_exists(#pk)`, silently **dropping the value check**, so AWS deleted the item regardless of its value while the mock (correctly) required both.

The `parity.test.ts` comment already flagged this exact hazard for `put`, but nothing covered AWS `delete`.

## Fix

Compose `delete()` conditions **conjunctively** — `attribute_exists(#pk) AND #value = :expected` — using `!== undefined` detection, matching the mock branch-for-branch (which checks existence, then value, and requires both). No condition is emitted when none is set, and an explicit `{ ifValueEquals: undefined }` is a no-op on both layers.

Behavior unchanged for the single-condition cases; the retriability mapping (`ifValueEquals` set && !`ifExists`) is untouched.

## Tests

`parity.test.ts` — a new `DeleteCommand`-shape describe asserting each combination: value-only, exists-only, **both (AND, value check retained)**, none, explicit `undefined` no-op, and `null` as a real condition. Verified **red→green** (pre-fix, the "compose with AND" and "undefined no-op" cases fail). bb-kv-store suite: 121/121.

`check:api` clean (no public API change); changeset added.


soberm (approved):
Correctness verified against the mock: the mock `delete()` checks `ifExists` then `ifValueEquals` and requires **both** to pass, so composing the AWS `ConditionExpression` with `AND` is genuinely branch-for-branch parity (put uses `OR` because put's semantics are OR — the PR borrows put's `!== undefined` *detection*, not its operator; the code comment gets this right). The `undefined`-marshalling bug and the dropped-value-check bug are both real.

packages/bb-kv-store/src/index.aws.ts
```diff
@@ -180,13 +180,29 @@ export class KVStore<T = string> extends Scope {
...
+			deleteConditions.push('attribute_exists(#pk)');
+			names['#pk'] = 'pk';
+		}
+		if (conditions?.ifValueEquals !== undefined) {
+			deleteConditions.push('#value = :expected');
+			names['#value'] = 'value';
+			attrValues[':expected'] = JSON.stringify(conditions.ifValueEquals);
+		}
+		if (deleteConditions.length > 0) {
+			command.ConditionExpression = deleteConditions.join(' AND ');
```
soberm, line 203: [minor] The tests here intercept the DynamoDB middleware and assert the command *shape* — they're unit tests, not over-the-wire. Per AGENTS.md ("Mock tests can't catch AWS-path serialization bugs — serialization/behavior changes also need a sandbox e2e" / "New behavior … ships a test, sandbox e2e if serialization changed"), this is a behavior change to what gets sent to DynamoDB.

The `both` case is the one that matters: `attribute_exists(#pk) AND #value = :expected` is an expression the old code **never emitted**, so nothing has confirmed real DynamoDB accepts and evaluates it end-to-end (both-hold → deletes; value-mismatch → ConditionalCheckFailed → 409). A single sandbox e2e in `test-apps/comprehensive/test/` exercising delete with both conditions set would close the gap the shape-only test can't.

Not blocking — the shape assertion is a solid guard and the mock parity is sound — but worth adding given the rule.

Simone319 (approved):

PR #555, opened 2026-09-16
feat(auth-oidc): let stubIdp() declare test users inline

osama-rizk:
## Problem (Bug Bash: "bb-auth-oidc: stub IdP user directory only seedable via gitignored file")

The stub IdP's identity directory (used by the login screen and the `users` handed to `onAuthorize`) could only be seeded from a `users.json` file in the mock data dir — otherwise a single deterministic default user. So declaring specific test users for E2E/fixtures meant shipping a data file (typically gitignored), with no way to pin them in code.

## Fix

`stubIdp()` gains an optional **`users?: StubUser[]`**. When provided (and non-empty), inline users are the authoritative directory, taking precedence over `users.json` and the built-in default:

```ts
stubIdp({
  name: 'corporate',
  users: [
    { sub: 'u-1', email: 'alice@example.com', name: 'Alice' },
    { sub: 'u-2', email: 'bob@example.com', name: 'Bob' },
  ],
});
```

Threaded through `StubIdpOpts` → `StubProvider` → `stubUserDirectory` (a new highest-precedence source) at all three call sites: the `onAuthorize` directory, the login screen, and the form-submit user lookup. Precedence is **inline `users` → `users.json` (mock only) → default**. Omitting it is unchanged, so existing apps and the AWS runtime are unaffected.

## Tests
`index.test.ts` (stub IdP /authorize): inline users render on the login screen and **replace** the default user; `onAuthorize` receives the inline directory (picks a user by `login_hint` → 302). bb-auth-oidc suite: 140/140.

`npm run lint`, `check:api` (no report change — the new field is on the `stubIdp` opts) clean; changeset added (`bb-auth-oidc` minor).


packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -143,11 +143,15 @@ function defaultUser(providerName: string): StubUser {
...
- * back to the single deterministic default when absent or malformed, so the
- * AWS runtime and un-configured apps are unchanged.
+ * Precedence: inline `users` declared on the provider (in code) → `users.json`
+ * from `dataDir` (mock runtime only) → the single deterministic default. Falls
+ * back on absent/malformed input, so the AWS runtime and un-configured apps are
+ * unchanged.
  */
-function stubUserDirectory(providerName: string, dataDir?: string): StubUser[] {
+function stubUserDirectory(providerName: string, dataDir?: string, inlineUsers?: readonly StubUser[]): StubUser[] {
+	if (inlineUsers && inlineUsers.length > 0) {
```
soberm, line 152: [nit] Inline `users` are returned as-is, while the `users.json` path filters malformed entries. TypeScript guards the shape at the call site, so that's fine — but there's no de-duplication on `sub`. Two inline users sharing a `sub` would make `handleAuthorizeSubmit`'s `.find(u => u.sub === ...)` silently pick the first. Harmless for a test-only stub; a one-line JSDoc note on `StubIdpOpts.users` ("`sub` should be unique") could save a confusing debug session. Non-blocking.

soberm (approved):

packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -237,15 +241,15 @@ export async function handleAuthorize(provider: StubProvider, ctx: BlocksContext
...
+	ctx.response.send(renderLoginPage(provider.name, stubUserDirectory(provider.name, dataDir, provider.users), req, authorizeAction));
 }
 
 /** Login-screen form submission: resolve the picked user and issue the code. */
 export async function handleAuthorizeSubmit(provider: StubProvider, ctx: BlocksContext, dataDir?: string): Promise<void> {
 	const form = new URLSearchParams(await ctx.request.text());
 	const req = parseAuthorizeRequest(form, ctx);
 	if (!req) return;
-	const user = stubUserDirectory(provider.name, dataDir).find((u) => u.sub === form.get('sub'))
+	const user = stubUserDirectory(provider.name, dataDir, provider.users).find((u) => u.sub === form.get('sub'))
```
sharonyajain, line 252: **Submit handler falls back to the built-in default for a `sub` not in the inline directory**

With inline `users` declared authoritative (the login screen no longer offers the default), `handleAuthorizeSubmit` still resolves an unmatched `sub` to `defaultUser(provider.name)` via `?? defaultUser(...)`. The two new tests only drive `handleAuthorize` (login-screen render + the `onAuthorize` directory); the submit call site — one of the three newly-threaded sites — has no inline-users test.

```ts
const user = stubUserDirectory(provider.name, dataDir, provider.users).find((u) => u.sub === form.get('sub'))
	?? defaultUser(provider.name);
```

**Why it matters:** inline users declared as the only directory (cause) -> the submit handler mints a code for `corporate-user@stub.invalid` on any submitted `sub` outside the set, and no test guards this path (mechanism) -> a fixture typo in `sub`, or a test intending to prove 'only inline users authenticate', silently signs in as the default identity instead of failing loudly, masking the fixture bug (consequence). Test-only, but the failure is invisible.

**Suggestion:** When inline `users` are set, drop the `?? defaultUser` fallback (or reject the submit) so an unknown `sub` fails explicitly; add a `handleAuthorizeSubmit` test covering both a `sub` in the inline set and one that isn't.

packages/bb-auth-oidc/src/engines/stub-idp.ts
```diff
@@ -143,11 +143,15 @@ function defaultUser(providerName: string): StubUser {
 /**
  * The local identity directory for `onAuthorize` and the login screen.
  *
- * Reads `users.json` from `dataDir` (mock runtime only) when present; falls
- * back to the single deterministic default when absent or malformed, so the
- * AWS runtime and un-configured apps are unchanged.
+ * Precedence: inline `users` declared on the provider (in code) → `users.json`
+ * from `dataDir` (mock runtime only) → the single deterministic default. Falls
+ * back on absent/malformed input, so the AWS runtime and un-configured apps are
```
sharonyajain, line 148: **Inline users aren't de-duplicated on `sub`, so duplicates make later entries unreachable**

Unlike the `users.json` path (which filters malformed entries), inline users are returned as-is with no `sub`-uniqueness check, and `handleAuthorizeSubmit`'s `.find(u => u.sub === ...)` returns the first match. (A human reviewer already flagged this as a non-blocking nit.)

```ts
if (inlineUsers && inlineUsers.length > 0) {
	return [...inlineUsers];
}
```

**Why it matters:** two inline users share a `sub` (cause) -> the login screen renders both but the submit `.find` always resolves to the first (mechanism) -> the duplicate is silently unreachable, costing a confusing debug session for the test author (consequence). Test-only.

**Suggestion:** Adding a one-line JSDoc note on `StubIdpOpts.users` that `sub` must be unique, or de-duplicate / warn on duplicate `sub` in `stubUserDirectory`.

sharonyajain (approved):

sarayev (approved):

PR #558, opened 2026-09-16
feat(create-blocks-app): front-load framework model into scaffolded AGENTS.md

sharonyajain:
## Problem

The scaffolded `AGENTS.md` (`packages/create-blocks-app/resources/AGENTS.md`, copied into every project by `copySharedResources()`) points a coding agent at the block docs folder instead of stating the load-bearing framework facts inline. A first-time agent then has to re-derive the same wiring on every project — the client-vs-server module split, RPC method namespacing, how auth works, and how the deployed frontend finds the API — before it can build.

Benchmarking this showed the rediscovery is a large, repeated per-run cost: building the same app on Blocks took an AI agent 2× the turns of raw CDK at equal quality. Classifying every turn, the biggest drivers were reading Blocks docs (+5.8 turns/run vs CDK, which reads none) and reverse-engineering API wiring (~+4.6 turns/run) — together roughly half the gap. Each new section below targets one of those measured buckets.

**Issue #, if available:** Fixes https://github.com/aws-amplify/private-amplify-backend-staging/issues/1127

## Changes

Front-load the framework model into the scaffolded `AGENTS.md` so the common path needs one file, not six READMEs. Every statement was verified against the current template + package source.

- **Backend/frontend client model + "don't import `index.ts`" pitfall** — targets the module-boundary confusion seen in traces (agent importing the server module in a test and getting the wrong object). Backend defines `new ApiNamespace(scope, '<name>', …)`; the frontend imports the same export names from `'aws-blocks'`, which resolves to the generated client proxy (`aws-blocks/client.js`).
- **Method namespacing** — calls are `namespace.method(...)`; a bare method name fails (a trace showed a retry after that exact mistake).
- **"Auth is a Building Block"** — targets the auth-doc reading in the `doc_read` bucket. Traces showed agents re-learning three auth facts every run: gating a method with `await auth.requireAuth(context)`, mounting the ready-made login UI from `@aws-blocks/blocks/ui` (`Authenticator`/`onAuthChange`), and that `AuthBasic` without `codeDelivery` auto-confirms sign-up (load-bearing when a test can't complete an email round-trip). Stated inline so the agent reads none of the auth READMEs.
- **"The deployed frontend finds the backend on its own"** — targets the `config_probe` bucket (~+4.6 turns/run, the #2 driver). Every Blocks run spent turns curl-probing `config.json` / the RPC endpoint / the JS bundle to discover that the client reads its endpoint from the hosting origin and needs no hardcoded URL. Stated as one rule: the client resolves the endpoint itself (dev server serves config locally; deployed reads same-origin config) — don't hardcode or curl-hunt.
- **Storage bullet** in Rules naming `KVStore` and `DistributedTable`, pointing to each block's own `README.md`/`API.md` for the exact API and the docs `README.md` for the full catalog (no catalog duplicated here).
- Softened the "read block docs" rule to "the framework model above covers the common path; read a block's own docs for its specific API surface."

Source-verified: `ApiNamespace`/`Scope`/`AuthBasic` exports (`packages/blocks/src/index.ts`), `auth.createApi()`/`requireAuth` (`packages/bb-auth-basic/src/index.ts`), `Authenticator`/`onAuthChange` from `@aws-blocks/blocks/ui` (`packages/blocks/src/ui.ts`), the generated `ApiNamespaceClient` proxy, and the endpoint-resolution order in `packages/core/src/client/index.ts` (`resolveApiUrl`). The `npm run dev` vs `npm run dev:server` distinction reflects that a fresh scaffold ships both while an existing-project integration adds only `dev:server`.

## Validation

- Docs-only change to one maintained file (`cp`'d into scaffolds; not imported, compiled, or asserted by any test — confirmed no test reads its content). No Building Block behavior changes.
- Cold-read check: an agent given only this file (no other repo access, no prior Blocks knowledge) answered the wiring, auth, config-discovery, and storage questions from the doc alone, where the previous version left most of them requiring the block READMEs.
- Local package build/tests could not be run in the authoring environment (workspace `npm install` is registry-auth-gated there); CI runs the suite.

## Checklist

- [x] PR description included


Simone319 (dismissed):
Can you update the same content to the blocks SKILL in AWS MCP and [agent toolkit](https://github.com/aws/agent-toolkit-for-aws)?  So that when customers using the SKILL without the AGENT.md have the some context.

packages/create-blocks-app/resources/AGENTS.md
```diff
@@ -7,6 +7,60 @@
...
+  import { api, authApi } from 'aws-blocks';
+  const todos = await api.listTodos();          // typed, awaited call
+  ```
+
+  At runtime that import resolves to an auto-generated **client proxy** (`aws-blocks/client.js`), not the server module — the types come from your backend, the transport is injected. **The JSON-RPC transport is invisible: never build request payloads by hand or `fetch()` the API directly** (only for one-off connectivity troubleshooting). Just import the namespace and call the method.
+- **Pitfall:** do not `import ... from './aws-blocks/index.ts'` in a script/test to call the API — that gives you the *server* definition object, which behaves differently from the client. Import from `'aws-blocks'` (the package name) so you get the client, exactly as `src/` does.
+
+### Methods are namespaced
+
+A call is always `namespace.method(...)` — e.g. `api.createTodo(title)`, `authApi.signUp(...)`. The namespace is the first argument you passed to `new ApiNamespace(scope, '<name>', …)`. A bare method name with no namespace will not resolve.
```
Simone319, line 29: **Blocking (correctness):** the namespace is the **second** argument, not the first. `ApiNamespace`'s constructor is `constructor(scope, name, handler)` (`packages/core/src/api.ts:194`) — `scope` is first, the `'<name>'` string is second. The snippet on this very line (`new ApiNamespace(scope, '<name>', …)`) shows this, so the prose contradicts it.

Suggest: "The namespace is the string you passed as the **second** argument to `new ApiNamespace(scope, '<name>', …)`."

Simone319, line 29: **Blocking (correctness):** `authApi.signUp(...)` is not a valid client call. `auth.createApi()` returns `new ApiNamespace(this, 'auth', …)` exposing only `getAuthState` / `setAuthState` (`packages/bb-auth-basic/src/index.ts:409-465`); sign-up goes through `setAuthState({ action: 'signUp', … })`. A bare `signUp` method only exists on the deprecated `buildApi()`.

As written, `await authApi.signUp(...)` sends an RPC for an unregistered method → runtime `ApiError` — the exact 'method-not-found retry' this section is meant to prevent. Suggest a real example (e.g. `api.createTodo(title)`) or referencing the `Authenticator(authApi)` UI shown above.

packages/create-blocks-app/resources/AGENTS.md
```diff
@@ -7,6 +7,60 @@
...
+- The backend (`aws-blocks/index.ts`) declares each API with `new ApiNamespace(scope, '<name>', (context) => ({ ...methods }))`, and auth with `auth.createApi()`. You `export` them (e.g. `export const api = …`, `export const authApi = auth.createApi()`).
+- The frontend imports those **same export names** from `'aws-blocks'`:
+
+  ```ts
+  import { api, authApi } from 'aws-blocks';
+  const todos = await api.listTodos();          // typed, awaited call
+  ```
+
+  At runtime that import resolves to an auto-generated **client proxy** (`aws-blocks/client.js`), not the server module — the types come from your backend, the transport is injected. **The JSON-RPC transport is invisible: never build request payloads by hand or `fetch()` the API directly** (only for one-off connectivity troubleshooting). Just import the namespace and call the method.
+- **Pitfall:** do not `import ... from './aws-blocks/index.ts'` in a script/test to call the API — that gives you the *server* definition object, which behaves differently from the client. Import from `'aws-blocks'` (the package name) so you get the client, exactly as `src/` does.
```
Simone319, line 25: **Nit (accuracy):** the direct-import path is off by one level for the script/test context this bullet names. In a scaffold, `src/` and `test/` are siblings of `aws-blocks/`, so from a script/test the path is `../aws-blocks/index.ts` — `./aws-blocks/index.ts` only resolves from the project root.

packages/create-blocks-app/resources/AGENTS.md
```diff
@@ -7,6 +7,60 @@
...
+
+- The backend (`aws-blocks/index.ts`) declares each API with `new ApiNamespace(scope, '<name>', (context) => ({ ...methods }))`, and auth with `auth.createApi()`. You `export` them (e.g. `export const api = …`, `export const authApi = auth.createApi()`).
+- The frontend imports those **same export names** from `'aws-blocks'`:
+
+  ```ts
+  import { api, authApi } from 'aws-blocks';
+  const todos = await api.listTodos();          // typed, awaited call
+  ```
+
+  At runtime that import resolves to an auto-generated **client proxy** (`aws-blocks/client.js`), not the server module — the types come from your backend, the transport is injected. **The JSON-RPC transport is invisible: never build request payloads by hand or `fetch()` the API directly** (only for one-off connectivity troubleshooting). Just import the namespace and call the method.
```
Simone319, line 24: **Nit (duplication):** this 'JSON-RPC transport is invisible / don't `fetch()` the API directly' guidance is also stated in the pre-existing Rules bullet (line 76). Consider stating it once so the two copies don't drift on future edits.

Simone319 (commented):
Reviewed the scaffolded `AGENTS.md` change against source at HEAD. The framework model is mostly accurate and well-verified (import/client model → `aws-blocks/client.js`, `Authenticator`/`onAuthChange` from `@aws-blocks/blocks/ui`, `requireAuth`, auto-confirm without `codeDelivery`, `/.blocks-sandbox/config.json` discovery + deployed same-origin, `KVStore`/`DistributedTable` — all check out).

Two **correctness** issues in the *Methods are namespaced* section (inline) should be fixed before merge — both sit in the section whose stated purpose is preventing exactly this kind of wiring mistake.

Minor nit not worth an inline: in fresh scaffolds `npm run dev` and `npm run dev:server` are identical aliases (`tsx watch aws-blocks/scripts/server.ts`), so the `dev` (fresh) vs `dev:server` (existing project) framing is about which script name ships, not a behavioral difference — a reader may infer `dev` does more than it does.

packages/create-blocks-app/resources/AGENTS.md
sharonyajain: Fixed in ab25a4a — reworded to 'the string you passed as the **second** argument to new ApiNamespace(scope, '\''<name>'\'', …)'. Confirmed against constructor(scope, name, handler) at packages/core/src/api.ts:194.

sharonyajain: Fixed in ab25a4a — replaced with api.createTodo(title) / api.listTodos(), and added a note that sign-up is authApi.setAuthState({ action: 'signUp', … }). Confirmed createApi() exposes only getAuthState/setAuthState (packages/bb-auth-basic/src/index.ts:409-465); bare signUp only exists on the deprecated buildApi().

packages/create-blocks-app/resources/AGENTS.md
sharonyajain: Fixed in ab25a4a — now ../aws-blocks/index.ts (script/test are siblings of aws-blocks/).

packages/create-blocks-app/resources/AGENTS.md
sharonyajain: Fixed in ab25a4a — removed the duplicate Rules bullet; the richer statement earlier in the doc is now the single source.

sharonyajain:
Thanks for the careful review, @Simone319 

**Blocking (both in *Methods are namespaced*, verified against source):**
- **Namespace arg order** — corrected to "the string you passed as the **second** argument to `new ApiNamespace(scope, '<name>', …)`". Confirmed `constructor(scope, name, handler)` at `packages/core/src/api.ts:194`.
- **`authApi.signUp(...)`** — removed; the example is now `api.createTodo(title)` / `api.listTodos()`, with a note that sign-up is `authApi.setAuthState({ action: 'signUp', … })`. Confirmed `createApi()` exposes only `getAuthState`/`setAuthState` (`packages/bb-auth-basic/src/index.ts:409-465`); the bare `signUp` client method only exists on the `@deprecated buildApi()`.

**Nits:**
- **Import path** — corrected to `../aws-blocks/index.ts` (script/test are siblings of `aws-blocks/`).
- **Duplication** — dropped the redundant "JSON-RPC transport is invisible" Rules bullet; the richer statement earlier in the doc is now the single source.

packages/create-blocks-app/resources/AGENTS.md
```diff
@@ -7,6 +7,60 @@
...
+import { Authenticator, onAuthChange } from '@aws-blocks/blocks/ui';
+authContainer.appendChild(Authenticator(authApi));
+onAuthChange(authApi, (user) => { /* re-render for signed-in/out */ });
+```
+
+`AuthBasic` without a `codeDelivery` config signs a user in immediately on sign-up (no email confirmation step).
+
+### The deployed frontend finds the backend on its own
+
+You do **not** wire an API URL into the frontend. The imported client resolves the API endpoint itself at runtime: locally, the dev server serves the backend and its config together (`npm run dev` in a fresh app, or `npm run dev:server` when Blocks was added to an existing project; it exposes `/.blocks-sandbox/config.json`); once deployed, the client reads the config served from the hosting origin, where the frontend and API sit same-origin. So build against the imported client and let the framework resolve the URL — don't hardcode one or curl-hunt for the API base.
```
soberm, line 45: [nit] The `npm run dev` (fresh) vs `npm run dev:server` (existing project) framing reads as a behavioral difference, but in the scaffolded `package.json`s the two are usually identical aliases (`tsx watch aws-blocks/scripts/server.ts`) — the distinction is which script name ships, not what it does. One template (`backend`) is the exception, where `dev` is `concurrently "npm:dev:server" "npm:dev:client"`. Consider rewording to something like "whichever dev script your app ships (`dev` or `dev:server`)" so a reader doesn't infer `dev` always does more. Doesn't affect the config-discovery point either way.

soberm (approved):
Docs-only front-loading of the framework model into the scaffolded AGENTS.md. Verified every load-bearing claim against canonical source at HEAD — all check out:

- `new ApiNamespace(scope, '<name>', handler)` → namespace **is** the second arg (`packages/core/src/api.ts:194`)
- `authApi.setAuthState({ action: 'signUp', … })` + `getAuthState` (`packages/bb-auth-basic/src/index.ts:411,423`)
- `AuthBasic` without `codeDelivery` → immediate confirm (`index.ts:247`)
- `Authenticator`/`onAuthChange` from `@aws-blocks/blocks/ui` (`packages/blocks/src/ui.ts:5`)
- `resolveApiUrl` + `/.blocks-sandbox/config.json` discovery (`packages/core/src/client/index.ts`)
- `ApiNamespaceClient` client proxy naming

The two correctness issues from the earlier review (namespace-arg order, bare `signUp`) are resolved in `ab25a4a`. Changeset present and correctly scoped (`@aws-blocks/create-blocks-app` patch). All CI green. One nit inline; not blocking — approve.

Simone319 (approved):

PR #559, opened 2026-09-16
feat(core): emit a machine-readable BLOCKS_DEPLOYED signal on deploy success

sharonyajain:
## Problem

`npm run deploy` streams CloudFormation events and ends with human-readable lines (`✅ Deployment complete!`, `📡 API URL`, `🌐 Frontend URL`). A programmatic caller — a coding agent, a CI step, a script — has no single stable line to detect "deploy finished, and here is the URL". So it polls the stack or greps the streamed output for a URL, which is slow and brittle.

This showed up as a real cost in agent benchmarking: agents spent multiple turns per run sleeping and probing (`describe-stacks`, curling `config.json`, grepping the bundle) just to learn the deploy was done and where the app lives.

**Issue #, if available:**

## Changes

Emit one machine-readable completion line as the LAST line of a successful `npm run deploy`:

```
BLOCKS_DEPLOYED url=<frontend> api=<backend>
```

`api=` is always present; `url=` is omitted for a backend-only app (no hosting). A caller greps that one line instead of parsing CloudFormation output or polling the stack. The existing `✅`/`📡`/`🌐` lines are unchanged — the signal is purely additive, printed only on the success path after `config.json` is written and the URLs are confirmed.

The formatting is a pure, exported `formatDeploySignal(apiUrl, hostingUrl?)` helper in `packages/core/src/scripts/deploy.ts`, called from `deploy()`, with unit coverage in `deploy.test.ts` (both branches + the stable-prefix contract). The scaffolded `AGENTS.md` documents the line so a coding agent greps it rather than polling.

## Validation

- `deploy.test.ts` covers `formatDeploySignal` for the hosting and backend-only branches and asserts the `BLOCKS_DEPLOYED ` prefix.
- No behavior change to the deploy itself: the signal is an added `console.log` on the existing success path; the human-readable lines are untouched. No test asserted the removed/old output (checked).
- The full workspace build/tests were not run in the authoring environment (`npm install` is registry-auth-gated there); CI runs the suite.

## Checklist

- [x] PR description included


svidgen (commented):
It's not clear to my why this is an issue or why a final line like this improve the situation. Wouldn't a AI/machine runner looking for exit 0 and then grep/filter for the two lines it needs? (Or maybe look for an output manifest.)

PR #560, opened 2026-09-16
ci(agent-bench): block PRs on failure/regression and comment the report

sharonyajain:
## Problem

The `PR Agent Bench` check is green even when cells fail. Every fallible cell step is `continue-on-error`, so a failed build, a timed-out agent, or an OIDC/infra failure still reports a green check. The only place a failure shows is the Actions run summary, which a reviewer has to open on purpose. There is no PR comment and nothing blocks the merge. A reviewer approving the PR cannot see whether the agent got better or worse against `main`.

## Changes

Three changes, all opt-in so an unconfigured repo keeps today's observational behaviour.

Add a merge gate. `scripts/agent-bench/steps/lib/gates.mjs` turns the run's headline numbers into a pass/fail verdict with three independent gates, each enabled only when its repo variable is set: an absolute floor (`BENCH_MIN_SCORE` — mean composite must clear it), a regression gate (`BENCH_MAX_REGRESSION` — the mean must not drop more than N points vs the `main` baseline; skipped on the first run when there is no baseline), and a harness gate (`BENCH_MAX_HARNESS_ERRORS` — at most N cells may be infra failures). The harness gate is what makes a broken measurement red the PR, kept separate from a low agent score, which is the signal the floor/regression gates read. `summary.mjs` calls the gate and exits 1 when any enabled gate fails.

Post the report as a PR comment. `summary.mjs` writes the same rendered report it puts in the Actions summary to a file, and a new step upserts it as one sticky PR comment keyed by a hidden marker (`scripts/agent-bench/steps/lib/pr-comment.sh`). The `summary` job gains `pull-requests: write` for this. The comment carries a Merge verdict section listing each gate's decision, and it posts even on a blocking run so the red check has an explanation on the PR.

Nothing changes until the repo variables are set — the gate module treats an unset/blank/non-numeric threshold as disabled, and the comment step only runs on a same-repo pull request.

## Validation

- Added `scripts/agent-bench/steps/lib/gates.test.mjs` (15 cases): each gate's disabled / pass / fail / skip branches, threshold parsing, the no-scored-cells skip, the no-baseline skip, a 0-point regression limit, and gate independence. Runs under the harness's `test:ci` (`node --test steps/lib/*.test.mjs`).
- Ran `summary.mjs` against fixtures: a run below the floor and 53 points under baseline exits 1 and writes a comment that says `BLOCKS the PR` with both reasons; an improved run exits 0 and says it does not block.
- `bash -n` on `pr-comment.sh`.
- Two pre-existing harness tests fail locally only (a `bedrock-retry` timing test and the issue-184 `benchagent` isolation test that needs a Linux `useradd`); both are in files this PR does not touch and pass on CI's Linux runner.

## Checklist

- [x] PR description included


sarayev (dismissed):

scripts/agent-bench/steps/summary.mjs
```diff
@@ -7,9 +7,29 @@ import { appendFileSync, readdirSync, readFileSync, writeFileSync } from 'node:f
...
 import { buildAggregate, cellComposite, deltaBall, diffAgainstBaseline, renderDetailed, renderOverview } from './lib/overview.mjs';
+import { evaluateGates } from './lib/gates.mjs';
 
 const RESULTS_DIR = process.env.RESULTS_DIR ?? 'results';
 
+// FINDING-4 breadcrumb: write a minimal fallback comment NOW, before the risky rendering below. If
+// summary.mjs throws mid-render, the always() comment step still finds a non-empty file and posts a
+// "failed to render" stub instead of silently leaving a red check with no explanation on the PR. The
+// full report overwrites this on the normal path.
+if (process.env.COMMENT_MD_PATH) {
```
soberm: [minor] This early fallback comment is written whenever `COMMENT_MD_PATH` is set, but the `Comment bench report on PR` step only runs `if: always() && github.event_name == 'pull_request'`. On a push-to-main run nothing consumes this file, so if `COMMENT_MD_PATH` is exported on non-PR events the fallback write is dead (harmless — a workspace temp file — just never read). Worth confirming `COMMENT_MD_PATH` is only exported on PR events, or noting the write is intentionally unconditional.

scripts/agent-bench/steps/lib/gates.mjs
```diff
@@ -0,0 +1,125 @@
...
+ * which is sensible below 0 — and a negative count would INVERT the harness gate (harnessErrors 0 > -1
+ * → a clean run fails). So a fat-fingered `-1` disables the gate rather than silently inverting it.
+ * @param {string|undefined|null} raw
+ * @returns {{enabled: boolean, value: number}}
+ */
+export function parseThreshold(raw) {
+	const s = (raw ?? '').trim();
+	if (s === '') return { enabled: false, value: NaN };
+	const n = Number(s);
+	return Number.isFinite(n) && n >= 0 ? { enabled: true, value: n } : { enabled: false, value: NaN };
```
soberm: [nit] `parseThreshold` accepts fractional values, so `BENCH_MAX_HARNESS_ERRORS=1.5` parses as enabled `value: 1.5`. Harmless for a count compared with `>`, but a count gate reading a fraction is slightly odd. Optional: an `Number.isInteger` guard for the harness (count) gate only — the floor/regression gates legitimately want fractions.

scripts/agent-bench/steps/lib/pr-comment.sh
```diff
@@ -0,0 +1,77 @@
...
+	if [ "${DUP_COUNT:-0}" -gt 1 ]; then
+		echo "pr-comment: ${DUP_COUNT} marker comments found — deleting $((DUP_COUNT - 1)) duplicate(s), keeping ${EXISTING_ID}" >&2
+		printf '%s\n' "$MARKER_IDS" | tail -n +2 | while read -r dup_id; do
+			gh api "repos/${REPO}/issues/comments/${dup_id}" -X DELETE >/dev/null 2>&1 \
+				&& echo "pr-comment: deleted duplicate ${dup_id}" >&2 \
+				|| echo "pr-comment: could not delete duplicate ${dup_id} (left for a human)" >&2
+		done
+	fi
+else
+	echo "pr-comment: creating new comment" >&2
```
soberm, line 73: [question] The duplicate-comment self-heal only runs in the `EXISTING_ID` (update) branch. If a race left two marker comments AND the list API then transiently failed on a prior run, this create path posts a third without reconciling. The API-failure guard above already bails rather than creating, so probability is low — is the reconcile deliberately update-only, or worth also running when `DUP_COUNT > 1` here?

soberm (commented):
[nit] The PR checklist leaves "Relevant documentation is changed or added" unchecked, but `README.md` got a solid 35-line update covering all three gates — just check the box.

scripts/agent-bench/steps/summary.mjs
sharonyajain: Fixed in eaa90a9. Thank you for pointing it out. `COMMENT_MD_PATH` was exported unconditionally. The workflow now sets it only on `pull_request` (`github.event_name == 'pull_request' && format(...) || ''`), so on push-to-main it's empty; `summary.mjs`'s `if (process.env.COMMENT_MD_PATH)` guard treats the empty string as unset and skips the write. No dead temp file on non-PR runs.

scripts/agent-bench/steps/lib/gates.mjs
sharonyajain: Leaving as-is. A fractional `BENCH_MAX_HARNESS_ERRORS=1.5` is compared with `>` against an integer count, so it behaves identically to `1` — no wrong verdict, just an odd-looking value the operator typed. An `Number.isInteger` guard would add a count/float branch to `parseThreshold` (which floor/regression legitimately share) to reject an input that can't misbehave; not worth the extra branch for a value only a human sets by hand. The README documents it as a count.

scripts/agent-bench/steps/lib/pr-comment.sh
sharonyajain: Deliberately update-only. The create branch only runs when the list returned zero marker comments, so by construction there is nothing to reconcile there. The scenario you describe (two markers exist, but a transient list failure hid them) can't reach create: the API-failure guard above `exit 0`s rather than falling through to create, so a failed list never posts a third. The only way to reach create with duplicates already present is a successful list that returned zero of them which is not possible. So the reconcile lives where duplicates are observable (the update branch), and the next successful run's sorted self-heal collapses any pair that a genuine race created.

scripts/agent-bench/steps/lib/gates.mjs
sharonyajain: Update: added the guard after all (ec929c0). `parseThreshold` now takes an optional `{integer:true}`, passed only by the harness count gate — a fractional `BENCH_MAX_HARNESS_ERRORS` is treated as unset, while floor/regression still accept fractions. Covered by 3 new tests; README notes the count must be whole.

soberm (approved):

sarayev (approved):

PR #564, opened 2026-09-17
feat(core): managed CloudFront API front door

Simone319:
## What

Production Blocks stacks get a single managed CloudFront distribution that presents the API on one stable domain. It is a single catch-all default behavior forwarding to the default compute's API Gateway origin (every API path resolves to that one compute today, so no per-path fan-out) — uncached, all methods, forwards everything except `Host`, HTTP→HTTPS.

- On in the production preset, off in sandbox (which reaches API Gateway directly). Override per app with `apiFrontDoor?: 'cloudfront' | 'none'`.
- Suppressed automatically when a same-stack `Hosting` distribution already fronts the API, so a CloudFront-hosted frontend doesn't get a second distribution.
- `ApiUrl` output resolves (lazily, at synth) to whichever front door fronts the API — the managed one, Hosting's, or API Gateway as a fallback — composed from `defaultEndpoint + /aws-blocks/api`.
- `Hosting` adds the API behaviors (reserved RPC + auth subtrees, plus one per app `RawRoute` outside those prefixes) to its own distribution.
- `LambdaCompute` exposes its origin base as a public `endpoint`.

## Breaking changes (repo is 0.x preview)

- `BlocksDefaults` gains a required `provisionApiFrontDoor: boolean`. Apps that spread a preset need no change.
- `Hosting`'s `api` prop takes `BlocksApiRouting` (`{ defaultEndpoint }`) instead of `BlocksStackApi` (`{ apiUrl }`).

## Test plan

Verified locally on Node 22:

- `npm run build` — pass
- `npm run lint` — pass (0 errors)
- `npm run lint:deps` — pass
- `npm test` — 4215 tests, 0 fail
- `npm run test:e2e:local` — 396 + 8 tests, 0 fail

Reviewed by the AIM multi-dimension code-review agent and the built-in `/code-review`; actioned findings (doc-accuracy corrections for the single-behavior design, `https://` scheme contract on `Compute.endpoint`, and the missing `@aws-blocks/bb-lambda-compute` changeset entry).

Draft to let CI run the full E2E suite.


packages/core/src/cdk/api-front-door.ts
```diff
@@ -0,0 +1,218 @@
...
+
+		// No default compute endpoint means no fallback origin, so there is nothing
+		// coherent to put behind the default behavior.
+		if (!this.defaultEndpoint) return;
+
+		const origin = httpOriginFromEndpoint(this.defaultEndpoint);
+		// Under the owner, not the stack: a `BlocksBackend` is a construct inside
+		// someone else's stack, and two of them in one stack would otherwise collide
+		// on this construct id.
+		const distribution = new Distribution(this.owner, FRONT_DOOR_ID, {
```
osama-rizk, line 189: Creating a whole `Distribution` (+ `CfnOutput`) *inside* a CDK Aspect's `visit()` is a fragile pattern. Aspects are designed to mutate/annotate existing constructs, not add new ones — adding constructs mid-traversal is version-sensitive (newer CDK can warn or throw "cannot add nodes during synthesis"), and a construct added this late may miss other aspects/validation passes. It works on the pinned CDK (build + e2e pass), so this is a forward-compat/maintenance risk rather than a live bug.

The deferral *reason* is sound (you need to run after `Hosting` is constructed, which is after `create()` returns, so the repo's `finalize*` registries — which run inside `create()` — are too early). It's the *mechanism* that's fragile. Worth a comment pinning the CDK-version assumption, and a note on whether a stack-synthesis hook or a `Lazy`-driven approach was considered instead of construct-creation-in-aspect.

packages/core/src/hosting.ts
```diff
@@ -850,62 +867,80 @@ export class Hosting extends Construct {
...
+    // Keyed on this construct's stack. For the usual `new Hosting(blocksStack, …)`
+    // that is the backend's stack, which is what the aspect reads.
+    //
+    // A Hosting in a *different* stack than its backend cannot suppress that
+    // backend's managed front door — the claim lands on this stack, and the aspect
+    // only reads its own. The app ends up paying for two distributions, with
+    // `ApiUrl` pointing at the managed one rather than this one. Pass
+    // `apiFrontDoor: 'none'` on the backend to make this distribution the only
+    // front door.
+    registerHostingDistribution(cdk.Stack.of(this), hosting.distributionUrl);
```
osama-rizk, line 894: The cross-stack limitation documented here (a `Hosting` in a different stack than its backend can't suppress the backend's managed front door → two distributions, `ApiUrl` points at the managed one) is honestly called out, so this is a confirm-the-tradeoff note rather than a defect. Since it's silent at deploy time (the app just pays for two distributions), consider emitting a synth-time `console.warn` when a Hosting claim lands on a different stack than the backend it fronts — a code comment won't reach the app author, but a warning will.

packages/core/src/cdk/api-front-door.ts
```diff
@@ -0,0 +1,218 @@
...
+ * plain-HTTP client recovers instead of failing.
+ *
+ * Exported so `Hosting` can apply the same settings when it fronts the API, and
+ * the two paths cannot drift.
+ */
+export const API_BEHAVIOR_OPTIONS = {
+	allowedMethods: AllowedMethods.ALLOW_ALL,
+	cachePolicy: CachePolicy.CACHING_DISABLED,
+	originRequestPolicy: OriginRequestPolicy.ALL_VIEWER_EXCEPT_HOST_HEADER,
+	viewerProtocolPolicy: ViewerProtocolPolicy.REDIRECT_TO_HTTPS,
```
osama-rizk, line 60: Minor/edge: `REDIRECT_TO_HTTPS` on an `ALLOW_ALL`-methods behavior means a plain-HTTP RPC `POST` gets a 301/302 and is re-issued by the browser as a GET with the body dropped. Only bites a client that initiates RPC over plain HTTP (SDKs use HTTPS), so it's marginal — but the docstring's "a plain-HTTP client recovers instead of failing" is true for GET raw routes and *not* for a POST RPC. Fine to keep the redirect (good UX for GETs); just don't over-claim recovery for the POST path.

osama-rizk (commented):
Well-built. I verified the load-bearing pieces: `ApiUrl` is genuinely `Lazy` (resolves after the aspect/Hosting run, not eagerly), the Hosting suppression ordering is correct (Hosting calls `registerHostingDistribution` at construction → the synth-time aspect reads the claim), `Compute.endpoint` is `https://…`, `API_BEHAVIOR_OPTIONS` is shared so the managed and Hosting paths can't drift, and `resolveApiFrontDoor` is the same two-tier `option ?? defaults` posture model as #302/#385. The `ALL_VIEWER_EXCEPT_HOST_HEADER` + `CACHING_DISABLED` posture is right for an API origin. Three findings, LOW/MED.

osama-rizk:
**Design posture — is the always-on decision (single-compute prod also gets a managed front door) settled?**

Reading this against the Multi-Compute Front Door one-pager (the design of record): the doc's stated position is **on-demand** — Scenario C says a single-compute app already has one origin and gets **no front door**, and §2.6 explicitly flags always-on (Track D) vs on-demand as an **open, unreconciled tradeoff** that "must be reconciled before Track D's `apiFrontDoor` work lands."

This PR takes the **always-on** side: it provisions the managed distribution for every prod stack, including single-compute ("every API path resolves to that one compute today"). That may well be the right call — origin stability across a later single→multi transition, avoiding the `Secure`-cookie loss / re-auth when the front door first appears — but right now it's decided implicitly by the preset default rather than as the explicit resolution of §2.6.

Two things worth making explicit here:
1. **State that §2.6 is resolved in favor of always-on, and why** (origin stability > the deploy tax on the single-compute majority), so the decision is on the record rather than a side effect of the default.
2. The cost side is real: every prod app — including the single-compute common case — now eats minutes of CloudFront propagation on deploy and an always-on distribution. If that's accepted, it just wants to be a conscious, documented choice.

Not blocking — a "make the decision explicit" ask, since the design doc names this as the one thing that must be reconciled before this lands.

packages/core/src/cdk/api-front-door.ts
```diff
@@ -0,0 +1,218 @@
...
+ * `endpoint` is `https://{host}/{stage}`, and CloudFront wants those two parts
+ * separately: the domain on the origin, the stage as `originPath` so it is
+ * prefixed onto every forwarded request. It is a CloudFormation token, so the
+ * split has to happen in the template via `Fn::Split` rather than in JS — string
+ * methods here would operate on the unresolved placeholder text.
+ */
+export function httpOriginFromEndpoint(endpoint: string): IOrigin {
+	const withoutScheme = cdk.Fn.select(1, cdk.Fn.split('https://', endpoint));
+	const hostname = cdk.Fn.select(0, cdk.Fn.split('/', withoutScheme));
+	const stage = cdk.Fn.select(1, cdk.Fn.split('/', withoutScheme));
```
osama-rizk, line 100: **[suggestion] `httpOriginFromEndpoint` deploy-fails on a path-less endpoint — latent, and load-bearing for the BYOC future this design targets.**

`Fn::Select(1, Fn::Split('/', withoutScheme))` assumes the endpoint always has a `/{stage}` segment. It does for `LambdaCompute` today, but a `Compute.endpoint` of `https://api.example.com` (a custom-domain origin, or a future container/BYOC HTTPS origin — exactly what the multi-compute one-pager says this helper must front) splits to `['api.example.com']`, so `Fn::Select(1, …)` is out of range → deploy fails with `Fn::Select cannot select nonexistent value`.

Safe now, but `Compute.endpoint`'s own docstring invites non-Lambda origins. Suggest either guarding the no-path case (empty `originPath`) or enforcing a "must carry a path segment" invariant on `Compute.endpoint`. (Pre-existing assumption lifted out of the old `addApiBehaviors`, now shared by the managed path.)

packages/core/src/cdk/api-front-door.ts
```diff
@@ -0,0 +1,218 @@
...
+	visit(_node: IConstruct): void {
+		if (this.done) return;
+		this.done = true;
+
+		const stack = cdk.Stack.of(this.owner);
+		const state = stateOf(stack);
+
+		// `Hosting` fronts the API itself, and has already published its origin.
+		// Provisioning a second distribution would double the hops and split the domain.
+		if (state.claimedByHosting) return;
```
osama-rizk, line 174: **[nit] An explicit `apiFrontDoor: 'cloudfront'` is silently ignored when a same-stack `Hosting` claims the role.**

This `if (state.claimedByHosting) return;` runs *before* the `if (!this.provision) return;` check below, so an app that sets `apiFrontDoor: 'cloudfront'` **and** has a same-stack `Hosting` builds no managed distribution — the API rides Hosting's. That's the intended outcome (one CloudFront still fronts the API), but the explicit opt-in is silently downgraded with no signal. A one-line doc note ("a present Hosting wins even over an explicit `'cloudfront'`") or a debug log would save a confused "why is there no `ApiFrontDoorUrl` output?" later.

packages/core/src/cdk/blocks-backend.ts
```diff
@@ -231,6 +242,22 @@ export class BlocksBackend extends Construct {
...
+	 * `/aws-blocks/api` suffix and no trailing slash.
+	 *
+	 * This is the origin the API front door forwards to, and the base a
+	 * client-facing RPC URL is composed from (`defaultEndpoint + BLOCKS_RPC_PREFIX`).
+	 */
+	get defaultEndpoint(): string {
+		const endpoint = this.requireDefaultCompute().endpoint;
+		if (!endpoint) {
+			throw new Error(
+				'Default compute has no HTTP endpoint — a worker-only compute cannot serve as the default.',
```
osama-rizk, line 256: **[nit/edge] `ApiUrl` surfaces this throw from an oblique location for a worker-only default compute.**

When no front door is provisioned/claimed, the `ApiUrl` Lazy falls back to `stack.defaultEndpoint`, which invokes this getter — and for a worker-only default compute (no HTTP endpoint) it throws here. Reachable only in the edge case of a worker-only default compute *and* no front door. Arguably correct (no HTTP compute ⇒ no API URL), but the error surfaces while producing the `ApiUrl` output rather than at a clear up-front validation point, so the message lands far from the cause. Consider validating this at backend setup so the failure points at the misconfiguration directly.

packages/core/src/cdk/api-front-door.ts
Simone319: Good question — the "create a construct from inside an aspect's `visit()`" shape is deliberate, and I've added a docstring on `ApiFrontDoorAspect` explaining why (a3063f6c):

- The distribution has to be built **after** `Hosting` is constructed, so a Hosting front-door claim can suppress it (`registerHostingDistribution`). Hosting is usually built *after* `create()` returns — which is after the in-`create()` `finalize*` registries have already run, so a finalize hook is too early.
- `Lazy` was considered but only defers *value* resolution, not construct *creation*.
- A stack-synthesis hook carries the same "add a node late" caveat this does.

So `visit()` is the least-bad hook. I've also committed to the CDK-version assumption in the docstring: aws-cdk-lib as pinned permits adding a construct from `visit()`; if a newer CDK warns/throws "cannot add nodes during synthesis", the fallback is an explicit post-`create()` finalize step the app invokes, preserving the Hosting-claim ordering. Build + e2e cover the current pin.

packages/core/src/hosting.ts
Simone319: Agreed a warning here would be worth it — a Hosting distribution in a *different* stack than the backend is a real footgun, since the claim silently wouldn't apply.

The catch is that the warn can't be emitted reliably at this claim site: `registerHostingDistribution` receives `cdk.Stack.of(this)` (the Hosting side) and a `distributionUrl` string, but the backend it's claiming for is reachable only through `BlocksApiRouting`, which is a structural interface (`{ readonly defaultEndpoint: string }`) — not a construct — so there's no backend stack to compare against here. A trustworthy cross-stack check needs the backend's stack identity threaded to this point, which is exactly the cross-stack Hosting↔backend follow-up already tracked (front-door / CORS / RawRoute ownership).

So leaving this as-is for now and folding the warn into that follow-up rather than adding a check that can't actually see both stacks. Not changing code in this PR.

packages/core/src/cdk/api-front-door.ts
Simone319: Right — the old wording over-claimed that "a plain-HTTP client recovers instead of failing," which isn't true for an RPC `POST`. Fixed the docstring (a3063f6c) to scope recovery to GETs:

> HTTPS-only via redirect rather than block: a plain-HTTP GET (a raw route) is recovered — the browser follows the 301. A plain-HTTP RPC `POST` is NOT recovered: the redirect re-issues it as a GET with the body dropped. That path is marginal (SDK clients always use HTTPS), so the redirect is kept for the GET UX rather than blocking plain-HTTP outright.

PR #565, opened 2026-09-17
feat(native): add version constants and CI drift checks for Swift and Dart

VarshithaPamisetty:
## Problem

The Swift and Dart libraries don't expose their own version in code, so it can't be read at runtime or reported for usage tracking.

## Fix

A new script generates a committed version constant for each library, with a `--check` mode CI runs to catch a stale value:

```
node scripts/generate-native-versions.mjs --swift --check
```

Swift `Version.swift` and Dart `version.dart` each expose `blocksRuntimeVersion` and `blocksUserAgentToken` (e.g. `aws-blocks-swift/0.1.1`), generated from the package's version. The constant is committed because Swift/Dart consumers build our source without running our scripts.

Note:
1. This PR covers Swift and Dart. Kotlin already syncs its version into Gradle (a build property) via `native/kotlin/scripts/sync-version.sh`; the runtime constant for Kotlin will follow in a separate PR.
2. `blocksUserAgentToken` is defined but not yet sent on any request. Wiring it into the outgoing User-Agent will come in a separate PR.

## Tests

- **Swift:** `BlocksRuntimeTests` 88/88 (5 new), swiftformat/swiftlint clean.
- **Dart:** `version_test` 5/5, format clean, analyze no new issues.
- **Generator:** `generate-native-versions.test` 14/14 (drift, arg handling, mistyped-flag, Swift header sourcing, CRLF tolerance, malformed version, +build rejection); a CI job runs the `--check` guard per language.

## Checklist

- [x] PR description included
- [x] Tests are changed or added
- [x] Documentation impact assessed; no change needed (no public API change yet)

PR #567, opened 2026-09-18
feat(native): add version constant and CI drift check for Kotlin

VarshithaPamisetty:
## Problem

The native Kotlin library syncs its version into Gradle (a build property) but doesn't expose it in code, so it can't be read at runtime or reported for usage tracking.

## Fix

A `generateVersion` Gradle task generates the runtime version constant at build time, with a `--check` mode CI runs to catch a stale value:

```
native/kotlin/scripts/sync-version.sh --check
```

The generated `Version.kt` exposes `blocksRuntimeVersion` and `blocksUserAgentToken` (e.g. `aws-blocks-kotlin/0.2.0`), generated from `gradle.properties` `VERSION_NAME`. The constant is generated at build time (not committed) because Kotlin ships a compiled artifact, unlike Swift/Dart which commit theirs.

Note:
1. This is the Kotlin follow-up to #565 (Swift and Dart).
2. `blocksUserAgentToken` is defined but not yet sent on any request. Wiring it into the outgoing User-Agent will come in a separate PR.

## Tests

- **Kotlin:** `VersionTest` 5/5 (token format/prefix/embeds-version, semver, not-placeholder), verified on JVM, Android host, and iOS simulator.
- **Script:** `sync-version.test.sh` 4/4 (added `--check` pass + drift cases, each asserting the file is unmodified).
- CI: a `version-sync` job runs the `--check` drift guard; a `build-and-test` job runs `:runtime:jvmTest`.

## Checklist

- [x] PR description included
- [x] Documentation impact assessed; no change needed (no public API change yet)


native/kotlin/runtime/build.gradle.kts
```diff
@@ -56,3 +56,37 @@ kotlin {
...
+        pkgDir.mkdirs()
+        pkgDir.resolve("Version.kt").writeText(
+            """
+            |package com.aws.blocks.kotlin
+            |
+            |/** Semver of this AWS Blocks Kotlin runtime library. */
+            |internal const val blocksRuntimeVersion = "$versionValue"
+            |
+            |/** User agent token for this runtime, e.g. `aws-blocks-kotlin/$versionValue`. */
+            |internal const val blocksUserAgentToken = "aws-blocks-kotlin/${'$'}blocksRuntimeVersion"
```
cadivus, line 84: Attach the generated user-agent token to outbound requests

The change generates and tests a user-agent token, but no runtime request path consumes it. BlocksClient still sends only Content-Type and the body, so released Kotlin clients never transmit the new SDK/version identifier.

native/kotlin/runtime/build.gradle.kts
```diff
@@ -56,3 +56,37 @@ kotlin {
...
+val versionOutputDir = layout.buildDirectory.dir("generated/version/commonMain/kotlin")
+
+val generateVersion by tasks.registering {
+    val versionValue = (project.findProperty("VERSION_NAME") as String?)
+        ?: error("VERSION_NAME is not set in gradle.properties")
+    inputs.property("version", versionValue)
+    outputs.dir(versionOutputDir)
+    doLast {
+        // Reject a non-semver VERSION_NAME rather than baking a malformed token.
+        require(Regex("""^\d+\.\d+\.\d+(?:-[0-9A-Za-z.]{1,20})?$""").matches(versionValue)) {
```
cadivus: Accept valid SemVer prerelease identifiers

This pattern rejects valid SemVer prerelease identifiers containing a hyphen, such as 1.2.3-rc-1, while accepting invalid forms such as 1.2.3-alpha..1. A valid release version can therefore make generateVersion fail, while malformed values can be baked into the constant.

.github/workflows/native-kotlin-analysis.yml
```diff
@@ -0,0 +1,65 @@
+name: Native Kotlin Analysis
+
+# Version-sync drift check + runtime unit tests for the native Kotlin SDK.
+# Path-filtered so it only runs for PRs that touch native/kotlin.
```
cadivus: Correct the workflow path-filter comment

The workflow also runs when this workflow file changes, even if native/kotlin does not. Replace these lines with one accurate description of the path filter.

```suggestion
# Runs for pull requests that change native/kotlin or this workflow.
```

native/kotlin/scripts/sync-version.sh
```diff
@@ -6,23 +6,47 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
+# pattern matches nothing, which would otherwise leave a stale version and
+# silently publish/tag the wrong number.
 if ! grep -q '^VERSION_NAME=' "$PROPS"; then
   echo "ERROR: no VERSION_NAME= line found in $PROPS" >&2
   exit 1
 fi
 
+# --check verifies VERSION_NAME matches package.json without writing. A stale
+# value silently bakes the wrong runtime constant at build time and publishes
+# the wrong version, so CI runs this on every native/kotlin change.
```
cadivus: Limit the CI claim to pull requests

CI does not run this check on every native/kotlin change. The workflow has a pull_request trigger but no push trigger, so direct branch changes are outside the automated check.

```suggestion
# --check compares VERSION_NAME with package.json without writing. CI runs it
# for native/kotlin pull requests to prevent stale runtime and release versions.
```

native/kotlin/scripts/sync-version.sh
```diff
@@ -6,23 +6,47 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # overridden (e.g. by tests) to point at a fixture directory.
 PROJECT_DIR="${PROJECT_DIR:-$(cd "$SCRIPT_DIR/.." && pwd)}"
```
cadivus: Why is this a shell script but a similar task in #565 is a node script?

cadivus (changes requested):

native/kotlin/runtime/build.gradle.kts
VarshithaPamisetty: Yeah that's intentional for now. This PR just lands the token. Wiring it onto the client requests comes in the follow-up.

native/kotlin/runtime/build.gradle.kts
VarshithaPamisetty: Thanks for pointing it out, fixed it so 1.2.3-rc-1 passes and 1.2.3-alpha..1 is rejected.

.github/workflows/native-kotlin-analysis.yml
VarshithaPamisetty: Applied

native/kotlin/scripts/sync-version.sh
VarshithaPamisetty: Applied

native/kotlin/runtime/src/commonTest/kotlin/com/aws/blocks/kotlin/VersionTest.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+import io.kotest.matchers.string.shouldContain
+import io.kotest.matchers.string.shouldMatch
+import kotlin.test.Test
+
+class VersionTest {
+
+    // The token format this library commits to: aws-blocks-<lang>/<semver> —
+    // a bounded lowercase-alphanumeric language segment plus a strict semver.
```
cadivus: Do not call this grammar strict SemVer

The regex rejects build metadata and accepts numeric prerelease identifiers with leading zeroes, so it does not enforce strict SemVer. The comment overstates the contract and can mislead future compatibility checks.

```suggestion
    // Token format: aws-blocks-<lang>/<version>. The language is 1-16 lowercase alphanumeric
    // characters starting with a letter; version is numeric X.Y.Z with an optional prerelease and no build metadata.
```

.github/workflows/native-kotlin-analysis.yml
```diff
@@ -0,0 +1,103 @@
+name: Native Kotlin Analysis
+
+# Version-sync drift check + runtime unit tests for the native Kotlin SDK.
+# Runs for pull requests that change native/kotlin or this workflow.
```
cadivus: Suggestion: Remove the workflow narration

The workflow name, job names, pull_request trigger, and path list already state both facts. This block adds no constraint or rationale and makes readers scan the same information twice.

.github/workflows/native-kotlin-analysis.yml
```diff
@@ -0,0 +1,103 @@
...
+
+      - uses: actions/setup-node@v7
+        with:
+          node-version-file: '.nvmrc'
+
+      # The runtime version constant is generated at build time from
+      # gradle.properties VERSION_NAME; this asserts that stays in sync with
+      # package.json (the changesets source of truth). Token-semver validation
+      # lives in the generateVersion Gradle task, exercised by the runtime
+      # unit tests below.
```
cadivus: Suggestion: Condense the version-check explanation

This five-line block can preserve its source-of-truth and validation details in two direct sentences. The current wording buries those facts in build-time narration and indirect references.

```suggestion
      # Keep the generated runtime version aligned with package.json, the changesets source of truth.
      # Runtime tests exercise the generateVersion task's token-semver validation.
```

.github/workflows/native-kotlin-analysis.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Check gradle.properties version matches package.json
+        working-directory: native/kotlin
+        run: scripts/sync-version.sh --check
+
+      # Run the drift guard's own tests so its behavior cannot silently regress.
```
cadivus: Suggestion: Remove the generic test rationale

The adjacent step is already named Run version-sync script tests. Preventing regressions is the purpose of any test, so this comment adds no project-specific contract or rationale.

native/kotlin/runtime/build.gradle.kts
```diff
@@ -56,3 +56,37 @@ kotlin {
...
+// from VERSION_NAME (unlike Swift/Dart, which commit the constant).
+val versionOutputDir = layout.buildDirectory.dir("generated/version/commonMain/kotlin")
+
+val generateVersion by tasks.registering {
+    val versionValue = (project.findProperty("VERSION_NAME") as String?)
+        ?: error("VERSION_NAME is not set in gradle.properties")
+    inputs.property("version", versionValue)
+    outputs.dir(versionOutputDir)
+    doLast {
+        // Reject a non-semver VERSION_NAME rather than baking a malformed token.
```
cadivus: Do not describe the relaxed grammar as SemVer validation

The adjacent regex accepts leading zeroes in core and numeric prerelease identifiers, so it does not reject every version that SemVer forbids. The comment should describe the narrower token grammar actually enforced.

```suggestion
        // Reject versions outside the N.N.N[-prerelease] token format.
```

.github/workflows/native-kotlin-analysis.yml
```diff
@@ -0,0 +1,103 @@
...
+      - name: Run version-sync script tests
+        working-directory: native/kotlin
+        run: bash scripts/sync-version.test.sh
```
cadivus: Can we run all future `.test.sh` too, something like `bash scripts/*.test.sh`?

Similar to the other PR:
```suggestion
      - name: Test the Kotlin scripts
        run: |
          for test in native/kotlin/scripts/*.test.sh; do
            echo "== $test =="
            bash "$test"
          done
```

cadivus (changes requested):

native/kotlin/runtime/src/commonTest/kotlin/com/aws/blocks/kotlin/VersionTest.kt
VarshithaPamisetty: Applied

.github/workflows/native-kotlin-analysis.yml
VarshithaPamisetty: Applied

.github/workflows/native-kotlin-analysis.yml
VarshithaPamisetty: Removed it

.github/workflows/native-kotlin-analysis.yml
VarshithaPamisetty: Removed it

native/kotlin/runtime/build.gradle.kts
VarshithaPamisetty: Applied

.github/workflows/native-kotlin-analysis.yml
VarshithaPamisetty: Added

native/kotlin/runtime/build.gradle.kts
```diff
@@ -56,3 +56,36 @@ kotlin {
...
+    val versionValue = (project.findProperty("VERSION_NAME") as String?)
+        ?: error("VERSION_NAME is not set in gradle.properties")
+    inputs.property("version", versionValue)
+    outputs.dir(versionOutputDir)
+    doLast {
+        // Reject versions outside the N.N.N[-prerelease] token format.
+        require(Regex("""^\d+\.\d+\.\d+(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?$""").matches(versionValue)) {
+            "VERSION_NAME \"$versionValue\" is outside the N.N.N[-prerelease] token format (no build metadata)."
+        }
+        val pkgDir = versionOutputDir.get().dir("com/aws/blocks/kotlin").asFile
```
cadivus: generateVersion breaks the Gradle configuration cache

`doLast` captures versionOutputDir, a script-level val, which pulls the build-script object into the task's serialized state:

    $ ./gradlew :runtime:generateVersion --configuration-cache
    - Task `:runtime:generateVersion`: cannot serialize Gradle script
      object references
    Configuration cache entry discarded with 1 problem.

origin/main stores the entry cleanly, so this is the first such problem. CI stays green today because the cache is not enabled, but it blocks enabling it, and Gradle 10 makes it the default. Hoisting the provider into a task-local fixes it (verified).

```suggestion
val generateVersion by tasks.registering {
    val outputDir = versionOutputDir
    val versionValue = (project.findProperty("VERSION_NAME") as String?)
        ?: error("VERSION_NAME is not set in gradle.properties")
    inputs.property("version", versionValue)
    outputs.dir(outputDir)
    doLast {
        // Reject versions outside the N.N.N[-prerelease] token format.
        require(Regex("""^\d+\.\d+\.\d+(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?$""").matches(versionValue)) {
            "VERSION_NAME \"$versionValue\" is outside the N.N.N[-prerelease] token format (no build metadata)."
        }
        val pkgDir = outputDir.get().dir("com/aws/blocks/kotlin").asFile
```

cadivus (changes requested):

cadivus (approved):

soberm (approved):

PR #568, opened 2026-09-18
ci(agent-bench): provision Playwright chromium before the agent phase, capture command output

sharonyajain:
## Problem

The agent-bench runner has environment instability that inflates cell variance far beyond the effect a PR is measured against, so single-run bench comparisons can't be trusted. Two concrete faults, plus a diagnosis gap:

- **Chromium is provisioned too late.** The verifier (`3-build-and-test.sh`) installs the Playwright browser, but that is step 3 — *after* the agent's edit phase (step 2). When the agent tries a browser smoke-test while editing, no chrome binary exists yet, so it burns cycles hunting for one (observed in traces: `find / -name chrome-linux`, guessing `/usr/bin/google-chrome`, `npx playwright install chromium` mid-build). On one cell this drove a 4× token/cycle blow-up.
- **The dev server does not reliably stay up**, so agents loop on restart/poll churn — but this can't be root-caused because…
- **Traces record tool CALLS but not shell OUTPUT.** The `commands.jsonl` this PR adds is what makes the dev-server failure diagnosable next time.

**Issue #, if available:** #1132

## Changes

1. **Chromium before the agent phase.** Add a job-level `PLAYWRIGHT_BROWSERS_PATH` (a fixed dir under `github.workspace`) and a `Provision Playwright chromium` step between `1. Init` and `2. Agent run`, pinned to the verifier's `1.60.0`, `continue-on-error`. Step 2's agent shell inherits the job env, so it and step 3 resolve chromium from the same cache. `3-build-and-test.sh` keeps its install as a guarded fallback (a cache-hit no-op when pre-provisioned), so it still works standalone.

2. **Capture command output.** In `WorkspaceSandbox.executeStreaming` (`lib/run-shell.ts`), append one JSON line per command — `{ts, cwd, command, exitCode, stdout≤4000, stderr≤4000}` — to the path in `BENCH_CMD_LOG`, best-effort in `try/catch` so it can never break a command. `2-agent-run.ts` points `BENCH_CMD_LOG` at a `commands.jsonl` beside the trace, and the `6b. Upload trace` step uploads it with the bench-trace artifact.

3. **Dev-server stability is deliberately out of scope here.** A scope note + `TODO` in `3-build-and-test.sh` records that the deep fix (removing the deploy-poll loops) is delivered by the separate `BLOCKS_DEPLOYED` readiness signal; the verifier's reap/free-port logic is untouched, and no `NODE_OPTIONS` memory ceiling is guessed without a repro.

## Validation

- `bash -n` passes on the edited `3-build-and-test.sh`; `agent-bench.yml` parses as YAML.
- The two `.ts` edits are two stdlib imports (`node:path`, `node:fs`) plus guarded string ops; they were **manually type-reviewed** — `tsc` could not run in the authoring environment (no offline `tsc`; `npx` hit the internal-registry auth wall). CI's typecheck confirms on this PR.
- Runtime behavior (the new workflow step, the command-log append) was not executed locally — it needs a full Actions run; this PR's own bench run is the first live exercise, and if part 1 works its cells should show fewer dev-server/chromium churn cycles than baseline.
- Caveat: `npx playwright install --with-deps chromium` runs `apt-get` for OS libs (needs root). Fine on GitHub-hosted runners; on a self-hosted runner without passwordless apt it may partial-fail, but the step is `continue-on-error` and step 3's fallback covers it.

## Checklist

- [x] PR description included


scripts/agent-bench/steps/install-chromium-shim.sh
```diff
@@ -0,0 +1,55 @@
...
+seen=" "
+for name in chromium chromium-browser google-chrome; do
+  bin="$(command -v "$name" 2>/dev/null || true)"
+  [ -n "$bin" ] || continue
+  real="$(readlink -f "$bin")"
+  [ -n "$real" ] || continue
+  case "$seen" in *" $real "*) continue ;; esac
+  seen="$seen$real "
+  wrap_real "$real"
+done
```
soberm, line 80: [minor] The shim wraps only chromium resolved from PATH (`command -v chromium|chromium-browser|google-chrome`), but the preceding `npx playwright@1.60.0 install --with-deps chromium` places its browser under `PLAYWRIGHT_BROWSERS_PATH` (the job's `.pw-browsers` dir), not on PATH. On a stock GitHub-hosted ubuntu image there may be no system `chromium`/`google-chrome` on PATH, in which case this loop finds nothing and the shim becomes a no-op — the agent's ad-hoc `/usr/bin/chromium` launches could still hit the crashpad abort the shim exists to prevent. The step is `continue-on-error` and the header comment acknowledges an unwrapped chromium still runs, so this isn't fatal, but the stated benefit (fewer wasted agent turns hunting/repairing chromium) may not materialize.

Fix: On this PR's own bench run, confirm the `chromium shim installed:` output actually lists a wrapped binary. If PATH resolution is empty on the runner, consider also wrapping the Playwright-installed binary at `$PLAYWRIGHT_BROWSERS_PATH/chromium-*/chrome-linux/chrome`, or symlinking it onto PATH so the agent's absolute-path launches resolve to a wrapped file.

scripts/agent-bench/steps/install-chromium-shim.sh
```diff
@@ -0,0 +1,55 @@
...
+  [ -e "$saved" ] && return 0
+  # Never wrap a file that is itself a saved-aside real binary.
+  case "$real" in *.real) return 0 ;; esac
+
+  mv "$real" "$saved"
+  cat > "$real" <<EOF
+#!/usr/bin/env bash
+# Auto-generated chromium wrapper (agent-bench): inject startup flags, then exec the real binary.
+exec "${saved}" ${SHIM_FLAGS} "\$@"
+EOF
```
soberm, line 45: [minor] By replacing the real chromium file in place and prepending `SHIM_FLAGS` to every exec (L33, no `--type` guard), the flags are also injected when chromium re-launches itself for zygote/renderer/GPU child processes (chromium spawns children via its own binary path, now the wrapper). Those subprocesses generally tolerate the extra `--disable-*` flags, but re-injecting startup flags on every internal fork is a slightly wider blast radius than wrapping just the top-level launch.

Fix: Worth verifying on the live run that a wrapped chromium completes a headless navigation without child-process argv complaints; if it does, no change needed. Otherwise, gate the flag injection to top-level launches (e.g. only prepend when `--type` is absent from the incoming args).

scripts/agent-bench/steps/2-agent-run.ts
```diff
@@ -35,6 +36,15 @@ const TASK_PROMPT_PATH = required('TASK_PROMPT', '[bench]');
 const OUTPUT = required('OUTPUT', '[bench]');
 const TRACE_PATH = process.env.TRACE;
 const METRICS_PATH = process.env.METRICS;
+// Point the sandbox's per-command log at commands.jsonl beside the trace (so the
+// bench-trace upload picks it up; absent on a local run with no TRACE). Set before any invoke() so
+// every cycle is captured.
+if (TRACE_PATH && !process.env.BENCH_CMD_LOG) {
+	process.env.BENCH_CMD_LOG = join(dirname(TRACE_PATH), 'commands.jsonl');
```
soberm: [nit] `BENCH_CMD_LOG` is derived as `join(dirname(TRACE_PATH), 'commands.jsonl')`, while the workflow uploads the hardcoded `/tmp/commands.jsonl` (agent-bench.yml:376). These agree only because `TRACE` is set to `/tmp/trace.json`. Quiet coupling: if `TRACE` ever moves off `/tmp`, the upload path and the actual log path silently diverge and the artifact goes missing with no error.

Fix: Either drop a one-line comment at agent-bench.yml:376 noting the path is derived from TRACE's dirname, or key the upload off the same derivation, so the two can't drift apart unnoticed.

scripts/agent-bench/steps/install-chromium-shim.sh
sharonyajain: Fixed. The shim now also wraps the Playwright-installed binary: it globs $PLAYWRIGHT_BROWSERS_PATH/chromium-*/chrome-linux/chrome (guarded by [ -x ] and PLAYWRIGHT_BROWSERS_PATH being set) alongside the three PATH names, deduped by realpath.Header comment updated to reflect the wider coverage.

scripts/agent-bench/steps/install-chromium-shim.sh
sharonyajain: Fixed by scoping injection to top-level launches. The generated wrapper now scans its args and, on any --type=…/--type (which chromium always passes when it re-launches itself for zygote/renderer/gpu children), execs the real binary with no added flags; SHIM_FLAGS are prepended only when --type is absent. That keeps the crashpad/breakpad/dev-shm/gpu flags on the top-level launch we care about and drops the wider per-fork blast radius.

scripts/agent-bench/steps/2-agent-run.ts
sharonyajain: Fixed with the documentation option. Added a comment at the upload step in agent-bench.yml noting that commands.jsonl is written by 2-agent-run.ts as dirname(TRACE)/commands.jsonl, that the hardcoded /tmp/commands.jsonl matches only because TRACE is /tmp/trace.json, and that both must move together.

soberm (approved):

sarayev (approved):

PR #569, opened 2026-09-18
feat(core): registry-driven routing fan-out (path-addressed namespaces)

Simone319:
> **Stacked on #564** — base branch is \`zimzha/api-front-door\`, not \`main\`, so this PR's diff shows only the fan-out change. Review/merge #564 first; GitHub will auto-retarget this PR to \`main\` when #564 lands. (Draft until #564 merges.)

## What

Build the API front door's CloudFront behaviors from **one route table** — the existing \`RawRoute\` registry (\`getRegisteredRoutes()\`) — instead of hardcoding the reserved behaviors and looping over \`RawRoute\`s separately. Each registered route carries the \`endpoint\` of the compute that serves it, so a single \`addRouteBehaviors()\` pass emits one behavior per distinct \`(path, endpoint)\` and dedupes origins by endpoint. Both front-door paths (the managed distribution and Hosting's own distribution) now route identically from that table.

## Wire protocol change

An API namespace is now **addressed by path**: a call to namespace \`foo\` posts to \`/aws-blocks/api/foo\` (was \`/aws-blocks/api\` with the namespace only in the JSON-RPC body). The namespace still travels in the body and the server still dispatches on it; the path is a CloudFront routing hint. The whole \`/aws-blocks/api\` subtree is treated as RPC via \`isRpcPath()\` (Lambda handler + local dev server).

## No-op on the wire today

Every route currently resolves to the default compute, so \`addRouteBehaviors()\` collapses to exactly the behaviors #564 emitted by hand. This is deliberately a no-op until the assignment surface (a later PR) gives a namespace a non-default compute — at which point the *same* registration carries a non-default endpoint and the *same* front-door code fans that path out, with **no change here**.

## Deferred (no observable effect while every route is default-endpoint)

- Per-stack \`stackKey\` registry scoping
- Routability guard rejecting assignment to an ingress-less compute
- Native (Swift/Kotlin/Dart) codegen path-addressing

## Public surface (\`@aws-blocks/core\`)

- \`RegisteredRoute\` gains optional \`endpoint\` and \`subtree\`.
- \`registerRoutingEntry()\` — registers a handler-less, routing-only entry (bypasses the reserved-path guard; user \`RawRoute\`s still cannot claim \`/aws-blocks/api*\`). \`matchRoute()\` skips handler-less entries; \`isDispatchRoute()\` narrows to an invokable route.
- \`isRpcPath(pathname)\` — true for the bare RPC prefix and its subtree.
- Removes the unused \`Compute.namespaces\` field.

Minor bump for \`@aws-blocks/core\` (pre-1.0 minor = behavior change): the client's RPC URL and the synthesized CloudFront behavior set both change.


packages/core/src/cdk/api-front-door.ts
```diff
@@ -98,6 +100,106 @@ export function httpOriginFromEndpoint(endpoint: string): IOrigin {
...
+
+	for (const route of routes) {
+		const endpoint = route.endpoint ?? defaultEndpoint;
+
+		// Routing-only namespace entry (`/aws-blocks/api/{ns}` + subtree). Always
+		// emitted: on a default-compute namespace this is redundant with the RPC
+		// catch-all added last, but a redundant behavior points at the same origin,
+		// so it is inert — and emitting unconditionally is what lets both front-door
+		// paths share this code with no mode flag.
+		if (route.subtree) {
```
osama-rizk, line 175: This emits the per-namespace pair (`/aws-blocks/api/{ns}` + `/aws-blocks/api/{ns}/*`) **unconditionally** — with no skip when `endpoint === defaultEndpoint`. Since `api.ts` registers one subtree routing-only entry per API namespace, the **managed** front door goes from #564's **1 behavior** (just `defaultBehavior`) to `defaultBehavior + 2N + 3` (N = namespace count).

Two consequences:
- **Deploy fails at scale.** CloudFront's default quota is **25 cache behaviors per distribution**; at ~11 namespaces (2×11+3 = 25) you hit it and CloudFormation errors creating the distribution. #564 never approached this — an app with a dozen `ApiNamespace`s couldn't deploy.
- **Not a no-op.** Today every namespace is default-endpoint, so these 2N behaviors route *identically* to the `/aws-blocks/api/*` catch-all added last — pure redundancy. The PR/changeset claim that `addRouteBehaviors()` "collapses to exactly the behaviors #564 emitted by hand" is inaccurate; it emits 2 extra per namespace.

Fix: emit the per-namespace pair **only when `endpoint !== defaultEndpoint`** (i.e. actually fanned out to a non-default compute). That preserves the shared-code / no-mode-flag design, makes it a *true* no-op today (just the 3 reserved catch-alls), and defers the quota pressure to real multi-compute apps — where you'd also document raising the CloudFront behavior quota.

test-apps/comprehensive/test/auth-cookie-attrs.test.ts
```diff
@@ -45,7 +45,7 @@ async function rpcSetCookies(
   method: string,
   args: unknown[],
 ): Promise<string[]> {
-  const resp = await fetch(`${baseUrl}/aws-blocks/api`, {
+  const resp = await fetch(`${baseUrl}/aws-blocks/api/api`, {
```
osama-rizk, line 48: This test was changed from POSTing the **bare** `/aws-blocks/api` to `/aws-blocks/api/api`, which removes the only integration-level exercise of the bare-prefix path. Bare-prefix backward-compat is exactly the contract the deferred native (Swift/Kotlin/Dart) codegen depends on — and it's now covered only by the `isRpcPath(BLOCKS_RPC_PREFIX) === true` unit assertion. Worth keeping (or adding) one handler/integration test that POSTs the bare prefix and asserts it still dispatches, so a future change can't silently break native clients.

Separately (same PR-wide gap): nothing asserts the load-bearing **path/body-mismatch** dispatch — the two updated integ tests use a path segment that *matches* the body namespace, so they wouldn't catch a regression that started keying dispatch off the path. A handler-level test with a mismatched path vs body would lock that invariant.

packages/core/src/client/index.ts
```diff
@@ -231,10 +231,36 @@ function processResponse(data: unknown): unknown {
...
+ * serves every namespace path from the same origin, so this is safe there too. The
+ * namespace also stays in the JSON-RPC body, which is what the server dispatches on;
+ * the path is purely for routing.
+ *
+ * `baseUrl` may be relative (Hosting publishes `/aws-blocks/api` for same-origin
+ * browser calls), so this does deliberate string concatenation rather than going
+ * through `new URL()`.
+ */
+function namespaceRequestUrl(baseUrl: string, namespace: string): string {
+  return `${baseUrl.replace(/\/+$/, '')}/${namespace}`;
```
osama-rizk, line 261: Minor: `namespaceRequestUrl` interpolates the raw `namespace` into the path with no `encodeURIComponent` (same for `registerNamespaceRoute` in `api.ts`). Namespaces are developer-chosen export identifiers, so in practice URL-safe — but a name containing a space / `?` / `#` / unicode would produce a malformed path (and a path/behavior mismatch once fan-out is live). A one-line "namespace names must be URL-path-safe" note, or an assertion at registration, closes it cheaply.

osama-rizk (commented):
The wire protocol is sound — I verified (with a deep pass) that the client appends `/{namespace}` exactly once (trailing-slash safe, no double-prefix), both the Lambda handler and the dev server classify the whole subtree as RPC via the shared `isRpcPath`, dispatch stays body-driven (a path/body mismatch still dispatches on the body — the path is advisory), and bare-prefix callers (native, deferred codegen) keep working since `isRpcPath` matches the bare prefix too. The reserved-path guard still rejects user `RawRoute`s under `/aws-blocks/api*`, and `matchRoute`/`isDispatchRoute` never dispatch a handler-less routing entry. Nice single-table design shared by both front-door paths.

The one thing to fix before merge is the CloudFront behavior *count* (F1 inline) — it isn't the no-op it's described as. Plus two body-level notes:

**F-registry (MED, inferred):** `getRegisteredRoutes()` reads a `globalThis`-scoped registry (`raw-route.ts:~202`), not a per-stack one, so in a multi-stack synth every stack's front door emits behaviors for every *other* stack's namespaces — pointing `/aws-blocks/api/{other-ns}` at another stack's endpoint (wrong routing) and compounding F1's count. You explicitly defer per-stack `stackKey` scoping, so it's acknowledged — but note it *ships* incorrect/extra behaviors for multi-stack apps rather than being inert, and it interacts with F1's quota math. Same class as the per-CDK-stack scoping gap raised on #391. Single-stack (the common case) is unaffected.

**F-handler (LOW):** `lambda-handler.ts:539` still guards with raw `requestPath.startsWith(BLOCKS_RPC_PREFIX)` instead of `isRpcPath`, so a prefix-sibling like `/aws-blocks/apix` (which `isRpcPath` deliberately excludes, per `constants.test.ts`) is RPC-dispatched in Lambda but 404'd in the dev server — the exact "works in one runtime, 404s in the other" divergence this PR centralizes `isRpcPath` to prevent. A nonsense path no real caller hits, but use `isRpcPath(requestPath)` (or `startsWith(prefix + '/')`) here too.

packages/core/src/hosting.ts
```diff
@@ -893,44 +905,18 @@ export class Hosting extends Construct {
...
-          `[Hosting] ⚠️  RawRoute '${route.path}' creates CloudFront behavior '${behaviorPattern}' ` +
-            'which may shadow SSR/frontend routes under the same prefix. ' +
-            `Consider placing this route under ${BLOCKS_RPC_PREFIX}/ to avoid conflicts.`,
-        );
-      }
-
-      addedPatterns.add(behaviorPattern);
-      hosting.distribution.addBehavior(behaviorPattern, apiOrigin, API_BEHAVIOR_OPTIONS);
-    }
+    addRouteBehaviors(hosting.distribution, getRegisteredRoutes(), api.defaultEndpoint, new Map([[api.defaultEndpoint, apiOrigin]]));
```
osama-rizk, line 911: **[suggestion] Lifecycle asymmetry undercuts the "both front doors route identically" invariant.** This reads `getRegisteredRoutes()` at **construction** time (as the docstring above now admits — a `RawRoute` registered after `new Hosting(...)` gets no behavior), whereas the Blocks-owned `ApiFrontDoorAspect.visit` reads the registry at **synth** time.

Scenario: a `RawRoute`/`ApiNamespace` constructed *after* `new Hosting(...)` lands on the managed distribution but **not** on Hosting's — so the two paths diverge despite the PR's headline claim. Inert today (all routes resolve to the default origin, and routes are usually declared first), but a real divergence once fan-out is load-bearing. Worth either reading the registry lazily/at-synth on both paths, or documenting the ordering requirement as a hard constraint.

packages/core/src/cdk/api-front-door.ts
```diff
@@ -98,6 +100,106 @@ export function httpOriginFromEndpoint(endpoint: string): IOrigin {
...
+	distribution: Distribution,
+	routes: readonly RegisteredRoute[],
+	defaultEndpoint: string,
+	sharedOrigins?: Map<string, IOrigin>,
+): void {
+	const originFor = sharedOrigins ?? new Map<string, IOrigin>();
+	const added = new Set<string>();
+
+	const addBehavior = (pattern: string, endpoint: string): void => {
+		if (added.has(pattern)) return;
```
osama-rizk, line 153: **[suggestion] Dedupe is by CloudFront pattern only, not `(pattern, endpoint)`.** `added.has(pattern)` collapses two routes that truncate to the same wildcard.

Scenario: `/users/{id}` and `/users/{id}/posts` both truncate to `/users/*`; today they share the default endpoint so it's harmless, but once the assignment surface gives them **different** endpoints, the second route's traffic silently routes to the first's origin — a misroute with no warning. Consider keying the set on `(pattern, endpoint)` and emitting a warning (or error) when the same pattern maps to two distinct endpoints, before the assignment PR makes this reachable.

packages/core/src/raw-route.ts
```diff
@@ -277,6 +303,58 @@ export function registerRoute(options: RawRouteOptions & { path: string }): void
...
+export function registerRoutingEntry(options: RoutingEntryOptions): void {
+  const state = getState();
+  if (state.locked) {
+    throw new Error('Routes must be registered during initialization. Cannot register routes after handler creation.');
+  }
+
+  const normalizedPath = options.path.replace(/\/+/g, '/').replace(/\/$/, '') || '/';
+
+  const existing = state.routes.find((r) => r.handler === undefined && r.path === normalizedPath);
+  if (existing) return;
```
osama-rizk, line 347: **[nit] Routing-entry idempotency keeps the first endpoint; a re-registration with a changed endpoint is silently ignored.** The guard matches on `(handler === undefined && path)` and returns early, discarding a new `endpoint`/`subtree`.

Fine for re-imports (identical values). But when the future assignment surface re-registers a namespace with a non-default endpoint, the stale entry wins unless the surface clears the registry first. Latent, tied to the same future work — worth a note here so the assignment PR knows it must clear/replace rather than re-register.

osama-rizk (commented):
Second pass — the routing/guard/parity areas I re-checked are clean (reserved-path guard holds, handler-less entries provably can't reach dispatch, `isRpcPath` parity between the Lambda handler and dev server is intact, `Compute.namespaces` removal is complete, native clients keep working). Three latent findings, all of which only bite once the future assignment surface gives a namespace a non-default endpoint — flagging them now so they're tracked for that PR rather than discovered there.

packages/core/src/cdk/api-front-door.ts
Simone319: Good catch — this is a real deploy-blocker at scale, not just redundancy. Emitting the `/{ns}` + `/{ns}/*` pair for **every** namespace adds two CloudFront behaviors each, so an app with ~11 namespaces would blow past CloudFront's default 25-behaviors-per-distribution quota and fail to deploy — for no routing gain, since on the default compute every one of those behaviors points at the same origin the RPC catch-all already covers.

Fixed: `addRouteBehaviors` now emits the pair **only when the namespace fans out to a non-default origin** (`endpoint !== defaultEndpoint`). A default-compute namespace gets no dedicated behavior and falls through to the `/aws-blocks/api/*` catch-all. The behavior count is now bounded (RawRoutes + the two reserved subtrees) regardless of namespace count, and the fan-out path is unchanged — an assigned namespace still gets its exact/subtree pair ahead of the catch-all.

This is still a true no-op on the wire today (every namespace resolves to the default compute), and it keeps the no-mode-flag, shared-code design. Follow-ups landed in the same change:
- inverted the unit assertion in `api-front-door.test.ts` (a default-endpoint namespace now gets **no** dedicated behavior);
- corrected the `registry-driven-routing-fan-out` changeset, which had described the per-path emission as unconditional.

test-apps/comprehensive/test/auth-cookie-attrs.test.ts
Simone319: Agreed — changing this call site to the namespaced path (`/aws-blocks/api/api`) lost the coverage that the **bare** `/aws-blocks/api` prefix still dispatches, which is exactly the backward-compat guarantee this PR must not silently drop.

Rather than re-broaden this cookie-focused test, I added a dedicated, handler-level block in `packages/core/src/lambda-handler.test.ts` (`RPC path is advisory, dispatch is body-driven`) that pins the contract directly:
1. a bare `/aws-blocks/api` POST dispatches from the JSON-RPC body;
2. a request whose **path** names one namespace but whose **body** method is `api.echo` still dispatches `api.echo` — proving the path is only a CloudFront routing hint and the body is authoritative;
3. `/aws-blocks/apiXYZ` (shares the prefix but is not the RPC subtree) 404s instead of falling through to RPC dispatch — the prefix-boundary guard (see the handler fix below).

Handler-level rather than an integration test on the dev server: it controls both the path and the body namespace directly, needs no deploy, and reads the exact seam the reviewer flagged.

packages/core/src/client/index.ts
Simone319: The namespace name flows into two places that must agree: the client builds `${baseUrl}/${namespace}` and the front door matches a CloudFront behavior on that exact `/aws-blocks/api/${namespace}` path. `encodeURIComponent` at the client would paper over a bad name on the client side while the synthesized behavior pattern kept the raw name — silently diverging the two, so a "safe" client URL would route to nothing.

In practice namespace names are JS identifiers (they're property keys on the exported API object), so they're already URL-path-safe; the gap was only that nothing enforced it. So instead of encoding, I added a fail-fast assertion at registration (`registerNamespaceRoute`): a name must match `^[A-Za-z0-9_-]+$`, else it throws at definition time (local/synth) with a message pointing at the `/aws-blocks/api/<name>` path segment. That catches a malformed name before deploy and keeps the client URL and the CloudFront pattern identical by construction.

On the breaking-change flag: this isn't a regression. A name outside `^[A-Za-z0-9_-]+$` (e.g. a dot) was already broken before this PR — RPC dispatch splits `<ns>.<method>` on the first `.`, so a dot-name never dispatched. The assertion just surfaces that at definition time instead of failing mysteriously at runtime; it rejects only input that was already non-functional, and hyphens/underscores (the realistic non-alphanumerics) are explicitly allowed.

Covered by two new tests in `packages/core/src/api.test.ts`: one asserts a dot-name and a slash-name both throw `/is not URL-path-safe/` and register nothing; the other pins that `my-api_v2` is accepted.

Simone319:
Two responses to points raised in the review summary (not tied to a single line):

## RPC prefix-boundary handling in `lambda-handler.ts`
Right — using a raw `startsWith(BLOCKS_RPC_PREFIX)` here was inconsistent with the `isRpcPath()` guard just above it, and it left a boundary hole. Inside `if (!isRpcPath(requestPath))` we already know the path is not the RPC subtree, so the inner `!requestPath.startsWith(BLOCKS_RPC_PREFIX)` was both redundant *and* wrong for a look-alike like `/aws-blocks/apiXYZ`: that path is not RPC (so it enters the branch) but *does* start with the prefix, so the old check skipped the 404 and let it fall through to RPC dispatch.

Fixed by dropping the redundant inner guard entirely: within the `!isRpcPath` branch, an unmatched path always 404s (and logs the diagnostic). `BLOCKS_RPC_PREFIX` is no longer needed in this file, so its import is removed. Covered by the new `/aws-blocks/apiXYZ` → 404 case in `lambda-handler.test.ts`.

## `globalThis` routing registry
Acknowledged — the routing registry hanging off `globalThis` is a real smell. It's addressed project-wide by #572, which de-globalizes the registry (keys it under the resolved Blocks root via `getBlocksRoot()` instead of a bare `globalThis` symbol), so I'd rather not add a second, divergent fix here.

Note the merge-ordering dependency: #572 should land before this branch merges, so the two don't fight over the registry's home. For the single-stack case this PR targets, the current placement is behaviorally correct — the concern is multi-copy / cross-stack isolation, which is exactly what #572 fixes. No code change here.

PR #573, opened 2026-09-18
feat(compute): declare a compute by what a workload needs

Simone319:
## What

PR3 of the multi-compute rollout. Introduces `ComputeProvider.provide(id, requirements?, scope?)` so a workload declares **what it needs** — `{ timeoutSeconds, memoryMb }` — rather than naming an AWS service. Today every declaration resolves to a serverless (Lambda) compute; when another fulfillment exists the same declaration resolves differently with no call-site change.

```ts
import { ComputeProvider } from '@aws-blocks/blocks';

const reports = ComputeProvider.provide('reports', { timeoutSeconds: 60 * 4, memoryMb: 1024 });
```

## How

- **`packages/blocks/src/compute-provider.ts`** — one module holds `ComputeRequirements`, `SERVERLESS_COMPUTE_LIMITS`, `validateComputeRequirements` (pure, runs in **every** condition; **fails rather than clamps**, naming field/value/ceiling), and the shared `provideCompute` returning the condition-resolved compute as `unknown`.
- **Two conditional entries share that one JS** and differ only in the type they promise (the repo sets no `customConditions`, so `tsc` resolves `bb-lambda-compute` to its mock types everywhere — hence the cast):
  - `index.ts` → runtime `Scope` handle, **no guard** (tolerates a missing stack so a declared compute is inert in Lambda / local dev).
  - `index.cdk.ts` → `Compute`, **adds** the "declared before the stack existed" guard (under CDK a missing parent is a real mistake).
- **`bb-lambda-compute`** — `timeout` (`Duration | number`) and `memorySize` become configurable props, defaulting to the previous hardcoded 15 min / 2048 MiB, so an app that never configures a compute is unaffected.
- **`Compute`** is now exported **type-only** from `@aws-blocks/core/cdk` (the runtime value stays on `/cdk/internal` to avoid a module-init TDZ cycle with `class Scope`).

## Tests

Validation boundaries; the runtime/mock inert-handle path; CDK synth (requirements→props mapping, single registration, the pre-stack guard); and the direct-construction `cdk.Duration` timeout path.

## Scope

Nothing consumes the provider yet — the assignment surface (`{ compute }` on namespaces/workers) ships in a later PR. Standalone and independently mergeable; touches no routing.

DoD gate green: `build`, `lint:deps`, `lint`, unit tests, `test:e2e:local`. Reviewed with both the code-review agent and the built-in `/code-review`; all agreed findings applied.

Draft — opening for early review.


packages/blocks/src/index.cdk.ts
```diff
@@ -194,6 +199,70 @@ export { Tracer } from '@aws-blocks/bb-tracer';
...
+	 *   Validated against the limits of the compute it resolves to, so a workload
+	 *   that cannot be hosted fails at synth rather than deploying and then timing
+	 *   out. Omitted fields take the resolved compute's defaults.
+	 * @param scope - Where to attach. Defaults to the ambient `BlocksStack`, which
+	 *   is where the framework parents its own default compute — so resource names
+	 *   never depend on where in the app tree the compute was declared. Pass this
+	 *   explicitly when the ambient stack is ambiguous (two `BlocksBackend`s in one
+	 *   parent stack).
+	 */
+	provide(id: string, requirements?: ComputeRequirements, scope?: ScopeParent): Compute {
```
soberm, line 242: [major] This provisions a real `LambdaCompute` (function + log group + execution role) that nothing can route to yet — the assignment surface ships in a later PR. A user who follows the docstring example and deploys gets a billed, idle Lambda with no way to use it, and there's no synth-time guard flagging a declared-but-unassigned compute.

The `internal.ts` comment this PR deletes explicitly committed to avoiding exactly this: *"a compute a customer can declare but that is silently ignored is a worse experience than not having the feature."* Can we either (a) hold `provide()` until the assignment surface lands so it never exists in a "creates a Lambda that does nothing" state, or (b) add a synth-time guard/warning for a declared compute with no route?

Relatedly: this is PR3 of the rollout and there's no second fulfillment on `main`, so today `provide()` is a more constrained equivalent of `new LambdaCompute(scope, id, { timeout, memorySize })` (which this PR also keeps). What's the reasoning for landing the public API before the consumer that makes it functional, rather than together with it?

packages/blocks/src/compute-provider.ts
```diff
@@ -0,0 +1,208 @@
...
+			throw new Error(
+				`Invalid compute requirement${where}: timeoutSeconds must be a positive whole number of ` +
+					`seconds, got ${timeoutSeconds}.`,
+			);
+		}
+		// Subsumed by the positive-integer check while `minTimeoutSeconds` is 1 (an
+		// integer that passed that check is already >= 1), so this cannot fire today.
+		// Kept coupled to the limit so raising the floor makes it enforce, matching the
+		// reachable `memoryMb` floor below rather than leaving the two fields asymmetric.
+		if (timeoutSeconds < limits.minTimeoutSeconds) {
```
soberm, line 130: [minor] This branch can't fire today: `minTimeoutSeconds` is 1 and the `isPositiveInteger` check above already guarantees >= 1, as the comment notes. Keeping it coupled to the limit for symmetry is reasonable, but it's untested dead code — no test can exercise it. Suggest either an explicit coverage-ignore marker to signal the untestable-by-design intent, or dropping it until the floor is actually raised.

packages/blocks/src/compute-provider.ts
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * The limits of the serverless fulfillment — the only compute Phase 1 can
+ * resolve requirements to.
+ *
+ * Stated here rather than inside the compute package so the requirements surface
+ * can reject an impossible request before any concrete compute is constructed,
+ * and so the error text is identical wherever the check runs.
+ */
+export const SERVERLESS_COMPUTE_LIMITS = {
```
soberm, line 83: [minor] Two gaps in the validation vs. the fulfillment it validates against:

- `memoryMb` isn't checked for Lambda's 64 MiB step granularity — a value like `200` passes here but Lambda rejects it at deploy, which reintroduces the "deploys then fails" outcome this validation exists to prevent.
- `SERVERLESS_COMPUTE_LIMITS` is a single module-level const, but the stated goal is a second fulfillment that maps the same requirements differently — with different limits. Worth a note on how these limits become per-fulfillment, since the current single-const shape doesn't anticipate it.

packages/blocks/src/index.ts
```diff
@@ -425,6 +433,28 @@ export type { AnnotationValue, Segment, TracerOptions } from '@aws-blocks/bb-tra
...
+ * compute still points at something, and no CDK is pulled into the Lambda or the
+ * browser.
+ *
+ * Shares its implementation with the `cdk` entry (./compute-provider.ts); only the
+ * declared return type differs. Must stay export-compatible with that entry — a
+ * customer's backend module imports this name under both conditions.
+ */
+export const ComputeProvider = {
+	/** See the `cdk` entry for the full contract. Inert at runtime. */
+	provide(id: string, requirements?: ComputeRequirements, scope?: ScopeParent): Scope {
```
soberm, line 453: [question] At runtime this returns an inert handle typed as `Scope`, and under CDK it returns a `Compute`. Since nothing consumes the return value yet, is there a case today where a customer's backend module actually reads the returned handle at runtime? If not, it may be worth documenting that the runtime return is currently only "the import resolves and the reference is non-null" — so a reviewer doesn't assume the two conditions return interchangeable capabilities.

packages/blocks/src/index.cdk.ts
sharonyajain: 🔴 **Public, documented provide() ships a PR early: a declared compute silently stands up a real, unroutable Lambda + public API Gateway**

The new public `ComputeProvider.provide()` (and the README section "Declaring a compute", which tells customers to call it now) provisions a full compute under CDK: `provide()` -> `provideCompute()` -> `new LambdaCompute(parent, id, {...})`, whose constructor builds a Lambda NodejsFunction, a handler LogGroup, an execution role, self-registers on the stack, AND creates a public `apigateway.RestApi` ("Blocks API") with a root/`/aws-blocks/api` `anyMethod` proxy integration and a live `apiUrl`. No assignment/routing surface exists yet (the PR itself says "Nothing consumes the provider yet"), so nothing can direct traffic to this compute through the framework. In the same diff, internal.ts's explicit commitment is deleted: "It must NOT be made public before then: a compute a customer can declare but that is silently ignored is a worse experience than not having the feature." The PR description makes `Compute` public but never acknowledges or justifies reversing that commitment, and adds no synth-time guard rejecting a declared-but-unassigned compute. (The CI checks on this SHA are green, but a green build cannot detect a premature-public-surface / silent-no-op design defect — it is a semantic gap, not a lint/type gap.)

```typescript
provide(id: string, requirements?: ComputeRequirements, scope?: ScopeParent): Compute {
		...
		return provideCompute(id, requirements, parent) as Compute;
	},
```

**Why it matters:** cause: a public, README-documented provider is shipped one PR before any assignment/routing surface and with no unassigned-compute guard -> mechanism: a customer follows the docs, calls `ComputeProvider.provide('reports', {...})`, and `cdk synth` provisions a real Lambda plus its own public API Gateway (anyMethod proxy, public apiUrl) that no framework routing points at and no error/guard flags -> consequence: (1) a deployed feature that silently does nothing for the workload — the exact below-bar UX the codebase committed to avoid; and (2) a security/attack-surface impact via an explicit threat chain: entry point = the auto-created public `apiUrl` for the declared compute; trust boundary = public internet -> backend handler through the anyMethod proxy; mechanism = the app author never opted into or is tracking this second public ingress and cannot manage it (no assignment/routing to disable or scope it); impact = unintended, undocumented public exposure of a handler endpoint plus wasted/duplicated infrastructure, shipped by a call whose stated purpose was merely "give this workload more memory".

**Suggestion:** Until the assignment+routing surface exists, either keep the provider on the internal path (`@aws-blocks/core/cdk/internal`) / undocumented (revert the README + public `Compute` exposure), or add the synth-time guard the deleted internal.ts comment promised: fail `cdk synth` when a declared compute is never assigned/routed, so a declared-but-unusable compute can never deploy silently. If public exposure now is deliberate, state in the PR why the previously-documented "must not be public before assignment" commitment no longer applies and have the appropriate owners sign off (posture-bearing public-API change).

_[code-review-sage]_

PR #582, opened 2026-09-21
feat: forward native client user-agent token to AWS SDK

VarshithaPamisetty:
## Problem

Native runtime libraries (Swift, Kotlin, Dart) now carry a version token, but nothing forwards it to AWS. So server-side traffic cannot be attributed to the client library that produced it.

## Fix

`@aws-blocks/core` reads a custom `x-blocks-user-agent` header on the RPC request and validates it against a strict grammar (length-capped at 128 bytes; malformed values are dropped). It carries the token per request in an `AsyncLocalStorage`, and a new `installClientUserAgent` middleware appends it to the outgoing AWS SDK user agent as a `client/<lang>/<version>` pair. A DynamoDB call from a Swift app renders as:

```
aws-sdk-js/3.x ... aws-blocks/0.1.1 client/swift/0.1.1 bb/KVStore/0.1.1
```

Every Building Block that configures an SDK client installs the middleware, and a CI check (`check:client-user-agent`) fails if a `customUserAgent` site is missing it. The header value stays raw (`aws-blocks-swift/0.1.1`); the server maps it to the `client/...` pair before appending, matching the existing `bb/...` convention.

## Validation

- Unit tests: grammar (accepts a not-yet-shipped language, rejects injection, malformed, and oversized input), the middleware append plus `x-amz-user-agent` fallback plus no-op-when-absent, and no token leak between requests.
- A `lambda-handler` end-to-end test: inbound header, validate, store, middleware, outgoing user agent.
- A `bb-kv-store` integration test drives a real `KVStore.get()` and asserts the token lands on the outgoing DynamoDB user agent.

Note:
1. The native clients (Swift, Kotlin, Dart) that actually send `x-blocks-user-agent` are not in this PR; that is Phase 3. Until then nothing sets the header, so this is inert with no behavior change on its own.


packages/core/src/server/client-user-agent.ts
```diff
@@ -0,0 +1,94 @@
...
+  // single structural type fits every client.
+  middlewareStack: { add: (middleware: any, options: any) => void };
+}): void {
+  client.middlewareStack.add(
+    (next: (args: { request?: unknown }) => Promise<unknown>) =>
+      async (args: { request?: unknown }): Promise<unknown> => {
+        const token = getClientUserAgentToken();
+        if (token && hasHeaders(args.request)) {
+          const headers = args.request.headers;
+          const key = 'user-agent' in headers ? 'user-agent' : 'x-amz-user-agent';
```
soberm: [minor] The key pick `'user-agent' in headers ? 'user-agent' : 'x-amz-user-agent'` assumes only one of the two is ever present. SDK v3 sets one or the other at the `build` step, so in practice this is fine — but if a request somehow carried both, only `user-agent` would be appended to and `x-amz-user-agent` (the header AWS telemetry often actually reads) would be left untouched.

Suggestion: either a one-line comment stating the "SDK v3 sets exactly one" assumption, or appending to whichever is present.

packages/core/src/server/client-user-agent.ts
```diff
@@ -0,0 +1,94 @@
...
+
+/** Defensive cap on header length (characters), applied before parsing. */
+const MAX_CLIENT_USER_AGENT_LENGTH = 128;
+
+/**
+ * Full-string grammar for the `aws-blocks-<lang>/<version>` token; the fixed
+ * shape blocks injection.
+ */
+const CLIENT_USER_AGENT_PATTERN =
+  /^aws-blocks-([a-z][a-z0-9]{0,15})\/(\d+\.\d+\.\d+(?:-[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?)$/;
```
soberm: [minor] The `{0,15}` cap on the language segment means any language name longer than 16 chars is silently dropped as malformed. That's a fine defensive limit, but it's an undocumented constraint a future client author could trip over.

Suggestion: add the constraint to the JSDoc on `validateClientUserAgentToken`, e.g. "language: 1–16 lowercase alphanumerics, leading letter".

packages/bb-data/src/engines/data-api-engine.ts
```diff
@@ -153,6 +154,7 @@ export class DataApiEngine implements DatabaseEngine {
     this.client = config.client ?? new RDSDataClient({
       ...(config.customUserAgent ? { customUserAgent: config.customUserAgent } : {}),
     });
+    if (!config.client) installClientUserAgent(this.client);
```
soberm: [nit] The `!config.client` gate is correct — an injected client is the caller's responsibility. A short inline comment (e.g. `// Skip for injected test clients — caller's responsibility`) would save the next reader a double-take.

scripts/check-client-user-agent-consistency.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (!rel.startsWith("bb-")) return false; // only Building Blocks consume the helper
+  if (rel.includes("/templates/")) return false; // commented example, not a real client
+  return true;
+}
+
+/**
+ * Counts `customUserAgent` client-construction sites in a file, ignoring type
+ * fields, reassignments, and options merely forwarded to a non-client helper.
+ */
+function countClientUserAgentSites(content: string): number {
```
soberm: [nit] The line-based heuristic (`/\bcustomUserAgent\s*[:,}]/` + a file-level `new XxxClient(` match) works for the current one-client-per-file BBs but could miscount if a file later has two SDK clients where only one sets `customUserAgent`, or if the property spreads across lines. Acceptable as a guard — just flagging that AST-level checking would be more robust if this ever gets noisy.

soberm (commented):
Reviewed the full diff. The validation grammar, ALS-scoped isolation, call-time middleware read (correct for the realtime cached-client path), and test coverage all look solid. A few minor suggestions below — nothing blocking.

scripts/check-client-user-agent-consistency.ts
VarshithaPamisetty: Left it as-is for now since it counts correctly across all current BBs. 
Yes, AST-level parsing would be the more robust approach if a file later introduces multiple clients or a multi-line customUserAgent.

packages/bb-data/src/engines/data-api-engine.ts
VarshithaPamisetty: Added a 1 liner

packages/core/src/server/client-user-agent.ts
VarshithaPamisetty: Yes, modified the comment.

packages/core/src/server/client-user-agent.ts
VarshithaPamisetty: Thanks for pointing out. SDK v3 sets both `user-agent` and `x-amz-user-agent`, so switched to append to each present header + added a both-present test.

soberm (dismissed):

packages/core/src/lambda-handler.ts
```diff
@@ -521,7 +530,14 @@ function createHandler(backend: any) {
     // APIs server-side gets forwarded cookies.
     const inboundCookies = event.headers?.cookie || event.headers?.Cookie || '';
 
-    return requestCookies.run(inboundCookies, async () => {
+    // Validate the attacker-controlled header before storing, so malformed
+    // input never reaches the AWS user agent.
+    const clientUserAgent = validateClientUserAgentToken(
+      event.headers?.[CLIENT_USER_AGENT_HEADER] || event.headers?.['X-Blocks-User-Agent'],
+    );
```
cadivus: Allow the client user-agent header through CORS preflight

An allowed cross-origin browser client that sends `x-blocks-user-agent` must preflight the request, but the OPTIONS response still allows only `Content-Type, Authorization`. The browser therefore blocks the POST before this new header-handling path can run.

packages/core/src/lambda-handler.test.ts
```diff
@@ -1364,3 +1365,50 @@ describe('handleEventSourceRecords — SQS partial batch responses', () => {
...
+
+  it('appends a valid inbound token to the outgoing SDK user agent', async () => {
+    const result = await invoke(uaProbeBackend(), probeEvent('aws-blocks-swift/9.9.9'));
+    const body = JSON.parse(result.body);
+    assert.strictEqual(body.result, 'aws-sdk-js/3.700.0 aws-blocks/0.5.0 client/swift/9.9.9');
+  });
+
+  it('drops a malformed inbound token without changing the outgoing user agent', async () => {
+    // Invalid token (trailing garbage): the validator rejects it, so nothing
+    // is appended.
```
cadivus: Suggestion: Delete the repeated malformed-token explanation

The test title states that a malformed token is dropped, and the adjacent input visibly contains trailing garbage. This comment adds no contract or rationale.

scripts/check-client-user-agent-consistency.ts
cadivus: Describe the heuristic instead of exact site detection

The function counts matching property lines whenever the file contains any client constructor. It can count forwarded options in such a file, so the claim that those are ignored is false.

```suggestion
/**
 * Estimates client construction sites by counting `customUserAgent` property
 * lines in files that contain an SDK client constructor.
 */
```

cadivus (changes requested):
Mainly
> Allow the client user-agent header through CORS preflight

looks critical to me, the other is just a nit about a comment.

packages/core/src/lambda-handler.ts
VarshithaPamisetty: True. I added `x-blocks-user-agent` to the preflight Access-Control-Allow-Headers so browser clients aren't blocked.

scripts/check-client-user-agent-consistency.ts
VarshithaPamisetty: Applied

PR #583, opened 2026-09-21
fix(bb-auth-cognito): requireRole reads live group membership

osama-rizk:
## Problem

`requireRole` authorized against the signed-in user's `cognito:groups` **token claim** — a snapshot taken at sign-in. Consequences:

- **Grant doesn't apply:** after `auth.admin.addUserToGroup(user, 'admins')`, that user's live session kept getting **403** until their token refreshed or they re-logged in. This is the reported *"requireRole returns 403 for everyone / admin surface unreachable"* symptom.
- **Revoke doesn't apply:** `removeUserFromGroup` did **not** lock out a live session — the removed member kept access until their stale token expired.

Ref: `aws-amplify/private-amplify-backend-staging#417`

## Fix

`requireRole` now reads membership **live** at call time, mirroring how `fetchUserAttributes` already reads live rather than trusting the token:

- **AWS:** `AdminListGroupsForUser` (paginated).
- **Mock:** in-process `state.groups`.
- Returns a `CognitoUser` whose `.groups` reflects the live read.

Grants and revocations take effect on the user's **next request**, no re-login.

### Notable decisions
- **IAM:** `cognito-idp:AdminListGroupsForUser` is now granted to the execution role **unconditionally** (it backs a client-facing guard), independent of the opt-in `admin` surface. Every other `Admin*` action still requires opt-in — the CDK least-privilege test was updated to encode this new baseline.
- **Cost/scope:** one extra Cognito call per guarded request. The cheaper identity reads (`requireAuth` / `getCurrentUser` / `signIn`) deliberately still surface the cached claim; README/DESIGN document the split.

## Tests
- `admin.test.ts`: two new cases proving the bug both directions — sign-in-*then*-grant now passes; sign-in-then-*revoke* now locks out. Both failed before the fix.
- Full `bb-auth-cognito` suite (243) + full monorepo unit suite pass; `biome lint`, `lint:deps`, `check:api` clean; changeset added.

## Follow-up
Sandbox e2e against the real `AdminListGroupsForUser` path — running separately; will report results on this PR.


osama-rizk:
### Sandbox e2e ✅ (real Cognito, us-west-2)

Deployed the `comprehensive` backend to a sandbox and ran the full e2e against real Cognito, then tore it down. Added the previously-missing scenario (the existing sandbox test only did *add-then-fresh-sign-in*):

- **`requireRole reads live membership — group change hits an existing session without re-login`** → **pass**
  - sign in → `requireRole('admins')` 403 → admin `addUserToGroup` → `requireRole('admins')` **succeeds with no re-login** (returned `groups` includes `admins`) → admin `removeUserFromGroup` → `requireRole('admins')` 403 again.
  - This exercises the real `AdminListGroupsForUser` path and the new unconditional IAM grant.
- The whole `AuthCognito admin surface (sandbox)` suite passed. No auth-cognito regressions.

Also added a small `authCAdminRemoveFromGroup` route to the test app (it had `addToGroup`/`listGroupsForUser` but no `remove`) in commit `945e7b0`.

The run's only failures were the unrelated **EmailClient** suite (`send` / `sendBatch`) — the sandbox account has no verified SES domain (`noreply@example.com`), flagged as a prerequisite at deploy time.

packages/bb-auth-cognito/src/index.aws.ts
```diff
@@ -1604,10 +1604,42 @@ export class AuthCognito<const O extends AuthCognitoOptions = AuthCognitoOptions
...
+		try {
+			do {
+				const resp = await this.client.send(new AdminListGroupsForUserCommand({
+					UserPoolId: this.adminUserPoolId(), Username: username, NextToken: nextToken,
+				}));
+				for (const g of resp.Groups ?? []) if (g.GroupName) out.push(g.GroupName);
+				nextToken = resp.NextToken;
+			} while (nextToken);
+		} catch (e) {
+			throw asApiError(e);
```
soberm, line 1640: [major] This routes a Cognito SDK error through `asApiError`, which sets `cause: e` on the `ApiError` (see its definition ~L569). Per AGENTS.md Core Rule #5, `Error.cause` must not be attached enumerably — it serializes `$metadata`/ARNs to the client.

The leak in `asApiError` is pre-existing, but until now it only fired on **admin-gated** surfaces. `requireRole` is an unconditional, client-facing guard, so this is the first time a normal (non-admin) request can serialize SDK metadata to an ordinary client — a Cognito throttle/config error on any `requireRole` call now leaks. That's an expansion of the exposure surface, not a like-for-like reuse.

Options: sanitize on this hot path — `throw new ApiError(e instanceof Error ? e.message : 'AdminListGroupsForUser failed', 500, { name: e instanceof Error ? e.name : undefined })` (no `cause`) — or fix `asApiError` repo-wide to attach `cause` non-enumerably (`Object.defineProperty(err, 'cause', { value: e, enumerable: false })`), which also closes it on the existing admin paths.

packages/bb-auth-cognito/src/index.aws.ts
```diff
@@ -1604,10 +1604,42 @@ export class AuthCognito<const O extends AuthCognitoOptions = AuthCognitoOptions
...
 		const user = await this.requireAuth(context);
-		if (!user.groups.includes(role)) {
+		// Read membership live via `AdminListGroupsForUser` rather than trusting
+		// the session token's `cognito:groups` claim — the same reasoning as
+		// `fetchUserAttributes`. An admin who runs `addUserToGroup` /
+		// `removeUserFromGroup` against an already-signed-in user would otherwise
+		// not take effect (grant) or not revoke (removal) until that user's token
+		// next refreshed or they re-logged in. Costs one extra Cognito call per
+		// guarded request; the returned user's `groups` reflects the live read.
+		const liveGroups = await this.liveGroupsForUser(user.username);
```
soberm, line 1614: [minor] `requireRole` now issues a live `AdminListGroupsForUser` on every guarded request. Two consequences worth a note (and possibly a small mitigation):

- **Throttling → cascading auth failures.** Cognito's admin API quota is low and shared. Under load, throttling on this call turns into `requireRole` failures across all guarded routes at once. Because it's now on the hot auth path (not an occasional admin op), the blast radius is larger than before.
- **Cost/latency:** one extra round-trip per authorization; a request that calls `requireRole` twice pays twice.

Consider request-scoped memoization keyed by `username` so repeated `requireRole` calls within a single request dedupe to one Cognito call. The README/DESIGN "one extra Cognito call per guarded request" note is good — worth adding the throttle-amplification caveat there too.

packages/bb-auth-cognito/src/index.ts
```diff
@@ -1285,10 +1285,20 @@ export class AuthCognito<const O extends AuthCognitoMockOptions = AuthCognitoMoc
...
+		// Read membership live from the source of truth rather than the session
+		// token's `cognito:groups` claim — the same reasoning as
+		// `fetchUserAttributes`. An admin who runs `addUserToGroup` /
+		// `removeUserFromGroup` against an already-signed-in user would otherwise
+		// not take effect (grant) or not revoke (removal) until that user's token
+		// next refreshed or they re-logged in. The returned user's `groups`
+		// reflects the live read so callers never see a stale list. The mock
+		// derives from `this.state.groups`; the AWS runtime calls
+		// `AdminListGroupsForUser`.
+		const liveGroups = Object.keys(this.state.groups).filter((g) => this.state.groups[g].includes(user.username));
```
soberm, line 1297: [minor] `this.state.groups[g].includes(...)` assumes every group key holds an array. In practice the mock's own API always initializes keys to `[]` (createGroup) and only ever `filter()`s them, so this is safe today — but `loadState` rehydrates `raw.groups ?? {}` from a persisted blob, so a malformed/hand-edited `.bb-data` file with a null membership would throw `TypeError` here instead of behaving like the AWS path.

Cheap hardening: `(this.state.groups[g] ?? []).includes(user.username)`. Note the identical unguarded pattern already exists at L327 (`buildAdminSurface`) and L359 — this PR just propagates it, so fixing all three (or leaving all three) keeps them consistent. Not blocking.

packages/bb-auth-cognito/src/index.aws.ts
osama-rizk: Thanks — I dug in, and I don't think this reaches the client, so I've left `asApiError` in place. Two things close it:

1. `ApiError` passes `cause` to the native `Error(message, { cause })` constructor (`errors.ts:60`). Per ES2022 that installs `cause` **non-enumerable** — verified locally: `Object.getOwnPropertyDescriptor(err, 'cause').enumerable === false`, and `JSON.stringify(err) === '{}'`.
2. The wire encoder never serializes the error object anyway. `encodeRpcError` (`rpc.ts:205-210`) builds an explicit allowlist: `code` (= `status`), `message`, and `data.name`. `cause` / `$metadata` / ARNs are never read.

So a throttle/config error on `requireRole` sends the client only `{ status, message, name }` — e.g. `name: 'TooManyRequestsException'`, which is genuinely useful for clients to branch on. Keeping `asApiError` preserves that typed name; sanitizing to a generic 500 would drop it.

Core Rule #5 targets *enumerable* `cause` specifically, and this path is non-enumerable, so it already complies. If you'd still like defense-in-depth I'm glad to do it — either sanitize on this hot path, or make `asApiError`'s `cause` explicitly non-enumerable via `Object.defineProperty` as a repo-wide belt-and-suspenders. And if you had a serialization path other than `encodeRpcError` in mind where this reproduces, point me at it and I'll fix it there.

packages/bb-auth-cognito/src/index.aws.ts
osama-rizk: Done in `0d4b984`. `requireRole`'s `AdminListGroupsForUser` is now memoized per request on a `BlocksContext`-keyed `WeakMap` (then by username, caching the in-flight Promise so concurrent guards share one round trip). Multiple guards in one request → one Cognito call; entries GC with the per-request context so the shared Lambda singleton doesn't accumulate. Scope is deliberately one request, so a membership change is still observed on the next.

Also added the throttle-amplification caveat to both README and DESIGN as you suggested.

I did not add a dedupe-count unit test: this package's AWS test file only covers pure helpers — there's no Cognito-client mock harness in it, and pulling in `aws-sdk-client-mock` just for this felt like more surface than the optimization warrants. Happy to add it if you'd like the regression guard.

packages/bb-auth-cognito/src/index.ts
osama-rizk: Done in `0d4b984` — hardened all three sites (`requireRole`, `toAdminUser`, `listGroupsForUser`) with `(this.state.groups[g] ?? []).includes(...)`, so a malformed/hand-edited `.bb-data` with a null membership behaves like the AWS path instead of throwing `TypeError`. Fixed them together for the consistency you flagged.

PR #584, opened 2026-09-22
fix(bb-app-setting): validate option combinations in local mock (#577)

### LINKED ISSUE #577: AppSetting: validate deploy-invalid options in local mock
## Problem

`AppSetting` validates several invalid option combinations only in its CDK constructor. The same application can therefore start in local mock mode and fail later during sandbox or production synth.

A minimal example is accepted by the mock (which initializes an empty string) but rejected by CDK:

```ts
new AppSetting(scope, "client-id", { name: "google-client-id" });
```

The CDK layer rejects it because non-secret settings require an initial `value`. The same gap applies to `secret` plus `value`, `secret` plus `schema`, and `kmsKeyArn` without `secret: true`: the mock accepts or ignores these configurations while CDK rejects them.

This is especially easy to encounter in a local-first workflow: `npm run dev` appears healthy, but the deploy path is the first place that reports the configuration error.

## Proposed direction

Share the synchronous option-combination validation between the mock and CDK constructors, so local startup fails with the same `ValidationFailedException` before persisting mock data. The actual schema validation of a supplied value can remain where it is, since Standard Schema validation may be asynchronous.

## Why open this before a PR

Rejecting configurations that the local mock previously accepted tightens local behavior. The documented combinations are already invalid at CDK synth, but I would like maintainer guidance on whether matching that validation in local development is the intended compatibility policy before changing it.

## Related

The generated `db pull` AuthOIDC guide exposed this gap by creating a non-secret client-ID setting without a value; PR #575 corrects that generated example independently.

sarayev:
Fixes #577.

## Problem

`AppSetting`'s CDK constructor throws on several invalid option combinations at synth time, but the local mock (`index.mock.ts`) had none of those checks. So `npm run dev` stayed green on config that later failed at `ampx sandbox` / deploy, and a non-secret setting with no `value` silently defaulted to an empty string instead of erroring.

The combinations the CDK rejected but the mock did not:
- `secret` + `schema` together
- a `schema` with no `value`
- `kmsKeyArn` without `secret: true`
- an empty/whitespace `kmsKeyArn`
- a `secret` carrying a hardcoded `value`
- an `external` setting with a `value`, or without a `name`
- a non-secret setting with no `value`

## Change

Extracted the synchronous option-combination checks into a shared `validateAppSettingOptions()` in `packages/bb-app-setting/src/validation.ts`, following the existing `bb-metrics` / `bb-distributed-data` `validation.ts` pattern. Both the CDK and mock constructors now call it, so there's a single source of truth and local dev fails fast with the same `ValidationFailedException` the CDK throws.

Async Standard Schema *value* validation is unchanged and stays in each variant's `put()` path (it can be async and depends on the supplied value, not just the option shape).

## Behavior change

The local mock now **rejects** these combinations instead of accepting them. Any config relying on the old lax behavior (most notably a non-secret setting declared with no `value`) will now throw locally, exactly as it always would have at synth. Called out in the changeset.

## Tests

- Added an `Option validation (parity with CDK)` block to `index.test.ts`: one case per invalid combination, plus valid combinations asserting they do **not** throw.
- Full `bb-app-setting` suite passes locally (80/80, includes the existing CDK synth tests).

## Notes

Biome lint could not run in my local environment (the prebuilt biome binary needs a newer GLIBC than the box has); CI will lint. The code matches the surrounding files' style and the TS build is green.


soberm (dismissed):

packages/bb-app-setting/src/validation.ts
```diff
@@ -0,0 +1,98 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { AppSettingErrors } from './errors.js';
+import type { InternalAppSettingOptions } from './types.js';
+
+// Unlike the `blocksError` in index.mock.ts / index.aws.ts (which prefix the
+// message with `${name}: `), this one leaves the message unprefixed on purpose:
+// these checks moved out of the CDK constructor, which threw a plain message, so
+// keeping it unprefixed preserves the exact pre-refactor CDK synth error text.
+function blocksError(name: string, message: string): Error {
```
osama-rizk, line 16: There are now three `blocksError` helpers that diverge intentionally: this one is deliberately **unprefixed** (to preserve the pre-refactor CDK synth message), while `index.mock.ts` and `index.aws.ts` each keep a `blocksError` that **prefixes** `${name}: `. The comment explains it well, but three near-identical helpers with a subtle intentional difference is a small trap — a future "DRY these up" edit could re-prefix the synth text and break whatever pins it. Consider a one-line `@see` cross-reference to the other two (or a single shared helper taking an explicit `prefix` param) so the intent survives a later refactor.

packages/bb-app-setting/src/index.mock.ts
```diff
@@ -104,6 +105,12 @@ export class AppSetting<T = string> extends Scope {
 	constructor(scope: ScopeParent, id: string, options: AppSettingOptions<T>) {
 		super(id, { parent: scope, bbName: BB_NAME, bbVersion: BB_VERSION });
 		this.log = options?.logger ?? new Logger(this, 'logger', { level: 'error' });
+
+		// Synchronous option-combination checks, shared with the CDK variant
+		// (see validation.ts), so local dev rejects configs that would fail at
+		// synth instead of silently defaulting a missing value to ''.
+		validateAppSettingOptions(id, options as InternalAppSettingOptions<T>);
```
osama-rizk, line 115: Nit / awareness: after this change the mock emits two error-message shapes for the same construct — option-combination errors arrive as `AppSetting '<id>': …` (from `validation.ts`), while schema/value errors from the mock's local `blocksError` are `<name>: …`. Both identify the setting, so it's cosmetic, but a consumer asserting on the message prefix would see inconsistency. Not worth changing unless you want uniform mock error formatting; flagging for awareness.

osama-rizk (commented):
Model small refactor — I verified extraction fidelity line-by-line against base `main`: all 7 checks reproduced with identical conditions, message text, and first-match order (no check dropped, none silently added to the CDK path), the CDK now delegates at the constructor, and the mock calls it at `index.mock.ts:112` *before* the empty-string default at `:124` — so the non-secret-no-value bug is actually fixed. `fromExisting` routes through the constructor with `external:true`, so the external checks fire, and the parity tests exercise the mock and assert `err.name === ValidationFailed`. The behavior change only rejects configs that were never deployable anyway (the CDK always rejected them). Two cosmetic notes inline.

osama-rizk (approved):

packages/bb-app-setting/src/validation.ts
sarayev: Good catch — added `@see` cross-references on all three helpers plus a 'don't merge these into one prefixing helper' note on the validation.ts one, since re-prefixing it would change the CDK synth text that's pinned. Pushed in 7e1824d.

packages/bb-app-setting/src/index.mock.ts
sarayev: Agreed it's cosmetic — leaving the two shapes as-is for this PR since both identify the setting and unifying mock error formatting is a broader change. Thanks for flagging.

soberm (approved):

PR #588, opened 2026-09-22
feat(kotlin): OIDC sign-in on the iOS and JVM targets

mattcreaser:
## Problem

The Kotlin runtime's OIDC client only worked on Android. iOS and JVM/desktop had no launcher, so `signIn()` was unusable there. The Gradle property naming the app's relay target (`oidc { redirectUrl }`) also didn't match the backend's `allowedRelayOrigins`.

**Issue #, if available:**

## Changes

- **iOS:** sign-in presents an `ASWebAuthenticationSession`.
- **JVM:** opens the system browser and receives the relay redirect on a loopback address bound per sign-in, so desktop apps need no relay configuration at all.
- **`oidc { redirectUrl }` → `relayTo`**, matching the backend's `allowedRelayOrigins`. The old name still works with a deprecation warning; setting both to different values fails the build.
- The relay target is validated at build time, and codegen only stubs out OIDC when every target registers a scheme, so a misconfigured project fails at build rather than at sign-in.
- The platform launcher now owns the relay target, which is what lets JVM substitute its loopback address.
- **Example apps:** both the Android and KMP apps now sign in with OIDC instead of the password provider, and seed their initial state from `AuthApi.getAuthState()` so a launch following an earlier sign-in starts signed in.

## Validation

- Unit tests added for the relay-target gate (codegen), the `oidc` DSL and requirement resolution (plugin), and the JVM loopback session, browser opener, and iOS launcher (runtime). `:codegen:test :plugin:test :runtime:jvmTest` pass.
- `OidcE2ETest` covers the flow end to end.
- Manually verified: sign-in completes on the iOS simulator against the example backend with a real Google IdP.

## Checklist

- [x] PR description included

---


mattcreaser:
Note: There will be a follow-up PR to improve the post-sign in landing page experience on JVM

